PR #3, opened 2015-07-20
Switch submodule links to https

nulltoken:


shana:
Good catch, ssh urls were done on purpose for an internal CI but shouldn't be needed for the (upcoming) public CI

shana:
:sparkles:

zeusro:
support SSH?It's really a good news.

PR #21, opened 2015-07-30
Handle Enterprise http urls

### LINKED ISSUE #13: GitHub Enterprise Repo Links are always SSL
First of all, thanks! Really liking the plugin so far!

But, one minor issue is our internal github enterprise install is not SSL and the repo in Visual Studio 2015 is http://, but the plugin converts all the links "Pull Request", "Pulse", "Graphs", "Issues" to https:// breaking the feature for us.

Haacked:
Fixes #13

This PR not only fixes #13, but consolidates all the URL parsing of repository information by leveraging the UriString class adapted from GHfW.


src/GitHub.Exports/Primitives/UriString.cs
```diff
@@ -197,7 +166,7 @@ public override UriString Combine(string addition)
             if (url != null)
             {
                 var urlBuilder = new UriBuilder(url);
-                if (!String.IsNullOrEmpty(urlBuilder.Query))
+                if (!string.IsNullOrEmpty(urlBuilder.Query))
```
shana: Is there an advantage to doing this with the value type instead of the class type?

Haacked: Nope. I think it's my overeager R# that does this. We should pick an approach and stick to it. Do you have a preference?

BTW, this has nothing to do with value type or class type. `string` is just the shorthand for `System.String` but it's the exact same class. :smile:

src/GitHub.Exports/Services/Services.cs
```diff
@@ -144,21 +145,13 @@ public static Repository GetRepoFromSolution(this IVsSolution solution)
...
-            if (url.StartsWith("git@github.com:", StringComparison.Ordinal))
-                url = url.Replace("git@github.com:", "https://github.com/");
-            if (!Uri.TryCreate(url, UriKind.Absolute, out uri))
-                return null;
-            return uri;
+            return repo
+                ?.Network
+                .Remotes
+                .FirstOrDefault(x => x.Name.Equals("origin", StringComparison.Ordinal))
+                ?.Url;
```
shana: Gotta love them null conditional operators! :heart:

src/GitHub.VisualStudio/Base/TeamExplorerNavigationItemBase.cs
```diff
@@ -28,7 +28,14 @@ public TeamExplorerNavigationItemBase(ISimpleApiClientFactory apiFactory, ITeamE
...
 
-            OnThemeChanged();
+            try
+            {
+                OnThemeChanged();
+            }
+            catch (ArgumentNullException)
+            {
+                // This throws in the unit test runner.
+            }
```
shana: I think I prefer having this try/catch in the body of the method instead of here.

Haacked: Oh, within `OnThemeChanged`? I can do that.

src/GitHub.Exports/Primitives/UriString.cs
shana: > Nope. I think it's my overeager R# that does this. We should pick an approach and stick to it. Do you have a preference?

Weeell, I always type String.Something out of habit, so I'm sure the code is way more sprinkled with the uppercase version than the lowercase one (and I doubt I can teach my fingers otherwise...) :smile:

> BTW, this has nothing to do with value type or class type. string is just the shorthand for System.String but it's the exact same class. :smile:

Yeah, I keep forgetting we're in .NET land where there are no crazy peeps doing weird stuff with jit and string/String differences...

src/GitHub.VisualStudio/Base/TeamExplorerNavigationItemBase.cs
shana: Yeah, in `OnThemeChanged`.
I can imagine a situation where the event is actually triggered during a manual unit test run inside VS (the theme changed event is sometimes triggered automatically when VS syncs settings in ze cloud), so we might end up blowing up in the handler that's right after this line, as well.

Haacked:
:pineapple:

shana:
:ice_cream:

PR #26, opened 2015-07-31
Fix packaging of libgit2sharp binaries

### LINKED ISSUE #25: LibGit2Sharp binaries are not properly packaged
The native libraries aren't being properly packaged during the build, which causes pinvoke errors when certain features are used.

shana:
Fixes #25

libgit2sharp.csproj is including a custom rule during the AfterBuild step to manually copy the native libs over. Unfortunately, this doesn't plug in very nicely to the dependency discovery logic of msbuild, so other projects that reference libgit2sharp have no clue that there are these extra files it depends upon. It assumes all you want to do is copy files to the destination build directory, which it assumes is the same for all projects, which might work fine for a normal project, but definitely fails miserably for packaging a vsix. Because (all together now) "Assumptions are the mother of all [...]s".

I imagine it was done so the native libraries don't show up in the Solution Explorer, but meh, don't care much about that m'self. Also, some of my build tweaks might be causing the rule not to run properly (so files don't get copied), so it's not all libgit2sharp's fault. Oh well...


Haacked:
selfie-1

Issue #16, opened 2015-07-29
List of GitHub repositories in Connect Page


shana:
Imported from https://github.com/github/VisualStudioInternal/issues/372

Add list of known local GitHub repo clones in the Connect page, under the clone/create/sign out links.

![githubsection](https://cloud.githubusercontent.com/assets/9871368/7401775/85a9768c-ee7b-11e4-83a6-a9689bbf8da7.jpg)

shana:
Fixed in #32

PR #32, opened 2015-08-05
Add list of repos to the connect page

shana:
I'm not sure it's worth caching the known GitHub repositories list in our connection cache instead of just poking the registry every time the Connect page is loaded. I'll have to monitor the registry for changes so I can update our cache when new repos get added to VS (because there's no API to notify us of that)


shana:
### This is what it looks like now
screen shot 2015-08-24 at 11 35 11 pm

### Highlighting
screen shot 2015-08-24 at 11 35 34 pm

### Opening a repo
If an entry is double clicked or cloning is finished, file explorer with solution filter is run:

screen shot 2015-08-24 at 11 36 12 pm

shana:
Fixes #16

shana:
Since I'm already hooking up into a property change that tells me the cloning is done, we can also do it for the repo creation and maybe switch out the message. Could be a nice thing to have.

shana:
:pineapple: 

I haven't added tests for the new classes yet, doing that tomorrow, probably on a separate branch off of this one.

/cc @Haacked

shana:
Should I order the repo list?

Haacked:
> If an entry is double clicked or cloning is finished, file explorer with solution filter is run:

Shouldn't we do the same thing that double clicking a repository listed in the "Local Git Repositories" section?

> Should I order the repo list?

Yeah, I think so.

shana:
>> If an entry is double clicked or cloning is finished, file explorer with solution filter is run:

>Shouldn't we do the same thing that double clicking a repository listed in the "Local Git Repositories" section?

I wish, but there's no API to trigger that at all. It's switching the source control context (closing opened solutions and changing to another repo) without actually opening anything in the new repo. It's all private logic that we have no access to.

TBH, I think offering to open a solution in the selected repo when you double click is better than blindly closing the current one and switching to the selected repo without doing anything else or asking the user. It's a very confusing experience, you don't know what happened and VS enters into this limbo state where nothing is loaded but there's some sort of source control context that's active, which changes all the default paths for things (and you only notice if you pay particular attention to the default path in the project creation dialog). The fact that it closes the currently opened solution without asking annoys me particularly, I don't like things happening without warning, but that's just me :stuck_out_tongue:

Haacked:
:thumbsup: to that reasoning.

shana:
I've added sorting to the list, fixed up the styles.

I'm also displaying a notification when a repo is created that replaces the cloning notification. It's hacky, but should fix #54 

Another thing I did was invoke the Create New Project dialog when you create a new repo - if you're on VS creating a new github repo, it's highly likely that you're going to put a VS project in it. This is a change of behavior in regards to what Create repo does now. Opinions, yay, nay?

Haacked:
Trying it out now and will let you know. :smile:

Haacked:
> Another thing I did was invoke the Create New Project dialog when you create a new repo - if you're on VS creating a new github repo, it's highly likely that you're going to put a VS project in it. This is a change of behavior in regards to what Create repo does now. Opinions, yay, nay?

I say "nay" to this for the moment. I think this PR is taking on too much and anything like this should be done separately.

Also, here's what happened when I ran through the flow, I created a repository named "MyGreatRepo".

new project

Notice that the default path is wrong. Also, there's a checkbox "Add to source control". What would happen if that's checked?

Also, when I hit "Cancel", the same dialog popped up again. It seemed like it did it once per each repository.

> I'm also displaying a notification when a repo is created that replaces the cloning notification. It's hacky, but should fix #54

:thumbsup: This is a nice touch. I think this should be done in a separate PR though. :smile:

> I've added sorting to the list, fixed up the styles

Looks great. I noticed that after logging in, it wasn't populated. It wasn't till I switched away and switch back that the list was populated. Known issue?

shana:
> I say "nay" to this for the moment. I think this PR is taking on too much and anything like this should be done separately.

Right, good point. I'll take it out.

> Notice that the default path is wrong. Also, there's a checkbox "Add to source control". What would happen if that's checked?

Adding to source control means it will start tracking the files (doing a git add on everything)

> Also, when I hit "Cancel", the same dialog popped up again. It seemed like it did it once per each repository.

Oh, ah, bug. I thought I fixed that... That's likely why the path isn't populated correctly, it's not supposed to be showing the dialog multiple times.

>  This is a nice touch. I think this should be done in a separate PR though. 

Yeah, I wanted to, but it basically comes for free with this PR (which already needs to track when the cloning is finished so it can update the repo list). I can take out the notification call and put it on a separate PR based on this one, so we can discuss the message to show there.

> Looks great. I noticed that after logging in, it wasn't populated. It wasn't till I switched away and switch back that the list was populated. Known issue?

Nope, it should be populated, that's a bug, I forgot to call Refresh when we're done logging in.

Haacked:
> Yeah, I wanted to, but it basically comes for free with this PR (which already needs to track when the cloning is finished so it can update the repo list). I can take out the notification call and put it on a separate PR based on this one, so we can discuss the message to show there.

Yeah. I want to try the experience out without it popping up the dialog. I found it to be a bit jarring.

src/GitHub.App/Models/RepositoryModel.cs
```diff
@@ -1,28 +1,68 @@
...
-                    : Octicon.repo;
+            SetIcon(isPrivate, isFork);
         }
 
-        public string Name { get; private set; }
-
         public IAccount Owner { get; private set; }
+        public override int GetHashCode()
+        {
+            return (Owner != null ? 0 : Owner.GetHashCode()) ^ base.GetHashCode();
```
Haacked: I think you meant `(Owner == null ? 0 : Owner.GetHashCode()) ^ base.GetHashCode();` on this line. Otherwise you could get a `NullReferenceException`

shana:
> Yeah. I want to try the experience out without it popping up the dialog. I found it to be a bit jarring.

Yeah, I agree. The message is just a notification, not a popup, so that's not going to be jarring. I've set up a separate PR (#59) for it anyway, it doesn't need to be here (but it's based on this one so reviewing that one means only looking at the one commit that has the message)

src/GitHub.App/Models/RepositoryModel.cs
shana: Sharp :eyes:!

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -86,11 +89,60 @@ public string GetActiveRepoPath()
...
+            return Registry.CurrentUser.OpenSubKey(TEGitKey + "\\" + path, true);
+        }
+
+        static IEnumerable<ISimpleRepositoryModel> PokeTheRegistryForRepositoryList()
+        {
+            using (var key = OpenGitKey("Repositories"))
+            {
+                return key.GetSubKeyNames().Select(x =>
+                {
+                    using (var subkey = key.OpenSubKey(x))
```
Haacked: Isn't it possible that `subkey` could be null here? Or that `OpenGitKey` could throw an exception such as a `SecurityException`? We should make sure we're handling exceptions in the calls to `PokeTheRegistryForRepositoryList`

src/GitHub.VisualStudio/Services/ConnectionManager.cs
```diff
@@ -81,17 +91,27 @@ public ConnectionManager(IProgram program, Rothko.IOperatingSystem os)
...
             return true;
         }
 
+        void AddConnection(Uri hostUrl, string username /*, [AllowNull] IEnumerable<RepositoryCacheItem> repositories */)
+        {
+            var address = HostAddress.Create(hostUrl);
+            if (Connections.FirstOrDefault(x => x.HostAddress == address) != null)
+                return;
+            var conn = SetupConnection(address, username);
+            //repositories?.ForEach(r => conn.Repositories.Add(new SimpleRepositoryModel(r.Name, r.CloneUrl, r.LocalPath)));
```
Haacked: Was this comment a leftover from something else? Could it be removed?

src/GitHub.VisualStudio/Services/ConnectionManager.cs
```diff
@@ -169,7 +209,13 @@ void SaveConnectionsToCache()
 
             var cache = new CacheData();
             cache.connections = Connections.Select(conn =>
-                new ConnectionCacheItem { HostUrl = conn.HostAddress.WebUri, UserName = conn.Username });
+                new ConnectionCacheItem
+                {
+                    HostUrl = conn.HostAddress.WebUri,
+                    UserName = conn.Username,
+                    /* Repositories = conn.Repositories.Select(x =>
```
Haacked: Ditto this comment. Comments like this make me think the code was commented out for testing and that the developer forgot to uncomment it. :stuck_out_tongue:

src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs
```diff
@@ -38,16 +49,32 @@ bool LoggedIn
...
+            [return:AllowNull]
+            get { return repositories; }
+            set { repositories = value; this.RaisePropertyChange(); }
         }
 
+        ISimpleRepositoryModel selectedRepository;
+        [AllowNull]
+        public ISimpleRepositoryModel SelectedRepository {
+            [return: AllowNull] get { return selectedRepository; }
+            set { selectedRepository = value; this.RaisePropertyChange(); } }
```
Haacked: Nitpick Neymar: brace formatting here is inconsistent with the rest of the code. :smile:

src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs
```diff
@@ -129,23 +182,70 @@ void UpdateConnection()
         void OnPropertyChange(object sender, System.ComponentModel.PropertyChangedEventArgs e)
```
Haacked: You could probably omit the `System.CompenentModel.` part here since it's already declared in the usings section.

src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs
```diff
@@ -129,23 +182,70 @@ void UpdateConnection()
...
+            {
+                e.NewItems
+                    .Cast<ISimpleRepositoryModel>()
+                    .Select(async r =>
+                {
+                    if (String.Equals(Holder.ActiveRepo?.RepositoryPath, r.LocalPath, StringComparison.CurrentCultureIgnoreCase))
+                        SelectedRepository = r;
+                    var repo = await ApiFactory.Create(r.CloneUrl).GetRepository();
+                    r.SetIcon(repo.Private, repo.Fork);
+                }).ToList();
```
Haacked: The use of `ToList` here is confusing because we're not using the return value.

I think `ForEach` would more clearly express the intent here.

```csharp
                e.NewItems
                    .Cast<ISimpleRepositoryModel>()
                    .ForEach(async r =>
                {
                    if (String.Equals(Holder.ActiveRepo?.RepositoryPath, r.LocalPath, StringComparison.CurrentCultureIgnoreCase))
                        SelectedRepository = r;
                    var repo = await ApiFactory.Create(r.CloneUrl).GetRepository();
                    r.SetIcon(repo.Private, repo.Fork);
                });
```

src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs
```diff
@@ -156,11 +256,59 @@ public void Login()
             StartFlow(UIControllerFlow.Authentication);
         }
 
+        public void OpenRepository()
+        {
+            var old = Repositories.FirstOrDefault(x => String.Equals(Holder.ActiveRepo?.RepositoryPath, x.LocalPath, StringComparison.CurrentCultureIgnoreCase));
+            // open the solution selection dialog when the user wants to switch to a different repo
+            // since there's no other way of changing the source control context in VS
+            if (SelectedRepository != old)
```
Haacked: I haven't thought this through, but this is a reference equality check. Should this be:

```csharp
if (!Equals(SelectedRepository, old))
```

src/GitHub.VisualStudio/UI/Views/GitHubConnectContent.xaml.cs
```diff
@@ -39,23 +41,74 @@ void signOut_Click(object sender, RoutedEventArgs e)
...
+        public object Convert(object[] values, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (values.Length != 2 || parameter as string != "FormatRepositoryName")
+                return String.Empty;
+
+            if (!(values[1] is IGitHubConnectSection))
+            {
+                if (values[0] is ISimpleRepositoryModel)
+                    return ((ISimpleRepositoryModel)values[0]).Name;
+                return String.Empty;
```
Haacked: This is probably more a code style issue than an actual perf issue since array access and casting types is pretty cheap, but I would probably write the above as:

It probably doesn't matter for perf, but I tend to replace double cast checks with something like:

```csharp
var model = values[0] as ISimpleRepositoryModel;
return model != null ? model.Name : String.Empty;
```

That accesses the array only once and type checks only once. However, more importantly to me, I think that's a bit more readable.

Haacked: Or even better.

``csharp
var model = values[0] as ISimpleRepositoryModel;
return model?.Name ?? String.Empty;
```

**Comment by Haacked** [comment]
> Yeah, I agree. The message is just a notification, not a popup, so that's not going to be jarring. I've set up a separate PR (#59) for it anyway, it doesn't need to be here (but it's based on this one so reviewing that one means only looking at the one commit that has the message)

Sorry, i don't think i was being clear. I thin that popping up the New Project dialog after creating a new repository should be done in a separate PR too. I don't think it belongs in this PR which should solely be focused on listing repositories.

File: src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs (hunk 8 above)
Comment (shana) [inline src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs]: It's on purpose, the equality operators are overriden in SimpleRepositoryModel to perform comparisons based on content (because the instances are different, maybe not their content). There's comments in SimpleRepositoryModel to that effect.

File: src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs (hunk 7 above)
Comment (shana) [inline src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs]: Yup, ForEach is better! :+1:

File: src/GitHub.VisualStudio/Services/ConnectionManager.cs (hunk 3 above)
Comment (shana) [inline src/GitHub.VisualStudio/Services/ConnectionManager.cs]: Yeah, I need to take this out, I don't think we want to cache this information in our cache file (that's what it was doing)

**Comment by shana** [comment]
> Sorry, i don't think i was being clear. I think that popping up the New Project dialog after creating a new repository should be done in a separate PR too. I don't think it belongs in this PR which should solely be focused on listing repositories.

Oh yeah, I'm taking that part out too :smile:

File: src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs (hunk 8 above)
Comment (shana) [inline src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs]: nvm, I took the overloaded operators out and switched to Equals on this

File: src/GitHub.Exports/Services/VSServices.cs (hunk 2 above)
Comment (shana) [inline src/GitHub.Exports/Services/VSServices.cs]: Added the exception checks on the callers!

**Comment by shana** [comment]
Ok, I think I caught all the issues and fixed all the things! :ice_cream:

File: src/GitHub.VisualStudio/UI/Views/GitHubConnectContent.xaml.cs [hunk 10]
```diff
@@ -39,23 +41,70 @@ void signOut_Click(object sender, RoutedEventArgs e)
...
+        {
+            if (values.Length != 2 || parameter as string != "FormatRepositoryName")
+                return String.Empty;
+
+            var item = values[0] as ISimpleRepositoryModel;
+            var context = values[1] as IGitHubConnectSection;
+            if (context == null)
+                return item?.Name ?? String.Empty;
+            
+            if (context.SectionConnection.Username == item.CloneUrl.Owner)
```
Haacked: If `context` not null, but `item` was null, we'd get a `NullReferenceException` here. Perhaps line 79 should be `if (context == null || item == null)`

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -86,11 +89,68 @@ public string GetActiveRepoPath()
...
+            }
+        }
+
+        const string PathsKey = @"Software\Microsoft\VisualStudio\14.0\NewProjectDialog\MRUSettingsLocalProjectLocationEntries";
+        public string SetDefaultProjectPath(string path)
+        {
+            string old;
+            using (var key = Registry.CurrentUser.OpenSubKey(PathsKey, true))
+            {
+                old = (string)key.GetValue("Value0", string.Empty, RegistryValueOptions.DoNotExpandEnvironmentNames);
```
Haacked: Potential `NullReferenceException` here with `key`.

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -86,11 +89,68 @@ public string GetActiveRepoPath()
...
+
+        static IEnumerable<ISimpleRepositoryModel> PokeTheRegistryForRepositoryList()
+        {
+            using (var key = OpenGitKey("Repositories"))
+            {
+                return key.GetSubKeyNames().Select(x =>
+                {
+                    using (var subkey = key.OpenSubKey(x))
+                    {
+                        var path = subkey.GetValue("Path") as string;
```
Haacked: Potential `NullReferenceException` here. If we want to throw an exception when `subkey` is null (since the caller is handling it), we probably should throw something better and more informative than `NullReferenceException`.

Haacked:
I went ahead and fixed the potential nullrefs. I think this branch is good.

Haacked:
high-five-to-self

PR #40, opened 2015-08-06
Better login failed error tooltip

### LINKED ISSUE #14: Improve error message for login failed
Imported from https://github.com/github/VisualStudioInternal/issues/440

User used a personal token as password and got the following error:

image

Haacked:
When the user provides a personal access token instead of their password, the API responds with a 403. We now show a more helpful tooltip.

Fixes #14


shana:
yay ^_^

PR #49, opened 2015-08-13
Fix libgit2sharp packaging

shana:
Go back to using a reference to VS's version of libgit2sharp. We don't bundle it, just reference it, so we use the version that's loaded at runtime in VS. This is because something is wonky with the current version of libgit2sharp and it's not authenticating correctly, but I don't want to delay the release to figure out why (since it fixes a crasher).

This PR also contains some changes to how things are bundled in the vsix so we don't corlib and other class libraries in it, making the vsix even smaller than before :smile:


shana:
@Haacked branch beta-1.0.12 already contains these changes, and I've done a full test run of this in testpad (https://github.ontestpad.com/project/7) with the 1.0.12.1 build (that I haven't yet released as beta because I'm going to release it in production)

Haacked:
:+1:

PR #60, opened 2015-08-27
:art: Remove rendundant null check

Haacked:
In `GitClient.cs` we checked whether `repository.Head == null` twice in
a row. I wanted to remove this redundancy and just ended up using the
fun null propagation operator where we could.

### DIFF
--- a/src/GitHub.App/Services/GitClient.cs
+++ b/src/GitHub.App/Services/GitClient.cs
@@ -18,10 +18,7 @@ namespace GitHub.Services
 
             return Observable.Defer(() =>
             {
-                if (repository.Head != null
-                    && repository.Head != null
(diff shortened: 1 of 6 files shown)


src/GitHub.App/Services/GitClient.cs
```diff
@@ -18,10 +18,7 @@ public IObservable<Unit> Push(IRepository repository, string branchName, string
 
             return Observable.Defer(() =>
             {
-                if (repository.Head != null
-                    && repository.Head != null
-                    && repository.Head.Commits != null
-                    && repository.Head.Commits.Any())
+                if (repository.Head?.Commits != null && repository.Head.Commits.Any())
```
shana: :heart:

Issue #62, opened 2015-08-31
Create Private Repo creates public repository


jongalloway:
This blog post says that the author checked the "Create Private Repo" checkbox, but a public repository was created and the author inadvertently published AWS keys: [How a bug in Visual Studio 2015 exposed my source code on GitHub and cost me $6,500 in a few hours](https://www.humankode.com/security/how-a-bug-in-visual-studio-2015-exposed-my-source-code-on-github-and-cost-me-6500-in-a-few-hours)

I'm fairly certain this is an issue in the GitHub extension, rather than Visual Studio itself.

Here's how the Visual Studio sync dialog appears without the GitHub extension installed:
![Without GitHub extension](http://i.imgur.com/EGjYziR.png)

And with the extension:
![Without GitHub extension](http://i.imgur.com/shRBwXM.png)

gaui:
:+1:

shiftkey:
cc @thecarlo

NickCraver:
I have just signed up for private repros and **I can reproduce this**, specifically from the Team Explorer -> Sync route. The other route of creating a repo correctly sets it as private.

shiftkey:
@NickCraver thanks for the repro steps. I'm looking at this now to see if I can organize a fix.

shiftkey:
I've also been able to recreate this. Talking with @haacked about how to solve this, but will open something soon with details.

ghuntley:
Reddit discussion over at https://www.reddit.com/r/programming/comments/3j4ydl/how_a_bug_in_visual_studio_2015_exposed_my_source/

shiftkey:
The updated build has landed on the gallery - https://visualstudiogallery.msdn.microsoft.com/75be44fb-0794-4391-8865-c3279527e97d.

Haacked:
First of all, I'm very sorry for this bug. It's inexcusable and I feel horrible.

As @shiftkey points out, we updated the extension with a fix on the Visual Studio Gallery and at https://visualstudio.github.com/.

The problem only affected publishing repositories, not in creating repositories which used pretty much the same code, but in a slightly different context.

You can see the fix in this PR: https://github.com/github/VisualStudio/pull/64

Based on our preliminary investigation, checkboxes in a Team Explorer pane appear to behave differently than checkboxes in a standard WPF control. We're not sure why that is and we'll continue to investigate.

Once we have a root cause down, we'll post a post-mortem somewhere appropriate.

niik:
[...]

PR #64, opened 2015-09-01
workaround for checkbox not propagating changes to KeepPrivate property

shiftkey:
Proposed fix for #62 by observing the events directly

 - [x] still have to bind the property to the VM somehow.
 - [x] testing testing testing

### DIFF
--- a/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryPublishControl.xaml.cs
+++ b/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryPublishControl.xaml.cs
@@ -31,11 +31,21 @@ namespace GitHub.VisualStudio.UI.Views.Controls
                 
                 d(this.Bind(ViewModel, vm => vm.Description, v => v.description.Text));
 
+                // BEGIN DANGER ZONE
+                //
+                // This replaces the default Bind behaviour as the Checkbox control
+                // does not raise an event here when it is hosted in the Team Explorer
+                // view.
+                //
+                // We've used this.Bind in other places (popups) so it's something
+                // we need to investigate further
+                //
                 d(Observable.FromEventPattern<RoutedEventArgs>(makePrivate, "Checked")
                     .Subscribe(_ => ViewModel.KeepPrivate = true));
-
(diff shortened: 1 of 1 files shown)


src/GitHub.VisualStudio/UI/Views/Controls/RepositoryPublishControl.xaml.cs
```diff
@@ -30,7 +30,13 @@ public RepositoryPublishControl()
...
                 
                 d(this.Bind(ViewModel, vm => vm.Description, v => v.description.Text));
-                d(this.Bind(ViewModel, vm => vm.KeepPrivate, v => v.makePrivate.IsChecked));
+
+                d(Observable.FromEventPattern<RoutedEventArgs>(makePrivate, "Checked")
+                    .Subscribe(_ => ViewModel.KeepPrivate = true));
+
+                d(Observable.FromEventPattern<RoutedEventArgs>(makePrivate, "Unchecked")
+                    .Subscribe(_ => ViewModel.KeepPrivate = false));
+
```
Haacked: Let's add a comment on why this abomination is required as opposed to the mores straightforward code that works elsewhere. :smile:

shiftkey: :thumbsup:

Haacked:
> does your binding need an UpdateSourceTrigger maybe

@gulbanana We're not using XAML bindings. We're using ReactiveUI bindings. Note that the `Bind` method we used here is the same one we use in the repository create control: https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCreationControl.xaml.cs#L50

It works over there, but not here for some reason.

shiftkey:
Ready for review.

Haacked:
:cry:

Thanks for the fix @shiftkey! :sparkles:

gulbanana:
ah, ok. i've definitely experienced weird partial-supportedness with the IsChecked dependencyproperty before but that was using {Binding}, not reactiveui.

PR #67, opened 2015-09-01
Checkbox styling fix for #62

### LINKED ISSUE #62: Create Private Repo creates public repository

shana:
Fixes #62

This fixes the checkbox styling so it still matches Team Explorer but doesn't break our reactive bindings.

The previous style was using an inner checkbox for the rendering, and when reactiveui binds the view to the viewmodel, it likely looks for changed properties in the RoutedEventArgs Source or possibly the sender object that goes along with the event, and both of these represent the outer checkbox, which is not the originator of the Checked event, so it doesn't actually have the IsChecked property set.

This caused the UI to appear correct (inner checkbox is set), but the binding to not pick up the change (outer checkbox not set).

### DIFF
--- a/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryPublishControl.xaml
+++ b/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryPublishControl.xaml
@@ -123,7 +123,24 @@
 
         <CheckBox x:Name="makePrivate"
                   Margin="0,8,0,0"
-                  Style="{StaticResource VSStyledCheckBox}">Private Repository</CheckBox>
+                  Style="{StaticResource VSStyledCheckBox}">
+          <TextBlock TextWrapping="Wrap"
+                     Foreground="{DynamicResource VsBrush.ToolWindowText}"
+                     Background="{DynamicResource VsBrush.ToolWindowBackground}">
+            <TextBlock.Resources>
+              <Style TargetType="{x:Type TextBlock}">
+                <Style.Triggers>
+                  <Trigger Property="IsEnabled" Value="False">
+                    <Setter Property="Opacity" Value="0.5" />
+                  </Trigger>
+                </Style.Triggers>
+              </Style>
+            </TextBlock.Resources>
+            <TextBlock.Inlines>
+              <Run>Private Repository</Run>
+            </TextBlock.Inlines>
+          </TextBlock>
+        </CheckBox>
         
         <Label x:Name="upgradeToMicroPlanWarning" HorizontalAlignment="Left" Visibility="Collapsed">
             <StackPanel Orientation="Horizontal" HorizontalAlignment="Stretch">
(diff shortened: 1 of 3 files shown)


shana:
triple-facepalm

shiftkey:
Pulled this down to test, and yes I can see the checkbox state updating the VM property correctly :sparkles:

@shana anything else you'd like us to focus on with the review?

shana:
@shiftkey One of the reasons things broke is that I tried to make a style combining the CheckBox and the TextBlock without recreating the checkbox style (because I have no clue what it is). Is there a better way of doing this than what I'm doing now (adding the textblock inside the checkbox directly in the xaml publish control)?

src/GitHub.VisualStudio/UI/Views/Controls/RepositoryPublishControl.xaml
```diff
@@ -123,7 +123,24 @@
...
+            <TextBlock.Resources>
+              <Style TargetType="{x:Type TextBlock}">
+                <Style.Triggers>
+                  <Trigger Property="IsEnabled" Value="False">
+                    <Setter Property="Opacity" Value="0.5" />
+                  </Trigger>
+                </Style.Triggers>
+              </Style>
+            </TextBlock.Resources>
+            <TextBlock.Inlines>
```
shiftkey: You can strip this back to just the literal `Private Repository` element - this will get rendered as an inline `TextBlock` and pick up the styling...

shana: Ah, right, I can move the trigger up to the checkbox

/me fixicates

shana: Sigh, can't merge the TextBlock. If I do, then the Hover color in the checkbox (turning the square border a different color) also gets applied to the text.

shiftkey: Uggghh :thumbsup:

shana:
@Haacked :eyes: :+1: :-1:?

Haacked:
If @shiftkey gives a :+1: then I'm a :+1:!

PR #42, opened 2015-08-07
Localization

jau228:
This is basically pulling strings from xaml and place them in the resource file. And using static binding to link those strings. I also added an exception in the GitHubVS.ruleset to ignore CA1701 for the string renaming error("UserName" instead of "username", "Github" instead of "GitHub", etc...) I am not seeing a whole lot of strings beside the GitHub.VisualStudio project. So this is what I focus so far.

PR #75, opened 2015-09-08
Localization

shana:
Merge https://github.com/github/VisualStudio/pull/42 with master, clear up string duplication and unused code, and add new strings.


src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml
```diff
@@ -1,226 +1,234 @@
 ﻿<uirx:SimpleViewUserControl x:Class="GitHub.VisualStudio.UI.Views.Controls.RepositoryCloneControl"
```
shana: The whitespace and alignment was all screwed up on this file, so I reformatted it, hence all the apparent changes :/

src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCreationControl.xaml
```diff
@@ -217,41 +218,12 @@
                     x:Name="makePrivate"
                     Grid.Column="1"
                     Grid.Row="9"
-                    Content="Private Repository"
+                    Content="{x:Static prop:Resources.makePrivateContent}"
                     Style="{DynamicResource BlueRoundedCheckBox}"
                     Padding="8,0,0,0" />
-                <Grid Grid.Column="1" Grid.Row="10">
```
shana: We've never shown this section (code was commented out in the cs file)

shana:
Fixed the rest of the view model strings and updated from master. This should be good for now.

src/common/GitHubVS.ruleset
```diff
@@ -37,6 +37,7 @@
         <Rule Id="CA1501" Action="None" />
         <Rule Id="CA1502" Action="None" />
         <Rule Id="CA1506" Action="None" />
+        <Rule Id="CA1701" Action="None" />
```
Haacked: At this point, could we remove `<Rule Id="CA1303" Action="None" />` so that we get build failures if we add literal strings?

Haacked: I should have linked that. https://msdn.microsoft.com/en-us/library/ms182187.aspx?f=255&MSPPError=-2147217396

> CA1303: Do not pass literals as localized parameters
> A method passes a string literal as a parameter to a constructor or method in the .NET Framework class library and that string should be localizable.
> This warning is raised when a literal string is passed as a value to a parameter or property and one or more of the following cases is true:

Haacked:
selfie-0

Haacked:
We can deal with https://github.com/github/VisualStudio/pull/75#discussion_r39582119 in another PR if there's anything to be done.

src/common/GitHubVS.ruleset
shana: Yeah, this is a good idea, those conditions won't trigger for our other usages of string literals (logging and such), so should be helpful!

PR #76, opened 2015-09-09
Show proper message when repo is created

shana:
Fixes #54

Also updates the icon in the list.

Third time's the charm!

PR #77, opened 2015-09-09
Fixes to the repo list in the connect page

shana:
Found a few bugs that needed fixing.

- Cloning or creating when logged out was throwing a NRE
- Cloning or creating when logged out would trigger multiple "Open Solution" dialogs.

I didn't like the open solution thing popping up automatically, so I changed the notifications to show links to the Open/Create project dialogs when the cloning/creating process is done.
Now it checks whether a cloned repo has commits and content (files or directories) to decide whether to link to the Create or the Open project dialog. It's not too smart, but it should work well for most cases.

### DIFF
--- a/src/GitHub.App/Controllers/UIController.cs
+++ b/src/GitHub.App/Controllers/UIController.cs
@@ -236,6 +236,14 @@ namespace GitHub.Controllers
                     .IsLoggedIn(hosts)
                     .Do(loggedin =>
                     {
+                        if (!loggedin && currentFlow != UIControllerFlow.Authentication)
+                        {
+                            connectionManager.Connections.CollectionChanged += (s, e) => {
+                                if (e.Action == System.Collections.Specialized.NotifyCollectionChangedAction.Add)
+                                    uiProvider.AddService(typeof(IConnection), e.NewItems[0]);
+                            };
+                        }
+
                         machine.Configure(UIViewType.None)
                             .Permit(Trigger.Auth, UIViewType.Login)
                             .PermitIf(Trigger.Create, UIViewType.Create, () => loggedin)
--- a/src/GitHub.App/Controllers/UIController.cs
+++ b/src/GitHub.App/Controllers/UIController.cs
@@ -17,6 +17,7 @@ using GitHub.ViewModels;
 using NullGuard;
 using ReactiveUI;
 using Stateless;
+using System.Collections.Specialized;
 
 namespace GitHub.Controllers
 {
@@ -36,6 +37,7 @@ namespace GitHub.Controllers
         readonly StateMachine<UIViewType, Trigger> machine;
         Subject<UserControl> transition;
         UIControllerFlow currentFlow;
+        NotifyCollectionChangedEventHandler connectionAdded;
 
         [ImportingConstructor]
         public UIController(IUIProvider uiProvider, IRepositoryHosts hosts, IExportFactoryProvider factory,
@@ -238,10 +240,11 @@ namespace GitHub.Controllers
                     {
                         if (!loggedin && currentFlow != UIControllerFlow.Authentication)
                         {
-                            connectionManager.Connections.CollectionChanged += (s, e) => {
-                                if (e.Action == System.Collections.Specialized.NotifyCollectionChangedAction.Add)
+                            connectionAdded = (s, e) => {
+                                if (e.Action == NotifyCollectionChangedAction.Add)
                                     uiProvider.AddService(typeof(IConnection), e.NewItems[0]);
                             };
+                            connectionManager.Connections.CollectionChanged += connectionAdded;
                         }
 
                         machine.Configure(UIViewType.None)
@@ -278,6 +281,9 @@ namespace GitHub.Controllers
                 Debug.WriteLine("Disposing ({0})", GetHashCode());
                 disposables.Dispose();
                 transition?.Dispose();
+                if (connectionAdded != null)
+                    connectionManager.Connections.CollectionChanged -= connectionAdded;
+                connectionAdded = null;
                 disposed = true;
             }
         }
--- a/src/GitHub.App/Controllers/UIController.cs
+++ b/src/GitHub.App/Controllers/UIController.cs
@@ -1,9 +1,11 @@
 ﻿using System;
 using System.ComponentModel.Composition;
 using System.Diagnostics;
+using System.Reactive.Concurrency;
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
+using System.Windows;
 using System.Windows.Controls;
 using GitHub.Authentication;
 using GitHub.Exports;
@@ -15,8 +17,6 @@ using GitHub.ViewModels;
 using NullGuard;
 using ReactiveUI;
 using Stateless;
-using System.Windows;
-using System.Reactive.Concurrency;
 
 namespace GitHub.Controllers
 {
@@ -265,10 +265,7 @@ namespace GitHub.Controllers
         public void Stop()
         {
             Debug.WriteLine("Stop ({0})", GetHashCode());
-            if (machine.IsInState(UIViewType.End))
-                Fire(Trigger.Next);
-            else
-                Fire(Trigger.Cancel);
+            Fire(machine.IsInState(UIViewType.End) ? Trigger.Next : Trigger.Cancel);
         }
 
         bool disposed; // To detect redundant calls
@@ -291,6 +288,6 @@ namespace GitHub.Controllers
             GC.SuppressFinalize(this);
         }
 
-        public bool IsStopped { get { return machine.IsInState(UIViewType.Finished); } }
+        public bool IsStopped => machine.IsInState(UIViewType.Finished);
     }
 }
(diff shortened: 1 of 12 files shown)


shana:
This incorporates and adds to #76, so that one should either go in or be closed before this one goes.

shana:
screen shot 2015-09-09 at 5 07 49 pm

screen shot 2015-09-09 at 5 19 14 pm

shana:
We can also always show links to the create and open dialogs in all situations. Dunno if that's more confusing to people.

Haacked:
@shana I like the use of a link in the success message. I'm wondering if we should add a bit of context there. For example,

"Open an existing project or solution in this repository."

/cc @github/desktop for any feels.

src/GitHub.App/Controllers/UIController.cs
```diff
@@ -236,6 +236,14 @@ public void Start([AllowNull] IConnection connection)
                     .IsLoggedIn(hosts)
                     .Do(loggedin =>
                     {
+                        if (!loggedin && currentFlow != UIControllerFlow.Authentication)
+                        {
+                            connectionManager.Connections.CollectionChanged += (s, e) => {
```
Haacked: Do we need to unsubscribe this when you log out?

src/GitHub.Exports/Exports/ExportMetadata.cs
```diff
@@ -42,4 +44,12 @@ public interface IViewModelMetadata
     {
         UIViewType ViewType { get; }
     }
+
+    public static class ExportViewAttributeExtensions
+    {
+        public static bool IsViewType(this UserControl c, UIViewType type)
+        {
+            return c.GetType().GetCustomAttributesData().Any(x => x.AttributeType.Equals(typeof(ExportViewAttribute)) && (UIViewType)x.NamedArguments[0].TypedValue.Value == type);
```
Haacked: For types we should probably use reference equality: `x.AttributeType == typeof(ExportViewAttribute)`

Also, it seems like we have a potential `NullReferenceException` here with `x.NamedArguments` possibly being null, `x.NamedArguments[0]` possibly being null, and `x.NamedArguments[0].TypedValue` possibly being null.

If that's an impossibility, we should assert that. Otherwise we should handle the case.

src/GitHub.App/Controllers/UIController.cs
shana: I need to test to make sure, but I believe this subscription gets killed when this instance dies (when the flow is done) and doesn't keep anything alive that shouldn't be. I'll check.

src/GitHub.Exports/Helpers/SimpleRepositoryModelExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace GitHub.Extensions
+{
+    using VisualStudio;
+
+    public static class SimpleRepositoryModelExtensions
+    {
+        public static bool HasCommits(this ISimpleRepositoryModel repository)
+        {
+            var repo = Services.GetRepoFromPath(repository.LocalPath);
+            return repo?.Commits.Count() > 0;
```
Haacked: I generally avoid using `Count` because it has to enumerate the entire enumerable. In this case, we're talking about a LibGit2Sharp queryable which could potentially have costly rev walking implications.

How about?

```csharp
return repo?.Commits != null && (repo.Commits).Any();
```

src/GitHub.Exports/Exports/ExportMetadata.cs
shana: If it's an ExportViewAttribute, it has to have a named argument list with one entry, it's part of the class

src/GitHub.Exports/Helpers/SimpleRepositoryModelExtensions.cs
shana: Oh yes, that's way better :+1:

src/GitHub.Exports/Helpers/SimpleRepositoryModelExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+        }
+        public static bool HasContent(this ISimpleRepositoryModel repository)
+        {
+            var dir = new DirectoryInfo(repository.LocalPath);
+            return dir.EnumerateFileSystemInfos("*", SearchOption.TopDirectoryOnly)
+                      .Any(x => ((x.Attributes.HasFlag(FileAttributes.Directory) || x.Attributes.HasFlag(FileAttributes.Normal)) &&
+                                !x.Name.StartsWith(".", StringComparison.Ordinal) && !x.Name.StartsWith("readme", StringComparison.OrdinalIgnoreCase)));
```
Haacked: What if there's a file named `readme-but-not-a-readme.md`? Certainly an edge case and maybe we don't care. But I think a safer check is `!x.Name.StartsWith("readme.") && !x.Name.Equals("readme", StringComparison.OrdinalIgnoreCase)`

Also, what's the purpose of this method? Seems like if it has a `readme`, then it _does_ have content. Maybe clarify in the method name and/or comment why we ignore those files when determining if it has content.

src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs
```diff
@@ -279,7 +326,15 @@ void StartFlow(UIControllerFlow controllerFlow)
         {
             var uiProvider = ServiceProvider.GetExportedValue<IUIProvider>();
             uiProvider.GitServiceProvider = ServiceProvider;
-            uiProvider.RunUI(controllerFlow, SectionConnection);
+            var ret = uiProvider.SetupUI(controllerFlow, SectionConnection);
+            ret.Subscribe((c) =>
```
Haacked: Parens around `c` are redundant. Yeah, I'm a nitpicker. Sorry!

src/GitHub.Exports/Helpers/SimpleRepositoryModelExtensions.cs
shana: By content I mean a solution or project file that VS can open, mainly. I don't want to waste time scanning for explicit files, this is just a simple naive check to see if the repo has more than just the default content of .git files and the readme, which is a decent clue to use in deciding whether to show the create new or the open project dialog.

It could be that someone has created a repo with no subdirectories and a project and solution files are both called Readme.csproj/sln - that would totally fool this check :smile:

Otherwise, if the user is cloning a repo in VS, it's either because they want to open an existing project (in which case there will be something more in there besides the readme) or create a new solution (so the repo will be initialized but only with our default stuff). Either case should be decently covered by this best guess check.

Haacked: Ok, so it's a heuristic. Maybe a better name for the method is `MightContainSolution` or `ShouldAllowOpeningInVisualStudio` so the method name is scoped to the one specific use case. I worry that I might later see this method name and call it when I need to determine whether a repo has content (ok, I won't, but maybe someone else will).

src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs
shana: hehe where's my nitpick nancy pic when I need it.

src/GitHub.Exports/Exports/ExportMetadata.cs
Haacked: Ah, ok. Maybe we should assert that just in case the contract ever changes?

shana: Sure! The contract won't ever change for the existence of named arguments (the sole purpose of the class is to have that named argument) but the type might.

src/GitHub.Exports/Helpers/SimpleRepositoryModelExtensions.cs
shana: Yup, a better name makes sense, I can see it misleading people.

shiftkey:
> /cc @github/desktop for any feels.

Had a quick chat with @shana about this - [start here](https://chat.githubapp.com/messages/1509554905/context) - and to TL;DR it:

 - the messages are great
 - having clickable links seems to be The Right Way™
 - the differences between popping a modal for creating/cloning are different to how VSO does it - but let's tackle that later as it's a UX issue

shana:
:heart: @shiftkey

shana:
Ok, I'm happy with this for now :sparkles:

src/GitHub.Exports/Exports/ExportMetadata.cs
```diff
@@ -42,4 +45,13 @@ public interface IViewModelMetadata
     {
         UIViewType ViewType { get; }
     }
+
+    public static class ExportViewAttributeExtensions
+    {
+        public static bool IsViewType(this UserControl c, UIViewType type)
+        {
+            Debug.Assert(c.GetType().GetCustomAttributesData().Any(x => x.AttributeType == typeof(ExportViewAttribute) && x.NamedArguments.Any() && x.NamedArguments[0].TypedValue.ArgumentType == typeof(UIViewType)), "Someone broke the ExportViewAttribute contract...");
```
Haacked: This assert doesn't actually assert the part that could be a null ref, namely x.NamedAttributes. I went ahead and added the proper assert in a commit.

Haacked:
@shana could you take a look at https://github.com/github/VisualStudio/commit/c04ed0961254a887864a404e8782f4df397b8881? If that looks good to you, feel free to merge this or I will. Everything else looks :sparkles:

shana:
Looks good, thanks!

PR #79, opened 2015-09-15
Fix publish form not getting hidden after publishing is done

### LINKED ISSUE #48: Publish stays on Synchronization page
After publishing a new repository to GitHub, the TE view stays on the publish page (pictured), instead of switching to the Home view.
screenshot 25

shana:
Fixes #48

This PR grew a bit more than I expected. The main fix is that we were only looking at the local path to see if VS had loaded a new repository, but it turns out that when you publish a repository, the url change also needs to be monitored so the UI is updated accordingly. This required some changes to how repositories are compared (via SimpleRepositoryModel) and how changes to the active repo are tracked.

This led to having ISimpleRepositoryModel be used everywhere instead of using IGitRepositoryInfo, so we can compare things better.

- Adds IView Error observable so we can expose errors and let the UI handle them instead of showing error messages directly from the view model

- Switches to ISimpleRepositoryModel for all objects representing a repository, doing away with direct usage of IGitRepositoryInfo (which is VS's and doesn't have much information we can use)

- Checks and updates ActiveRepo.CloneUrl whenever the underlying repository url changes (user changes it manually on the command line or in VS)

- Hides the publish form properly (and disposes it) when publishing is done.


src/GitHub.App/Controllers/UIController.cs
```diff
@@ -160,22 +160,22 @@ void SetupView(UIViewType viewType, IView view)
...
             else if (viewType != UIViewType.TwoFactor)
             {
-                view.Done
+                disposables.Add(view.Done
                     .ObserveOn(RxApp.MainThreadScheduler)
-                    .Subscribe(_ => Fire(Trigger.Next));
+                    .Subscribe(_ => Fire(Trigger.Next)));
             }
-            view.Cancel.Subscribe(_ => Stop());
+            disposables.Add(view.Cancel.Subscribe(_ => Stop()));
```
shana: This is not strictly required, but might as well.

Haacked:
Hmm, build is broken. Also, could this be broken up into smaller PRs or is this all logically the same fix?

shana:
I broke some of the tests I recently added for the SimpleRepositoryModel because I changed how arguments get validated, and I was too tired to fix it up last night, fixing today.

shana:
I took out the extension classes cleanup #89, the disposable calls #88, the get started fix #87 and the error handling #90 to cut down on the amount of clutter in this PR.

src/GitHub.VisualStudio/Base/TeamExplorerServiceHolder.cs
```diff
@@ -156,11 +178,16 @@ void CheckAndUpdate(object sender, System.ComponentModel.PropertyChangedEventArg
             if (service == null)
                 return;
 
-            var repo = service.ActiveRepositories.FirstOrDefault();
-            // this comparison is safe, the extension method supports null instances
-            if (!repo.Compare(ActiveRepo))
+            var repo = service.ActiveRepositories.FirstOrDefault()?.ToModel();
+            if (repo != ActiveRepo)
```
shana: FWIW, this comparison is the actual fix to #48.

src/GitHub.Exports/Models/SimpleRepositoryModel.cs
```diff
@@ -18,6 +21,31 @@ public SimpleRepositoryModel(string name, UriString cloneUrl, string localPath =
             Icon = Octicon.repo;
         }
 
+        public SimpleRepositoryModel(string path)
+        {
+            if (path == null)
+                throw new ArgumentNullException("path");
```
Haacked: Wouldn't `NullGuard.Fody` catch this? If not, you could use the `nameof` operator here: `throw new ArgumentNullException(nameof(path));`

Yay C# 6!

src/GitHub.Exports/Models/SimpleRepositoryModel.cs
```diff
@@ -18,6 +21,31 @@ public SimpleRepositoryModel(string name, UriString cloneUrl, string localPath =
...
+            if (!Directory.Exists(path))
+                throw new ArgumentException("Path does not exist", path);
+            var uri = VisualStudio.Services.GetRepoFromPath(path)?.GetUri();
+            var name = uri?.NameWithOwner;
+            if (name == null)
+                name = Path.GetDirectoryName(name);
```
Haacked: Huh? If `name` is null, then `Path.GetDirectoryName(name)` will also return null.

Did you mean `if (name != null)` perhaps? Or `if (!string.IsNullOrEmpty(name))`?

shana: ugh, no, Path.GrtDirectoryName(path). Need tests for this.

Haacked: Trololol

Haacked:
So the original issue suggests we should load the Home page after a successful publish. I tested this and we stay on the Sync tab, but we update the state to that of a published repository. Staying in the Sync tab feels right to me.



@abuchholtzau do you have strong feelings either way?

Haacked:
vader-haack-thumbsup

PR #87, opened 2015-09-16
Fix Get Started link showing up with a delay in the Connect page

shana:


Haacked:
BIG CHANGE! :stuck_out_tongue:

PR #88, opened 2015-09-16
Adds some dispose calls

shana:
Some things live longer than the subscriber, so clean up. Not strictly required, but better safe than sorry.


Haacked:
Always Be Disposing

PR #97, opened 2015-09-18
Add repository model tests

shana:
This PR is based on #79, so that one should be merged first before looking at this.

Add repository model tests and support for testing things that rely on the GetUri and GetRepo extension methods that use libgit2#.

Testing RepositoryModel creation via the path constructor implies hitting libgit2sharp codepaths, and libgit2sharp doesn't expose nice interfaces for testing (why??). So, move things around and add a
GitService that replaces the extension methods that depend on libgit2sharp, so we can at least mock at a higher level.

The things that require this GitService are not created via mef, so expose it via the Services static class.

### DIFF
--- a/src/GitHub.Exports/Services/Services.cs
+++ b/src/GitHub.Exports/Services/Services.cs
@@ -5,7 +5,6 @@ using GitHub.Services;
 using LibGit2Sharp;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.ComponentModelHost;
-using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using GitHub.Info;
 using GitHub.Primitives;
@@ -13,16 +12,16 @@ using GitHub.Extensions;
 
 namespace GitHub.VisualStudio
 {
+
     public static class Services
     {
         public static IServiceProvider PackageServiceProvider { get; set; }
 
         /// <summary>
-        /// Three ways of getting a service. First, trying the passed-in <paramref name="provider"/>,
-        /// then <see cref="PackageServiceProvider"/>, then <see cref="T:Microsoft.VisualStudio.Shell.Package"/>
-        /// If the passed-in provider returns null, try PackageServiceProvider or Package, returning the fetched value
-        /// regardless of whether it's null or not. Package.GetGlobalService is never called if PackageServiceProvider is set.
-        /// This is on purpose, to support easy unit testing outside VS.
+        /// Two ways of getting a service. First, trying the passed-in <paramref name="provider"/>,
+        /// then <see cref="PackageServiceProvider"/>
+        /// If the passed-in provider returns null, try PackageServiceProvider, returning the fetched value
+        /// regardless of whether it's null or not.
         /// </summary>
         /// <typeparam name="T"></typeparam>
         /// <typeparam name="Ret"></typeparam>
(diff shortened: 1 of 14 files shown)


src/GitHub.Exports/Services/Services.cs
```diff
@@ -101,18 +98,6 @@ public static IVsSolution GetSolution(this IServiceProvider provider)
             return GetGlobalService<SVsSolution, IVsSolution>(provider);
         }
 
-        public static T GetExportedValue<T>(this IServiceProvider serviceProvider)
```
shana: This got moved up to VSExtensions 'cause it makes sense there.

src/GitHub.Exports/Services/Services.cs
```diff
@@ -141,5 +126,10 @@ public static Repository GetRepoFromSolution(this IVsSolution solution)
                 return null;
             return new Repository(repoPath);
         }
+        static UriString GetUri(IRepository repo)
```
shana: This is the same logic as IGitService.GetUri(repo), but since this is only called from GetRepoFromSolution (which is static), I'm just leaving this bit here because it's a one liner and faster than looking up the service.

src/GitHub.Exports/Services/Services.cs
```diff
@@ -81,15 +71,9 @@ public static IVsOutputWindowPane OutputWindowPane
...
-        {
-            get { return GetGlobalService<DTE, DTE>(); }
-        }
+        public static DTE Dte => GetGlobalService<DTE, DTE>();
 
-        public static DTE2 Dte2
-        {
-            get { return Dte as DTE2; }
-        }
+        public static DTE2 Dte2 => Dte as DTE2;
```
Haacked: @shana: So according to R#, there is no type in the solution that inherits from both `EnvDTE.DTE` and `EnvDTE80.DTE2`. That leads me to believe that the `Dte2` property is always null. Also, I can't find any usages of the `Dte2` property. Do we need this?

src/UnitTests/Helpers/TestBaseClass.cs
```diff
@@ -7,14 +8,35 @@
...
+
+    public override void OnEntry()
+    {
+        var f = Path.GetTempFileName();
+        var name = Path.GetFileNameWithoutExtension(f);
+        File.Delete(f);
+        Directory = new DirectoryInfo(Path.Combine(Path.GetTempPath(), name));
+        Directory.Create();
+        base.OnEntry();
+    }
```
Haacked: I'm a little wary of having unit tests that touch the file system. I'm fine with them now if they're fast. It'd be a pain to mock out all the fs operations, although that is what I built https://github.com/haacked/rothko for. :smile:

Or should we add an `IntegrationTests` project and move our tests of git operations there?

Haacked:
:pineapple:

Haacked:
@shana: If you're ok with my changes, this is good to merge. The only outstanding question is the `Dte2` property. I have a concern about touching the file system in the tests, but let's not worry about that right now.

src/GitHub.Exports/Services/Services.cs
shana: Oh, don't believe everything R# says. These are COM interfaces, the cast is 

It's going to be used in the future, so I keep it around, along with some other handy magical vs types (that are all probably used now, but weren't in the beginning).

src/UnitTests/Helpers/TestBaseClass.cs
shana: Yeah, this one is stretching the unit test moniker. There's no way around it, the SimpleRepositoryModel is used in places where rothko is not yet loaded. And instantiated often enough that querying a service would impact performance if it were available, so we'd need to start caching services more aggressively to be safe using them in models.

We can move these to integration tests, I'm fine with that.

shana:
@Haacked Changes look :+1:!

Haacked:
:sparkles:

src/UnitTests/Helpers/TestBaseClass.cs
Haacked: If we keep the tests that touch the file system limited, I'm fine with this as is for now.

shana: :+1:

PR #98, opened 2015-09-18
Caching the UIProvider

shana:
This PR is based on #79 and #97, those need to be merged first before this one.

The IUIProvider service provides access to everything else, no point in doing a lookup every time we need to access it manually (not via mef constructors)

### DIFF
--- a/src/GitHub.Exports/Extensions/VSExtensions.cs
+++ b/src/GitHub.Exports/Extensions/VSExtensions.cs
@@ -7,6 +7,8 @@ namespace GitHub.Extensions
 {
     public static class VSExtensions
     {
+        static IUIProvider cachedUIProvider = null;
+
         public static T TryGetService<T>(this IServiceProvider serviceProvider) where T : class
         {
             return serviceProvider.TryGetService(typeof(T)) as T;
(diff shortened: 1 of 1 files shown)


src/GitHub.Exports/Extensions/VSExtensions.cs
```diff
@@ -33,20 +38,42 @@ public static object TryGetService(this IServiceProvider serviceProvider, Type t
...
+            if (type == typeof(CacheType))
+                cache = (CacheType)ret;
+            return ret;
+        }
+
+        static T GetExportedValueAndCache<T, CacheType>(ref CacheType cache)
+        {
+            var ret = VisualStudio.Services.ComponentModel.DefaultExportProvider.GetExportedValue<T>();
+            if (typeof(T) == typeof(CacheType))
+                cache = (CacheType)(object)ret;
```
Haacked: What's with the double cast?

shana: The type system doesn't know that it's the same type at compile time inside this condition :wink:

Haacked:
:sparkles:

PR #101, opened 2015-09-18
Update octokit to 0.16.0

shana:
Let's not merge this one quite yet, it's not needed for the next release.

### DIFF
--- a/src/GitHub.App/Services/ImageDownloader.cs
+++ b/src/GitHub.App/Services/ImageDownloader.cs
@@ -30,7 +30,6 @@ namespace GitHub.Services
                 BaseAddress = new Uri(imageUri.GetLeftPart(UriPartial.Authority)),
                 Endpoint = new Uri(imageUri.PathAndQuery, UriKind.RelativeOrAbsolute),
                 Method = HttpMethod.Get,
-                AllowAutoRedirect = true
             };
 
             return HttpClient.Send(request)
--- a/src/GitHub.App/Infrastructure/ExportWrappers.cs
+++ b/src/GitHub.App/Infrastructure/ExportWrappers.cs
@@ -1,5 +1,7 @@
 ﻿using System.ComponentModel.Composition;
 using Octokit.Internal;
+using System;
+using System.Net.Http;
 
 namespace GitHub.Infrastructure
 {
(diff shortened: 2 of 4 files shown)


src/GitHub.App/Services/ImageDownloader.cs
```diff
@@ -30,7 +30,6 @@ public IObservable<byte[]> DownloadImageBytes(Uri imageUri)
                 BaseAddress = new Uri(imageUri.GetLeftPart(UriPartial.Authority)),
                 Endpoint = new Uri(imageUri.PathAndQuery, UriKind.RelativeOrAbsolute),
                 Method = HttpMethod.Get,
-                AllowAutoRedirect = true
```
Haacked: So this isn't needed any more?

shana: It's deprecated and ignored due to redirect work, says the comment coming from this commit: https://github.com/octokit/octokit.net/commit/3acdd10be4800f376f5f008fd202b3a776409695. The code analysis bails on deprecation warnings, so I took all the calls out. 

I have *not* tested everything, so we might want to hold on merging this for now.

shiftkey: This is a side-effect of the repository redirects from the API, and because `AllowAutoRedirect` doesn't set the `Authorization` header [citation](https://msdn.microsoft.com/en-us/library/system.net.httpwebrequest.allowautoredirect%28v=vs.110%29.aspx) when redirecting, it would break for most of our scenarios.

> The Authorization header is cleared on auto-redirects and HttpWebRequest automatically tries to re-authenticate to the redirected location. In practice, this means that an application can't put custom authentication information into the Authorization header if it is possible to encounter redirection.

We're now handling the redirects in Octokit and setting the right authorization headers. See https://github.com/octokit/octokit.net/pull/808 for the whole backstory.

shana: Thanks for the info @shiftkey!

shana:
We're good to go on this one, redirects are handled internally in HttpClient in a proper fashion so we don't have to worry about that :pineapple:

Haacked:
:sparkles:

PR #105, opened 2015-09-23
Fix IGitService

shana:
**TL;DR**: Create static methods that mirror the IGitService instance methods, and use either a real mef instance or a fake one to get at the instance methods. This way we can expose the service in unit tests and use it at runtime without worrying about different codepaths or untestable static methods.

**Long Version**
Things are slightly broken getting a reference to IGitService - in real life, it's not a service, it's an exported value, but in unit tests it's treated as a service (because this silly distinction between services
and exported values is silly).

`IGitService` is nothing but a bunch of helper methods, and we want to use it as an instance and not just static methods, so that we can bypass hitting the filesystem and libgit2# methods in unit tests. Some bits that need `IGitService` during runtime are running in places where there's no MEF.

Therefore, the only time where it really makes a difference when/how the `IGitService` instance is created is in unit tests. When running for real, it makes zero difference if the instance is coming from mef or if it's just created on the spot, so that's what we're doing now.

`IGitService` can be obtained via MEF as usual, or through `ComponentModel.DefaultExportProvider.GetExportedValue<IGitService>()`, or `IUIProvider.GetService<IGitService>()` and variants (which end up going through `ComponentModel`, so it's all the same), and now the unit tests substitutes also provide all these paths (before they'd only provide `GetService()`).


Haacked:
Nice work @shana!

PR #111, opened 2015-10-05
Changes-tracking collection for list views

shana:
TrackingCollection gets items from an observable sequence and updates its contents in such a way that two updates to the same object (as defined by an Equals call) will result in one object on the list being updated (as opposed to having two different instances of the object added to the list). 

This allows us to grab a bunch of objects from cache, load a collection with them, and then request updated versions of the objects from the server, which when returned will update the collection in place (and the UI with it), seamlessly.

The collection supports ordering and filtering, and filtering includes enough information to allow limiting the number of items to return (and display). It can be resorted and refiltered in place.

This PR also adds the IPullRequestModel model interface, because that's where this list will be used and might as well test things on it.

Also now that I've tagged 1.0.15, we can put these into master. Any further bug fixes on the 1.0 series can be done on a branch if needed.


shana:
This is actually passing. The first unit test run failed in the xunit runner itself on completely unrelated bits, and a re-run ran correctly (the jenkins/build_log entry), but the checks are not refreshing correctly and it erroneously shows as failing.

Haacked:
Is `TrackingCollection` similar to `TrackedRepositories` that we have in GitHub for Windows?

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+using System.Reactive.Concurrency;
+using System.Reactive.Disposables;
+using GitHub.Helpers;
+using System.Reactive;
+using System.Linq;
+using ReactiveUI;
+
+namespace GitHub.Collections
+{
+    public class TrackingCollection<T> : ObservableCollection<T>, ITrackingCollection<T>
```
Haacked: Let's document this class like you did in the PR.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+    {
+        enum TheAction
+        {
+            None,
+            Move,
+            Add,
+            Insert
+        }
+
+        CompositeDisposable disposables = new CompositeDisposable();
```
Haacked: This should be `readonly`

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+        IObservable<Unit> source;
+        Func<T, T, int> comparer;
+        Func<T, int, IList<T>, bool> filter;
+        IScheduler scheduler;
```
Haacked: This should be `readonly` too.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+                // the move caused the object to become visible in the live list, insert it
+                // and recalculate all the other things on the live list after this one
+                else if (idx < 0)
+                    InsertAndRecalculate(obj, liveListCurrentIndex, position);
+
+                // move the object
+                else
+                    MoveAndRecalculate(obj, idx, liveListCurrentIndex, oldPosition, position);
+
+                return;
```
Haacked: This `return` statement is redundant. It's at the end of the method anyways.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+        /// <param name="right"></param>
+        static void Swap(IList<T> list, int left, int right)
+        {
+            var l = list[left];
+            list[left] = list[right];
+            list[right] = l;
+        }
+
+#if !DISABLE_BACKING_DICT
+        Dictionary<T, int> liveListMap = new Dictionary<T, int>();
```
Haacked: This should be `readonly`

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+
+            // for all cases except add, we need to know where the touched
+            // element is going to be located in the live list, so grab that
+            var liveListCurrentIndex = GetLiveListPivot(position);
+
+            //if the element was changed but didn't move, then sorting is correct but we 
+            //need to check the filter anyway to determine whether to hide or show the element.
+            //if that changes, then other elements after it might be affected too, so we need to check
+            //all elements from the live list from the element that was changed to the end
+            if (action == TheAction.None)
```
Haacked: Seems like we should use a `switch` statement here for the next three conditionals.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+        public TrackingCollection(IObservable<T> source, Func<T, T, int> comparer = null, Func<T, int, IList<T>, bool> filter = null, IScheduler scheduler = null)
+            : this(comparer, filter, scheduler)
+        {
+            Listen(source);
+        }
+
+        public ITrackingCollection<T> Listen(IObservable<T> obs)
+        {
+            source = obs
+                .ObserveOn(filter != null ? TaskPoolScheduler.Default : scheduler)
```
Haacked: Why does the filter affect the scheduler? Perhaps a comment is warranted because it's not immediately clear to me.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+                    {
+                        var old = list[idx];
+                        var comparison = 0;
+                        if (comparer != null)
+                            comparison = comparer(t, old);
+                        list[idx].CopyFrom(t);
+
+                        if (comparer == null || comparison == 0)
+                            return Tuple.Create(TheAction.None, idx, idx); // no sorting to be done, just replacing the element in-place
+                        else
```
Haacked: This `else` is redundant.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+                            if (filter == null)
+                            {
+                                OnPropertyChanged(new PropertyChangedEventArgs("Item[]"));
+                                OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Move, old, i, idx));
+                            }
+                            return Tuple.Create(TheAction.Move, i, idx); // pass the new position and old position
+                        }
+                    }
+                    // the element doesn't exist yet. If there's a comparer and other items, add it in the correct spot
+                    else if (comparer != null && list.Count > 0)
```
Haacked: The `else` here is redundant.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+                        for (; i < list.Count && comparer(t, list[i]) > 0; i++) { }
+                        if (i == list.Count)
+                        {
+                            list.Add(t);
+                            return Tuple.Create(TheAction.Add, list.Count - 1, -1);
+                        }
+                        list.Insert(i, t);
+                        return Tuple.Create(TheAction.Insert, i, -1);
+                    }
+                    else
```
Haacked: As is this `else`

src/GitHub.Exports/Helpers/ICopyable.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace GitHub.Helpers
+{
+    public interface ICopyable<in T>
```
Haacked: Please document the interface and the method. What's its purpose, etc. I didn't see any implementations of this interface. I assume that comes later? :smile:

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+                            return Tuple.Create(TheAction.None, idx, idx); // no sorting to be done, just replacing the element in-place
+                        else
+                        {
+                            var i = idx;
+                            if (comparison < 0) // replacing element has lower sorting order, find the correct spot towards the beginning
+                                for (var pos = i - 1; i > 0 && comparer(list[i], list[pos]) < 0; i--, pos--)
+                                    Swap(list, i, pos);
+                            else if (comparison > 0) // replacing element has higher sorting order, find the correct spot towards the end
+                                for (var pos = i + 1; i < list.Count - 1 && comparer(list[i], list[pos]) > 0; i++, pos++)
+                                    Swap(list, i, pos);
```
Haacked: The above block hurts my head a bit. I get it now. Maybe extract that into its own private method to make this larger method more readable. :smile:

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+
+                            if (filter == null)
+                            {
+                                OnPropertyChanged(new PropertyChangedEventArgs("Item[]"));
+                                OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Move, old, i, idx));
```
Haacked: I don't understand why we don't change the property event args if filter is null. Is that because a filter causes us to change another list as opposed to `this` list?

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
Haacked: Also, is there a reason we don't just use a `SortedList` to handle all this sort and positioning logic?

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
shana: Sorting can technically happen off the main thread if there's a filter, because no collection change events will be triggered during sorting if there's a filter (sorting in this case happens on a separate list than the one that triggers collection change events). However, there might be a race condition between sorting and filtering, so I might drop this.

Haacked: Would it make sense to extract this inner list into our own sorted list? Perhaps `NotifyingSortedList` with a single method for merging items that returns information about what happened?

That'd be a nice way of keeping that responsibility separate from this class.

src/GitHub.Exports.Reactive/Collections/ITrackingCollection.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using GitHub.Helpers;
+
+namespace GitHub.Collections
+{
+    public interface ITrackingCollection<T> : IDisposable, IList<T>  where T : ICopyable<T>
```
Haacked: Please document this interface. :smile:

src/GitHub.Exports/Helpers/ICopyable.cs
shana: Oh yeah, anything that wants to get tracked by this collection needs to implement this interface, but this PR only contains the collection and test interfaces, and this is mocked in the tests. Implementations are on subsequent PRs :smile:

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+                        idxSmallest = i;
+                if (idxSmallest == start) continue;
+                Swap(list, start, idxSmallest);
+
+                // if there's a filter, then it's going to trigger events and we don't need to manually trigger them
+                if (filter == null)
+                {
+                    OnPropertyChanged(new PropertyChangedEventArgs("Item[]"));
+                    OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Move, this[start], start, idxSmallest));
+                }
```
Haacked: If I understand correctly, this method does a completely new sort, right? Could we just call `list.Sort` and then raise the `NotifyCollectionChangedAction.Reset` event rather than raise an event for each item?

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,472 @@
...
+            }
+        }
+
+        void UpdateFilter(Func<T, int, IList<T>, bool> newFilter)
+        {
+            if (filter == null && newFilter == null)
+                return; // nothing to do
```
Haacked: Is this even valid to do this? Should we throw an exception to indicate that this is a mistake? Also, what about the case where `newFilter == filter`? An edge case, but we should probably just handle it too.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
shana: This is an implementation of insertion sort - http://www.sorting-algorithms.com/insertion-sort
It's the fastest if the list is already almost sorted and spends minimal memory doing it.
In the case where there's no filter, we're sorting the underlying `Items` property of the collection, which triggers collection changed events for the UI, so we're a bit stuck with it. If we are using a filter, we need to figure out what happens to the object (moved from x to y, removed, inserted, appended, not changed) so we can tell the filter what happened and which items to process, so we end up having to control the sorting process ourselves.

We can optimize the `var idx = list.IndexOf(t);` on line 56 for the (likely) case that the observable source is coming in sorted with the same sorting conditions that we're using, and save some time there.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
shana: If the filter is null, then sorting is happening directly on the `Items` property, and in the case of a move, we need to raise move events because the swapping we do to sort things doesn't trigger them.

If filter is not null, then sorting is happening on a backup list, and the UpdateFilter step is going to apply the move on the `Items` collection, triggering the events there.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
shana: Yeah, `list.Sort` makes sense here.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
shana: If the sorting comparer is reset, it triggers an update to the filter which goes through this method. If the filter is null, it will hit this codepath, and it's valid to just return as a noop.

If the new filter is the same as the existing filter, it will be recalculated as if it's a new filter - again, because if the sorting comparer is changed, everything needs to be recalculated regardless of whether the filter has changed or not.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
Haacked: :cool: Maybe we could extract that logic into an `InsertionSortedList`. :smile:

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
Haacked: :+1:

Haacked:
Ok, I'm done with my first pass. :ice_cream:

shana:
I made it always use two lists (one with everyty, one filtered), both sorted. It handles adding and removing items outside of the source observable now. It also regulates the interval betweyitem processing so it doesn't overload the UI too much.

The tests are red due to a stack overfyin xunit itself whicthas nothing to do with the tests (it randomly does it locally too, with the stacktrace full of xunit RunAsync). Not sure if it's threading or something to do with the alternate ctor I'm using on the test class for grabbing test output. :-/

I'll see what's up with xunit tomorrow, but in the meantime, this is ready for review.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+            Add,
+            Insert
+        }
+
+        readonly CompositeDisposable disposables = new CompositeDisposable();
+        IObservable<T> source;
+        IObservable<T> sourceQueue;
+        Func<T, T, int> comparer;
+        Func<T, int, IList<T>, bool> filter;
+        IScheduler scheduler;
```
Haacked: This could be `readonly`

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+        // lookup optimizations
+        // for speeding up IndexOf in the unfiltered list
+        readonly Dictionary<T, int> sortedIndexCache = new Dictionary<T, int>();
+
+        // for speeding up IndexOf in the filtered list
+        readonly Dictionary<T, int> filteredIndexCache = new Dictionary<T, int>();
+
+        TimeSpan delay;
+        TimeSpan requestedDelay;
+        TimeSpan fuzziness;
```
Haacked: This could be `readonly` too.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+
+        public TrackingCollection(Func<T, T, int> comparer = null, Func<T, int, IList<T>, bool> filter = null, IScheduler scheduler = null)
+            : this()
+        {
+#if DISABLE_REACTIVEUI
+            this.scheduler = GetScheduler(scheduler);
+#else
+            this.scheduler = scheduler ?? RxApp.MainThreadScheduler;
+#endif
+            this.comparer = comparer ?? new Func<T, T, int>((o, p) => Comparer<T>.Default.Compare(o, p));
```
Haacked: The `new Func<T, T, int>` isn't necessary here.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+            var index = GetIndexFiltered(item);
+            InternalRemoveItem(item);
+            RecalculateFilter(original, index, position, original.Count);
+        }
+
+        void SetAndRecalculateSort(Func<T, T, int> compare)
+        {
+            comparer = compare;
+            var list = filter != null ? original : Items as List<T>;
+            RecalculateSort(list, 0, list.Count);
```
Haacked: `list` in theory could be null here because `Items` doesn't have to be a `List<T>`. In practice, it always will be in this specific case unless the implementation of `ObservableCollection<T>` or `Collection<T>` changes. So I think we should assert that `list != null` before this line.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+            // the move caused the object to not be visible in the live list anymore, so remove
+            if (!data.IsIncluded && data.Index >= 0)
+                RemoveAndRecalculate(data.List, data.Item, filteredListChanged ? 0 : data.Index, filteredListChanged ? startPosition : start);
+
+            // the move caused the object to become visible in the live list, insert it
+            // and recalculate all the other things on the live list from the start position
+            else if (data.IsIncluded && data.Index < 0)
+            {
+                start = startPosition < start ? startPosition : start;
+                end = endPosition > end ? endPosition : end;
```
Haacked: This `end` is not used anywhere.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+
+        /// <summary>
+        /// Insert an object into the live list at liveListCurrentIndex and recalculate
+        /// positions for all objects from the position
+        /// </summary>
+        /// <param name="list">The unfiltered, sorted list of items</param>
+        /// <param name="item"></param>
+        /// <param name="index"></param>
+        /// <param name="position">Index of the unfiltered, sorted list to start reevaluating the filtered list</param>
+        void InsertAndRecalculate(IList<T> list, T item, int index, int position, bool rescanAll)
```
Haacked: There's no `<param/>` description of `rescanAll`.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+        /// <summary>
+        /// Move an object in the live list and recalculate positions
+        /// for all objects between the bounds of the affected indexes
+        /// </summary>
+        /// <param name="list">The unfiltered, sorted list of items</param>
+        /// <param name="from">Index in the live list where the object is</param>
+        /// <param name="to">Index in the live list where the object is going to be</param>
+        /// <param name="start">Index in the unfiltered, sorted list to start reevaluating the filter</param>
+        /// <param name="end">Index in the unfiltered, sorted list to end reevaluating the filter</param>
+        /// <param name="obj"></param>
```
Haacked: There's no `obj` parameter for this method. :smile:

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+
+
+        /// <summary>
+        /// Go through the list of objects and adjust their "visibility" in the live list
+        /// (by removing/inserting as needed). 
+        /// </summary>
+        /// <param name="index">Index in the live list corresponding to the start index of the object list</param>
+        /// <param name="start">Start index of the object list</param>
+        /// <param name="end">End index of the object list</param>
+        void RecalculateFilter(IList<T> list, int index, int start, int end)
```
Haacked: Please document `list`.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+                }
+            }
+
+            // there was no element to the left of the one we want, start at the beginning of the live list
+            if (index < 0)
+                index = 0;
+            return index;
+        }
+
+        void RaiseMoveEvent(T item, int from, int to)
```
Haacked: This method is not used. Should we remove it?

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+        {
+            OnPropertyChanged(new PropertyChangedEventArgs("Item[]"));
+            OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Move, item, to, from));
+        }
+
+        void RaiseResetEvent()
```
Haacked: Ditto. Should this method be removed?

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+            UpdateIndexCache(index, Count, Items, filteredIndexCache);
+            base.InsertItem(index, item);
+        }
+
+        /// <summary>
+        /// Removes an item from the filtered list
+        /// </summary>
+        void InternalRemoveItem(T item)
+        {
+            int idx = -1;
```
Haacked: No need for the assignment. The assigned value is not used since the next method is an `out` not a `ref`.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+            base.MoveItem(oldIndex, newIndex);
+        }
+
+        /// <summary>
+        /// The filtered list always has a cache filled up with
+        /// all the items that are visible.
+        /// </summary>
+        int GetIndexFiltered(T item)
+        {
+            int idx = -1;
```
Haacked: No need for this assignment.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+
+        /// <summary>
+        /// The unfiltered has a lazy cache that gets filled
+        /// up when something is looked up.
+        /// </summary>
+        /// <param name="item"></param>
+        /// <returns></returns>
+        int GetIndexUnfiltered(T item)
+        {
+            int ret = -1;
```
Haacked: No need for this assignment.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+                ret = original.IndexOf(item);
+                if (ret >= 0)
+                    sortedIndexCache.Add(item, ret);
+            }
+            return ret;
+        }
+
+        /// <summary>
+        /// When items get moved/inserted/deleted, update the indexes in the cache.
+        /// If start < end, we're inserting an item and want to shift all the indexes
```
Haacked: XML doc comments must be valid XML. Can't use `<` here. Need to use `&lt;`.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+            if (cache.ContainsKey(list[pos]))
+                cache[list[pos]] += change;
+        }
+
+        static int BinarySearch(List<T> list, T item, Func<T, T, int> comparer)
+        {
+            return list.BinarySearch(item, new LambdaComparer<T>(comparer));
+        }
+
+        static IScheduler GetScheduler(IScheduler scheduler)
```
Haacked: Method is not used.

src/UnitTests/GitHub.Collections/Thing.cs
```diff
@@ -0,0 +1,84 @@
...
+    public event PropertyChangedEventHandler PropertyChanged;
+
+    void OnPropertyChanged(string propertyName)
+    {
+        PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+    }
+
+    public override string ToString()
+    {
+        return string.Format("id:{0} title:{1} created:{2:u} updated:{3:u}", Number, Title, CreatedAt, UpdatedAt);
```
Haacked: Here's a case where you can use the new string interpolation syntax if you wanted to. `$"id:{Number} title:{Title} created:{CreatedAt:u} updated:{UpdatedAt:u}"`

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
Haacked: Actually, it is used if `DISABLE_REACTIVEUI` is set. We should probably wrap this method in the same conditional.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+            if ((data.IsIncluded && data.Index >= 0) || !data.IsIncluded && data.Index < 0)
+                return data;
+
+            // wasn't on the live list, but it is now
+            if (data.IsIncluded && data.Index < 0)
+                InsertAndRecalculate(data.List, data.Item, data.IndexPivot, data.Position, false);
+
+            // was on the live list, it's not anymore
+            else if (!data.IsIncluded && data.Index >= 0)
+                RemoveAndRecalculate(data.List, data.Item, data.Index, data.Position);
```
Haacked: The unit tests don't cover this code path. There should probably be a test for this case.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+
+                // this happens if the original observable is not sorted, or it's sorting order doesn't
+                // match the comparer that has been set
+                else
+                {
+                    idx = BinarySearch(list, item, comparer);
+                    if (idx < 0)
+                        idx = ~idx;
+                    if (idx == list.Count)
+                        ret = new ActionData(TheAction.Add, item, null, list.Count, -1, list);
```
Haacked: This code path is never touched by the unit tests. We should make sure a test exercises it.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,842 @@
...
+            queue.Enqueue(item);
+        }
+
+        public void RemoveItem(T item)
+        {
+            if (disposed)
+                throw new ObjectDisposedException("TrackingCollection");
+            var position = GetIndexUnfiltered(item);
+            if (position < 0)
+                return;
```
Haacked: This code path also untouched by unit tests.

Haacked:
Looking good. I ran unit test coverage on it. Given how fundamental `TrackingCollection` is, we should aim for really high test coverage if we can.

src/UnitTests/GitHub.Collections/Thing.cs
shana: Oh nice, gotta remember that!

src/UnitTests/GitHub.Collections/Thing.cs
```diff
@@ -0,0 +1,84 @@
...
+            return true;
+        return other != null && other.Number == Number;
+    }
+
+    public override bool Equals(object obj)
+    {
+        var other = obj as Thing;
+        if (other != null)
+            return Equals(other);
+        return base.Equals(obj);
```
Haacked: I think we can return `false` here because in this case, we know that `obj` is not a `Thing`, but we definitely _are_ a `Thing` so we can't be equal. `base.Equals` in this case tests for reference equality and that can't be true here.

Haacked:
My current theory is that the use of `ManualResetEvents` in the tests are causing xUnit issues. When I debug the stack trace, it seems to be happening here: https://github.com/xunit/xunit/blob/master/src/xunit.execution/Sdk/MaxConcurrencySyncContext.cs#L93

Haacked:
By the way, `src/UnitTests/GitHub.Collections/TrackingCollectionTests2.cs` is so large, it doesn't show the diff in this PR. So I haven't reviewed it yet. I think it might help to move `TestBase` into its own file since it's also used by `TrackingCollectionTests.cs`. Also document it and explain how it's different from `TestBaseClass`. It's a little confusing that we have both. Can they be combined?

Also, many of the tests in `TrackingCollectionTests2` are really hard to follow. With code like 

```csharp
        var list1 = new List<Thing>(Enumerable.Range(1, count).Select(i =>
            new Thing() { Number = i, Title = "Run 1", CreatedAt = now + TimeSpan.FromMinutes(i), UpdatedAt = now + TimeSpan.FromMinutes(count - i) })).ToList();
```

It might be useful to have helper methods that create these things and explain what and why it's creating it.

Haacked:
Also, in `TrackingCollectionTests2` there are several places where you define a variable `now` but don't go on to use it. Is it a mistake that it's not being used?

Haacked:
I went ahead and moved `TestBase` into its own file. It's still an open question on whether this could be combined with `TestBaseClass` or not.

Haacked:
I logged an xunit issue just in case there's some light they can shed. https://github.com/xunit/xunit/issues/635

I tried converting tests to `Task` and using `await` instead of `ManualResetEvent` but still get the crashes. I'm at a loss here too. I'm starting to think it's something with xUnit. We could downgrade to know for sure.

shana:
The reasons why those tests inherit from TestBase and not the other base class we have is because 1) I'm mainly developing and testing this in https://github.com/shana/handy-things and copying things over, so keeping things isolated is easier, and 2) the TestBaseClass implements the EntryExitDecorator, which I don't need for these tests.

I'm moving these tests to nunit, which I'm sure can handle this stuff much better.

shana:
It's GREEN!

rejoice

All the tests except for two timing tests are running on ci. The timings tests are iffy and likely to fail on a server, so better not have that stuff there.

Also, looks like we can't use your string interpolation thing on the server. Not sure why, but it just won't build and I'm not in the mood to fight it any more.

:package:

Haacked:
Excellent! I'll do another round of reviews tonight!

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,889 @@
...
+
+            var position = GetIndexUnfiltered(item);
+            if (position < 0)
+                return;
+
+            var data = new ActionData(TheAction.Remove, item, null, position - 1, position, original);
+            data = CheckFilter(data);
+            data = CalculateIndexes(data);
+            data = SortedRemove(data);
+            data = FilteredRemove(data);
```
Haacked: `data` isn't returned. So why the assignments? Could these be void methods? Or should `data` be returned?

shana: These methods are part of a pipeline of methods that get called sequentially and that process an item (see Listen method). They might or might not return a different ActionData object as part of the processing. SortedRemove and FilteredRemove aren't technically a part of the pipeline of the Listen method (because they'd always be noop there), but for all intents and purposes they follow the same pattern.
Just pretend these 4 calls are each a Select call on an observable :smile: 

I debated whether to return the removed item (hence the final assignment doing nothing). Might stil do that. In the meantime, the final assignment is just keeping the pattern.

Haacked: If it's a pipeline, shouldn't each method return a new instance of `ActionData` instead of mutating the passed in instance? I think this calling pattern is really confusing.

Also, `SortedRemove` and `FilteredRemove` actually mutate the current `TrackingCollection`. They are very different from the methods called before them, which do some calculations and return an ActionData with the calculations. I think that difference and intent would be better communicated by making those two methods return `void`.

So this method would look like.

```csharp
var data = new ActionData(TheAction.Remove, item, null, position - 1, position, original);
data = CheckFilter(data);
data = CalculateIndexes(data);
SortedRemove(data);
FilteredRemove(data);
```

Where `data` is never mutated. Thoughts?

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,889 @@
...
+
+            filteredIndexCache.Add(item, index);
+            UpdateIndexCache(index, Count, Items, filteredIndexCache);
+            base.InsertItem(index, item);
+        }
+
+        /// <summary>
+        /// Removes an item from the filtered list
+        /// </summary>
+        int InternalRemoveItem(T item)
```
Haacked: The return value is never used. Could we just make this a `void` method since it's private.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,889 @@
...
+    /// <summary>
+    /// TrackingCollection is a specialization of ObservableCollection that gets items from
+    /// an observable sequence and updates its contents in such a way that two updates to
+    /// the same object (as defined by an Equals call) will result in one object on
+    /// the list being updated (as opposed to having two different instances of the object
+    /// added to the list).
+    /// It is always sorted, either via the supplied comparer or using the default comparer
+    /// for T
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
```
Haacked: Might be good to document that this class is not thread safe. So it's up to the consumer to synchronize access to it. For example, if you call `RemoveItem` on two threads, you could run into a race condition because access to `sortedIndexCache` is not synchronized.

Is that going to be a problem? Should we consider using a `ConcurrentDictionary`? Or will it be fine in normal usage patterns?

Haacked:
Ugh, `TrackingCollectionsTests` are too big to render in the PR diff so I can't provide comments and feedback on it.

Haacked:
Just a little bit more to go! Back to you @shana. :smile:

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
shana: `RemoveItem` is the only method that's not threadsafe, because it's the only method that doesn't go through the pipeline that's set up in the `Listen` method. This is because I haven't yet thought of a way to encode the "Remove" action into a stream of objects coming in through the observable (an item read from the source observable is either something we haven't seen before, so it's going to be added, or something we've seen before, so it's going to be updated).

`RemoveItem` also triggers events from the ObservableCollection, which need to be on the main thread (because otherwise WPF will seriously complain). Being a subclass of ObservableCollection, this class has the same threading restrictions as that one - actions that change the class have to be on the main thread.

A call to `RemoveItem` is an edge case, so no need to add code just for it.

shana: `RemoveItem` is the only method that's not threadsafe, because it's the only method that doesn't go through the pipeline that's set up in the `Listen` method. This is because I haven't yet thought of a way to encode the "Remove" action into a stream of objects coming in through the observable (an item read from the source observable is either something we haven't seen before, so it's going to be added, or something we've seen before, so it's going to be updated).

`RemoveItem` also triggers events from the ObservableCollection, which need to be on the main thread (because otherwise WPF will seriously complain). Being a subclass of ObservableCollection, this class has the same threading restrictions as that one - actions that change the class have to be on the main thread.

A call to `RemoveItem` is an edge case, so no need to add code just for it.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
shana: > If it's a pipeline, shouldn't each method return a new instance of ActionData instead of mutating the passed in instance? I think this calling pattern is really confusing.

Yes, if this was a purely functional thing, it would return a new `ActionData` for every call. I don't do that because I don't want to be creating thousands of objects and putting pressure on the GC when I don't have to.

`ActionData` has non-mutable data, which is the data regarding the item that's being processed and where it is in the sorted, unfiltered list (the item, it's position in the list, where it was before, and a reference to the list itself). This data is built on `ProcessItem`. After that, other information is added to `ActionData` so I don't have to keep looking it up, like whether it's visible according to the current filter, it's index in the filtered list, etc. All this (mutable) information is related to the filtered list (the one that is visible via the `Items` property of the base class).

I tried to strike a balance between immutability and the fact that every item processed would have at least 3 "updates" to it's `ActionData` object (or more), and go through 5 methods, and I don't like needlessly throwing pressure at the GC. Whenever data that is essential to the object is calculated, a new `ActionData` is created. Other data merely updates the existing `ActionData` (for instance, if there's no filter, only the data for the unfiltered list is needed)

> Also, SortedRemove and FilteredRemove actually mutate the current TrackingCollection and do not mutate data.

Neither do the other `SortedXXX` and `FilteredXXX` methods that are called via the `Select` calls in the `Listen` method. By the time they're called, the data in `ActionData` is ready for processing by these methods, so they merely use it and pass it through.

In a future where I can figure out how to include Remove actions into the pipeline that's set up in the `Listen` methods, `SortedRemove` and `FilteredRemove` will have to conform to the same pattern as the other `SortedXX` and `FilteredXX` methods. The only reason they're not included in the `Listen` method list of `Select` calls right now is because they would be noops and I don't see the point in having method calls that currently are known to be noops.

shana: I change this so that RemoveItem actually returns the item that was removed. It's a common pattern and useful thing to have (because the item you pass in might not be the exact same instance that is removed). It returns null when it couldn't find anything to remove, so that's a nice thing to have too.

shana:
> Ugh, TrackingCollectionsTests are too big to render in the PR diff so I can't provide comments and feedback on it.

I really don't know what to do about this. I can't understand why we have a limit on PR size without a way to see the full diff on the website, it makes it really pointless for feature work. :cry: 

I'm hacking on a diff commenter in VS so you can comment there on the PR, but can't make that work fast enough to get this thing out the door. If only we had PR reviewing in VS... :stuck_out_tongue_winking_eye:

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
Haacked: > A call to RemoveItem is an edge case, so no need to add code just for it.

Ok.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
Haacked: *bump*

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
Haacked: > and I don't like needlessly throwing pressure at the GC

But you wouldn't be if `ActionData` was a struct, right? It really feels like a struct to me. It's just a value type that defines what action should happen.

shana: > But you wouldn't be if ActionData was a struct, right? It really feels like a struct to me. It's just a value type that defines what action should happen.

You're absolutely correct, it should be a struct, and that makes things make more sense. I'll fix that up right away.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
shana: Oh, I changed that already. So weird, I could have sworn I had pushed...

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -0,0 +1,934 @@
...
+                OldItem = other.OldItem;
+                Position = other.Position;
+                OldPosition = other.OldPosition;
+                List = other.List;
+                Index = index;
+                IndexPivot = indexPivot;
+                IsIncluded = other.IsIncluded;
+            }
+
+            public ActionData(ActionData other, int position)
```
Haacked: This `ctor` is never used. Should we remove it?

Haacked: It seems like some of these `ctors` could be defined in terms of the others.

shana: Cleaned up the constructors. The `ctor` above is being used now (before it was passing default args and hitting another `ctor`, that's fixed now)

shana:
Back to you! :pineapple:

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
Haacked: Did you remember to push those changes? The only change I see is a merge to master. The constructors haven't changed.

shana: I did, but it went to the wrong branch because the deskop app is just not liking me today. All fixed, everything pushed correctly now :P

Haacked:
This is great! I know it's been a long trudge, but I appreciate you humoring my nitpicky reviews.

spin-thumbs

PR #112, opened 2015-10-05
Turn menu creation methods into extension methods

shana:
We're going to need to create menus from more places.

Also consolidate extensions some more, and remove unused menu command id

### DIFF
--- a/src/GitHub.Exports/Extensions/VSExtensions.cs
+++ b/src/GitHub.Exports/Extensions/VSExtensions.cs
@@ -1,11 +1,17 @@
 ﻿using System;
+using System.ComponentModel.Design;
 using System.Diagnostics;
+using GitHub.Primitives;
 using GitHub.Services;
+using LibGit2Sharp;
 using Microsoft.TeamFoundation.Controls;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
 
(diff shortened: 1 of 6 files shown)


src/GitHub.Exports/Extensions/VSExtensions.cs
```diff
@@ -75,5 +81,64 @@ static object GetServiceAndCache<CacheType>(IServiceProvider provider, Type type
...
+                    ((OleMenuCommand)s).Visible = canEnable();
+                },
+                id);
+            mcs.AddCommand(item);
+        }
+    }
+
+    public static class ISolutionExtensions
+    {
+        public static UriString GetRepoUrlFromSolution(IVsSolution solution)
```
Haacked: This is not an extension method. Is that intentional?

Haacked:
:peach: This looks good! Just one minor question.

src/GitHub.Exports/Extensions/VSExtensions.cs
shana: It's an oversight. This method isn't being currently used, and the way it stands right now, it doesn't have valuable logic that's worth keeping it around (GetRepoFromSolution does the same exact thing), so might just kill it.

Haacked: :fire: it!

shana:
:tada:

Haacked:
Excellente!

PR #113, opened 2015-10-05
:art: Add explicit part creation policy, for completeness sake

shana:
Being explicit about such things is nice.


Haacked:
I like being explicit.

PR #114, opened 2015-10-05
Simplify inotify code

shana:
Anything inheriting from `NotificationAwareObject` also needs functionality from `INotifyPropertySource`, so might as well put that in `NotificationAwareObject`


Haacked:
I concur!

PR #115, opened 2015-10-05
Fix the generation of the octicons resources file

shana:
Build misconfigurations annoy the heck out of me.


Haacked:
Well this is an easy enough one to review. :sparkles:

PR #116, opened 2015-10-05
:fire: Cut down on duplicated code

shana:
Tired of stuffing a bunch of ViewModel copies every time I do a new view. Every time you duplicate code, god kills a kitten (and every time you deduplicate code, god restores a kiten, so we should be striving for positive kitten numbers here)


Haacked:
:clap:

PR #117, opened 2015-10-06
GitHub pane, toolbar and menu item

shana:
This PR depends on #112 and #116 

This PR adds:

- a tool window to host the GitHub pane, where GitHub specific features will live. This has no functionality beyond loading the hosting view.
- a GitHubView and GitHubViewModel to host the specific UI views (pr list, detail, creation, etc).

The GitHubView/ViewModel handle hosting the views for specific features (pull request list, pull request creation, detail, etc), as well as loading and navigating between them, using a long-lived instance of UIController (ui logic to be added there in a separate PR). Toolbar commands are controlled by the view model.

The view model gets informed of contextual repository changes in the same way that sections and navigation items do - by hooking up to TeamExplorerServiceHolder (via the constructor) and getting notifications. These notifications are now also sent via property events, as well as `RepoChanged()` calls.


src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -48,9 +52,17 @@ public GitHubPackage(IServiceProvider serviceProvider)
         protected override void Initialize()
         {
 
-            base.Initialize();
-
             ServiceProvider.AddTopLevelMenuItem(GuidList.guidGitHubCmdSet, PkgCmdIDList.addConnectionCommand, (s, e) => StartFlow(UIControllerFlow.Authentication));
+            ServiceProvider.AddTopLevelMenuItem(GuidList.guidGitHubCmdSet, PkgCmdIDList.showGitHubPaneCommand, (s, e) =>
+            {
+                var window = FindToolWindow(typeof(GitHubPane), 0, true);
+                if (window == null || window.Frame == null)
```
Haacked: `if (window?.Frame == null)`

Haacked:
Nice work!

MistyKuu:
@shana @Haacked  hey, this maybe an accident but we are just developing Team Explorer Extension and this pull request is the only page that is found when googling "ConnectArrowBrush".  It seems that VisualStudio is having the same resource name in their code. Have you ever had an issue which caused visual studio static resource holder to throw an exception that resource "ConnectArrowBrush" cannot be found?

shana:
@MistyKuu We define it in our own xaml dictionary (https://github.com/github/VisualStudio/blob/d92ca6f4c5bdb8ea800ef1263a8c052e740049fc/src/GitHub.VisualStudio.UI/SharedDictionary.xaml#L18-L27) that gets included in all our xaml files, so regardless of whether VS has it defined, ours would override it, so never got exceptions from that.

MistyKuu:
@shana  as I suspect Visual Studio sections are loaded BEFORE any extension, aren't they?. Why is it missing if it's a VIsual Studio style?

shana:
@MistyKuu Pages and sections are loaded on demand when they're needed, so if something doesn't require that style, it won't get loaded. The Team Explorer home page is usually the first one to get loaded (but it doesn't have any arrows in it). When a page is navigated away, it gets unloaded. xaml is usually cached so it's not unloaded, but it highly depends on what got loaded when. Don't trust that anything that you might need from Team Explorer is actually loaded at any point in time (all MEF parts are non-shared, services only get created when they're being used, etc).

MistyKuu:
@shana thank you for clarification. Team Explorer actually has arrows in it - in Visual Studio Team invitation section and this is the place where the exception is being thrown, we don't actually use this style in our code. In addition, it doesn't happen on every machine so it's hard to say what is causing missing vs style.

shana:
@MistyKuu I've never seen that exception, and xaml usually doesn't throw exceptions unless dependent DLLs fail to load. Is it an exception or a warning of a missing style? Do the arrows still show up?

MistyKuu:
it's an exception while loading team services which forces visual to close. We catch all unhandled dispatcher exceptions so we get that message in our logs.

shana:
Do you know where it's coming from, what's throwing it?

MistyKuu:
https://gist.github.com/MistyKuu/016943f273933c5aeb4a6a6517c647ab

shana:
That's really odd, and no clue what's trying to load it either. :cry:

MistyKuu:
Thanks for quick response. We will investigate it further.

shana:
What we can do is change the name of our brush for our next release. That might help determine what exactly is crashing there.

MistyKuu:
@shana ah, you misunderstood me. We are creating our own extension for Team Explorer and that is not your extension that is throwing this exception. Your extension works fine!

shana:
@MistyKuu Ah, let me see if I understand correctly: VS is crashing trying to load the ConnectArrowBrush resource. Your extension does not use a resource with this name. As far as you can tell, the only extensions that use a resource with this name is us and Team Explorer itself. You know it's not us crashing, so it must be Team Explorer. Is that correct?

MistyKuu:
yep, exactly. We've uninstalled your extensions just in case and it's still failing.

PR #118, opened 2015-10-06
Pull request list apis with cache

shana:
This PR depends on #111

This PR adds a cache index for akavache objects, and implements getting lists of pull requests from github into the cache and index and streaming them out via TrackingCollection.

CacheIndex stores a list of string keys of all the objects that are a part of a query - all pull requests for a repository (regardless of state); all repositories for a user, etc. When a query is performed (GetPullRequests for instance), the CacheIndex object is obtained from akavache and all the objects referenced by the index keys are returned. If it needs to be refreshed, the fetch function is called and each returned object will be added/refreshed in akavache and the index will also be updated.

This way, no matter what type of filtering is performed on the query, only one object of each type will be saved (keys are in the form `user|repo|objecttype|id`), and we can leverage the fact that GitHub list apis return full objects and not trimmed-down indexes, avoiding doing the same query twice (once for list, once for detail), since the object will be fully cached during listing and immediately available for detail.

We will want to use sql queries directly at some point, for searches and other more advanced uses, but for now this works. We can use these index objects to create real sql indexes later on.


src/GitHub.App/Extensions/AkavacheExtensions.cs
```diff
@@ -34,7 +38,7 @@ public static class AkavacheExtensions
         {
             return Observable.Defer(() =>
             {
-                var absoluteExpiration = blobCache.Scheduler.Now + maxCacheDuration;
+                var absoluteExpiration = blobCache.Scheduler.Now.ToUniversalTime() + maxCacheDuration;
```
Haacked: Given that `Scheduler.Now` returns a `DateTimeOffset`, the call to `ToUniversalTime` is unnecessary. The value returned represents the _same point in time_ as the original value. Here's some code to demonstrate.

```csharp
var localTime = new DateTimeOffset(new DateTime(2007, 6, 15, 
12, 0, 0), TimeSpan.FromHours(8));
var universalTime = localTime.ToUniversalTime();
Console.WriteLine("Local offset: " + localTime.Offset);
Console.WriteLine("Universal offset: " + universalTime.Offset
);
Console.WriteLine("Difference: " + (localTime - universalTime
));
```

__Output__
```
Local offset: 08:00:00
Universal offset: 00:00:00
Difference: 00:00:00
```

Haacked:
I'm still reviewing, but I got tired and decided to :zzz:

src/GitHub.App/Extensions/AkavacheExtensions.cs
shana: Ah, yeah, I just replaced all the calls and didn't notice some didn't need it.

shana: So I poked further, and it's not that Scheduler.Now returns a `DateTimeOffset`, it's that the  implementation actually returns `SystemClock.UtcNow`, which is why there's no difference between the two. Because the property is called `Now` and not `UtcNow`, I thought it returned `DateTimeOffset.Now`, which is local time and definitely not what we want. Sigh at API inconsistencies...

As long as that doesn't change, it's all good.

shana:
I figured out how to delete items from the collection when the cache is refreshed from live data, so now whenever the cache is refreshed from the api, if anything is deleted (probably not pull requests, but when we start tracking repositories with this, it's going to happen), whatever items were deleted remotely will be removed from the cache index and from the collection.

shana:
This branch is a-:+1: and ready for :eyes:!

src/GitHub.App/Models/PullRequestModel.cs
```diff
@@ -0,0 +1,111 @@
...
+        public PullRequestModel(int number, string title, IAccount author, DateTimeOffset createdAt, DateTimeOffset? updatedAt = null)
+        {
+            Number = number;
+            Title = title;
+            Author = author;
+            CreatedAt = createdAt;
+            UpdatedAt = updatedAt ?? CreatedAt;
+        }
+
+        public void CopyFrom(IPullRequestModel other)
```
Haacked: This method is confusing. I was wondering why it doesn't update `Number` and `Author` and it occurred to me, this isn't a "copy" method, it's an "update" method, right? That's why it only updates fields that could possibly change for a given pull request.

Should we perhaps ensure that the `other.Id` matches the current `Id`?

src/GitHub.App/Services/ModelService.cs
```diff
@@ -134,6 +135,35 @@ public IObservable<AccountCacheItem> GetUserFromCache()
...
+                .Select(user => string.Format(CultureInfo.InvariantCulture, "{0}|{1}|pr", user.Login, repo.Name));
+
+            TrackingCollection<IPullRequestModel> col = null;
+
+            var source = Observable.Defer(() => keyobs
+                .SelectMany(key =>
+                    hostCache.GetAndFetchLatestFromIndex(key, () =>
+                        apiClient.GetPullRequestsForRepository(repo.CloneUrl.Owner, repo.CloneUrl.RepositoryName)
+                                 .Select(PullRequestCacheItem.Create),
+                        item => col.RemoveItem(Create(item)),
```
Haacked: `col` could be null here, no? I don't see where it's set to anything but null.

Haacked: Oh wait, I see it's after the call. Maybe protect against null just in case?

shana: It's set 7 lines down, on line 163, but this method doesn't get called until the collection has been subscribed to, which only happens after we return from this method.

shana: Yeah, can do a col? there

Haacked: Actually, in thinking about this, I'm a little bothered by this pattern. I think we should instantiate the `TrackingCollection` above. Then after we create `source` call `col.Listen(source);`

If you look at the ctor of `TrackingCollection` that is called after this, it effectively does exactly that anyways.

Haacked: Just to clarify, `TrackingCollection` has the following ctor.

```csharp
        public TrackingCollection(IObservable<T> source,
            Func<T, T, int> comparer = null,
            Func<T, int, IList<T>, bool> filter = null,
            IScheduler scheduler = null)
            : this(comparer, filter, scheduler)
        {
            this.source = source;
            Listen(source);
        }
```

But `Listen` then overwrites `source`. So this ctor seems pointless. I think we should remove it.

Haacked:
:pear: back to you! There's no visible change to the UI in this PR, right? Still laying groundwork?

shana:
Yeah, no UI here yet, this is the last groundwork PR! After this, we'll have all the pieces in place, just add paint!

src/GitHub.App/Services/ModelService.cs
shana: I fixed the calling code, it was needlessly convoluted as you noted.

The constructor is handy for the tests right now, and will also be handy in other use cases where there's no need for an explicit `ctor` + `Listen` call (it's just required here because the observable itself references the collection, but that might not necessarily be true in other situations).

src/GitHub.App/Models/PullRequestModel.cs
shana: Yeah, makes sense for this implementation to check that the id matches. I've pushed a fix for that.

In this case the copy is only copying what it wants updated, yes. In other cases the copy might be a full deep copy, it's for the implementation to decide. I've documented the interface to clarify this.

src/GitHub.App/Models/PullRequestModel.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            if (Number != other.Number)
+                throw new ArgumentException("Instance to copy from must match", nameof(other));
```
Haacked: Probably would be useful to print out what the current ID is and what the other one was.

src/GitHub.App/Models/PullRequestModel.cs
Haacked: Sorry to nitpick, but would `CopyPropertiesFrom` be a more descriptive name?

shana: Well, you can technically copy fields too, if you want 😄  It depends on the actual implementation, it could grab values via methods to copy, too.

Haacked: True, but this particular implementation only copies properties. :smile:

shana: Well yes, but changing the name of the method in the interface because this implementation happens to only copy properties doesn't really feel logical. 😜

Haacked:
:sparkles:

PR #119, opened 2015-10-06
Pull Request list view and viewmodel

shana:
This PR depends on ~~#111~~ ~~#112~~ ~~#113~~ ~~#114~~ ~~#116~~ ~~#117~~ ~~#118~~

### Tasks
- [x] GitHub pane
  - [x] Navigation
  - [x] Refresh
  - [x] PR button

- [ ] PR list
  - [ ] Filtering
   - [ ] State
     - [ ] UI
     - [ ] Data
   - [x] Assignee
     - [x] UI
     - [x] Data
   - [x] Author
     - [x] UI
     - [x] Data
  - [ ] PR details
  - [x] Jump to detail on the website
  - [x] Jump to PR creation

- [ ] PR creation
 - [ ] Target branch list
    - [ ] UI
    - [ ] Data
 - [ ] Source branch
 - [ ] Title
 - [ ] Description
 - [ ] Optional assignee user
    - [ ] UI
    - [ ] Data
 - [ ] Cancel button
 - [ ] Refresh button
 - [ ] PR creation


shana:
Progress so far:

image

Haacked:
:heart_eyes:

shana:
The filtering is not quite working right, but the dropdown selections are pretty neat.

pr-list-view

/cc @donokuda 👀

donokuda:
Nice! I can also do a pass and add the comment octicon next to the number on the right (based on this mockup from a while back):

image

Should we show a tooltip when someone hovers over the filter dropdown? Right now, if you have an assignee and author selected, it could be confusing to remember which dropdown referred to which.

image

shana:
> I can also do a pass and add the comment octicon next to the number on the right (based on this mockup from a while back):

That would be awesome! I'm flying to SF tomorrow so I won't be able to touch this before wednesday.

> Should we show a tooltip when someone hovers over the filter dropdown? Right now, if you have an assignee and author selected, it could be confusing to remember which dropdown referred to which.

Oh yes, definitely!

Haacked:
I'll start to to look at this. I just noticed that the build failed though. :stuck_out_tongue: :

shana:
@Haacked One of the tests is being silly for some odd reason. The build should be fine though (CI has been having random build problems and I don't know why)

src/GitHub.App/Factories/UIFactory.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    /// <summary>
+    /// This class holds ExportLifetimeContexts (i.e., Lazy Disposable containers) for IView and IViewModel objects
+    /// A view type (login, clone, etc) is composed of a pair of view and viewmodel, which this class represents.
+    /// </summary>
+    public class UIPair : IUIPair
+    {
+        ExportLifetimeContext<IView> view;
+        ExportLifetimeContext<IViewModel> viewModel;
+        CompositeDisposable handlers = new CompositeDisposable();
```
Haacked: This is set here _and_ in the constructor on line 64. Seems like you could remove line 64 and make this `readonly` too.

src/GitHub.App/Factories/UIFactory.cs
```diff
@@ -0,0 +1,103 @@
...
+        UIViewType viewType;
```
Haacked: `viewType`, `view`, and `viewmodel` could be `readonly`.

Haacked: Whoops, I see that you `view` and `viewModel` to null in the `Dispose` method. That's probably not necessary though, is it?

src/GitHub.UI/Behaviours/AddEmptyItemToList.cs
```diff
@@ -0,0 +1,73 @@
...
+                    if (previousSelection == null)
+                    {
+                        previousSelection = AssociatedObject.SelectedItem;
+                        if (defaultValue == null)
+                        {
+                            var binding = AssociatedObject.GetBindingExpression(ListBox.ItemsSourceProperty);
+                            if (binding == null)
+                            {
+                                var m = BindingOperations.GetMultiBindingExpression(AssociatedObject, ListBox.ItemsSourceProperty);
+                                binding = (BindingExpression)m.BindingExpressions[0];
```
Haacked: It wasn't clear to me that `m` is guaranteed not to be null here. Maybe we should assert that.

src/GitHub.UI/Controls/DropDownActionLinkBehaviour.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            Button source = sender as Button;
+            if (source != null && source.ContextMenu != null)
+            {
+                // Only open the ContextMenu when it is not already open. If it is already open,
+                // when the button is pressed the ContextMenu will lose focus and automatically close.
+                if (!isContextMenuOpen)
+                {
+                    source.ContextMenu.AddHandler(ContextMenu.ClosedEvent, new RoutedEventHandler(ContextMenu_Closed), true);
+                    Interlocked.Increment(ref attachedCount);
```
Haacked: Why is the `Interlocked` necessary? Given this is a click handler, I would assume this could only be called on the UI thread.

src/GitHub.VisualStudio/UI/Views/GitHubPaneViewModel.cs
```diff
@@ -0,0 +1,299 @@
...
+using System.Collections.Generic;
+using Microsoft.VisualStudio.Shell;
+
+namespace GitHub.VisualStudio.UI.Views
+{
+    [ExportViewModel(ViewType = UIViewType.GitHubPane)]
+    [PartCreationPolicy(CreationPolicy.NonShared)]
+    public class GitHubPaneViewModel : TeamExplorerItemBase, IGitHubPaneViewModel
+    {
+        CompositeDisposable disposables = new CompositeDisposable();
```
Haacked: `disposables` here could be `readonly`

src/GitHub.VisualStudio/UI/WindowController.xaml.cs
```diff
@@ -7,27 +7,83 @@
...
 namespace GitHub.VisualStudio.UI
 {
-    public partial class WindowController : DialogWindow
+    public partial class WindowController : DialogWindow, IDisposable
     {
-        public WindowController(IObservable<UserControl> controls)
+        IDisposable subscription;
+        IObservable<LoadData> controls;
+        Func<IView, bool> shouldLoad;
+        Func<IView, bool> shouldStop;
```
Haacked: `controls`, `shouldLoad`, and `shouldStop` could all be `readonly`.

src/GitHub.VisualStudio/UI/Views/GitHubPaneViewModel.cs
```diff
@@ -0,0 +1,299 @@
+using System;
+using System.ComponentModel.Composition;
+using System.Reactive.Disposables;
+using System.Reactive.Linq;
+using System.Threading;
+using System.Windows;
+using System.Windows.Input;
+using GitHub.Api;
```
Haacked: This code doesn't compile for me. Missing a `using GitHub.App.Factories;` here.

Haacked: I fixed it myself so I could feel like a real big time code contributor.

Haacked:
:sparkles: Really excited to see this come along! :palm_tree: 

I found a few minor code issues (stuff that could be `readonly` that's not `readonly`).

When I first ran the code, the Experimental instance crashed when I clicked "Pull Requests". When I tried again in the debugger, I could not reproduce the crash. Possible race condition? Perhaps something unrelated? When I shut down the experimental instance, I did get:

> System.StackOverflowException was unhandled
Message: An unhandled exception of type 'System.StackOverflowException' occurred in Microsoft.VisualStudio.Shell.Interop.dll


I noticed that if none of the issues are assigned, we show a slim empty dropdown when we try to filter on assignee. 



We should have a default value in there like `[None]` always show up. When I click the "Open" dropdown nothing happens. What's supposed to happen?

Lastly, the hit target for the dropdowns seems to only be the text. Notice how I can get the arrow between the text and the chevron and it goes back to being an arrow rather than a hand.



We should make the hit target much larger with a bit of padding. @niik or @shiftkey probably could tell you the exact right WPF incantation to make that work correctly.

Haacked:
> We should make the hit target much larger with a bit of padding. @niik or @shiftkey probably could tell you the exact right WPF incantation to make that work correctly.

It occurred to me that I believe our newest team member @grokys is a bit of a WPF expert and might have ideas here too. :smile:

src/GitHub.UI/Controls/DropDownActionLinkBehaviour.cs
shana: Good question, this is copied over straight from the Desktop app, haha!
I'm actually going to :fire: this since I have a better way of doing it without having to use Context Menus, so there 😄

grokys:
I'd have to take a proper look but if you're overriding the control template, this can happen when no background is set on the container Panel. Just setting `Background="Transparent"` will usually make it work.

PR #122, opened 2015-10-08
ToRepositoryUrl should never throw, it's a conversion helper

### LINKED ISSUE #121: Invalid origin url crashes Visual Studio
1. Install the Github extension (1.0.14.0) on VS 2015
2. Add a remote 'origin' with url 'asdf'
3. Watch the fireworks

Any further attempts to connect to that repo will crash Visual Studio.
If you restart Visual Studio, it will remain connected, and opening Team Explorer will cause a crash.

Relevant parts of the call stack:

    System.dll!System.Uri.CreateThis(string uri, bool dontEscape, System.UriKind uriKind) Line 49
    System.dll!System.Uri.Uri(string uriString) Line 400
    System.dll!System.UriBuilder.Uri.get() Line 320
    GitHub.Exports.dll!GitHub.Primitives.UriString.ToRepositoryUrl()	
    GitHub.Exports.dll!GitHub.VisualStudio.Services.GetUri(LibGit2Sharp.Repository repo)	
    GitHub.VisualStudio.dll!GitHub.VisualStudio.Base.TeamExplorerItemBase.RepoChanged()	
    GitHub.VisualStudio.dll!GitHub.VisualStudio.Base.TeamExplorerNavigationItemBase.UpdateRepo(Microsoft.VisualStudio.TeamFoundation.Git.Extensibility.IGitRepositoryInfo repo)	
    GitHub.VisualStudio.dll!GitHub.VisualStudio.Base.TeamExplorerServiceHolder.NotifyActiveRepo()	
    GitHub.VisualStudio.dll!GitHub.VisualStudio.Base.TeamExplorerServiceHolder.ActiveRepo.set(Microsoft.VisualStudio.TeamFoundation.Git.Extensibility.IGitRepositoryInfo value)	
    GitHub.VisualStudio.dll!GitHub.VisualStudio.Base.TeamExplorerServiceHolder.CheckAndUpdate.AnonymousMethod__2(object r)	

It’s ultimately throwing a `UriFormatException`.

shana:
Fixes #121

### DIFF
--- a/src/GitHub.Exports/Primitives/UriString.cs
+++ b/src/GitHub.Exports/Primitives/UriString.cs
@@ -131,7 +131,7 @@ namespace GitHub.Primitives
         /// <summary>
         /// Attempts a best-effort to convert the remote origin to a GitHub Repository URL.
         /// </summary>
-        /// <returns></returns>
+        /// <returns>A converted uri, or the existing one if we can't convert it (which might be null)</returns>
         public Uri ToRepositoryUrl()
         {
             // we only want to process urls that represent network resources
--- a/src/GitHub.Exports/Primitives/UriString.cs
+++ b/src/GitHub.Exports/Primitives/UriString.cs
@@ -135,8 +135,7 @@ namespace GitHub.Primitives
         public Uri ToRepositoryUrl()
         {
             // we only want to process urls that represent network resources
-            if (!IsValidUri && !IsScpUri) return url;
-            if (IsValidUri && IsFileUri) return url;
+            if (!IsScpUri && (!IsValidUri || IsFileUri)) return url;
 
             var scheme = url != null && IsHypertextTransferProtocol
                 ? url.Scheme
--- a/src/GitHub.Exports/Primitives/UriString.cs
+++ b/src/GitHub.Exports/Primitives/UriString.cs
@@ -108,6 +108,7 @@ namespace GitHub.Primitives
                 Host = match.Groups["host"].Value.ToNullIfEmpty();
                 Owner = match.Groups["owner"].Value.ToNullIfEmpty();
                 RepositoryName = GetRepositoryName(match.Groups["repo"].Value);
+                IsScpUri = true;
                 return true;
             }
             return false;
@@ -123,6 +124,8 @@ namespace GitHub.Primitives
 
         public bool IsFileUri { get; private set; }
 
+        public bool IsScpUri { get; private set; }
+
         public bool IsValidUri => url != null;
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/GitHub.Exports/Primitives/UriString.cs
```diff
@@ -131,7 +134,9 @@ bool ParseScpSyntax(string scpString)
         /// <returns></returns>
         public Uri ToRepositoryUrl()
         {
-            if (url != null && IsFileUri) return url;
+            // we only want to process urls that represent network resources
+            if (!IsValidUri && !IsScpUri) return url;
```
Haacked: In this case, url is effectively null, right given the definition of `IsValidUri`.

shana: Yes, there's no real change between `url != null` and `IsValidUrl`, but it's more explicit what it's checking for (if it's a valid url that was successfully parsed from a string value).
And you never know, maybe the definition of what constitutes a valid url might change in the future ¯\_(ツ)_/¯

shana: (i.e., it might be a non-null uri that's invalid for other reasons)

src/GitHub.Exports/Primitives/UriString.cs
```diff
@@ -131,7 +134,9 @@ bool ParseScpSyntax(string scpString)
...
+            if (IsValidUri && IsFileUri) return url;
```
Haacked: I think the previous two conditionals could be combined and simplified to:

`if (!IsScpUri && (!IsValidUri || IsFileUri))) return url;`

src/GitHub.Exports/Primitives/UriString.cs
Haacked: I guess my point is does this change simply push the nullref exception further up the stack?

src/GitHub.Exports/Primitives/UriString.cs
shana: I opted for clarity instead of terseness here because I wanted to make the logic easy to read here, since the two conditions depend on three somewhat conflicting states.

shana: Actually, nevermind, there's something else in the uri parsing that's making this check more complicated than it should be, let me fix that.

src/GitHub.Exports/Primitives/UriString.cs
shana: Ah, no, all the callers are prepared to get a null back from this method, null is a valid return.

src/GitHub.Exports/Primitives/UriString.cs
shana: bleh, nevermind, the fix I was thinking of requires a bit more refactoring than I want to do right now.

PR #123, opened 2015-10-08
Fix potential NullReferenceExceptions

Haacked:
I ran R# analysis and found these potential nullrefs.

### DIFF
--- a/src/GitHub.Exports/Primitives/HostAddress.cs
+++ b/src/GitHub.Exports/Primitives/HostAddress.cs
@@ -98,7 +98,7 @@ namespace GitHub.Primitives
             if (ReferenceEquals(this, obj))
                 return true;
             var other = obj as HostAddress;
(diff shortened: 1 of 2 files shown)


src/GitHub.Exports/Primitives/HostAddress.cs
```diff
@@ -98,7 +98,7 @@ public override bool Equals(object obj)
             if (ReferenceEquals(this, obj))
                 return true;
             var other = obj as HostAddress;
-            return obj != null && WebUri.IsSameHost(other.WebUri) && ApiUri.IsSameHost(other.ApiUri);
+            return obj != null && other != null && WebUri.IsSameHost(other.WebUri) && ApiUri.IsSameHost(other.ApiUri);
```
shana: Oh, this is totally a bug, it should be `other != null` right after the cast!

shana:
Nice catch! :sparkles:

PR #126, opened 2015-10-09
Await the Vacuum call

Haacked:
I just happened to notice this in the build output. We're supposed to await that call.


shana:
This is blowing up the test.

Haacked:
I'll look into this.

shana:
This is fixed now. Received() doesn't work well with GetAwaiter/await calls.

Haacked:
:sparkles:

PR #133, opened 2015-10-23
Fix icon color switch on theme change

### LINKED ISSUE #110: Invitation Card: Octocat logo is black on dark gray on the Visual Studio dark theme
After installing the GitHub extension on a new machine, the invitation card shows the logo in black on a very dark background. (This used to invert to white and should have continued to do so.) 

black_on_black

shana:
Fixes #110

### DIFF
--- a/src/GitHub.VisualStudio/Base/TeamExplorerNavigationItemBase.cs
+++ b/src/GitHub.VisualStudio/Base/TeamExplorerNavigationItemBase.cs
@@ -29,6 +29,7 @@ namespace GitHub.VisualStudio.Base
             VSColorTheme.ThemeChanged += _ =>
             {
                 OnThemeChanged();
+                Invalidate();
             };
 
             holder.Subscribe(this, UpdateRepo);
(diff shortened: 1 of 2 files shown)


Haacked:
This does fix #110 in that the dark theme looks good. But now the light theme is too light!

shana:
Darn it! It switched colors just fine for me when I tested it! Why the heck is it not refreshing?? *grumble*

Haacked:
It switched colors just fine. I even switched to dark and back to light. It was still that faded color. Weird.

shana:
What non-dark theme did you try? Light or Blue? I might only have tested with the Blue one.

Haacked:
I did "Light"

shana:
@Haacked I can tell this is on the connect page (because of the text), and this PR doesn't touch the connect page at all (only the invitation area on the sync page and the navigation items on the home page). 

I can't reproduce this problem though. What color scheme did you start visual studio with, before switching to the light scheme? Or did this happen when it started up?

Haacked:
> I can tell this is on the connect page (because of the text), and this PR doesn't touch the connect page at all (only the invitation area on the sync page and the navigation items on the home page).

Doh! My bad.

> I can't reproduce this problem though. What color scheme did you start visual studio with, before switching to the light scheme? Or did this happen when it started up?

And I can no longer repro it either. I tested everything and this change looks good!

PR #135, opened 2015-10-26
Login in the sync page

### LINKED ISSUE #57: Sync Button should launch login dialog for unauthenticated users
Today, if a user clicks the "Sync" button and tries to sync a GitHub.com repository, they'll receive the standard basic authentication dialog. This will work fine if the user is not using two factor authentication (2fa). But if they are using 2fa, this is not a nice experience.

Instead, if the user clicks Sync, we should launch our own login dialog. And after the user logs in, we should take them to the Sync page (if possible) though it'd be fine to take them back to the home screen and let them click Sync again.

There's a few ways we might be able to accomplish this.

1. Hide the Sync button for unauthenticated users viewing a GitHub.com repository and add our own "Sync" button that launches the login dialog and then navigates to the Sync page.
2. Hook into the Sync button and override the event handling to launch our dialog. (I have no idea if this is possible).
3. We don't show the Team Explorer section and rebuild the whole section ourselves (ewww!)
4. Something else that people more knowledgeable about Team Explorer might think up.

Also, we need to make sure we don't forget to also hijack the menu if possible.

shana:
Fixes #57

If the current repository is a github/github enterprise repo but the user isn't logged in to its instance, then network git operations might fail. Show a warning on the Sync page if this is the case, and allow the
user to login from there.

### DIFF
--- a/src/GitHub.Api/SimpleApiClient.cs
+++ b/src/GitHub.Api/SimpleApiClient.cs
@@ -66,7 +66,12 @@ namespace GitHub.Api
                     }
                 }
             }
-            // it'll throw if it's private
+            // it'll throw if it's private or an enterprise instance requiring authentication
+            catch (ForbiddenException)
+            {
+                if (!HostAddress.IsGitHubDotComUri(OriginalUrl.ToRepositoryUrl()))
+                    isEnterprise = true;
+            }
             catch {}
             finally
             {
--- a/src/GitHub.Api/SimpleApiClient.cs
+++ b/src/GitHub.Api/SimpleApiClient.cs
@@ -71,7 +71,7 @@ namespace GitHub.Api
                 }
             }
             // it'll throw if it's private or an enterprise instance requiring authentication
-            catch (ForbiddenException)
+            catch (ApiException)
             {
                 if (!HostAddress.IsGitHubDotComUri(OriginalUrl.ToRepositoryUrl()))
                     isEnterprise = true;
--- a/src/GitHub.Api/SimpleApiClient.cs
+++ b/src/GitHub.Api/SimpleApiClient.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using GitHub.Primitives;
 using GitHub.Services;
 using Octokit;
+using GitHub.Extensions;
 
 namespace GitHub.Api
 {
@@ -23,10 +24,13 @@ namespace GitHub.Api
         bool? isEnterprise;
         bool? hasWiki;
 
-        public SimpleApiClient(HostAddress hostAddress, UriString repoUrl, IGitHubClient githubClient,
+        public SimpleApiClient(UriString repoUrl, IGitHubClient githubClient,
             Lazy<IEnterpriseProbeTask> enterpriseProbe, Lazy<IWikiProbe> wikiProbe)
         {
-            HostAddress = hostAddress;
+            Guard.ArgumentNotNull(repoUrl, nameof(repoUrl));
+            Guard.ArgumentNotNull(githubClient, nameof(githubClient));
+
+            HostAddress = HostAddress.Create(repoUrl);
             OriginalUrl = repoUrl;
             client = githubClient;
             this.enterpriseProbe = enterpriseProbe;
@@ -48,7 +52,7 @@ namespace GitHub.Api
             await sem.WaitAsync();
             try
             {
-                if (owner == null && OriginalUrl != null)
+                if (owner == null)
                 {
                     var ownerLogin = OriginalUrl.Owner;
                     var repositoryName = OriginalUrl.RepositoryName;
--- a/src/GitHub.Exports.Reactive/Extensions/ConnectionManagerExtensions.cs
+++ b/src/GitHub.Exports.Reactive/Extensions/ConnectionManagerExtensions.cs
@@ -12,7 +12,8 @@ namespace GitHub.Extensions
         {
             return cm.Connections.ToObservable()
                     .SelectMany(c => c.Login())
-                    .Select(c => hosts.LookupHost(c.HostAddress)).Any(h => h.IsLoggedIn);
+                    .Select(c => hosts.LookupHost(c.HostAddress))
+                    .Any(h => h.IsLoggedIn);
         }
 
         public static IObservable<bool> IsLoggedIn(this IConnectionManager cm, IRepositoryHosts hosts, HostAddress address)
@@ -20,7 +21,8 @@ namespace GitHub.Extensions
             return cm.Connections.ToObservable()
                     .Where(c => c.HostAddress.Equals(address))
                     .SelectMany(c => c.Login())
-                    .Select(c => hosts.LookupHost(c.HostAddress)).Any(h => h.IsLoggedIn);
+                    .Select(c => hosts.LookupHost(c.HostAddress))
+                    .Any(h => h.IsLoggedIn);
         }
     }
 }
--- a/src/GitHub.Exports.Reactive/Extensions/ConnectionManagerExtensions.cs
+++ b/src/GitHub.Exports.Reactive/Extensions/ConnectionManagerExtensions.cs
@@ -2,6 +2,7 @@
 using GitHub.Models;
 using System;
 using ReactiveUI;
+using GitHub.Primitives;
 
 namespace GitHub.Extensions
 {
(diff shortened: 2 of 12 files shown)


Haacked:
Very nice! I tried this out with a github.com repository and it worked great. However, I tried it with a GitHub Enterprise repository and it didn't work. Here's the steps I took.

1. Log out of both GitHub.com and GitHub Enterprise.
2. Open a solution that has its remote pointing to a GitHub Enterprise instance.
3. Go to the Sync tab.

Expected: An error that shows I'm not logged into the GitHub Enterprise instance corresponding to the remote.
Actual:



Note that the tooltip says "api.github.com". It should probably have the domain of the GitHub Enterprise instance. When I click "Login" it launched the login dialog, but the dialog had "GitHub" selected instead of GitHub Enterprise.

src/GitHub.Exports.Reactive/Extensions/ConnectionManagerExtensions.cs
```diff
@@ -13,5 +14,13 @@ public static IObservable<bool> IsLoggedIn(this IConnectionManager cm, IReposito
                     .SelectMany(c => c.Login())
                     .Select(c => hosts.LookupHost(c.HostAddress)).Any(h => h.IsLoggedIn);
         }
+
+        public static IObservable<bool> IsLoggedIn(this IConnectionManager cm, IRepositoryHosts hosts, HostAddress address)
+        {
+            return cm.Connections.ToObservable()
+                    .Where(c => c.HostAddress.Equals(address))
+                    .SelectMany(c => c.Login())
+                    .Select(c => hosts.LookupHost(c.HostAddress)).Any(h => h.IsLoggedIn);
```
Haacked: I think putting the `.Any(...)` call here on a new line aligned with `.Where` `.SelectMany` and `Select` would be more consistent with how we chain methods elsewhere.

src/GitHub.VisualStudio/Base/EnsureLoggedInSection.cs
```diff
@@ -0,0 +1,68 @@
...
+            CheckLogin();
+        }
+
+        protected override void RepoChanged()
+        {
+            base.RepoChanged();
+            CheckLogin();
+        }
+
+        async void CheckLogin()
```
Haacked: Unless this is used as an event handler delegate (which it doesn't appear to be), we should [avoid using `async void`](http://haacked.com/archive/2014/11/11/async-void-methods/) and return a `Task` instead.

Haacked:
:ice_cream:

shana:
Looks like our detection of enterprise instances when not authenticated was not quite correct. Now it should correctly detect them, the message should use the correct domain name and the home page should show the information about the repository if it's an enterprise instance repo (when not logged in).

shana:
> When I click "Login" it launched the login dialog, but the dialog had "GitHub" selected instead of GitHub Enterprise.

There's no way right now of telling the login dialog which tab to show. That's going to take a bigger change of how we instantiate views in general (to pass contextual information to them), and I need to do that for the PR views as well (part of the UIController refactor I'm doing), so I'll fix that once I have a system to do it properly in place.

src/GitHub.Api/SimpleApiClient.cs
```diff
@@ -66,7 +66,12 @@ public class SimpleApiClient : ISimpleApiClient
                     }
                 }
             }
-            // it'll throw if it's private
+            // it'll throw if it's private or an enterprise instance requiring authentication
+            catch (ForbiddenException)
+            {
+                if (!HostAddress.IsGitHubDotComUri(OriginalUrl.ToRepositoryUrl()))
```
Haacked: Flow analysis seems to think `OriginalUrl` could be `null`. However, it looks to me that's not legitimate. So maybe add an `Assert` here that it's not null?

src/GitHub.Extensions/TaskExtensions.cs
```diff
@@ -20,5 +20,8 @@ public static class TaskExtensions
                 return default(T);
             }
         }
+        public static void Forget(this Task task)
```
Haacked: Huh? What's this for? Shouldn't we just call `task.Wait` where we'd call this method?

src/GitHub.VisualStudio/Base/EnsureLoggedInSection.cs
```diff
@@ -0,0 +1,73 @@
...
+        public override void Initialize(IServiceProvider serviceProvider)
+        {
+            base.Initialize(serviceProvider);
+            CheckLogin().Forget();
+        }
+
+        protected override void RepoChanged()
+        {
+            base.RepoChanged();
+            CheckLogin().Forget();
```
Haacked: Shouldn't this be `.Wait()`?

src/GitHub.VisualStudio/Base/EnsureLoggedInSection.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        async Task CheckLogin()
+        {
+            // this is not a github repo, or it hasn't been published yet
+            if (ActiveRepo == null || ActiveRepoUri == null)
+                return;
+
+            var isgithub = await IsAGitHubRepo();
+            if (!isgithub)
+                return;
```
Haacked: What if it's an Enterprise repo? If it's not a GitHub repository, shouldn't we use the `EnterprisePrope.Probe` method to see if it's a GitHub Enterprise instance?

src/GitHub.VisualStudio/Base/EnsureLoggedInSection.cs
shana: The callers don't really care when `CheckLogin` is done, since `CheckLogin` contains all the logic to show the notification information. It's a a fire-and-forget thing.

src/GitHub.VisualStudio/Base/EnsureLoggedInSection.cs
shana: `IsAGitHubRepo` does all the necessary checks. IsAGitHubRepo doesn't mean it's a github.com repo, it means the repository is hosted in a github instance (public or enterprise).

src/GitHub.VisualStudio/Base/EnsureLoggedInSection.cs
shana: I think having a `Forget` method makes it clear that we really don't care about waiting for this method, when it's done it's done.

src/GitHub.Api/SimpleApiClient.cs
shana: Yeah, flow analysis is completely wrong here, this exception can only be thrown inside the condition that guarantees that `OriginalUrl` is not null (and `OriginalUrl` is readonly it's impossible for it to be changed in a way that makes it non-null to go into the condition and null in the catch block). I'm doing some changes to the constructor so this it not an issue anymore.

shana:
I've fixed things up and this branch is totally :100: ready! :sparkles:

Haacked:
I tested it with a GitHub Enterprise repository and it still doesn't work for me.

shana:
Do you get information about the repo on the team explorer home page when you're not logged in? I've tried with garage and with a test enterprise instance, what instance are you trying with? Maybe there's something different about the instance you're using that's tripping the check wrongly.

Haacked:
> Do you get information about the repo on the team explorer home page when you're not logged in?

Nope. Just the basic Git stuff.

> I've tried with garage and with a test enterprise instance, what instance are you trying with? Maybe there's something different about the instance you're using that's tripping the check wrongly.

I'm trying with a repo in our internal GitHub Enterprise instance. You know the one. :smile:

shana:
Ok, there's some creepy threading action going on which is making things not load properly. Fixicating...

shana:
@Haacked Ok, fixed it. Looks like different enterprise instances throw different types of exceptions when you're not logged in, and the exception I was catching was a little too specific.

src/GitHub.Extensions/TaskExtensions.cs
shana: BTW, this is the same as https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.threading.tplextensions.forget.aspx (turns out I'm not the only one with the awesomely imaginative method naming conventions :P)

Haacked: I keep thinking `Tasks` are "cold" (aka lazy) when they're not, which is why I was confused that this would actually work. This name is appropriate.

Haacked:
Ok, I'll test this right now.

Haacked:
> There's no way right now of telling the login dialog which tab to show. That's going to take a bigger change of how we instantiate views in general (to pass contextual information to them), and I need to do that for the PR views as well (part of the UIController refactor I'm doing), so I'll fix that once I have a system to do it properly in place.

Ok, I'll log a separate issue for this.

Also, I noticed that after I logged in, the message didn't automatically go away. And if I log out and then go back to the Sync button, I don't see the message. I'll log separate issues for that too. I think the current behavior is good enough for now.

Haacked:
thumbs-up-phil-dunphy

PR #136, opened 2015-10-26
Fix the order of the connect section

shana:
Make sure the second connection shows up above the hosted service providers section, when both are visible.


Haacked:
Screenshot of before and after?

shana:
## Before
screen shot 2015-10-27 at 1 30 42 pm

## After
screen shot 2015-10-27 at 1 33 41 pm

Haacked:
:sparkles:

PR #141, opened 2015-11-03
Update manifest.com

amytruong:
Added scenarios to expand test coverage.

### DIFF
--- a/documentation/manifest.md
+++ b/documentation/manifest.md
@@ -82,6 +82,15 @@
         - [ ] Progress bar appears above Synchronization title
         - [ ] Team Explorer view changes to Home page
         - [ ] Notification appears: "Repository published successfully"
+        - [ ] Publish a private repo and verify on .com that it's private 
+- [ ] **Project section (Home button)*** 
+    - [ ] Click on "Home" icon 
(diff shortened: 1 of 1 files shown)


documentation/manifest.md
```diff
@@ -82,6 +82,15 @@
         - [ ] Progress bar appears above Synchronization title
         - [ ] Team Explorer view changes to Home page
         - [ ] Notification appears: "Repository published successfully"
+        - [ ] Publish a private repo and verify on .com that it's private 
+- [ ] **Project section (Home button)*** 
+    - [ ] Click on "Home" icon 
+        - [ ] Verify Project has the following sections/buttons: Changes, Pull Requests, Branches, Sync, Pulse, Graphs, Wikis, Issues, Wikis, and Settings
```
shana: Thanks for adding this!

Wikis and issues will depend on whether the repository has them enabled, so they might not show up all the time.
The Changes, Branches, Sync and Settings buttons are from Microsoft, so probably no need to validate those. I imagine they aren't going anywhere, but that depends on what MS does with VS.

amytruong: 👍 thanks @shana. I'll clarify in the manifest that Wikis and Issues _should_ display when you're signed into github and the repo has it enabled.

shana:
:sparkles:

PR #146, opened 2015-11-09
Do not export IServiceProvider

### LINKED ISSUE #137: Do not export IServiceProvider
VS extensions should not export `System.IServiceProvider` (without an explicit, non-default contract name). This is because this is a globally shared type name and if two folks exported it, it would cause everyone who imports it to fail since they expect only one export but got more than one. Generally speaking, you should only export your own types, or BCL types with an explicit contract name that has your namespace in it. 

For example, this is wrong:

```csharp
[Export(typeof(IServiceProvider))]
```

But this is OK:

```csharp
[Export("Github.ExtensionForVS.IServiceProvider", typeof(IServiceProvider))]
```

This is where [GitHub's extension violates this rule](https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio/Services/UIProvider.cs#L25).

Now in this particular case, if the `IServiceProvider` that you actually want is the VS global service provider, you don't need to export anything, but simply import `SVsServiceProvider` (which derives from IServiceProvider):

```csharp
[Import]
SVsServiceProvider GlobalServiceProvider { get; set; }
```

shana:
Fixes #137

### DIFF
--- a/src/GitHub.App/Models/ConnectionRepositoryHostMap.cs
+++ b/src/GitHub.App/Models/ConnectionRepositoryHostMap.cs
@@ -1,7 +1,6 @@
-﻿using System;
-using System.ComponentModel.Composition;
-using GitHub.Extensions;
+﻿using System.ComponentModel.Composition;
 using GitHub.Models;
+using GitHub.Services;
 
 namespace GitHub.ViewModels
 {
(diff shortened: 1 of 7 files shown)


Haacked:
applause-crowd

PR #148, opened 2015-11-09
Fix setting create project dialog path when keys don't exist

### LINKED ISSUE #147: Path is not correct when creating a new project after cloning/creating repo
Conditions:
- New installation of VS
- Create Project dialog has never run

Steps:
- Clone or create a new github repository
- Create a new project from the link in the notification message that shows up once cloning/creation is done

Symptoms:
- Path is incorrect the first time the dialog runs

Yes, it's convoluted, I'm amazed I stumbled onto it :stuck_out_tongue_winking_eye:

shana:
Fixes #147

### DIFF
--- a/src/GitHub.Exports/Services/VSServices.cs
+++ b/src/GitHub.Exports/Services/VSServices.cs
@@ -147,35 +147,47 @@ namespace GitHub.Services
         const string MRUKeyPath = "MRUSettingsLocalProjectLocationEntries";
         public string SetDefaultProjectPath(string path)
         {
-            string old;
-            using (var newProjectKey = Registry.CurrentUser.OpenSubKey(NewProjectDialogKeyPath, true))
+            var old = String.Empty;
+            try
             {
-                using (var mruKey = newProjectKey?.OpenSubKey(MRUKeyPath, true))
+                var newProjectKey = Registry.CurrentUser.OpenSubKey(NewProjectDialogKeyPath, true);
+                if (newProjectKey == null)
+                    newProjectKey = Registry.CurrentUser.CreateSubKey(NewProjectDialogKeyPath);
+
+                using (newProjectKey)
                 {
+                    var mruKey = newProjectKey.OpenSubKey(MRUKeyPath, true);
                     if (mruKey == null)
-                        return String.Empty;
+                        mruKey = Registry.CurrentUser.CreateSubKey(MRUKeyPath);
 
-                    // is this already the default path? bail
-                    old = (string)mruKey.GetValue("Value0", string.Empty, RegistryValueOptions.DoNotExpandEnvironmentNames);
-                    if (String.Equals(path.TrimEnd('\\'), old.TrimEnd('\\'), StringComparison.CurrentCultureIgnoreCase))
-                        return old;
-
-                    // grab the existing list of recent paths, throwing away the last one
-                    var numEntries = (int)mruKey.GetValue("MaximumEntries", 5);
-                    var entries = new List<string>(numEntries);
-                    for (int i = 0; i < numEntries - 1; i++)
+                    using (mruKey)
                     {
-                        var val = (string)mruKey.GetValue("Value" + i, String.Empty, RegistryValueOptions.DoNotExpandEnvironmentNames);
-                        if (!String.IsNullOrEmpty(val))
-                            entries.Add(val);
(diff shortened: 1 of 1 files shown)


src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -147,36 +147,48 @@ static string PokeTheRegistryForLocalClonePath()
...
         {
-            string old;
-            using (var newProjectKey = Registry.CurrentUser.OpenSubKey(NewProjectDialogKeyPath, true))
+            var old = String.Empty;
+            try
             {
-                using (var mruKey = newProjectKey?.OpenSubKey(MRUKeyPath, true))
+                var newProjectKey = Registry.CurrentUser.OpenSubKey(NewProjectDialogKeyPath, true);
+                if (newProjectKey == null)
+                    newProjectKey = Registry.CurrentUser.CreateSubKey(NewProjectDialogKeyPath);
```
Haacked: Why not

```csharp
                var newProjectKey = Registry.CurrentUser.OpenSubKey(NewProjectDialogKeyPath, true)
                    ?? Registry.CurrentUser.CreateSubKey(NewProjectDialogKeyPath);
```

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -147,36 +147,48 @@ static string PokeTheRegistryForLocalClonePath()
...
+
+                using (newProjectKey)
                 {
+                    var mruKey = newProjectKey.OpenSubKey(MRUKeyPath, true);
```
Haacked: `newProjectKey` could still be null in theory. Perhaps `newProjectKey?.OpenSubKey(...`?

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -147,36 +147,48 @@ static string PokeTheRegistryForLocalClonePath()
...
                     if (mruKey == null)
```
Haacked: Let's use the null propagating operator here `??`

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -147,36 +147,48 @@ static string PokeTheRegistryForLocalClonePath()
...
-                    var entries = new List<string>(numEntries);
-                    for (int i = 0; i < numEntries - 1; i++)
+                    using (mruKey)
                     {
-                        var val = (string)mruKey.GetValue("Value" + i, String.Empty, RegistryValueOptions.DoNotExpandEnvironmentNames);
-                        if (!String.IsNullOrEmpty(val))
-                            entries.Add(val);
-                    }
+                        // is this already the default path? bail
+                        old = (string)mruKey.GetValue("Value0", string.Empty, RegistryValueOptions.DoNotExpandEnvironmentNames);
```
Haacked: `mruKey` could still be null here.

Haacked:
:sake:

src/GitHub.Exports/Services/VSServices.cs
shana: It can't be null, Create will either return something or throw due to permissions or whatever else fails there, which will land in the exception handler.

src/GitHub.Exports/Services/VSServices.cs
shana: Why not indeed :smile:

src/GitHub.Exports/Services/VSServices.cs
shana: Can't be null, it's locked in the using statement, and same logic as the newProjectKey, CreateSubKey will either return something or throw, in which case it will land on the exception handler.

src/GitHub.Exports/Services/VSServices.cs
Haacked: Ah, I see. Could you assert that then both for future code reviewers and to help out static flow analysis? :smile:

src/GitHub.Exports/Services/VSServices.cs
Haacked: Ok, let's assert that then.

Haacked:
Back to you! :chestnut:

shana:
Asserted and null propagated! :football:

Haacked:
:sparkles:

PR #154, opened 2015-11-16
Experimental feature: synchronize forks

seraphshine:
When you fork a repo, it's hard to keep it in sync with the upstream
repository. This experimental feature goes through all the remote
branches of a configured "upstream" remote, grab all local branches
which match by name, have not diverged in history but are behind the
upstream remote branch, updates the local branches via a fast forward
fetch, and pushes the updated local branches to the "origin" remote, so
that the fork is synchronized on the server with upstream.


shana:
@seraphshine Why are you opening a PR for this? It's not ready, not building, not mergeable and it's not your code.

PR #156, opened 2015-11-25
[WIP] Gist support

georgebearden:
Okay, so this is def. a work-in-progress but I wanted to get feedback on a couple of things.

* Octokit.ApiConnection.Connection.Credentials.AuthenticationType is always returning Anonymous, even though the GitHubCredentialStore returns my username/password when GetCredentials is called.  Because of this, the create gist code is creating anonymous gists...  I thought the create gist command was not working because after a few tests I did not see any new gists on my account, but they were being creating anonymously.  Sorry to GitHub for the burst of non-sense anon gists :stuck_out_tongue_winking_eye: Here is the offending code snippet.
```csharp
var githubClient = ServiceProvider.GetExportedValue<IGitHubClient>();
return await githubClient.Gist.Create(newGist);
```
* Should the create gist feature be worked into the UIController workflow?  I can foresee at least the case where a user goes to create a gist but is not logged in to GitHub, which needs to prompt the user for login before creating the gist.
* Right now I am using `IServiceProvider.GetExportedValue` to get the various GitHub services I need, but once the gist code gets refactored out of the `GitHubPackage` class I will switch everything to to use IoC like is being done throughout the rest of the code.

In the mean time, I will be working on the Popup Window the user will see after clicking the 'Create Gist' context menu item.

Stay tuned for more!  Thanks.


georgebearden:
Okay, so Gist Creation has been MEF*ized* :metal:

Onto the next part, which is the Create Gist Window and making sure the user is logged in.

After studying the code for a while, I definitely think it makes sense to incorporate the 'Create Gist' window into the UIController workfow.  I *think* I have a good handle on what needs to be done to achieve this at the code level.  A high-level overview is: 
* User presses the 'Create Gist' context menu item
* A window appears 
  * If the user is not logged in, they go through the log in process, then transition to the Create Gist Window
  * If the user is logged in, they immediately go to the Create Gist Window
  * The Create Gist Window will contain the following
    * A `Public` Checkbox, when true will create a the gist publicly, otherwise privately.  Should this default to public?  Can the value specified last time be preserved?
    * A `Filename` TextBox, which is required.  Is this something that we can auto-generate for the user with some generic name, like 'GistFromVisualStudio_DateTime.Now'.  This would not be very much work, and may help with the "Context Switching" issue that some users report.  The less for the user to do when creating the gist the better.  Plus they can always override the default filename in the TextBox if they want to.
    * A `Description` TextBox, which is optional.
    * A `Create` Button
    * A `Cancel` Button.  Not sure if we need a cancel button if there is already close window button at the top of the window.

shana:
This is great, thanks so much for taking this on! I just came back from the US and am recovering from jetlag, so I can't look at this properly until tomorrow.

In the meantime, regarding authentication, you need to either get an ApiClient instance via the ApiClientFactory and then add Gist methods in the ApiClient, or you need to create a GitHub client in the same way ApiClientFactory does in https://github.com/github/VisualStudio/blob/master/src/GitHub.App/Factories/ApiClientFactory.cs#L34, to have auth correctly configured there.

As for the UI flow, you might want to do the same that the publish flow on the sync page does, which is fire up the auth dialog if it's not authenticated - https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio/TeamExplorer/Sync/GitHubPublishSection.cs#L99

georgebearden:
Thanks for the tips. There is now a `CreateGist` method on ApiClient.  After testing, the credentials are all good now but another small issue came up leading to a failure when creating the gist.  It looks like the GitHub Extensions for Visual Studio only has OAuth scopes for `user, repo, write:public_key`.  This was returned in one of the headers when trying to post the new gist to the api url.  Checking the ApiClient code https://github.com/github/VisualStudio/blob/master/src/GitHub.App/Api/ApiClient.cs#L30 confirms this.

So... I thinks we need to authorize the `gist` scope so we can get write access for gists.

shana:
Feel free to add the `gist` scope to that scope list. I assume you registered a new application on GitHub and updated your local copy of https://github.com/github/VisualStudio/blob/master/src/GitHub.App/Api/ApiClientConfiguration.cs with the clientId/clientSecret data of the application? That also needs to have the `gist` scope to make this work.

georgebearden:
I went ahead and explicitly added the gist authorization scope here https://github.com/georgebearden/VisualStudio/commit/ceb986838192737bd00acc4f4fe6b8114e5c511a.  However, it turns out that specifying it there has no affect on the end result.  Once I got the scopes in my developer application authorization callback url right, the code was creating gists just fine!

Auth UI flow and Gist popup window coming soon!

shana:
@georgebearden Yeah, the scope list on the developer app configuration in GitHub is the one that controls whether you can do something or not. The scope list that you request in the code can specify less access, but never more than what the app is configured to allow on GitHub. 

> Once I got the scopes in my developer application authorization callback url right, the code was creating gists just fine!

Yay!

src/GitHub.App/Services/GistCreator.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            var loggedInComplete = new TaskCompletionSource<bool>();
+
+            var authUI = uiProvider.SetupUI(UIControllerFlow.Authentication, null);
+            authUI.Subscribe(_ => { }, async () =>
+            {
+                var isLoggedIn = await connectionManager.IsLoggedIn(repositoryHosts);
+                loggedInComplete.SetResult(isLoggedIn);
+            });
+            uiProvider.RunUI();
```
shana: Actually, it might be better to subclass from  https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio/Base/EnsureLoggedInSection.cs for this. It already supports launching the UI and showing notifications if you're not logged in. It will require a manual `Initialize(uiProvider)` call to trigger its logic.

src/GitHub.App/Services/GistCreator.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        public async Task<Gist> CreateFromSelectedText()
+        {
+            await EnsureLoggedIn();
+
+            var selectedText = selectedTextProvider.GetSelectedText().ToTask().Result;
+
+            // Show the Create Gist Window, and on Create button clicked execute the following
+            return await apiClient.CreateGist("NameWillBeEnteredInThePopup", true, selectedText)
```
shana: This can be `selectedTextProvider.GetSelectedText().Select(selectedText => apiClient.CreateGist(....)`, no reason to go from an observable to a task back to an observable.

shana: Actually, there's little reason to return a `Task<Gist>` here instead of the original `IObservable<Gist>`. The only reason certain APIs deal with `Task<T>` instead of `IObservable<T>` is because instances of classes created in the Team Explorer home page are constrained to load as few dependencies as possible, and that means no reactive extensions. That's not an issue for this, since this loads on demand only when a user requests a gist to be created, so we can use rx here just fine.

georgebearden: I was wondering about the Task vs. Rx APIs you have, and now that makes sense!

shana: Yeah, we're a bit limited in some areas... This is why there's a big split in the Exports and UI libraries, and also why there's an ApiClient implementation in `GitHub.App` and a separate api library in `GitHub.Api`... `GitHub.Api`, `GitHub.UI`, `GitHub.Exports` and `GitHub.Extensions` are loaded up front as soon as VS and Team Explorer load, and they have the minimum dependencies required to do things. `GitHub.App`, `GitHub.Exports.Reactive`, and `GitHub.UI.Reactive` load when other things get called - namely the popup dialogs, which are "on demand" when the user does some specific action/request.

Unless something is needed in the home page, it's best to avoid adding things to the constrained libraries, and just add them to the non-constrained ones instead.

src/GitHub.App/Services/GistCreator.cs
shana: To clarify: the point is to split up the "am I logged in" logic from the actual gist creation logic. The login logic needs to happen at multiple points all across the extension (ever more as we add more functionality). Right now it's true that it's mostly triggered from the section and navigation items in the team explorer pages, which makes the current code a bit less generic than I would like.

I'm currently rewriting the UIController to allow for ui flows that trigger authentication dialogs automatically, so this flow will fit in that logic. In the meanwhile, the gist creation service should not care about whether it's logged in or not, just like all the other services don't care (see https://github.com/github/VisualStudio/blob/master/src/GitHub.App/Services/RepositoryCloneService.cs for an example). Assume we're logged in already when we get here.

georgebearden: Awesome, thanks for all the feedback today!  I will get on refactoring the points you mentioned :)

shana:
Thinking about your IGistCreator, the functionality it has (removing the login bits) are more of a view model than a service. Since at some point there might be a desire to show the selected text in the gist popup, pretty much everything that the IGistCreator takes in the constructor are things that a view model for the gist popup will also likely need. There's no real need for a gist creation service, as the apiclient call is pretty much that, at least for now.

georgebearden:
I am working on the `GistCreationViewModel` now and was thinking the same thing now that the login is being abstracted out.  While on the subject, should I wait to integrate anything into the `UIController` until you have had a chance to refactor the authentication dialogs?  My plan was to create a new `Gist` enum value on `UIViewType` and then add in the appropriate `machine.Configure(UIViewType.Gist)` stuff.  I probably won't have the `GistCreationView/ViewModel` ready until tomorrow-ish anyways.

shana:
The refactor is mostly done on another branch, but that's going to take a while to go in, so I might have to split it up to get the UIController in sooner. In any case, nothing is going in to master until the next release goes out (which will be this week).

Don't bother with the UIController for now, assume that you're already logged in when doing things, while I figure out the best way for this to work.

georgebearden:
I started working on adding the `View/ViewModel` and one thing led to another and I ended up integrating it into the `UIController`.  I think what happened was I went to show the view model in the `GitHubPackage.cs`, and to be consistent with everything else, decided to use the `UIProvider` which led me down to the `UIController`.  The good news is everything works great :+1, and it only took about 30 minutes so if it needs to get moved somewhere else no big deal.  

I think the last thing is to fill out the stubs in the `GistCreationViewModel` and `GistCreationControl`.  Almost there!

shana:
> I started working on adding the View/ViewModel and one thing led to another and I ended up integrating it into the UIController.

Aaaaah, yes, that's totally fine, it'll just pop up the login UI in the window before it shows the gist creation UI. Very likely that won't affect the SelectedTextProvider because the window is not an editor window with a TextView so... it should automagically work :+1:

shana:
I added the [WIP] prefix for the title while it's still undergoing work. When you feel that this is ready, you can remove the prefix.

georgebearden:
I am going to be working on some unit tests to finish up the work in this branch, but try out the feature when you get a chance!  AFAIK everything is working as expected.

I modeled the view after the create repository view, and was not sure if the create gist view should also contain the account selector.  I left it out for now, but let me know if it should and I can add it in!  

P.S. Thanks for all the guidance over the last couple weeks :smile:

georgebearden:
@shana This PR should be :thumbsup: now.

shana:
@georgebearden Thanks! I ran out of internets and there's a bunch of stuff I need to sort out before vacation, so it's going to take a bit to go through this again, but not to worry, it's not forgotten!

shana:
Hey hey! I've not forgotten this! Just came back from the holidays and slowly digging through the backlog pile and getting things back on track. :construction_worker:

georgebearden:
I hear you there :smile:.  Let me know when you have had a chance to go through things.  I have more features that I would like to implement around the gist creation that will make it pretty slick, but I think I will hold off until we get this baby merged in.

src/GitHub.VisualStudio/GitHub.VisualStudio.vsct
```diff
@@ -134,6 +145,12 @@
     <GuidSymbol name="guidManageConnections" value="{0A014553-A0AA-46DD-8D6B-B8E3178CA435}">
       <IDSymbol name="idManageConnections" value="0x1009"/>
     </GuidSymbol>
+
+    <!-- This is the guid used to group the menu commands together -->
+    <GuidSymbol name="guidCreateGistCommandPackageCmdSet" value="{D265E7EB-F3FC-411D-B4DC-1C6AD9264C4B}">
+        <IDSymbol name="MyMenuGroup" value="0x1020" />
```
shana: Probably want to give this id a slightly more descriptive name, for our future selves.

src/GitHub.App/ViewModels/GistCreationViewModel.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        readonly ISelectedTextProvider selectedTextProvider;
+        readonly IApiClient apiClient;
+
+        [ImportingConstructor]
+        public GistCreationViewModel(ISelectedTextProvider selectedTextProvider, IApiClientFactory apiClientFactory)
+        {
+            Title = Resources.CreateGistTitle;
+            this.selectedTextProvider = selectedTextProvider;
+            this.apiClient = apiClientFactory.Create(HostAddress.GitHubDotComHostAddress);
```
shana: We actually want to do the gist on whatever server the user is connected to, which might be an enterprise instance. You can grab the correct `IApiClient` by copying what the other viewmodels are doing, like https://github.com/georgebearden/VisualStudio/blob/gist-support/src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs#L39

The (private) importing constructor there takes a `IConnectionRepositoryHostMap` and calls the public constructor with the host that matches the connection that's been set by the UIController in the UIProvider service list (either the connection that was passed to RunUI, or the first logged in connection). Just having the pattern of those two constructors will give you the `IRepositoryHost`, which exposes a `ApiClient` property for that particular host.

Something else to be considered is that we likely want to create the gist on the server that corresponds to the currently opened repository, instead of letting it pick whichever one is first (ideally, we'd have an account dropdown to pick where it should create it, but that's a nice extra for later methinks)
To get the correct `IConnection`, in the command handler, you can get an instance of `ITeamExplorerServiceHolder` and get the `ActiveRepo` from it, and then grab the connection from the `ConnectionManager` connections list and pass it on to `RunUI`, more or less like https://github.com/georgebearden/VisualStudio/blob/gist-support/src/GitHub.VisualStudio/Base/EnsureLoggedInSection.cs#L53 does when it calls `IsLoggedIn`. That goes to https://github.com/georgebearden/VisualStudio/blob/gist-support/src/GitHub.Exports.Reactive/Extensions/ConnectionManagerExtensions.cs.

We probably also want to show the server name in the view, so people know where it's going to be created on.

shana: Of course, if there's no connection that matches the `ActiveRepo` (or if `ActiveRepo` is null), passing null to `RunUI` is fine, it'll just have the user login to whatever they want, and then save that `IConnection` so that the view model can grab it when it loads from the (rather convoluted) constructor.

georgebearden:
Everything in your last few notes has been addressed except showing the server name in the view... thanks for pointing me in the right directions!.  I am adding the [WIP] prefix back though because
  1. I discovered that octokit returns a validation exception to us when the content of the file for the new gist is either null, empty or white space.  This was not something I was restricting before, so I need to add something to the Gist popup saying why the `Create` button is disabled.  The better solution would be to disable the `Create GitHub Gist` context menu item if no text is selected, just not sure if that can be done programmatically or not, but going to check there first.
  2. The `Create GitHub Gist` context menu item only displays for certain file types.  In my limited testing, I see the menu item for .cs and .config files, but not for .xaml or .json files.  Any ideas what could be going on there?

More to come soon :smile:

shana:
Ah, I was wondering about 1). You can use `AddDynamicMenuItem` instead of `AddTopLevelMenuItem` to hide the menu entry if it shouldn't be there. That means that the ISelectedText service implementation will need to move out of GitHub.App and live in GitHub.VisualStudio instead (which is fine, I'm trying to keep VS-specific APIs out of GitHub.App anyway. I think we'll need a vs helper assembly soon.). The interface needs to be in GitHub.Exports instead (which is fine, it has no complicated dependencies). You can then check if there's selected text in the `canEnable` callback and hide the menu if there isn't.

As for 2), hmmm, that's odd. The extension is loaded when you try with xaml files, I assume? It's set to load when the git service loads iirc, so there might be circumstances where it won't load until later in the VS lifecycle. We might need to make it load earlier for scenarios like these, if that's the case. But once it's loaded, it stays loaded, so if it's not showing up for "non-code" filetypes (which is what it sounds like), then need to investigate.

src/GitHub.VisualStudio/GitHub.VisualStudio.vsct
```diff
@@ -37,6 +37,10 @@
       <Group guid="guidGitHubToolbarCmdSet" id="idGitHubToolbarMenuGroup1" priority="0x0501">
         <Parent guid="guidGitHubToolbarCmdSet" id="idGitHubToolbar" />
       </Group>
+
+      <Group guid="guidCreateGistCommandPackageCmdSet" id="GistContextMenuGroup" priority="0x0600">
+        <Parent guid="guidSHLMainMenu" id="IDM_VS_CTXT_CODEWIN"/>
```
shana: CODEWIN is probably why the menu entry doesn't show up in non code windows. Some code windows are managed differently and have different GUIDs, so this is annoying.

I've pushed a couple of things to master to deal with this. There's a context menu group there now that hooks up into different types of code windows, so you can parent a command to it and it will show up everywhere.

src/GitHub.App/Services/SelectedTextProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+        public IObservable<string> GetSelectedText()
+        {
+            return Observable.Defer(() =>
+            {
+                var textManager = serviceProvider.GetService(typeof(SVsTextManager)) as IVsTextManager;
+                if (textManager == null)
+                    return Observable.Return(string.Empty);
+
+                IVsTextView activeView;
+                if (textManager.GetActiveView(1, null, out activeView) != VSConstants.S_OK)
```
shana: Do `if (ErrorHandler.Failed(textManager.GetActiveView(1, null, out activeView))` instead

src/GitHub.App/Services/SelectedTextProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+                    return Observable.Return(string.Empty);
+
+                // Maybe we should log here that no text was actually highlighted?
+                string highlightedText;
+                if (activeView.GetSelectedText(out highlightedText) != VSConstants.S_OK)
```
shana: `ErrorHandler.Failed` here too

src/GitHub.VisualStudio/GitHub.VisualStudio.vsct
georgebearden: Awesome!  I will try this out tonight.

shana:
For reference, if there's more windows where the context menu doesn't show up, enable vsip logging to get the guid and id of them, like this: http://blogs.msdn.com/b/dr._ex/archive/2007/04/17/using-enablevsiplogging-to-identify-menus-and-commands-with-vs-2005-sp1.aspx. Guid and CmdID is what you'll want.

georgebearden:
I was wondering how you were finding those random numbers for the different window types, thanks for the link.

georgebearden:
After refactoring the `GistCreationViewModel` to use the `IConnectionRepositoryHostMap` I am seeing a bug that happens roughly 50% of the time we start the `Gist UIControllerFlow`.  The issue appears to be when `ConnectionRepositoryHostMap` requests the `IConnection` here: https://github.com/georgebearden/VisualStudio/blob/master/src/GitHub.App/Models/ConnectionRepositoryHostMap.cs#L13.  When `UIProvider.TryGetService` is called to resolve `GitHub.Models.IConnection`, it reaches this line https://github.com/georgebearden/VisualStudio/blob/master/src/GitHub.VisualStudio/Services/UIProvider.cs#L117 which then throws the following error https://github.com/georgebearden/VisualStudio/blob/master/src/GitHub.VisualStudio/Services/UIProvider.cs#L127-L128

On startup, I verified that the `Connection` class is being instantiated with values you would expect, e.g., *UserName=georgebearden*.

One other note that may help.  Once this error happens it will continue to happen until I sign out of GHfVS.  After I sign back in everything thing will be good for a while, and then the process repeats.  I am going to look into it some more tonight.

georgebearden:
Hi @shana, 

Right now there is an initialization issue when requesting the `ActiveRepo` here: https://github.com/georgebearden/VisualStudio/blob/gist-support/src/GitHub.VisualStudio/GitHubPackage.cs#L80-L81.  `ActiveRepo` is always null until I interact with the `Team Explorer - Home` Pane.  After selecting the `Team Explorer - Home` Pane, `ActiveRepo` is set internally in the `TeamExplorerServiceHolder`. Because I can open the context menu without ever going through any of the `Team Explorer` code paths, I am seeing this issue while you may have not.

Is it possible to call the [Initialize](https://github.com/georgebearden/VisualStudio/blob/gist-support/src/GitHub.VisualStudio/Base/TeamExplorerSectionBase.cs#L89) method manually while passing in an `IServiceProvider` instance? You may have a better idea, because this does not seem like the best way to handle this :smile:

shana:
> After refactoring the GistCreationViewModel to use the IConnectionRepositoryHostMap I am seeing a bug that happens when we start the Gist UIControllerFlow. The issue appears to be when ConnectionRepositoryHostMap requests the IConnection here:

Aaaaaaah, I just spotted a hole in the UIController logic that could cause this. You're logged in and not in a github repo when this happens, right? In this particular combination of states, the IConnection service never gets stored so ConnectionRepositoryHostMap gets null. This never happens usually before all the sections that use this code only run when there's an active repo, so they don't trigger this.

I'll fix this so you can always be sure you'll have a connection when you get to your view model.

> It looks like this is because the current project is not a github repo. My plan is to extend the Hide/Show Create a GitHub Gist dynamic context menu item to also be triggered by if the project is a github repo or not. Does that sound reasonable?

Given that we allow people to login if there's no connection when trying to do a gist, we should good with always allowing gist creation regardless of whether the active repository is a github repo or not.

shana:
> Right now there is an initialization issue when requesting the ActiveRepo here:

Ah yes, I was afraid of this. The problem with this service is that it hooks up to another service which is asynchronous, so we can never be sure when we'll get the information. I'd rather not complicate this too much, so if there's no TeamExplorer service available, use the following fallback code:

```
var vsservices = ServiceProvider.GetExportedValue<IVSServices>();
string path = vsservices?.GetActiveRepoPath() ?? String.Empty;
try
{
  activeRepo = !String.IsNullOrEmpty(path) ? new SimpleRepositoryModel(path) : null;
}
catch (Exception ex)
{
  VsOutputLogger.WriteLine(string.Format(CultureInfo.CurrentCulture, "Error loading the repository from '{0}'. {1}", path, ex));
}
```

This will use a couple of alternate tricks to get at the repo. If it's still null, then we'll either pick whatever connection already exists or let the user log in.

georgebearden:
Creating the `activeRepo` manually using the code above works well for repos that are indeed github repos, but for other repos the `GistCreationViewModel` constructor still fails when `IConnectionRepositoryHostMap` fails to resolve an `IConnection`.  

For gist creation, there really should not be a limitation on the current repo being a github repo at all, only that the user has a github account that we can post the selected text to.  I bet this is why I am running into so many probs with this, as most of the work in this project was probably around the idea that the current repo is a github repo or should become a github repo!  

I feel like the previous way we were getting the `apiClient` from the `IApiClientFactory` was working better, but I understand that we need to make sure we create the `Gist` on the correct account.  I'll go ahead and finish up moving the `ISelectedTextProvider` into the right place so we can use it for the show/hide dynamic context menu while we figure out the right way to implement this.  Thanks @shana for all your help :exclamation:

georgebearden:
Great, thanks @shana.  I will be on the look out for the fix to make sure we always have an `IConnection` for the `GistCreationViewModel`.  I believe there are two more minor things to do then to call this :baby: done.
  1. Change the context menu item to be dynamic based on if text is selected
  2. Give visual feedback to what account the gist is being created on.

georgebearden:
[x] Change the context menu item to be dynamic based on if text is selected - Working as expected
[x] Give visual feedback to what account the gist is being created on. - For a first swing at it, it uses a ui similar to the `RepositoryCreationViewModel` except instead of an account selector combo box it just shows the first account.  Here is a screenshot of the current state of the popup.
create-gist-popup

I am wondering one thing though. I need to find a correlation between the `IApiClient` being returned from the `IRepositoryHost` and the `IAccount` being displayed in the window to make sure the account shown actually reflects which one the api client is going to create the gist on.  I am not sure how to connect the two after looking at them for a moment.

shana:
> I am wondering one thing though. I need to find a correlation between the IApiClient being returned from the IRepositoryHost and the IAccount being displayed in the window to make sure the account shown actually reflects which one the api client is going to create the gist on. I am not sure how to connect the two after looking at them for a moment.

Accounts are alternate profiles that can be associated with a particular user. A user always has at least one "account" (their own), and then possibly others if they are org admins. For our purposes here, we can always just use the user account, which is the first one on the list (returned here: https://github.com/github/VisualStudio/blob/master/src/GitHub.App/Services/ModelService.cs#L71).

This is all looking awesome! Thank you so much for your patience on this, this is great stuff :smile:

shana:
I've pushed a fix to the connection null issue to https://github.com/github/VisualStudio/tree/shana/fix-iconnection-when-logged-in

src/GitHub.Extensions/StringExtensions.cs
```diff
@@ -36,6 +36,16 @@ public static bool IsNullOrEmpty([AllowNull]this string s)
...
         }
 
+        public static bool IsNotNullOrEmptyOrWhiteSpace([AllowNull]this string s)
+        {
+            return !s.IsNullOrEmptyOrWhiteSpace();
+        }
+
+        public static bool IsNullOrEmptyOrWhiteSpace([AllowNull] this string s)
+        {
+            return string.IsNullOrEmpty(s) || string.IsNullOrWhiteSpace(s);
```
shana: `IsNullOrWhiteSpace` already returns true if the string is empty, so no need to do the extra checks

shana:
I'll be merging your branch to a branch in this repo tomorrow morning so I can switch this to the new menu system and resolve conflicts on some things before sending it off to master. That will cause this PR to appear merged, which is true! but not quite :smile:

I'll be opening a separate PR to merge the new branch to master. If you have more things to tweak in here, feel free to push to your (this PRs) branch and I'll grab them, or do a new PR against the branch I'll be creating.

src/GitHub.Extensions/StringExtensions.cs
georgebearden: Nice catch :smile: I missed that `IsNullOrWhiteSpace` returns true for null, empty or white space, I thought it was only true for null or white space...

georgebearden:
Thanks for the update @shana.  AFAIK, this branch is good to go with-holding the `IConnection` issue that you have a fix for. I'll keep an eye on the new branch to make sure things are good!  Whats next? :smile:

shana:
@georgebearden Next is lots of testing and putting out a new release! And then, hmmmm, maybe doing the reverse and opening a gist in VS? I wonder how useful that would be?

georgebearden:
@shana Well if the goal is to keep people in VS as much as people without having to go to other places like the browser, I could see a utility in having a `Gist Explorer` much like the `Solution Explorer` and `Team Explorer`. I would need to check what kind of functionality octokit and the GitHub API exposed.  Ideally, we could query GitHub for just the names of all the `Gist`s for a user, and then if they select one, we could query again to get the contents to view/edit  That sounds like a fun feature to work on!

shana:
@georgebearden So I'm thinking, how do people share gists with each other? They send the url, right? The next step after creating the gist is likely going to be giving the url to someone else, and if that other person is in VS, it probably makes sense that they would want to open the gist in VS, to copy. There could be a menu/section/something where we put the url in and it loads up the gist. Gists are basically repositories, so we can actually clone the repo and make changes to it, or use the octokit api for it if it supports it (haven't looked).

There's also the question of allowing comments on it. For that, I think we might need a separate comment pane, similar to what I'm planning for pull requests.

There's one other improvement to the gist creation dialog that I'm thinking now: embedding an editor window in the pane that shows the copied text. That kinda goes deep into the rabbit hole of projection buffers though, which is a pain.

Haacked:
I haven't been playing close attention to this so apologies for that.

> So I'm thinking, how do people share gists with each other? They send the url, right? The next step after creating the gist is likely going to be giving the url to someone else, and if that other person is in VS, it probably makes sense that they would want to open the gist in VS, to copy.

What does the feature do today?

At minimum, I would expect that after the gist is successfully created, we post the gist URL to the user's clipboard and notify them of that fact. That way they can easily paste it in a chat message or email.

georgebearden:
Currently, there is a dynamic `Create a GitHub Gist` context menu item that is displayed whenever text is selected and the right mouse button is pressed.  This action will open the following dialog: 
image

If the user has not logged into their GitHub account, the Login dialog will be displayed, followed by the Gist dialog after successful login.  After the user presses the `Create` Gist button, the dialog is closed.

It should be easy to copy the new gist's url to the clipboard by subscribing to the `IReactiveCommand` [here](https://github.com/georgebearden/VisualStudio/blob/gist-support/src/GitHub.Exports.Reactive/ViewModels/IGistCreationViewModel.cs#L12) and then getting the `HtmlUrl` of the returned `Gist`.

I cannot speak for what it would take to notify the user the URL was copied as I have not done any VS notifications yet.

The reason we only show the `Create a GitHub Gist` context menu item when text is selected is due to either the octokit.net library or the GitHub API returning an error if we try to post a new Gist with null, empty or white space for the value of a File [here](https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Request/NewGist.cs#L33), which makes sense I think. 

One other nice feature I have been thinking about would be to use the name of the file the selected text is coming from as the default File Name of the gist instead of the generic *GistFromVisualStudio.md` name I have in there right now.

Haacked:
> I cannot speak for what it would take to notify the user the URL was copied as I have not done any VS notifications yet.

One easy way to do this is to not close the dialog after the user presses the `Create` button but instead show a message in the dialog. We could even close the dialog after a few seconds so the user doesn't have to.

Alternatively, we have other places where we call the VS notification API. That might be the cleanest approach now that I think about it.

> One other nice feature I have been thinking about would be to use the name of the file the selected text is coming from as the default File Name of the gist instead of the generic *GistFromVisualStudio.md` name I have in there right now.

I like that. How about you file issues for these two things so we can discuss further. Thanks for your contribution!

[selfie-0](https://github.com/thieman/github-selfies/)

PR #161, opened 2015-12-08
Bump libgit2sharp and fix broken Publish

shana:
The publish repository flow is currently broken in VS Update 1 due to a mismatch between the libgit2sharp we compile against and the one that's served to the extension. The libgit2sharp switch was delayed due to auth issues, so the current version of the extension was still using an older copy of libgit2sharp from the first release of VS, which unfortunately matches the version of the library that runs in all VS versions (regardless of whether it is ABI compatible - it is not). This is totally my bad, I should have pressed on with the libgit2sharp update and switch after the dust settled down over the open sourcing of the extension.

So, this PR bumps the version of libgit2sharp to the latest of the vNext branch upstream, and hooks up the credential cache to libgit2 via a credential provider callback, so it can grab the credentials as it needs to do network operations. This also adds secure variants to the credential cache methods because libgit2sharp supports secure passwords and we already mostly supported it, so credentials are passed from the windows credential manager to libgit2 without having dropping down to insecure `string`

/cc @Haacked


shiftkey:
This all seems pretty great. Should I pull this down and test out specific things?

shana:
@shiftkey Oh! You have VS 2015 without Update 1, right? It would be really handy to see if it doesn't break on that one, since I'm now stuck on Update 1 and don't have another VM at hand to test it. What breaks is:

- When you clone, don't forget the submodule update, the script dir is a private submodule with the oauth tokens for the app (that are useful and you probably want them :) )
- Create a new project of any kind and add it to source control
- Go to Changes and create a commit
- Click on the "sync" link that shows up when the commit is done, in the informational message
- Click on "Get Started" in the GitHub section
- Go through the login dialog that pops up (if you're not logged in to github)
- Fill out and publish the repo in the form that shows up (replacing the GitHub section with the Get Started link)

It should say "Repository published" and possibly the message then disappears because of Other Reasons(tm). It should _not_ say "Blew up on libgit2sharp boo for you" (not a literal transcription of the message but close enough)

src/GitHub.App/Services/GitHubCredentialProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// </summary>
+        /// <returns></returns>
+        public Credentials HandleCredentials([AllowNull]string url, [AllowNull]string username, SupportedCredentialTypes types)
+        {
+            if (url == null)
+                return new DefaultCredentials();
+
+            var host = HostAddress.Create(url);
+            return secureCache.GetObject<Tuple<string, SecureString>>("login:" + host.CredentialCacheKeyHost)
+                .Select(x => new SecureUsernamePasswordCredentials()
```
shiftkey: :lipstick: `()` unnecessary

src/GitHub.App/Services/GitHubCredentialProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+                {
+                    Username = x.Item1,
+                    Password = x.Item2
+                } as Credentials)
```
shiftkey: This is messing with my OCD but I'm really not sure of a more elegant solution :disappointed:

shiftkey:
> Oh! You have VS 2015 without Update 1, right? 

I think one of my VMs hasn't been updated yet. Let me check.

src/GitHub.App/Services/GitHubCredentialProvider.cs
shana: I wanted to run this through GetLoginAsync in ILoginCache, but that has no SecureString handling on the types it returns and I didn't want to go down that rabbit hole tonight.

Also maybe I should be returning null here instead of DefaultCredentials (which takes the windows user credentials)?

I can also move this to an extension method so it's properly typed :smile:

src/GitHub.App/Services/GitHubCredentialProvider.cs
shana: Tell that to my fingers that are auto-typing () {} all the time...

shiftkey: lol

shana:
@shiftkey oh that sucks! I think I have a machine with 2015 so I'll test there. Thanks!

Issue #157, opened 2015-12-01
QA for Version 1.0.15.5


amytruong:
*Might* be bumped to 1.0.16.0. 
TestPad: https://github.ontestpad.com/script/77#2/1/

amytruong:
For: Line 83: 
- [ ] Team Explorer view changes to Home page after repo is published 
![screen shot 2015-12-01 at 2 32 59 pm](https://cloud.githubusercontent.com/assets/10725993/11516523/91f2c8d2-9838-11e5-81aa-578efca8071e.png)

The screen stays on the current page. I _don't_ think this a bug but just a change to the test manifest?

shana:
Nope, that's definitely a bug, or perhaps a regression. I vaguely remember talking about this before.

EDIT: yup, it's a regression: https://github.com/github/VisualStudio/issues/48

shana:
# Release Notes

## Bug fixes

- https://github.com/github/VisualStudio/issues/107 - Path field doesn't have default cloning path of C:\Users\[user]\Source\Repos 
- https://github.com/github/VisualStudio/issues/143 - Cloning from the non-GitHub section shows GitHub message
- https://github.com/github/VisualStudio/issues/147 - Path is not correct when creating a new project after cloning/creating repo

## Enhancements

- When a GitHub repo is open and the user is not logged in, a message shows up in the Sync area with a quick link to login to GitHub, so that git operations won't fail.

amytruong:
Thanks @shana I'll put it as failed with existing issue #48. 

We talked about this in slack but I _think_ this link is broken. :point_down: 
The evilamyt/beer-is-cool (i think it's a link to the repo on .com since the tooltip says
> u:http://github.com/evilamyt/beer-is-cool

![screen shot 2015-12-01 at 3 47 13 pm](https://cloud.githubusercontent.com/assets/10725993/11518216/673cf1c4-9844-11e5-9132-cce191d2d665.png)

I'll also test those bug fixes above and take a look at that new feature.

amytruong:
@shana I was testing out #107 for correct default cloning paths and noticed something a little strange. Wanted to confirm with you first. When logged out, I click on "Clone" and it brings up the Sign In screen. Sign in and I _think_:
Expected Results: The clone pop up displays and I can select/choose and clone as usual
Actual Results: The clone pop up displays but I also see a notification in the Window Explorer window. 
[...]

PR #162, opened 2015-12-09
Fix publish form not going away when done

shana:
Looks like the fix to #48 wasn't quite complete, @amytruong stumbled upon a regression of it in #157.

I tracked it down to the subscriptions going away when there's an error, because `NotifyDone` would be called as soon as the PublishRepository command was done, which would trigger the release of subscriptions. Unfortunately, if the publish error'd out, this would still trigger `NotifyDone`, so any further attempts at publishing would not have the listeners required to properly update the UI afterwards.

The reason we didn't notice this before is because the messages still show up - because those are being sent directly from the view model in GitHub.App into the VS Team Explorer notification area. This is a problem, they should be getting surfaced to the view and then displayed there. This PR doesn't change that, but it adds a bit of progress state reporting to start with (there's another PR around which surfaces error messages, I need to revive it)

Anyways, now it works as it should.


src/GitHub.App/Controllers/UIController.cs
```diff
@@ -142,6 +136,25 @@ public IObservable<UserControl> SelectFlow(UIControllerFlow choice)
...
+        {
+            completion = new Subject<bool>();
+            return completion;
+        }
+
+        void End(bool success)
+        {
+            uiProvider.RemoveService(typeof(IConnection));
+            transition.OnCompleted();
+            completion.OnNext(success);
```
Haacked: `completion` could be null `here` if you didn't call `ListenToCompletionState` first. Should it be an `InvalidOperationException` in that case?

shana: It's an optional listener, so being null is fine, I'll just add the ? there that I forgot.

Haacked: :thumbsup:

shana:
Fixed null things, should be good now!

src/GitHub.App/Controllers/UIController.cs
```diff
@@ -142,6 +136,27 @@ public IObservable<UserControl> SelectFlow(UIControllerFlow choice)
...
         }
 
+        /// <summary>
+        /// Allows listening to the completion state of the ui flow - whether
+        /// it was completed because it was cancelled or whether it succeeded.
+        /// </summary>
+        /// <returns>true for success, false for cancel</returns>
+        public IObservable<bool> ListenToCompletionState()
+        {
+            if (completion == null)
```
Haacked: As a stylistic thing, I tend to like:

```csharp
completion = completion ?? new Subject<bool>();
```

But it's a personal preference. I'll still merge this.

PR #165, opened 2015-12-18
Make really sure we detect an enterprise host correctly

### LINKED ISSUE #164: Incorrect UI while using a non-GitHub repo
I use this plugin for repos both on GitHub and our company's own Git repo.

When using the company repo, and I go to Synchronization, it gives me the following warning at the top: "You are not logged in to [server address], so certain git operations may fail. Login now"

The login now assumes I want to log into GitHub Enterprise, although I'm already logged in to my company's git server/repo. I can commit and everything to the company repo just fine.

So why is this message claiming otherwise? Why is it trying to get me to log in to GitHub Enterprise for a non-GitHub URL/server?

It also shows the company git server repo listed under GitHub on the Home tab, and it shows a bunch of GitHub-specific functionality under the Project toggle, like Pulse and Graphs, even though it's not a GitHub repo.

shana:
Fixes #164

### DIFF
--- a/src/GitHub.Api/SimpleApiClient.cs
+++ b/src/GitHub.Api/SimpleApiClient.cs
@@ -71,10 +71,10 @@ namespace GitHub.Api
                 }
             }
             // it'll throw if it's private or an enterprise instance requiring authentication
-            catch (ApiException)
+            catch (ApiException apiex)
             {
                 if (!HostAddress.IsGitHubDotComUri(OriginalUrl.ToRepositoryUrl()))
-                    isEnterprise = true;
+                    isEnterprise = apiex.HttpResponse?.Headers.ContainsKey("X-GitHub-Request-Id");
             }
             catch {}
             finally
--- a/src/GitHub.Exports/Services/EnterpriseProbeTask.cs
+++ b/src/GitHub.Exports/Services/EnterpriseProbeTask.cs
@@ -37,12 +37,18 @@ namespace GitHub.Services
             };
             request.Headers.Add("User-Agent", productHeader.ToString());
 
+            var success = false;
             var ret = await httpClient
                     .Send(request, CancellationToken.None)
-                    .Catch(ex => null);
+                    .Catch(ex => {
+                        var apiex = ex as ApiException;
+                        if (apiex != null)
+                            success = apiex.HttpResponse?.Headers.ContainsKey("X-GitHub-Request-Id") ?? false;
+                        return null;
+                    });
 
             if (ret == null)
-                return EnterpriseProbeResult.Failed;
+                return success ? EnterpriseProbeResult.Ok : EnterpriseProbeResult.Failed;
             else if (ret.StatusCode == HttpStatusCode.OK)
                 return EnterpriseProbeResult.Ok;
             return EnterpriseProbeResult.NotFound;


src/GitHub.Api/SimpleApiClient.cs
```diff
@@ -71,10 +71,10 @@ public class SimpleApiClient : ISimpleApiClient
                 }
             }
             // it'll throw if it's private or an enterprise instance requiring authentication
-            catch (ApiException)
+            catch (ApiException apiex)
             {
                 if (!HostAddress.IsGitHubDotComUri(OriginalUrl.ToRepositoryUrl()))
-                    isEnterprise = true;
+                    isEnterprise = apiex.HttpResponse?.Headers.ContainsKey("X-GitHub-Request-Id");
```
Haacked: Ooooh. Could you check GitHub Desktop to see if we're doing the same thing there?

src/GitHub.Exports/Services/EnterpriseProbeTask.cs
```diff
@@ -37,12 +37,18 @@ public EnterpriseProbeTask(IProgram program, IHttpClient httpClient)
...
             request.Headers.Add("User-Agent", productHeader.ToString());
 
+            var success = false;
             var ret = await httpClient
                     .Send(request, CancellationToken.None)
-                    .Catch(ex => null);
+                    .Catch(ex => {
+                        var apiex = ex as ApiException;
+                        if (apiex != null)
+                            success = apiex.HttpResponse?.Headers.ContainsKey("X-GitHub-Request-Id") ?? false;
```
Haacked: I know this is the tiniest bit of duplicate around, but maybe we should put `X-GitHub-Request-Id` in a constant?

Haacked: Or even better, but this check in a method in a common location and call it in both places. Something like: `IsGitHubApiException(Exception exception);`

Haacked:
A couple minor suggestions but I'm also OK merging as-is and having the suggestions done later. Should I just ignore the AppVeyor build failures?

src/GitHub.Exports/Services/EnterpriseProbeTask.cs
shana: Yeah, this code needs better cleanup/consolidation, there's yet another IEnterpriseProbe implementation that doesn't have this check (and isn't being used right now iirc, but it exists)

shana:
Yup, ignore build failures for appveyor, it requires scripts that this PR doesn't have. I agree on the clean up but can do that later with more time.

src/GitHub.Exports/Services/EnterpriseProbeTask.cs
Haacked: > Yeah, this code needs better cleanup/consolidation, there's yet another IEnterpriseProbe implementation that doesn't have this check (and isn't being used right now iirc, but it exists)

If you're busy, I can look at it too.

src/GitHub.Api/SimpleApiClient.cs
shana: I don't think so, desktop assumes that any exception means it's not an enterprise server, so if you're not authenticated, it isn't.
I'm being a bit more forceful and assuming that an exception might mean it's an authenticated enterprise server, which is why this causing doing false positives.

src/GitHub.Exports/Services/EnterpriseProbeTask.cs
shana: I just wanted to keep changes to a minimum for this pre-vacation release, and save that up for afterwards.

shana: Actually, nevermind, the other EnterpriseProbe is a subclass of this one so the code is not duplicated there.

Haacked: > I just wanted to keep changes to a minimum for this pre-vacation release, and save that up for afterwards.

:cool: Yeah, I think I'll do it as another PR after this release I mean. For funsies.

Haacked:
:sparkles: Nice work!

PR #166, opened 2015-12-18
Script going for doing public builds on AppVeyor

shana:
Getting builds going on appveyor (not fully usable builds, but good enough to run tests)


shana:
This is enough to build from the command line for whoever wants to do that without the `script` submodule, and also to support AppVeyor builds. It's not fully working on AppVeyor, but at least it builds things correctly on the command line, I'll need to struggle with AppVeyor some more to figure out what else it needs.

Haacked:
I can help out too. I have a bit of AppVeyor experience myself.

PR #168, opened 2015-12-18
Fix repository list not refreshing

shana:
This patches up refreshing the repository list so #6 gets fixed. It's not a proper fix (that would be actually using TrackingCollection to load the repo list), but it works until #167 is done

### DIFF
--- a/src/GitHub.App/Services/ModelService.cs
+++ b/src/GitHub.App/Services/ModelService.cs
@@ -98,8 +98,7 @@ namespace GitHub.Services
         {
             return hostCache.GetAndRefreshObject("user",
                 () => apiClient.GetUser().Select(AccountCacheItem.Create), TimeSpan.FromMinutes(5), TimeSpan.FromDays(7))
-                .PublishLast()
-                .RefCount()
+                .TakeLast(1)
                 .ToList();
         }
 
@@ -126,12 +125,8 @@ namespace GitHub.Services
         public IObservable<IReadOnlyList<IRepositoryModel>> GetRepositories()
         {
             return GetUserRepositories(RepositoryType.Owner)
-                .PublishLast()
-                .RefCount()
-                .Concat(GetUserRepositories(RepositoryType.Member)
-                    .PublishLast()
-                    .RefCount()
-                )
+                .TakeLast(1)
+                .Concat(GetUserRepositories(RepositoryType.Member).TakeLast(1))
                 .Concat(GetAllRepositoriesForAllOrganizations());
         }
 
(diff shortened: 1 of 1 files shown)


src/GitHub.App/Services/ModelService.cs
```diff
@@ -125,8 +126,12 @@ IObservable<IEnumerable<AccountCacheItem>> GetUserOrganizations()
         public IObservable<IReadOnlyList<IRepositoryModel>> GetRepositories()
         {
             return GetUserRepositories(RepositoryType.Owner)
-                .Take(1)
-                .Concat(GetUserRepositories(RepositoryType.Member).Take(1))
+                .PublishLast()
```
Haacked: I'm confused by this. Shouldn't we have the `Take(1)` before the `PublishLast` still?

shana: These methods only returns either 1 or 2 results (depending on whether they only return from cache or they hit the cache and then the server).

`Take(1)` implies that we only want the first result, so the code that refreshes the cache from the server never runs if there's anything in the cache to return, because it's all run on demand (and we never demand all the results back)

`PublishLast()` is an async `Last()` method, which is what we actually want (to return the most up to date result).

Haacked: > Take(1) implies that we only want the first result, so the code that refreshes the cache from the server never runs if there's anything in the cache to return, because it's all run on demand (and we never demand all the results back)

Ah! Ok. That's what you were seeing?



> PublishLast() is an async Last() method, which is what we actually want (to return the most up to date result).

`PublishLast` turns the observable into a connected observable and returns the last notification. And `RefCount` makes a connected observable behave like a normal one. That's why I'm confused by this. There's also a `LastAsync` method to return the last item asynchronously.

But I feel like I'm missing something here in my understanding. :smile:

shana: That's because I'm getting into the habit of returning connectable subscriptions that auto-dispose themselves when I'm returning blobs of data that one or more callers can subscribe to, so that's where my brain went to on this one. I'm replacing all of these methods with a single repository query on another branch, so I didn't put a lot of thought into it. `PublishLast` is probably overkill here. We could use `LastAsync` or we could do `TakeLast` here as well.

Haacked: Ah. I see. Yeah, seems like `TakeLast(1)` would be the right approach here.

shana: /me fixicates

Haacked:
I tested it and it works like a charm! I wonder if we could even decrease the cache interval more. However, in the interest of being careful, I'm fine with it as is.

Haacked:
:sparkles:

PR #170, opened 2015-12-18
Cleanup enterprise host detection code

shana:
@Haacked I ended up cleaning the code up really quick. Feel free to play with it if you want! The extension class is in GitHub.Exports and not in GitHub.Extensions because the latter doesn't reference Octokit. No particular reason why it doesn't, it just never needed to before. Maybe other extensions should be moved there as well, or do we want to keep GitHub.Extensions a library that only extends .NET base classes and not other things?


Haacked:
> GitHub.Extensions a library that only extends .NET base classes and not other things?

Yeah, for now. I like what you did here for now. :smile:

What we should _really_ do is add this as an extension method in Octokit proper!

Issue #106, opened 2015-09-25
Link from VS to Github in Browser


avanderhoorn:
**Request**
I would love for a feature to be added where I could right click on a given line and either open the browser up at that line for the given branch on Github and/or copy the link to my clipboard. 

**Background**
In a given day I've lost count of how often I am talking to someone remotely (via twitter/skype/slack/etc) and I'm wanting to reference a line of code and the easiest way to do that is get the file up on Github... from that decision to the point where I find the relevant file/line and get the link is a good 30-60sec of time and like at east 5-10 clicks and a major context switch.

**Current Workflow**
I have to:
 1. switch to the browser
 1. open a new tab
 1. type in the url for the repo
 1. switch to the right branch
 1. hit `t`, type in the name of the class
 1. hit enter (if its the first result)
 1. then scroll through a file which could have 200-500 lines of code for the specific line I'm referencing
 1. click the line that I'm interested in 
 1. copy the link
 1. finally I get to continue on

**Logic**
Github for VS already knows if I have a git repo, if that repo origin is github and what branch I'm on. From there, it should be reasonable to think that a URL for that exact spot should be able to be generated.

**Edge Cases**
 - If the github repo/branch doesn't have the same version of the file that I do or if the file/branch doesn't exist at all
     - To start with I would expect that this feature would just generate the link without doing any verification... For my cases, this is almost never the case and when this does happen I'm usually know whether I've pushed the changes up or not
     - In the future, you could do a diff between what the known state of the origin is locally and see if that is going to cause a conflict or affect the link you would generate... this still isn't perfect but percentage wise deal with most of the edge cases
     -  In the future future, you could do a check against what the actual state for the remote repo is... In practice this isn't likely to work all that well due to the latency of the check

I wouldn't let the above stop the feature, because as I mentioned, I usually am aware of the state of changes I've made and even if its not exact, usually getting me within range of the correct line saves me a ton of work.

**Other IDEs and Text Editors**
This isn't specific to VS, the same feature makes sense in @atom, VS Code, etc, etc

dalpert-korewireless:
I'd love this!

matthewkimber:
That would be awesome!

nosami:
Xamarin Studio has this 
<img width="395" alt="screen shot 2015-09-25 at 21 59 58" src="https://cloud.githubusercontent.com/assets/667194/10112144/d51d7a9a-63d0-11e5-9a05-57530e974eb2.png">

[...]

Issue #39, opened 2015-08-06
Create gist from Visual Studio


tkirill:
Is it possible to create gist from selected lines with this extension?

Haacked:
Not today, but it wouldn't be hard to add.

Haacked:
How often do you think you'd use that?

vorou:
+1
It's not mission-critical, but I'm sure I feel sad at least twice a week because I have to open the browser and open the site and copy-paste stuff and then copy link. Oh god, it's even more painful now when I described that.

Haacked:
I think we have some code for doing this laying around even. We had implemented this in GitHub for Windows and then removed it because it was so undiscoverable. We relied on a keyboard shortcut.

Would would be nice is if we could inject ourselves in the context menu when you select a snippet of text in the Visual Studio editor. I'm not sure if that's possible.

Alternatively, we could add a menu to _Create Gist from Selection_. We should think through the user experience before setting to work on it.

tkirill:
> How often do you think you'd use that?

I usually need it twice a week.

> We should think through the user experience before setting to work on it.

Both options are fine for me as long as I can reach them with keyboard.  Context menu may be more intuitive though.

shana:
### Notes for help-wanted label

This is a nice small feature that could be very interesting for someone wanting to contribute to the project! If you'd like to tackle this (hurrah!), here are the steps that should go into making it happen:

- [ ] Define the User Experience for the feature
  - How would users go about using it? Context menus? Keyboard shortcuts? Action tags? What is the discoverability path for it?
  - Mockup the flow of the experience - what happens from the point where a user wants to create a gist to the point where the action is completed. Does it popup a window? Show a notification? A dialog? What options would there be for creating the gist, if any? Is it a one-click thing or a more complex flow? Does the feature need user settings or configurations?
- [ ] Implement!
  - Fork and create a branch for the feature, and create a PR for the branch. You build the PR over time as you implement things, and practical implementation discussions happen there!
- [ ] :ship:

ChristopherHackett:
Thinking about the process to create a gist from code in the editor 
```
Select code
Ctrl + C
Ctrl + Tab (until browser)
Ctrl + T 
Start typing 'gist' and hit return when http://gist.github.com appears 
Name gist and hit tab
Ctrl + V
Review code doesn't contain sensitive information 
Mouse to 'Create secret gist' or 'Create public gist' 
Select URL
Ctrl + C
[...]

PR #183, opened 2016-01-10
Context menus

georgebearden:
Hi @shana, 

I tested the #181 work you put in a couple days ago and still had a few issues with custom menus showing up on (hopefully) all file types.  I was seeing custom context menu items on .xaml files now, but still not on .html or .json files.  Using the trick you showed me here http://blogs.msdn.com/b/dr._ex/archive/2007/04/17/using-enablevsiplogging-to-identify-menus-and-commands-with-vs-2005-sp1.aspx, I came up with a different set of Guid/CmdIds.<br/>  This is the debug popup for .html files.
html-extension-vsdebug-message
And here is the popup for .json files
json-extension-vsdebug-message

After updating the .vsct file accordingly, our custom context menu items were showing up correctly.  So far I have tested and can see the custom menus in the following file types: `.cs` `.xaml` `.config` `.json` `.js` `.html` `.h` `.c` `.cpp` `.gitignore` `.db` `.vb` `.manifest` `.tt` `.xml` `.xsd` `.xslt` `.yml`

I really wish there was an *add command to all editor/file types* because this list is still not exhaustive when considering that any file that exists in a github repo should get the features from #106 and #39.

@Austin94 This PR should help with your work on #106 also.


shana:
Thanks for the investigative work! My GUIDs came from another source (because the VS I was testing on didn't have the web sdk installed), so I'm not surprised they're different, the new language designers probably caused changes to that. I'm merging in this in #186 because this PR is a bit too noisy for what it needs to do (and the suppress warning is already in master).

PR #187, opened 2016-01-11
New menu handling system

shana:
We're rapidly accumulating menu handlers for the gist and copy link support, and our current code is less than ideal. This PR turns our current menu handling code into MEF components, so we can easily add as many handlers as we need.

There should be no behaviour change, everything should work exactly the same way.

### DIFF
--- a/src/GitHub.VisualStudio/Menus/MenuProvider.cs
+++ b/src/GitHub.VisualStudio/Menus/MenuProvider.cs
@@ -1,0 +1,16 @@
+﻿using System.Collections.Generic;
+using System.ComponentModel.Composition;
+
+namespace GitHub.VisualStudio.Menus
+{
+    [Export(typeof(IMenuProvider))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class MenuProvider : IMenuProvider
+    {
+        [ImportMany]
+        public IEnumerable<IMenuHandler> Menus { get; set; }
+
+        [ImportMany]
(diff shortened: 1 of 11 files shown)


src/GitHub.VisualStudio/Menus/MenuProvider.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.ComponentModel.Composition;
+
+namespace GitHub.VisualStudio.Menus
+{
+    [Export(typeof(IMenuProvider))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class MenuProvider : IMenuProvider
+    {
+        [ImportMany]
+        public IEnumerable<IMenuHandler> Menus { get; set; }
```
Haacked: I'm a little wary of making these settable properties. Could we change the interface to make these read only properties and use ctor injection here like so?

```csharp
[ImportingConstructor]
public MenuProvider([ImportMany]IEnumerable<IMenuHandler> menus, [ImportMany]IEnumerable<IDynamicMenuHandler> dynamicMenus)
{
    // ...
}
```

I believe that better represents our intentions here.

Also, I'm guessing we should expose these properties as `IReadOnlyCollection<T>` rather than `IEnumerable<T>` since these aren't lazy enumerables and we want to avoid any possibility of multiple enumerations.

Haacked:
This makes me happy! :smile:

shana:
It makes total sense, and I remember meaning to make those readonly once I was done with everything else and haha famous last words :yum:

Haacked:
:sparkles:

PR #188, opened 2016-01-12
Fix crash after publishing new repository

Lathejockey81:
Visual Studio crashed after publishing, but the publishing process completed before the crash. The completion.OnNext() subscriber changes the active UI, which disposes the UIController. The subsequent completion?.OnCompleted() call threw an exception because completion is not null, but it is disposed. Since the OnNext subscriber is not evaluating the parameter, I changed the subscriber to respond to the OnCompleted() event instead. This appears to fix the issue.


shana:
Yay race conditions :cry:

shana:
I'm going to fix this in a separate PR, as it requires a bit more logic to make sure it doesn't fail. Thanks for the help!

PR #189, opened 2016-01-12
Fix connections not being passed on to view models

shana:
Fix a bug where if we're logged in but don't pass a connection to Start, no connection object would be passed along to whatever viewmodel we're loading, which is a Bad Thing(tm).

If we're logged in but no connection object is passed in, we should retrieve the first available logged-in connection from the connections list and pass that one in, so that viewmodels *always* get a connection.

### DIFF
--- a/src/GitHub.App/Controllers/UIController.cs
+++ b/src/GitHub.App/Controllers/UIController.cs
@@ -18,6 +18,7 @@ using NullGuard;
 using ReactiveUI;
 using Stateless;
 using System.Collections.Specialized;
+using System.Linq;
 
 namespace GitHub.Controllers
 {
@@ -226,7 +227,11 @@ namespace GitHub.Controllers
         {
             if (connection != null)
             {
-                uiProvider.AddService(typeof(IConnection), connection);
+                if (currentFlow != UIControllerFlow.Authentication)
+                    uiProvider.AddService(connection);
+                else // sanity check: it makes zero sense to pass a connection in when calling the auth flow
+                    Debug.Assert(false, "Calling the auth flow with a connection makes no sense!");
+
                 connection.Login()
                     .Select(c => hosts.LookupHost(connection.HostAddress))
                     .Do(host =>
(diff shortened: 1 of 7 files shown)


src/GitHub.App/Controllers/UIController.cs
```diff
@@ -253,13 +258,21 @@ public void Start([AllowNull] IConnection connection)
...
-                        if (!loggedin && currentFlow != UIControllerFlow.Authentication)
+                        if (currentFlow != UIControllerFlow.Authentication)
                         {
-                            connectionAdded = (s, e) => {
-                                if (e.Action == NotifyCollectionChangedAction.Add)
-                                    uiProvider.AddService(typeof(IConnection), e.NewItems[0]);
-                            };
-                            connectionManager.Connections.CollectionChanged += connectionAdded;
+                            if (loggedin) // register the first available connection so the viewmodel can use it
+                                uiProvider.AddService(connectionManager.Connections.First(c => hosts.LookupHost(c.HostAddress).IsLoggedIn));
```
Haacked: Is there any possibility that there is no connection here and thus `First` would throw an exception?

Haacked: It should never happen, but could there be some sort of race condition or something here? Should we use `FirstOrDefault` and check for null to be defensive?

shana: The `IsLoggedIn(hosts)` call that gives us the `loggedin` variable goes through all the `connectionManager.Connections` and attempts to log them in. If it returns true, it's because there's at least one connection on the list and that connection is logged in (it uses the same exact check I'm using here), so we can safely grab it.

I can do some bigger changes and instead of returning bool for the `Do()` block, I return the connection itself or null...

Haacked: Thanks for the context!

Haacked:
:sparkles:

PR #190, opened 2016-01-12
Fix a delay in loading the connect page

shana:
Querying the registry for repositories so we can load the repository list in the Connect page is visibly delaying the UI, which is a no-no! Fixicate.

### DIFF
--- a/src/GitHub.Exports/Services/VSServices.cs
+++ b/src/GitHub.Exports/Services/VSServices.cs
@@ -124,10 +124,9 @@ namespace GitHub.Services
                             if (path != null)
                                 return new SimpleRepositoryModel(path);
                         }
-                        // no sense spamming the log, the registry might have ton of stale things we don't care about
                         catch (Exception)
                         {
-                            //VsOutputLogger.WriteLine(string.Format(CultureInfo.CurrentCulture, "Error loading the repository from the registry '{0}'", ex));
+                            // no sense spamming the log, the registry might have ton of stale things we don't care about
                         }
                         return null;
                     }
--- a/src/GitHub.Exports/Services/VSServices.cs
+++ b/src/GitHub.Exports/Services/VSServices.cs
@@ -124,9 +124,10 @@ namespace GitHub.Services
                             if (path != null)
                                 return new SimpleRepositoryModel(path);
                         }
-                        catch (Exception ex)
+                        // no sense spamming the log, the registry might have ton of stale things we don't care about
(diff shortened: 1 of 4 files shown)


src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -124,9 +124,10 @@ static IEnumerable<ISimpleRepositoryModel> PokeTheRegistryForRepositoryList()
                             if (path != null)
                                 return new SimpleRepositoryModel(path);
                         }
-                        catch (Exception ex)
+                        // no sense spamming the log, the registry might have ton of stale things we don't care about
+                        catch (Exception)
                         {
-                            VsOutputLogger.WriteLine(string.Format(CultureInfo.CurrentCulture, "Error loading the repository from the registry '{0}'", ex));
+                            //VsOutputLogger.WriteLine(string.Format(CultureInfo.CurrentCulture, "Error loading the repository from the registry '{0}'", ex));
```
Haacked: If we don't think we need to log here, let's just remove the commented out code and replace it with the comment.

src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs
```diff
@@ -377,7 +378,7 @@ public SectionStateTracker(ITeamExplorerSection section, Action onRefreshed)
                     .Ignore("Title")
                     .PermitIf("IsBusy", SectionState.Idle, () => !this.section.IsBusy)
                     .IgnoreIf("IsBusy", () => this.section.IsBusy)
-                    .OnExit(onRefreshed);
+                    .OnExit(() => onRefreshed());
```
Haacked: Does this need a `Forget` call here as in `onRefreshed().Forget()`?

Haacked:
:pineapple:

src/GitHub.VisualStudio/TeamExplorer/Connect/GitHubConnectSection.cs
Haacked: Nevermind. I'm being dumb.

shana: Nope, the compiler doesn't recognize this as being an potentially awaited usage so it doesn't complain.

shana:
:football: back to you!

Haacked:
:sparkles:

PR #191, opened 2016-01-12
Feature: Link from code to GitHub via context menus

### LINKED ISSUE #106: Link from VS to Github in Browser

shana:
Fixes #106 

This needs:

- [x] PR #187 - New menu handling system in master
- [x] PR #220 - Status bar notification service
- [x] PR #221 - Octicons update
- [x] Notification when link is copied on the status bar
- [ ] Code review


shana:
This is how it looks like currently.

@donokuda @Haacked Should we keep these context menu entries like this (considering we have potentially more context menus to add), or should we have a GitHub context menu entry with these as submenus?

copylink

shana:
@Haacked This is ready for :eyes:. A few things in here are part of PR #220 so might want to deal with that one first.

Haacked:
People gonna be upset if we add too much to the context menu. If there isn't already a grouping that makes sense, we should group our commands in a top level "GitHub" menu.

donokuda:
> People gonna be upset if we add too much to the context menu. If there isn't already a grouping that makes sense, we should group our commands in a top level "GitHub" menu.

:+1: I can see us having to do this anyway as we integrate more features (such as creating a gist.)

src/GitHub.Exports/Extensions/SimpleRepositoryModelExtensions.cs
```diff
@@ -30,5 +35,58 @@ public static bool MightContainSolution(this ISimpleRepositoryModel repository)
                       .Any(x => ((x.Attributes.HasFlag(FileAttributes.Directory) || x.Attributes.HasFlag(FileAttributes.Normal)) &&
                                 !x.Name.StartsWith(".", StringComparison.Ordinal) && !x.Name.StartsWith("readme", StringComparison.OrdinalIgnoreCase)));
         }
+
+        public static string CurrentSha(this ISimpleRepositoryModel repository)
+        {
+            var repo = GitService.GitServiceHelper.GetRepo(repository.LocalPath);
+            return repo.Commits.FirstOrDefault()?.Sha;
```
Haacked: Could `repo` be null here? Should this be `repo?.Commits.FirstOrDefault()?.Sha`?

src/GitHub.Exports/Extensions/SimpleRepositoryModelExtensions.cs
```diff
@@ -30,5 +35,58 @@ public static bool MightContainSolution(this ISimpleRepositoryModel repository)
...
+        }
+
+        public static Uri GenerateUrl(this ISimpleRepositoryModel repository, string path = null,
+            int startLine = -1, int endLine = -1)
```
Haacked: I'd like to avoid optional arguments in public APIs because of potential ambiguities and versioning issues: http://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/

src/GitHub.Exports/Extensions/SimpleRepositoryModelExtensions.cs
```diff
@@ -30,5 +35,58 @@ public static bool MightContainSolution(this ISimpleRepositoryModel repository)
...
+                    Debug.Assert(false, String.Format(CultureInfo.CurrentCulture, "GenerateUrl: path {0} doesn't match repository {1}", path, repository.LocalPath));
+                    path = null;
+                }
+                else
+                    path = path.Substring(repository.LocalPath.Length + 1);
+            }
+
+            var commitSha = repository.CurrentSha();
+            var uri = GenerateUrl(cloneUrl.ToRepositoryUrl().AbsoluteUri, commitSha, path, startLine, endLine);
+            return new UriString(uri).ToUri();
```
Haacked: On line 66 you generate a URL using `cloneUrl.ToRepositoryUrl().AbsoluteUri` so you know this is going to be an `http` or `https` URL (that's what `ToRepositoryUrl` guarantees. So you _could_ simplify the previous two lines to...

```csharp
var uri = GenerateUrl(cloneUrl.ToRepositoryUrl().AbsoluteUri, commitSha, path, startLine, endLine);
return new Uri(uri);
```

Alternatively, because we have an implicit conversion, you could also simplify it to...

```csharp
UriString uri = GenerateUrl(cloneUrl.ToRepositoryUrl().AbsoluteUri, commitSha, path, startLine, endLine);
return new uri.ToUri();
```

I leave it to you. There's nothing explicitly wrong with what you have here. :smile:

src/GitHub.Exports/Extensions/SimpleRepositoryModelExtensions.cs
```diff
@@ -30,5 +35,58 @@ public static bool MightContainSolution(this ISimpleRepositoryModel repository)
...
+        }
+
+        static string GenerateUrl(string basePath, string sha, string path, int startLine = -1, int endLine = -1)
+        {
+            var sb = new StringBuilder(basePath);
```
Haacked: For small number of string concatenations, I usually don't choose to use `StringBuilder` because the overhead of it isn't paid off until you get to a certain number of concatenations.

Given this is not performance critical code (it's not called often and not in a tight loop) I think it probably doesn't matter either way.

Jon Skeet has a great blog post about the subject. http://jonskeet.uk/csharp/stringbuilder.html

src/GitHub.VisualStudio/Helpers/ActiveDocumentSnapshot.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.VisualStudio.TextManager.Interop;
+using System;
+using System.ComponentModel.Composition;
+using System.Diagnostics;
+
+namespace GitHub.VisualStudio
+{
+    [Export(typeof(IActiveDocumentSnapshot))]
+    [PartCreationPolicy(CreationPolicy.NonShared)]
+    class ActiveDocumentSnapshot : IActiveDocumentSnapshot
```
Haacked: Can we export private classes?

src/GitHub.VisualStudio/Menus/CopyLink.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public void Activate()
+        {
+            if (!IsGitHubRepo())
+                return;
+
+            var link = GenerateLink();
+            if (link == null)
+                return;
+            Clipboard.SetText(link.AbsoluteUri);
```
Haacked: The documentation suggests `Clipboard.SetText` can only throw an `ArgumentNullException`. We've learned the hard way that's not true. For example, the [.NET source code](http://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/Clipboard.cs,139) shows it could throw a `SecurityException`.

Also, this interacts with COM and what happens in COM land, doesn't  stay in COM land but craps potential exceptions all over your best laid plans.

So we should probably wrap this in a `try/catch` so it doesn't crash the extension if adding to the Clipboard fails. In GitHub Desktop, we retry a couple times in case it's some weird intermittent failure. We don't have good data to know if it's possible that a retry would succeed though. So I'd just wrap it in a try/catch, write out a message, and the user can try again.

Haacked:
:fire_engine:

src/GitHub.VisualStudio/Menus/CopyLink.cs
shana: The `System.Windows` Clipboard behaves differently from the `System.Windows.Forms` one, but yeah, it can still throw completely random exceptions whenever it feels like it. I'll follow your suggestion and print out a message if it fails.

shana:
I've moved things to a submenu and cleaned up. The icons look like crap though, even though they're xaml icons. Feels like the xaml versions of the octicons that we have have strokes that are too wide?
screen shot 2016-02-09 at 2 04 53 pm
screen shot 2016-02-09 at 2 05 09 pm

Haacked:
I think we need an :octocat: icon for the top level GitHub menu item. What do you think @donokuda and @shana?

shana:
@haacked Would love to, but unfortunately it seems VS does not support icons on menu entries that lead to submenus :cry:

Haacked:
> @haacked Would love to, but unfortunately it seems VS does not support icons on menu entries that lead to submenus

:cry: /cc @acangialosi feature request!

src/GitHub.VisualStudio/Helpers/ActiveDocumentSnapshot.cs
Haacked: *bump* I'm surprised this works. Maybe I shouldn't be?

Haacked:
@shana I just have the one question about the private class that's exported.

src/GitHub.VisualStudio/Helpers/ActiveDocumentSnapshot.cs
shana: It works via reflection, so it doesn't actually care about the visibility of the implementation. But it should be public, yes, for consistency.

shana:
@Haacked I've gone and :hammer: all the things you mentioned. I still need to fix the icons so they look half decent, so don't merge this quite yet.

Haacked:
_Slowly takes finger off the big green button_

niik:
@shana A+

shana:
@Haacked 

Fixed the icon sizes. Also fixed the colors to match the colors of other context menus in VS. In the process, added a little extra code to support theming, so now we have three separate theme xaml files that define colors for the three themes VS comes with - Dark, Light and Blue.

Any xaml file that includes a SharedDictionaryManager resource dictionary with the uri "Theme.xaml" will get a themed dictionary included instead that tracks the VS theme and gets switched at runtime whenever the VS theme changes. If VS is running an unknown theme, the code detects whether it's a dark or light theme (based on brightness) and loads our dark or light theme file to match.

# Test run

green for dark theme, blue for blue theme, red for light theme
runtime-theme-change

# Dark
screen shot 2016-02-19 at 11 09 46 pm

# Light
screen shot 2016-02-19 at 11 11 18 pm

# Blue
screen shot 2016-02-19 at 11 11 04 pm

donokuda:
> Any xaml file that includes a SharedDictionaryManager resource dictionary with the uri "Theme.xaml" will get a themed dictionary included instead that tracks the VS theme and gets switched at runtime whenever the VS theme changes. If VS is running an unknown theme, the code detects whether it's a dark or light theme (based on brightness) and loads our dark or light theme file to match.

tumblr_nozh54mtna1s31h01o1_400

> green for dark theme, blue for blue theme, red for light theme

Are these colors just for demonstrating how theming works?

shana:
> Are these colors just for demonstrating how theming works?

@donokuda Yeah, I couldn't actually tell the difference. I'm actually using the same color for those icons on all three themes, and VS is apparently inverting the color on the dark theme when rendering the icons (got me so confused initially when I set it to white and it showed up black and vice-versa...), so I set them to RGB to make sure it was working (and recorded proof for future self)

donokuda:
:+1: That's what I thought. Love this!

Haacked:
This feature feels like magic!

Atcold:
How does one activate this?

PR #192, opened 2016-01-12
Fix race condition causing crash in Publish form

shana:
See #188


Haacked:
:sparkles: Nice catch!

PR #193, opened 2016-01-12
Fix login dialog not going away after it's done

### LINKED ISSUE #176: Stuck at Two-Step Authentication
After updating the extension to the latest, it gets stuck after I give the token. Tried in 4 pcs, happens in all of them.

image

shana:
Fixes #176


Haacked:
Wow. That's a simple fix!

shana:
This one is a total

PR #194, opened 2016-01-13
Feature: Gist support

shana:
Pulling in #156 to a staging branch

This needs:

- [x] #187 - New menu handling system in master
- [x] Integration with the new menu handling system
- [x] Fixing tests

This adds support for creating gists from selected text. It pops up a dialog to add filename and description to the gist (see https://github.com/github/VisualStudio/pull/156#issuecomment-170786587)

Right now, popping up a dialog is easier because it integrates straight into our authentication flow, but I'm wondering if in the long term we want this on a side pane instead of a dialog, so it doesn't obscure the text and integrates more naturally with VS. Thoughts?


georgebearden:
Hi @shana, 

Nice work on cleaning up everything! I could knock out #196 and #197 really quickly before testing for the next release starts if you think it is a good idea. I think both issues would add a lot to the feature for a little amount of work as they are both pretty localized.

Let me know and I can submit a PR against this branch for them ASAP. Thanks!

shana:
#196 is dependent on #201 so no point digging into that one. #197 would be nice to have! If you do a PR for it, target the feature/gist-support branch

shana:
We have a problem with this: the existing scopes don't include gist, which means that we'll need to invalidate/create a new authorization to have the added scope, or update the existing scope.

This will likely mean we need to force the users to relogin, because the authorizations API require the credentials. @Haacked, any ideas on how best to handle this?

georgebearden:
Hi @shana. Is there anything I can look into related to the new `gist` scope to help this along?

shana:
@georgebearden We've been talking about it, trying to think of a good flow for re-authorizing the user to get the new scope. Probably the most straightforward thing is to show a message saying you have to logout and in and again when we show the gist dialog, disabling the button and triggering the auth dialog from there (maybe by logging the user out and adding some extra steps to the UIController so that we can move to the login dialog and then back to the gist). Got any ideas?

I've been sick for the past few weeks so nothing has been advanced much on this, unfortunately. :/

georgebearden:
So I have been thinking about this for a while now, and I think I am missing something. When I download and install the GHfVS extension as a user, all I do is enter my username and password, and I never have to approve any oauth scopes through a redirect url or anything like that. On the other hand, as a developer on the GHfVS extension, I needed to create a developer application where I added the gist scope to the authorization callback url to get gists working in the extension. So what I am wondering is why do we need the users to re-login here, when the new gist scope is transparent to them anyway? Is it not enough to modify the "production" version of the developer application to add the gist support in the auth callback url?

Any clarification you can provide so I can understand this better is appreciated!

shana:
@georgebearden The login process takes your user and password and requests a token for use by the application for you. That token as a scope attached to it, so the application can't do more than what it initially requested . The requested scope isn't shown explicitly in the non-web oauth flow, but it's still there. To change the scope of the application, we either update the scope of the existing token, or we get a new token with the updated scope. The API for both these actions requires the original username and password, for security purposes. Since we don't store the password (only the token), the user has to login again so we can update the scope.

georgebearden:
Thank you for the clarification, it all makes sense now :smile: 

I agree that a simple solution would be to add a message to the gist dialog to re-login to get this working. The only reservation I am having is it seems like any time we need to add another scope to the existing token as the extension gets more features we would need to support more of these one-off login cases, not that there would be a ton off them being added or anything...

Another solution would be to just force the user to login back in after updating the extension. I feel like this is something that I have experienced before after updating applications, and did not mind too much. I understand though that not everyone would be using the gist feature so not everyone would actually **need** to log back in, but it does not seem too bad. If we did this, we could have a single place in the code base that checks to see if this is the first time being run, and if it is just force the login after the update. Then we could re-use this code anytime we need to force login in the future.

shana:
> Another solution would be to just force the user to login back in after updating the extension. I feel like this is something that I have experienced before after updating applications, and did not mind too much. I understand though that not everyone would be using the gist feature so not everyone would actually need to log back in, but it does not seem too bad

Yes, that's one of the alternatives we thought about. The problem is, of course, detecting that the extension has been updated and that it needs to re-auth. Also, how to tell the user to re-login - in normal apps, you own the UI so it's easy to inform the user, but here we don't really have a good place to do this. There's a few solutions I can think of, but I really dislike kicking the user out automatically because of an scope change that they might or might not ever use, so solutions need to be in the form of a message somewhere. And if they don't re-login, we need to detect that the scope is wrong if they do use the gist feature, and re-login them there - so in reality, we always have to do it specifically for the gist feature.

grokys:
Hi @georgebearden, I've just been added to the team working on this plugin to help @shana with the backlog, and getting this PR merged is going to be one of my first jobs. Apologies for the delay, but hopefully everything should be merged in the next week or so!

georgebearden:
No problem at all! I have been super busy with work related things over the past couple months, but it would be great to get the gist work into the plugin!. Let me know if you need anything from me or clarification on anything done so far.

Thanks 😄

Haacked:
selfie-0
[_GitHub Selfies_](https://github.com/thieman/github-selfies/)

shana:
Yes this totally deserves a ton of gifs!

@georgebearden Thank you SO MUCH for taking this on, and I'm so sorry it took so long to land this! I'll be putting up an alpha build of what is going to be 2.0 (or 1.1, haven't decided yet) in the releases page which includes the gist support.

thank-you

thumbs-up

rejoice

shana:
Note: Issue is #39

georgebearden:
yay! Gonna have to tackle something a little smaller next time :smile: Great work @shana and @grokys  on finishing this up and excited to start using it!

shana:
It would have been small id it wasn't for that silly little token scope detail... that made it really bad!

PR #198, opened 2016-01-14
[wip] :art: Cleanup GetService and TryGetService some more.

shana:
Cleanup some redundant and duplicated code. `UIProvider` should only use non-generic variants of GetService/TryGetService so that calls don't spill over to `ServiceProviderExtensions.cs` and potentially cause recursive calls. Everyone else goes through the extension methods.

### DIFF
--- a/src/GitHub.Exports/Extensions/ServiceProviderExtensions.cs
+++ b/src/GitHub.Exports/Extensions/ServiceProviderExtensions.cs
@@ -8,6 +8,7 @@ using Microsoft.TeamFoundation.Controls;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
+using System.Diagnostics.CodeAnalysis;
 
 namespace GitHub.Extensions
 {
@@ -15,6 +16,11 @@ namespace GitHub.Extensions
     {
         static IUIProvider cachedUIProvider = null;
 
+        /// <summary>
+        /// Safe variant of GetService that doesn't throw exceptions if the service is
+        /// not found.
+        /// </summary>
+        /// <returns>The service, or null if not found</returns>
         public static object TryGetService(this IServiceProvider serviceProvider, Type type)
         {
             if (cachedUIProvider != null && type == typeof(IUIProvider))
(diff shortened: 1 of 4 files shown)


src/GitHub.Exports/Extensions/ServiceProviderExtensions.cs
```diff
@@ -25,6 +31,20 @@ public static object TryGetService(this IServiceProvider serviceProvider, Type t
...
+        /// like <code>GetService(typeof(SVsTextManager) as IVsTextManager</code>
+        /// Does not throw exceptions
+        /// </summary>
+        [SuppressMessage("Microsoft.Design", "CA1004:GenericMethodsShouldProvideTypeParameter")]
+        public static Ret GetService<T, Ret>(this IServiceProvider provider)
+            where Ret : class
+            where T : class
+        {
+            return provider.TryGetService(typeof(T)) as Ret;
+        }
```
Haacked: Minor style nit. Needs a newline here. :smile:

src/GitHub.VisualStudio/Services/UIProvider.cs
```diff
@@ -35,14 +34,10 @@ public class UIProvider : IUIProvider, IDisposable
         bool initializingLogging = false;
 
         [AllowNull]
-        public ExportProvider ExportProvider { get; private set; }
+        public ExportProvider ExportProvider { get; }
```
Haacked: C# 6 in da house!

src/GitHub.VisualStudio/Services/UIProvider.cs
```diff
@@ -85,7 +80,7 @@ public object TryGetService(Type serviceType)
                 initializingLogging = true;
                 try
                 {
-                    var logging = TryGetService<ILoggingConfiguration>();
+                    var logging = TryGetService(typeof(ILoggingConfiguration)) as ILoggingConfiguration;
```
Haacked: It wasn't obvious to me why you're making this change and not just calling the cleaner looking generic method. Then I read your PR comment.

> UIProvider should only use non-generic variants of GetService/TryGetService so that calls don't spill over to ServiceProviderExtensions.cs and potentially cause recursive calls.

That's a bit tricky and nuanced and people looking at this code in the future might be really tempted to call that generic method. One idea is to add a private generic `TryGetService` to this class that then calls the non-generic one for you and document that one method with the comment you made. That way nobody will be tempted to change this call in a breaking manner.

Thoughts?

Haacked: @shana: ^^^ bump.

shana: Oh darn, how did I miss this? On it!

PR #201, opened 2016-01-14
Notifications service

shana:
Notifications service for dispatching notifications which can be then shown in Team Explorer or the status bar or somewhere else. View models dispatch messages so they don't care where they're going to show up - that's for whoever invoked the view to decide.

Right now it just replaces the existing notification system so that the behavior doesn't change. Another PR will add status bar message support on top of this.

@Haacked This is ready for :eyes: 

Requires:

- [ ] Review
- [ ] Testing for regressions


src/GitHub.App/ViewModels/RepositoryPublishViewModel.cs
```diff
@@ -157,18 +150,14 @@ IObservable<ProgressState> OnPublishRepository(object arg)
             var account = SelectedAccount;
 
             return repositoryPublishService.PublishRepository(newRepository, account, SelectedHost.ApiClient)
-                .Select(_ =>
-                {
-                    vsServices.ShowMessage("Repository published successfully.");
-                    return ProgressState.Success;
-                })
+                .Select(_ => ProgressState.Success)
```
Haacked: This no longer shows the "Repository published successfully." message that used to happen in this spot via the old call to `vsServices.ShowMessage("Repository published successfully.");`. Was that intentional?

Haacked:
I :heart: this change! I think it's a good idea to abstract notifications away to give us more control. Nice!

Just one question for you on something I didn't understand.

src/GitHub.App/ViewModels/RepositoryPublishViewModel.cs
shana: Oh, yeah, good catch. That message needs to be shown by the ui view, or by the controller that shows the form in the first place, based on the ProgressState result. I missed adding it there I think, need to fixicate.

shana:
So there's two types of notifications that can come up via the viewmodels: warnings from validators while filling out the form, and errors from the process once the user hits submit. Errors are sent via the notification system, and warnings come up via the validators.

I'm doing it like this for now because warnings are reactive based on user input (so they go away if the validator becomes valid), and I'm still not sure if the NotificationDispatcher should have a
`ClearNotifications` method. The dispatcher is just dispatching messages, is it its job to invalidate messages too? Perhaps, unsure for now.

Haacked:
:sparkles:

PR #205, opened 2016-01-15
Only process clone/create when the dialog is closed

### LINKED ISSUE #203: Clone/create message is showing before cloning/creating actually runs
From https://github.com/github/VisualStudio/issues/157#issuecomment-161386161

When showing the cloning/create dialog, if the authentication dialog runs first and the user logs in, the code watching for new repositories registered in VS will kick in because the list of repositories is populated on login, and it assumes that cloning/creating is done so the cloning/creating message is shown before the user has actually cloned or created anything.

shana:
Fixes #203 

Requires:

- [ ] Testing


Haacked:
:thumbsup:

PR #206, opened 2016-01-18
Allow the Account property to return null.

georgebearden:
@shana, because we get the [account asynchronously](https://github.com/georgebearden/VisualStudio/blob/092fc5b6cb6adcc57a137bbb45bb10dba32f439b/src/GitHub.App/ViewModels/GistCreationViewModel.cs#L45-L49) when instantiating the `GistCreationViewModel`, it can be null when we try to Bind it to the Gist popup which causes a fody nullguard assertion error like this one
nullguard_assertion.

Allowing the `Account` property to return null fixes the issue and I verified that everything works as expected as eventually the async call returns with our account and we have our nice avatar and username in the view.


shana:
:sparkles:

PR #209, opened 2016-01-21
Fix race condition when finishing ui flows

### LINKED ISSUE #203: Clone/create message is showing before cloning/creating actually runs

shana:
Let's decide that the `IUIController.ListenToCompletionState` observable finishes first, and then the `IUIController.SelectFlow` observable. Once the `SelectFlow` observable is completed, it's a fair bet that the `IUIController` is disposed and it isn't safe to access it for anything else.

Sanitize calls to `SelectFlow` and `ListenToCompletionState` so that UI navigation (which triggers `Dispose` to be called on everything) will only happen on completion of the observable returned by `SelectFlow`, as it is the last observable to be completed and things are safe to dispose after that.

Fixes #203 properly

### DIFF
--- a/src/GitHub.App/Controllers/UIController.cs
+++ b/src/GitHub.App/Controllers/UIController.cs
@@ -156,7 +156,6 @@ namespace GitHub.Controllers
             completion?.OnNext(success);
             completion?.OnCompleted();
             transition.OnCompleted();
-            completion = null;
         }
 
         void RunView(UIViewType viewType)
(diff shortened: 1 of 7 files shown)


Haacked:
Nice!

Issue #197, opened 2016-01-14
Copy the file name when creating a gist


georgebearden:
To further refine the `Gist` work being done in #194 it would be nice to set the default file name of the gist being created to the name of the file the text is being selected out of.  Here is a working snippet that has been tested on a single-pane and multi-pane text editor.

``` csharp
var dte = GetService(typeof (DTE)) as DTE;
if (dte == null)
    return;

var fileNameWithPath = dte.ActiveDocument.FullName;
var fileName = Path.GetFileName(fileNameWithPath);
```

shana:
@georgebearden Just FYI, `Services.DTE` is available for getting at these interfaces.

georgebearden:
Hi @shana, I just went to implement this feature and ran into one issue. Trying to reference DTE or DTE2 from the GitHub.App project leads to the following compile-time error. 

`error CS1762: A reference was created to embedded interop assembly 'EnvDTE, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' because of an indirect reference to that assembly created by assembly 'GitHub.Exports, Version=1.0.16.3, Culture=neutral, PublicKeyToken=null'. Consider changing the 'Embed Interop Types' property on either assembly.`

I don't want to go changing assembly info without knowing what is going on, but here is the code snippet we would need to add to the `GistCreationViewModel`'s constructor:

```csharp
var fullName = VisualStudio.Services.Dte2?.ActiveDocument?.FullName;
FileName = fullName != null ? Path.GetFileName(fullName) : Resources.DefaultGistFileName;
```

Don't you hate it when you think some feature is going to be super easy  and then you run into these kinds of issues.  

One workaround (but prob. not MVVM compliant) would be to set the `FileName` like we do above, but instead of settings in the `GistCreationViewModel` which lives in the `GitHub.App` project  we can set it in the code-behind of the `GistCreationView` as it lives in the `GitHub.VisualStudio` project which already has references to `EnvDTE` and `EnvDTE80`.

shana:
Oh yes, this happens often when pulling in a direct reference to EnvDTE.

**Warning**: Wall of text.

EnvDTE includes embedded interop types, which means that the assembly is wrapping a native com type in a .net type and it provides the complete information of the native type (basically the size of it, that's what really counts), and allowing others to save this information when they use the .net wrappers. This allows a user assembly to use a .net type that maps into a com type even if the version of the assembly where that type comes is different at runtime.

[...]

PR #214, opened 2016-02-02
Copies the file name of the active document to the gist popup, #197

georgebearden:
Went ahead and added the static method to the Services class for now. Thanks for the suggestion, and welcome back!


shana:
This is great, thank you so much!

PR #217, opened 2016-02-03
A handy date to duration converter

shana:
This will be used in the pr features.


Haacked:
What's with the Resource designer files? Were they intended to be part of this PR?

shana:
@Haacked They're for future eventual localization of the strings of the conversion. "x days ago", "x minutes ago", etc

Haacked:
Ah, cool! I'm currently reviewing.

src/GitHub.UI/Converters/DurationToStringConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+            const int month = 30;
+            const int day = 24;
+            const int hour = 60;
+            const int minute = 60;
+
+            TimeSpan duration;
+            if (value is TimeSpan)
+                duration = (TimeSpan)value;
+            else if (value is DateTime)
+                duration = DateTime.Now - (DateTime)value;
```
Haacked: Where would we be getting `value` from. Is it possible it'd be stored as `UTC` and we should use `UtcNow` here? Or can we assume it's always local time?

src/GitHub.UI/Converters/DurationToStringConverter.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Globalization;
+
+namespace GitHub.UI
+{
+    public class DurationToStringConverter : ValueConverterMarkupExtension<DurationToStringConverter>
+    {
+        public override object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            const int year = 356;
```
Haacked: Did you mean 365? I assume it just needs to be close so we can ignore leap years. :smile:

Haacked: Should  this be 365?

Haacked:
:pineapple:

src/GitHub.UI/Converters/DurationToStringConverter.cs
shana: lol, how the heck did that happen and obviously I'm not testing hard enough!

src/GitHub.UI/Converters/DurationToStringConverter.cs
shana: Good question... data can come from any number of sources, most often from the github api, so it's going to be UTC. /me fixicates

shana:
:truck:

Haacked:
:sparkles:

PR #221, opened 2016-02-10
Update octicons

shana:
Update the octicons to v3.4.1

## Before
screen shot 2016-02-10 at 9 22 14 pm

screen shot 2016-02-10 at 9 22 24 pm

## After
screen shot 2016-02-10 at 9 15 46 pm

screen shot 2016-02-10 at 9 16 00 pm


Haacked:
Tried it out and it looks so good!

shana:
:sparkles:

PR #224, opened 2016-02-22
Design time theme colors

shana:
Depends on

- [x] #191

This PR fixes theme colors so that it's possible to use the xaml designer in VS to see how things look. 

Seems like the xaml designer completely borks if a ResourceDictionary implements the Source property (which is kind of mandatory for caching resource dictionaries and doing the runtime theme swapping thing), so I've added a separate build configuration with a define that ifdefs the Source property in our `SharedDictionaryManager.cs` resource dictionary implementation.

To see things at design time, switch to the `XamlDesigner` build configuration in VS and build the solution (it should only build `GitHub.VisualStudio` if you're coming from a debug build, so it should be fast to switch). After that, xaml files should render properly and be designable in the VS xaml designer. Builds with the `XamlDesigner` configuration will always use the dark theme, so they won't reflect the current theme at runtime (use Debug or Release builds to see theming at runtime).

The `Styles\ThemeDesignTime.xaml` file in `GitHub.VisualStudio` defines hardcoded values for the VS colors that we use and loads the dark theme - this xaml file is switched at runtime with the `ThemeXXX.xaml` file that corresponds to what VS is running. If you prefer designing in another theme, change the theme file that `ThemeDesignTime.xaml` is loading and change the vs colors to the appropriate theme.


shana:
This is ready for review, but for some reason the PR view (at least for me) is not getting updated to reflect the commits that went into master as part of PR #191, which this branch is based upon. I've reported that bug, and in the meantime if the PR diff is still showing too much unrelated info, use https://github.com/github/VisualStudio/compare/master...shana/theme-colors to review.

PR #227, opened 2016-02-24
Collect anonymous metrics

shana:
This PR adds a settings service and options page to handle user settings. Also adds one setting: anonymous usage data collection opt-out.


shana:
@Haacked I've added the options page to save the user settings, with a `CollectMetrics` property (grab the `IPackageSettings` export anywhere with either `ServiceProvider.GetExportedValue` or `Services.DefaultExportProvider.GetExportedValue` if there's no available `IServiceProvider` available in the local context)

Haacked:
Why don't we split this PR into two. Add the options whether we're collecting data or not. Then a separate PR for the metrics. Thoughts?

src/GitHub.VisualStudio/UI/Options/OptionsControl.xaml
```diff
@@ -0,0 +1,82 @@
...
+            Padding="6 8 6 8"
+            Style="{DynamicResource GroupBoxFlat}">
+
+    <GroupBox.Header>
+      <Run Text="{x:Static prop:Resources.Options_PrivacyTitle}" />
+    </GroupBox.Header>
+    <DockPanel>
+      <WrapPanel DockPanel.Dock="Top">
+        <CheckBox x:Name="chkMetrics" VerticalAlignment="Center" />
+        <Label Content="{x:Static prop:Resources.Options_MetricsLabel}" />
```
Haacked: We should probably add a hyperlink here that points to https://visualstudio.github.com/samples/. That page shows examples of the type of data we collect.

Here's the [Desktop example](https://desktop.github.com/samples/). Obviously we'll need to add that page to the Visual Studio website, but lets add the link now. :smile:

src/GitHub.Exports/Helpers/SettingsStore.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace GitHub.Helpers
+{
+    internal class SettingsStore
+    {
+        readonly WritableSettingsStore store;
+        readonly string root;
+        public SettingsStore(WritableSettingsStore store, string root)
+        {
+            this.store = store;
```
Haacked: I noticed we're not using `NullGuard` for this project so we should probably validate that these arguments are not null here. Otherwise, we could get a `NullReferenceException` on line 34.

src/GitHub.Exports/Services/PackageSettings.cs
```diff
@@ -0,0 +1,104 @@
...
+            if (!settings.ContainsKey(name))
+            {
+                if (!SettingsProperties.ContainsKey(name))
+                    Debug.Assert(false, "PackageSettings: " + name + " setting missing. Did you add this setting to the SettingsProperties dictionary?");
+                else
+                    Debug.Assert(false, "PackageSettings: " + name + " setting missing. Registry might be corrupt.");
+#if !DEBUG
+                VsOutputLogger.WriteLine("PackageSettings: " + name + " setting missing.");
+#endif
+                value = SettingsProperties[name]; // default value
```
Haacked: Couldn't this throw in non-DEBUG builds? On line 54 we check to see if the `name` is in the `SettingsProperties` dictionary, but we don't bail out if it's not in there.

Haacked:
@shana a few code review comments for you. I'll look into the other checkboxes.

src/GitHub.VisualStudio/UI/Options/OptionsControl.xaml
shana: Excellent idea!

src/GitHub.Exports/Services/PackageSettings.cs
shana: If the first assert throws in release and we don't catch it in our testing, we should probably be taken out and have pastries thrown at our faces because it means we added a property and we didn't even bother testing it to make sure it works properly. I'll add some tests for this one to make sure it's covered.

shana:
> Why don't we split this PR into two. Add the options whether we're collecting data or not. Then a separate PR for the metrics. Thoughts?

Sounds reasonable. Should I add an issue to track these tasks instead so we don't lose the overall picture?

Haacked:
> Sounds reasonable. Should I add an issue to track these tasks instead so we don't lose the overall picture?

Seems legit to me.

shana:
Part 1 of #229

src/GitHub.VisualStudio/UI/Options/OptionsControl.xaml
Haacked: Turns out that we already have a samples page! But it looked terrible. I fixed it up with some placeholder content. Please add a link to https://visualstudio.github.com/samples.html

Thanks!

shana:
So instead of all this manual checking to make sure things are synchronized, I've switched the code to automatically generate the interface and the implementation of the settings class based on a json file that lists the settings we want with their types and default values. This way we only have to change things in the json file. VS regenerates the files on build if the json file is changed.

shana:
I also moved some other settings files that were spread around (colors, constants, etc) to Settings folder along with this stuff, to make things easier to find.

shana:
:soccer:

src/GitHub.Exports/Settings/IPackageSettings.tt
```diff
@@ -0,0 +1,31 @@
...
+*/
+namespace GitHub.Settings
+{
+    public interface IPackageSettings
+    {
+        void Save();
+<#
+foreach (var j in json["settings"].Children()) {
+#>
+        <#= j["type"] #> <#= j["name"] #> { get; set; }
```
Haacked: I fucking love this approach! Great use of text transformation templates!

I love it so much, I'm going to nitpick here and suggest that should be indented four spaces not eight.  :stuck_out_tongue:

src/GitHub.VisualStudio/Settings/PackageSettingsGen.cs
```diff
@@ -0,0 +1,34 @@
+﻿// This is an automatically generated file, based on settings.json and PackageSettingsGen.tt
```
Haacked: Do we even need to commit these? Will the CI and local build generate them as part of the build?

src/common/settings.json
```diff
@@ -0,0 +1,9 @@
+﻿{
+	"settings": [
```
Haacked: Are there `\t`abs in here? Trying to start a religious war? :stuck_out_tongue:

Haacked:
This is great! Had some nitpicky comments about spacing which aren't really important. The main question is whether we need to commit the generated files or not.

src/common/settings.json
shana: Not my fault, VS keeps putting them in! _blinks innocently_

src/GitHub.VisualStudio/Settings/PackageSettingsGen.cs
shana: They should be regenerated automatically on build if they're missing or stale. Can probably kill them from source control, yeah...

src/GitHub.Exports/Settings/IPackageSettings.tt
shana: But but but but... then the properties will be indented at the level of the interface definition!

Haacked: Don't worry about it. :smile: Trying to get generated code formatted correctly with TT is a time sink.

shana:
I just validated the build on a fresh clone, things are generated correctly and it only rebuilds if files are changed, so :+1:

Haacked:
We should probably add the generated files to `.gitignore` by either ignoring that directory or creating a new `generated` subdirectory to put the generated files in.

Sorry if I'm being nitpicky, but I want this to be a template for how I do this in more place. :smile:

Haacked:
Wonderful!

Issue #159, opened 2015-12-02
GitHub section of Home page is cut when large font is used


xperiandri:
With large font GitHub section of Home page is cut 

<img width="335" alt="github section height" src="https://cloud.githubusercontent.com/assets/2365592/11546149/22a971a2-9954-11e5-9160-417d9cd040d7.png">

Please, do not constraint its height, use appropriate panel for layout.

______________

You can debug this issue by installing Productivity Power tools.
Then:
* Invoke PresentEdit command
```
<?xml version="1.0"?>
<PresenterModeSettings>
  <EnvironmentFontFamily>Segoe UI</EnvironmentFontFamily>
  <EnvironmentFontSize>16</EnvironmentFontSize>
  <TextEditorFontFamily>Consolas</TextEditorFontFamily>
  <TextEditorFontSize>26</TextEditorFontSize>
</PresenterModeSettings>
```
* Save
* Invoke PresentOn command

xperiandri:
Is there a build to try?

shana:
@xperiandri This is included on the 1.0.18.4 release, out today!

PR #234, opened 2016-03-06
 Fixes Bug - GitHub section of Home page is cut when large font is used

gitexperience:
Bug link :- https://github.com/github/VisualStudio/issues/159 
This PR provides a ScrollViewer (hidden) so that all of the content can be seen just by scrolling. This is useful when the size of the font is increased large enough to fit the complete text inside StackPanel.


shana:
The panel should make the section grow so it doesn't require any scrolling no matter how big it gets. Scrolling is not an appropriate solution here, not unless there's no other way of getting the content visible.

shana:
@gitexperience Are you working on addressing the comment I left above? If not, I'm closing this PR, as it doesn't fix the issue correctly.

gitexperience:
@shana , if possible, please keep it open for a while, I am currently working on my GSoC proposal so unable to work in this right now , will try to solve this and address you once i am done with my proposal.

shana:
@gitexperience Sure!

gitexperience:
It will now look like this

shana:
@gitexperience Thanks for the fix! Your patch will be merged as part of #264

PR #241, opened 2016-03-16
[WIP] Add Pull Request Creation View

donokuda:
(Target branch: `feature/pr/views`)

This PR provides the static markup for the create pull request view. Would love some thoughts on the view as well as any feedback on improving the xaml. Also added new button and tab control styles. As a heads up, this isn't hooked up to live data. 

> create-pr-stubby-stub-stub
>
> Preview with example commits and file changes. (This gif includes hardcoded data for demo purposes, you won't see as much information if you were to pull down the branch and try it out.)

We show if the branch is able to merge in GHD and dotcom. Is this something we should also show on this screen?

I still have a few other outstanding things around theming to work on:

- [x] Convert hardcoded colors to theme variables.
  - [x] Light
  - [x] Blue
  - [x] Dark
- [x] Inconsistent padding

/cc @shana @niik


shana:
woohoo, this is awesome! :sparkles:
Don't worry about the failed tests, that's on my side.

shana:
The merge conflict is because I've made the pane title dynamic in `feature/pr/views`, so we can change it depending on the view. Now that that's in, what titles should we have for the creation and the detail view? 

https://github.com/atom/design/tree/master/wip/mvp has some interesting layout options, perhaps we should take another pass at the layout we have right now. Showing the commit and file count right now would probably be good, since we're not yet listing them.

simurai:
> Showing the commit and file count right now would probably be good, since we're not yet listing them.

In Atom's case, we haven't really decided what happens if you click those:

screen shot 2016-04-21 at 5 20 00 pm

Maybe initially it's just FYI. And then later have them open a list. In that case, the tabs shown above would work nicely. 👍

donokuda:
> Maybe initially it's just FYI. And then later have them open a list. In that case, the tabs shown above would work nicely. 

I like the direction Atom is going towards! Here's a mock of how it could look like in Visual Studio but following our tabbed direction:

image

Alternatively, we could use a different style for the tabs:

image

As an exploration, I also tried combining the merge status with the branch selection. There would be a ✅  if the branch can be merged and an alert symbol if there are conflicts. Hovering over the symbol will give more information in a tooltip.

image

shana:
> As an exploration, I also tried combining the merge status with the branch selection. There would be a ✅ if the branch can be merged and an alert symbol if there are conflicts. Hovering over the symbol will give more information in a tooltip.

I like this, it gives the branch information line more focus, which is important since users need to make sure they select the target branch correctly.

> Alternatively, we could use a different style for the tabs:

This design feels more crowded than with the "regular" tab style. But with the branch info + merge info all in one line, looks nice, I like it.

screen shot 2016-04-22 at 11 46 30 am

/cc @Haacked @niik for 👀

donokuda:
> This design feels more crowded than with the "regular" tab style. But with the branch info + merge info all in one line, looks nice, I like it.

Here's a version with the regular tab style with the branch information and status on one line. It does feel lighter and more open compared to the non-tab style.

image

shana:
I like it! Do you want to go ahead and implement that on this branch? I'll merge the combobox one in and you can continue the fun here.

donokuda:
@shana Sounds good. I'll do that right now! :rocket:

donokuda:
Whew, 99 commits. :sweat: I'm more than happy to squash this down if that's preferred. 

That said, I think this is ready for another review:

:sparkles: :movie_camera: :sparkles: 

create-blue
create-dark
create-light

I've also removed any font sizes that have been hardcoded. Not 100% visually ideal tho:

loud noises

src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml
```diff
@@ -3,33 +3,466 @@
...
+                                        RenderOptions.BitmapScalingMode="HighQuality"
+                                        Source="/GitHub.VisualStudio;component/Resources/default_user_avatar.png"/>
+                                </StackPanel>
+                            </Grid>
+
+                            <TextBlock Text="I am an example commit message"
+                                       ToolTip="I am an example commit message"
+                                       VerticalAlignment="Top"
+                                       Foreground="{DynamicResource GitHubVsToolWindowText}"
+                                       FontSize="14"
```
donokuda: Instead of this being hardcoded, is there a way to set this relative to the user's font settings?

src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml
```diff
@@ -3,33 +3,466 @@
...
+                                       Margin="10,-3,5,0"
+                                       Grid.Column="1"
+                                       Grid.Row="0" />
+
+                            <TextBlock Margin="10,0,0,0"
+                                       Text="Andreia Gaita committed 20 hours ago"
+                                       ToolTip="Andreia Gaita committed 20 hours ago"
+                                       VerticalAlignment="Top"
+                                       FontSize="11"
```
donokuda: Same question as above:

> Instead of this being hardcoded, is there a way to set this relative to the user's font settings?

shana:
Wow, this is great! Don't worry about the squashing, it's valuable history :smile:

I'll figure out if there are environment resources we can use in the same way we use VS colors, and whether that makes sense for these UIs. Regardless of that, I think the best way to go about this is to define the font sizes we want separately, the same way we define our custom colors. We don't need them per-theme, so they can go into the `SharedDictionary.xaml` file in the GitHub.VisualStudio project for now (we can put them in their own file in the Styles folder later). Then we can use them in the same way everywhere instead of hardcoding them on each ui element, and it's easy to change them across the board.

I should have thought about this when we were discussing it earlier, sorry about that :-P

shana:
Going to merge this in so we can continue more UI work directly off of `feature/pr/views`, and also start loading the data and doing the popups properly in the creation view.

@paladique You might need to do a new PR based off of `feature/pr/views` for loading the data in the PR creation view, since your branch is based off of this one and it's going to disappear.

shana:
Also amazing work 😍  :shipit: 🚀

PR #245, opened 2016-03-23
More stable to compare local paths in uppercase.

paladique:


src/GitHub.VisualStudio/UI/Views/GitHubConnectContent.xaml.cs
```diff
@@ -101,7 +101,7 @@ public object Convert(object[] values, Type targetType, object parameter, Cultur
 
             var item = values[0] as ISimpleRepositoryModel;
             var context = values[1] as IGitAwareItem;
-            return context?.ActiveRepoUri == item?.CloneUrl && context?.ActiveRepo?.LocalPath == item?.LocalPath;
+            return context?.ActiveRepoUri == item?.CloneUrl && context?.ActiveRepo?.LocalPath.ToUpper() == item?.LocalPath.ToUpper();
```
Haacked: `ToUpper` creates a new string. In this case, two throwaway strings just to compare equality. Also, you can run into the [Turkish I problem here](http://haacked.com/archive/2012/07/05/turkish-i-problem-and-why-you-should-care.aspx/) because no culture was specified.

The safe way to compare strings in a case insensitive manner is to use [`String.Equals`](https://msdn.microsoft.com/en-us/library/t4411bks(v=vs.110).aspx) and pass in a `StringComparison`. For example,

```csharp
String.Equals(context?.ActiveRepo?.LocalPath, item?.LocalPath, StringComparison.OrdinalIgnoreCase);
```

paladique: ah good point!

shana: Yeah, my bad, my brain was obviously not working properly yesterday. Phil's approach is totally the correct one for this!

Haacked: I have years of experience doing improper string operations. :stuck_out_tongue:

src/GitHub.VisualStudio/UI/Views/GitHubConnectContent.xaml.cs
```diff
@@ -101,7 +101,7 @@ public object Convert(object[] values, Type targetType, object parameter, Cultur
 
             var item = values[0] as ISimpleRepositoryModel;
             var context = values[1] as IGitAwareItem;
-            return context?.ActiveRepoUri == item?.CloneUrl && context?.ActiveRepo?.LocalPath == item?.LocalPath;
+            return context?.ActiveRepoUri == item?.CloneUrl && String.Equals(context?.ActiveRepo?.LocalPath, item?.LocalPath, StringComparison.OrdinalIgnoreCase);
```
Haacked: So, something that just occurred to me. Is `context?.ActiveRepoUri` a `UriString`? Is `item?.CloneUrl` too? If either is a string, then we'll get a string comparison here because of the implicit conversion of `UriString` to `string`. Even if they're both `UriString`, I'm not sure our `Equals` implementation does the right thing here. :frowning: Something worth testing out.

shana: The implementation uses [the Ordinal comparison as well](https://github.com/github/VisualStudio/blob/master/src/GitHub.Exports/Primitives/StringEquivalent.cs#L32), so == should actually be doing the correct thing here, they're both `UriString`.

Haacked:
Looks good. It'd be nice to have a unit test that fails without this fix and passes with it. But we can get to that later. :smile:

PR #263, opened 2016-04-14
Add combobox UI

donokuda:
**Destination branch:** `don/stub-ui`

This PR stubs out a filterable combobox ui for selecting branches and assignees. I originally went with a UI very similar to what we have in Desktop; however, it felt out of place with Visual Studio's UI. I ended up trying out a different approach with a look and feel closer to a native context menu.

**Desktopish (Mock-up)**
(I have a branch with this style but it has become so stale that it won't build on my machine without a bit of troubleshooting :sob:)

image

**Context menu-ish**
combobox-demo

There's still some refinement, polish, and theming, that needs to be done; but I would love some thoughts on this direction.

/cc @shana, @simurai, @niik for :eyes: and :thought_balloon: (although feedback from anyone is encouraged)

### DIFF
--- a/src/GitHub.UI/Assets/TextBlocks.xaml
+++ b/src/GitHub.UI/Assets/TextBlocks.xaml
@@ -71,7 +71,7 @@
                             </Border>
                         </Border>
 
-                        <Grid Margin="1,0,0,0">
+                        <Grid Margin="1,2,0,0">
                             <ScrollViewer x:Name="PART_ContentHost" Padding="{TemplateBinding Padding}" Focusable="false" HorizontalScrollBarVisibility="Hidden" VerticalScrollBarVisibility="Hidden" VerticalAlignment="Top" Margin="0"/>
                             <Label x:Name="PromptLabel" HorizontalAlignment="Left"
                                  Foreground="{DynamicResource GHTextBrush}"
--- a/src/GitHub.VisualStudio/Styles/GitHubComboBox.xaml
+++ b/src/GitHub.VisualStudio/Styles/GitHubComboBox.xaml
@@ -1,0 +1,56 @@
+﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+                    xmlns:ui="clr-namespace:GitHub.UI;assembly=GitHub.UI"
+                    xmlns:local="clr-namespace:GitHub.VisualStudio.Styles">
+    
+    <Style x:Key="GitHubComboBoxBorder" TargetType="{x:Type Border}">
+        <Setter Property="Margin" Value="10" />
+        <Setter Property="BorderThickness" Value="1" />
+        <Setter Property="BorderBrush" Value="{DynamicResource GitHubComboBoxBorderBrush}" />
+        <Setter Property="Effect">
+            <Setter.Value>
+                <DropShadowEffect Direction="315" ShadowDepth="5" BlurRadius="5" Opacity="0.25" />
+            </Setter.Value>
+        </Setter>
+    </Style>
+
+    <Style x:Key="GitHubComboBoxContainer" TargetType="{x:Type Grid}">
+        <Setter Property="Background" Value="{DynamicResource GitHubComboBoxBackgroundBrush}" />
+
+        <Style.Resources>
+            <Style TargetType="{x:Type TextBlock}" BasedOn="{StaticResource {x:Type TextBlock}}">
+                <Setter Property="Foreground" Value="{DynamicResource GitHubVsToolWindowText}" />
+            </Style>
+
+            <Style TargetType="{x:Type ui:OcticonImage}" BasedOn="{StaticResource {x:Type ui:OcticonImage}}">
+                <Setter Property="Foreground" Value="{DynamicResource GitHubVsToolWindowText}" />
+            </Style>
+            
+            <Style TargetType="{x:Type Separator}" BasedOn="{StaticResource {x:Type Separator}}">
+                <Setter Property="Background" Value="{DynamicResource GitHubComboBoxBorderBrush}" />
+                <Setter Property="Margin" Value="0" />
+            </Style>
+
+            <Style TargetType="{x:Type ui:FilterTextBox}" BasedOn="{StaticResource {x:Type ui:FilterTextBox}}">
+                <Setter Property="BorderBrush" Value="{DynamicResource GitHubComboBoxBorderBrush}" />
+                <Setter Property="Foreground" Value="{DynamicResource GitHubVsWindowText}" />
+                <Setter Property="Background" Value="{DynamicResource GitHubVsBrandedUIBackground}" />
+                <Setter Property="Margin" Value="5" />
+                <Setter Property="Height" Value="25" />
+            </Style>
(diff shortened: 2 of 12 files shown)


simurai:
The light grey makes it a bit stand out more. 👍 

Not sure about the border for selected items:

screen shot 2016-04-15 at 5 07 23 pm

Feels a bit crammed. I guess that's the default style? Maybe a bit more padding? Or no border, like for the assignee dropdown.

shana:
Agree with @simurai on the border. The assignee dropdown has no border, only the checkbox, should probably make both dropdowns consistent?

donokuda:
@simurai @shana Agree on the border. Going to :fire: that and see if it helps with the cramped feeling.

donokuda:
Not exactly sure how I "fixed" the gray box, but it's :fire: now:
cccccombobox

Assignee combo box
ccccombo-people

Here's the default hover state for a row. Not 100% sold on it so I'll most likely noodle on it while doing the different themes.
ccccombo-hover

src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml
```diff
@@ -381,43 +430,201 @@
...
+                            <Grid.RowDefinitions>
+                                <RowDefinition Height="*" />
+                                <RowDefinition Height="*" />
+                                <RowDefinition Height="Auto" />
+                            </Grid.RowDefinitions>
+
+                            <ui:FilterTextBox
+                                x:Name="filterAssignee"
+                                Grid.Row="0"
+                                Foreground="{DynamicResource GitHubVsWindowText}"
```
donokuda: @shana I was hoping to be able to move this to the style declarations around [line 62](https://github.com/github/VisualStudio/blob/add-combobox-ui/src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml#L62) but it doesn't seem to set the foreground in [`FilterTextBox.xaml`](https://github.com/github/VisualStudio/blob/add-combobox-ui/src/GitHub.UI/Assets/Controls/FilterTextBox.xaml#L54).

Any insights as to why this isn't happening?

shana: Hmmm, no clue 😕

shana: I haz clue! Take out https://github.com/github/VisualStudio/blob/add-combobox-ui/src/GitHub.UI/Assets/Controls/FilterTextBox.xaml#L126, it's resetting the foreground in such a way that it overrides dynamic resource'd settings.

donokuda: :sparkles: That did it! :sparkles:

Thanks for looking into it!

donokuda:
Added themes for the combo box. The light and blue themes share the same colors while the dark theme is it's own thing.

Blue:
combobox-blue

Dark:
combobox-dark

Light:
combobox-light

There's an issue where the prompt TextBox text is vertically off-centered which I plan on fixing next:

screen_shot_2016-04-19_at_3_39_06_pm

Would still love some :thought_balloon: and :eyes:

@shana This UI be moved to `GitHub.UI` instead of keeping it in the `PullRequestCreationView.xaml` file, right?

shana:
> This UI be moved to GitHub.UI instead of keeping it in the PullRequestCreationView.xaml file, right?

Let's put it in GitHub.VisualStudio.UI. That's going to require moving the style files as well, so just put the style in its own file in the Styles directory, and we'll move all of it later.

donokuda:
> Let's put it in GitHub.VisualStudio.UI. That's going to require moving the style files as well, so just put the style in its own file in the Styles directory, and we'll move all of it later.

:+1: Excellent, I'll go ahead and do that.

src/GitHub.VisualStudio/Styles/GitHubComboBox.xaml
```diff
@@ -0,0 +1,56 @@
...
+
+    <Style x:Key="GitHubComboBoxContainer" TargetType="{x:Type Grid}">
+        <Setter Property="Background" Value="{DynamicResource GitHubComboBoxBackgroundBrush}" />
+
+        <Style.Resources>
+            <Style TargetType="{x:Type TextBlock}" BasedOn="{StaticResource {x:Type TextBlock}}">
+                <Setter Property="Foreground" Value="{DynamicResource GitHubVsToolWindowText}" />
+            </Style>
+
+            <Style TargetType="{x:Type ui:OcticonImage}" BasedOn="{StaticResource {x:Type ui:OcticonImage}}">
```
donokuda: This line triggers a XAML parse error:

> {"Cannot find resource named 'GitHub.UI.OcticonImage'. Resource names are case sensitive."} 

I'm assuming I'm not resourcing correctly in Visual Studio on [Line 3](https://github.com/github/VisualStudio/blob/add-combobox-ui/src/GitHub.VisualStudio/Styles/GitHubComboBox.xaml#L3) :stuck_out_tongue: 

```xaml
<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                    xmlns:ui="clr-namespace:GitHub.UI;assembly=GitHub.UI">
```

Any chance of pointing me in the right direction of understanding how to get the right resource? I've been trying to follow how we're doing it in `PullRequestCreationView.xaml`.

shana: Hmmm, the `xmlns:ui` declaration is correct, the `OcticonImage` type is in the `GitHub.UI` namespace as declared here: https://github.com/github/VisualStudio/blob/master/src/GitHub.UI/Controls/Octicons/OcticonImage.cs#L5.
It might need the GitHubUI SharedDictionary.xaml file where the `OcticonImage.xaml` is included from (where the style is eventually declared).

Try adding to the top:

```
<ResourceDictionary.MergedDictionaries>
   <cache:SharedDictionaryManager Source="pack://application:,,,/GitHub.UI;component/SharedDictionary.xaml" />
</ResourceDictionary.MergedDictionaries>
```

donokuda: :bow: That did it! Added in 7a334de

src/GitHub.UI/Assets/TextBlocks.xaml
```diff
@@ -71,7 +71,7 @@
                             </Border>
                         </Border>
 
-                        <Grid Margin="1,0,0,0">
+                        <Grid Margin="1,2,0,0">
```
shana: Does this affect the UI in the Login/RepositoryCreation/RepositoryClone controls? Are they using this as well?

shana:
@donokuda I added a comment about the margin changes you did. Otherwise, things look ✨ ! I'll do a final pass and merge if you're happy with this branch 😄

src/GitHub.UI/Assets/TextBlocks.xaml
donokuda: This is a good question! I'll double check right now

donokuda: I believe those controls do so their text gets centered as well:
artboard 21

shana: purrrrfect!

PR #265, opened 2016-04-15
[wip] Fleshing out the data backend for PR creation

shana:
- [x] Covert collections of strings to objects in SampleData
- [x] Update xaml file to reflect Sample View Model
- [x] Confirm xaml file reflects sample data at compile time

- [x] Bind sample data to View Model
- [x] Update xaml file to reflect View Model
- [x] Confirm view reflects sample data at run time

- [ ] Bind real data to View Model


src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -10,9 +11,31 @@
 
 namespace GitHub.ViewModels
 {
+    //Add properties to interface first
+    //Ex would be login tab view model
     [ExportViewModel(ViewType = UIViewType.PRCreation)]
     [PartCreationPolicy(CreationPolicy.NonShared)]
     class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationViewModel
     {
+        public PullRequestCreationViewModel(IRepositoryHost repositoryHost, ISimpleRepositoryModel repository)
```
shana: The initialization code is failing because this constructor is not marked `[ImportingConstructor]`, so MEF can't create an instance of this. Every class that's exported to MEF (marked with `[Export..]`) has to either have an empty public constructor or a constructor marked with `[ImportingConstructor]`, so that MEF knows which constructor to invoke when instantiating things.

src/GitHub.App/SampleData/PullRequestCreationViewModelDesigner.cs
```diff
@@ -9,27 +14,45 @@ public class PullRequestCreationViewModelDesigner : BaseViewModelDesigner, IPull
...
-            Branches = new List<string>()
+            //TODO:Make String into an IBranch
+            Branches = new ReactiveList<IBranch>
             {
-                "don/stub-ui",
-                "feature/pr/views",
-                "release-1.0.17.0"
+               new Branch {Name = "don/stub-ui"},
+               new Branch {Name = "feature/pr/views"},
+               new Branch {Name = "release-1.0.17.0"}
```
shana: Make sure this follows the code design guidelines - spaces before `{` and `}` (same for code below)

src/GitHub.App/Services/ModelService.cs
```diff
@@ -370,4 +415,6 @@ public PullRequestCacheItem(PullRequest pr)
             public bool IsOpen { get; set; }
         }
     }
+
```
shana: Avoid introducing whitespace changes, they are prone to merge conflicts

shana:
I figured out why this failed to run, I've added a comment about it. You can go ahead and revert the eddd987 commit, or better yet, reset history to the commit before it and force push so we don't have it in the history, it's not particularly useful.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -1,18 +1,82 @@
...
     class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationViewModel
     {
+        static readonly Logger log = LogManager.GetCurrentClassLogger();
+        readonly IReactiveCommand<IReadOnlyList<IBranch>> loadBranchesCommand;
+        readonly ObservableAsPropertyHelper<bool> isLoading;
+        private bool IsLoading;
+
+
+
+        public PullRequestCreationViewModel(IRepositoryHost repositoryHost, ISimpleRepositoryModel repository)
```
shana: Needs `[ImportingConstructor]`

src/GitHub.App/GitHub.App.csproj
```diff
@@ -29,7 +29,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <RunCodeAnalysis>true</RunCodeAnalysis>
+    <RunCodeAnalysis>false</RunCodeAnalysis>
```
shana: This shouldn't be in production, don't forget to revert this change!

src/GitHub.VisualStudio/source.extension.vsixmanifest
```diff
@@ -10,7 +10,7 @@
     <PreviewImage>Resources\preview_200x200.png</PreviewImage>
     <Tags>GitHub git github.com team explorer</Tags>
   </Metadata>
-  <Installation AllUsers="true">
+  <Installation>
```
shana: This change should never be commited, change it back before you're done with this branch.

paladique:
No luck with getting sample data to populate at runtime, @shana suggested a sample side project/experiment with a simple example to try to understand why this isn't successful.

paladique:
Looks like the assignee list is populating with everyone from the whole org. `GetAllForRepository` grabs all possible users, so we might need to filter this out.

shana:
Yeah, it should only be listing the actual repository members. Could you check which GitHub API would return that? It should be somewhere on the API reference page on the web.

paladique:
>It should only be listing the actual repository members.

Searched the API for this, I'm currently using  `Issue.Assignee.GetAllForRepository` which is `GET /repos/:owner/:repo/assignees` and returns all possible assignees, which is org based. So a large org will have a long assignee list, seems at that point it's not repo specific since anyone in the org (or if not in an org, added as a collaborator) can be a collaborator.

There is `Repository.GetAllContributors` which is `/repos/:owner/:repo/contributors` which we could compare against and put all contributors at the top.

shana:
> There is Repository.GetAllContributors which is /repos/:owner/:repo/contributors which we could compare against and put all common contributors at the top, and would filter out non-collaborators.

Yup, that's a good idea!

paladique:
Development of the PR Creation is now [here](https://github.com/github/VisualStudio/pull/353)

PR #266, opened 2016-04-18
[WIP] Pull Requests: Logged out view

shana:
When the user is logged out, we should probably not try to show all the data, and instead inform them that they should log in. This PR (initially) contains a stub of a logged out view, to flesh out. To see it, log out of GitHub and make sure you have a GitHub repo selected.

### DIFF
--- a/src/GitHub.App/GitHub.App.csproj
+++ b/src/GitHub.App/GitHub.App.csproj
@@ -130,6 +130,7 @@
     </Compile>
     <Compile Include="SampleData\PullRequestCreationViewModelDesigner.cs" />
     <Compile Include="SampleData\PullRequestDetailViewModelDesigner.cs" />
+    <Compile Include="SampleData\LoggedOutViewModelDesigner.cs" />
     <Compile Include="SampleData\PullRequestListViewModelDesigner.cs" />
     <Compile Include="Services\AvatarProvider.cs" />
     <Compile Include="Services\GitHubCredentialProvider.cs" />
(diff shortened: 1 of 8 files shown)


shana:
@donokuda I rebased this branch against `feature/pr/views` and enabled the logged out view. This is what I get now:

screen shot 2016-04-20 at 10 38 30 am

We might want to dock it to the top so it doesn't appear floating in the middle of the pane, perhaps?

Since this is our UI and not Team Explorer, how about calling it "Login to GitHub" instead of "Connect to GitHub"? "Connect" is a Team Explorer thing which is specific to what it does in TFS (connecting to remote TFS projects).

Also, do we want to have more information on why the user should login? Would that be worth it?

@pmn @Haacked Thoughts?

shana:
On a side note, it is possible to have the login happen directly in the pane and not as a separate dialog, if we want to do that at some point. Just fyi.

shana:
Should we always set the padding on the container control, so all views get it automatically? Right now we are setting it in each view (like in https://github.com/github/VisualStudio/blob/feature/pr/views/src/GitHub.VisualStudio/UI/Views/PullRequestListView.xaml#L83).

donokuda:
Updated the logged out view (note: the panel title's color currently doesn't change when switching themes and probably should be fixed in a separate branch)

image

- Moved content to top
- Added margin around the content
- Synced terminology with what we have on GitHub.com (Login vs. Sign in)

> Also, do we want to have more information on why the user should login? Would that be worth it?

:+1: I copy-pasta'd the text from the Team Explorer but I definitely think the text should be clarified. We could iterate with what we have with something like "Get powerful collaboration, code review, and code management tools for..."

> On a side note, it is possible to have the login happen directly in the pane and not as a separate dialog, if we want to do that at some point. Just fyi.

I've thought about this but figured we could use what's already implemented in the extension. However, this might be worth visiting down the road.

pmn:
It looks great to me -- :heart: both the copy and the style.

Haacked:
:thumbsup:

shana:
This needs localization, but let's do a full localization pass on the side for the pr branch.

PR #270, opened 2016-04-22
Cache blob caches

shana:
SQLitePersistentBlobCache need to be cached, `ImageCache` creates a new one every time it needs to load an image from the cache, and they persist across the lifetime of the application, resulting in a million tasks just hanging there waiting for work, eventually exhausting system resources.


Haacked:
Flawless Victory!

PR #271, opened 2016-04-27
Polish Pull Request list view

donokuda:
**Destination branch:** `feature/pr/views`

Polished up the Pull Request list view! :sparkles: 

pull request list

Basically gave some of the list items a little breathing room and added some tooltips to the filter dropdowns!

tooltips

Styles aren't themed yet but I can either do them here or follow it up in a separate PR.

/cc @simurai for :eyes:


donokuda:
One thing I quickly noticed on my 13" screen is it's likely we'll run out of space for the title and description.

We currently have the PR title wrap and I have set the description text set to trim with ellipsis when we run out of space. Hovering over the text will give a tooltip:

tooltips

A similar UI on Desktop is the commit list. We truncate the title and the description text if it's too long (using ellipsis):

image

Here's a version of the PR list with the text treated like in Desktop:

clipped

Does it make sense to treat the list here like in Desktop?

simurai:
> Does it make sense to treat the list here like in Desktop?

Yeah, I like the truncating. 👍  Makes it easier to scan with all items having the same height.

Maybe to keep the names visible, remove the `opened`?

donokuda:
> Maybe to keep the names visible, remove the opened?

Good idea. Updated in 3d61dec

screen shot 2016-04-28 at 10 22 11 am

We can keep the "opened" in the tooltip in case we want to clarify for different states down the road ("opened," "closed," "updated," etc)

screen shot 2016-04-28 at 10 23 22 am

shana:
:+1: on removing the `opened`, we actually want to show opened or updated dates there anyway so it doesn't make much sense to have the verb.

shana:
> We can keep the "opened" in the tooltip in case we want to clarify for different states down the road ("opened," "closed," "updated," etc)

💯

donokuda:
This is good for a final review! :sparkles: :eyeglasses: :sparkles: 

Still unsure why the comment octicon disappears when we try to apply style triggers to it but I think that's something we could explore in a separate branch.

donokuda:
Also gave the the vertical separators a theming pass.

image

src/GitHub.App/SampleData/PullRequestListViewModelDesigner.cs
```diff
@@ -26,7 +26,7 @@ public PullRequestListViewModelDesigner()
                 new AccountDesigner { Login = "haacked", IsUser = true },
                 new AccountDesigner { Login = "shana", IsUser = true },
                 DateTimeOffset.Now - TimeSpan.FromMinutes(2)) { CommentCount = 4, HasNewComments = false });
-            prs.Add(new PullRequestModel(409, "Fix publish button style",
+            prs.Add(new PullRequestModel(409, "Fix publish button style and a really, really long name for this thing... OMG look how long this name is yusssss",
```
shana: 😍

shana:
This is so goood!!!!111eleventy!

PR #273, opened 2016-04-29
Fix dialog scaling

### LINKED ISSUE #238: Bad clone window layout
github clone

1. Please, never restrict height of elements because with larger font all the text will be cut.
2. Not all text elements use VS font settings
3. It is better to name this window "Clone my GitHub Repository" because it is very confusing how to clone someone's else repository

grokys:
For some reason seems to break layout in the clone dialog with large
font sizes. Fixes #238.

### DIFF
--- a/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml
+++ b/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml
@@ -137,7 +137,7 @@
           <Setter Property="Foreground"
                   Value="{DynamicResource GHTextBrush}" />
           <Setter Property="Margin"
-                  Value="0,0,12,0" />
+                  Value="0,6,12,6" />
         </Style>
       </Border.Resources>
       <Grid Margin="0">
--- a/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml
+++ b/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml
@@ -137,7 +137,7 @@
           <Setter Property="Foreground"
                   Value="{DynamicResource GHTextBrush}" />
           <Setter Property="Margin"
-                  Value="0,6,12,6" />
+                  Value="0,0,12,0" />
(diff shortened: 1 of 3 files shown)


Haacked:
Got any before and after screenshots?

grokys:
Sure! Before, using a large font in VS:

before-large-font

After:

after-large-font

Before using standard font size (everything fine here):

before-normal-font

After (no change here):

after-normal-font

grokys:
BTW I just noticed the path in the browse box. Is that normal? That is *not* my username ;)

shana:
So I figured out why it's getting cut off: the `repositoryBorderStyle` has a hardcoded height of 30, around line 44. Even if you adjust the margins, if the font gets bigger, it will still cut off.

facepalm

grokys:
Just to clarify: the hardcoded height is causing the clipping in the "Search Repositories" box rather than the clipping on the username?

shana:
@grokys I think both are suffering from hardcoded heights.

grokys:
I've updated `WindowController` to fix the font used in our dialogs - our theme uses various different font sizes for headers etc and there is no easy way to scale these based on the visual studio Environment font settings, so for now we decided to just hard-code the font settings in our dialogs.

The UI will still honor DPI scaling, which should be the preferred method to change the font size in VS.

src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml
```diff
@@ -137,7 +137,7 @@
           <Setter Property="Foreground"
                   Value="{DynamicResource GHTextBrush}" />
           <Setter Property="Margin"
-                  Value="0,6,12,6" />
+                  Value="0,0,12,0" />
```
shana: So maybe we want to keep the padding here, as it's not responsible for the clipping after all.

grokys: True, will revert that change.

shana:
🚀 ✨ 💃

Issue #277, opened 2016-05-02
Unable to login after logging out 


mponce1:
Steps to reproduce: 
1. Log into GHfVS via team explorer > manage connections > github
2. Log out
3. Login with the same user or any other user

Result: Unable to connect to server error message. 

Extension.log: 
```
2016-05-02 16:32:27.8143|INFO|thread: 1|LoginTabViewModel|Error logging into 'https://github.com/' as 'mergecity'
System.ObjectDisposedException: The cache 'SqlitePersistentBlobCache' was disposed.
```

![screen shot 2016-05-02 at 4 32 47 pm](https://cloud.githubusercontent.com/assets/6272408/14971075/850bb3cc-1083-11e6-80cf-9f33869cc79a.png)

grokys:
Can't repro this on master. Was it fixed by #280?

grokys:
Closing as this seems to be fixed.

PR #280, opened 2016-05-03
Fix premature cache disposal

shana:
Turns out #270 wasn't quite the flawless victory... it causes premature disposal of caches, as seen in #277

Now that we're not creating a million copies of the same cache every time something wants something from it, `ModelService` can't just willy-nilly dispose of the `hostCache` object when it feels like it,
because that cache object will be reused by other services, if needed.

The factory is the one responsible for keeping the lifetime of its cached objects (and yes, now it should probably not be called a factory anymore, it should be a service but oh well), and so now it needs to be
disposable and take care of things when it itself is disposed of, usually when MEF shuts down.

### DIFF
--- a/src/GitHub.App/Factories/HostCacheFactory.cs
+++ b/src/GitHub.App/Factories/HostCacheFactory.cs
@@ -39,5 +39,23 @@ namespace GitHub.Factories
 
         IOperatingSystem OperatingSystem { get { return operatingSystem.Value; } }
         IBlobCacheFactory BlobCacheFactory { get { return blobCacheFactory.Value; } }
+
+        bool disposed;
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                if (disposed) return;
+                disposed = true;
+                if (blobCacheFactory.IsValueCreated)
+                    blobCacheFactory.Value.Dispose();
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


src/GitHub.App/Factories/HostCacheFactory.cs
```diff
@@ -39,5 +39,23 @@ public IBlobCache Create(HostAddress hostAddress)
 
         IOperatingSystem OperatingSystem { get { return operatingSystem.Value; } }
         IBlobCacheFactory BlobCacheFactory { get { return blobCacheFactory.Value; } }
+
+        bool disposed;
+        protected virtual void Dispose(bool disposing)
```
Haacked: Given that this class doesn't have a finalizer, is it even necessary to have a `Dispose` that takes in `disposing` since it's always true? I guess a derived class could add a finalizer, but we could solve that by making this sealed. I don't feel too strongly about it though.

Haacked:
Code looks good to me. I made one comment, but it's not that important. Code is correct as is. Nice work!

src/GitHub.App/Factories/HostCacheFactory.cs
shana: Yeah, it's a bit of a waste of lines of code, but code analysis rules will always complain if this pattern isn't followed for non-sealed classes, and given that this is an annoyingly hard pattern to get right, I'm taking the approach of copy-pasting the pattern across all our implementations to make sure it's always correct. I've had more problems with random copy-pasting particular implementations of Dispose that were only appropriate for some specific scenario, which then fail subtly, than I've had just using the more verbose pattern everywhere, so that's what I'm doing here 😄 

Since the blob cache factory now needs to be disposed, we should actually be using `ExportLifetimeProvider` instead of `Lazy`, which is the proper way of controlling MEF disposals and object lifetime. I might throw you another PR with that.

Issue #276, opened 2016-05-02
Make entries in Clone dialog collapsible.


grokys:
Make entries collapsible by user/organization. Start collapsed by default if there is >1.

shana:
Weird, I wasn't this closed automatically when the PR was merged?

PR #288, opened 2016-05-04
Initial support for collapsing in clone dialog.

grokys:
Fix for #276 

Makes groups in the clone dialog collapsible. If there is only a single
group, then it will be expanded; as more groups are created, the first
group is collapsed and further groups will be added collapsed.

Only problem with this is that you can see an initial flash of expanded
items when loading with multiple groups, however I felt this was
preferable to everything being collapsed until the whole list is
loaded...

### DIFF
--- a/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml.cs
+++ b/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml.cs
@@ -31,7 +31,7 @@ namespace GitHub.VisualStudio.UI.Views.Controls
     [PartCreationPolicy(CreationPolicy.NonShared)]
     public partial class RepositoryCloneControl : GenericRepositoryCloneControl
     {
-        private Dictionary<string, RepositoryGroup> groups = new Dictionary<string, RepositoryGroup>();
+        readonly Dictionary<string, RepositoryGroup> groups = new Dictionary<string, RepositoryGroup>();
 
         public RepositoryCloneControl()
         {
@@ -73,7 +73,7 @@ namespace GitHub.VisualStudio.UI.Views.Controls
 
         class RepositoryGroupDescription : GroupDescription
         {
-            RepositoryCloneControl owner;
+            readonly RepositoryCloneControl owner;
 
             public RepositoryGroupDescription(RepositoryCloneControl owner)
             {
@@ -82,7 +82,9 @@ namespace GitHub.VisualStudio.UI.Views.Controls
 
             public override object GroupNameFromItem(object item, int level, System.Globalization.CultureInfo culture)
             {
-                var name = ((IRepositoryModel)item).Owner.Login;
+                var repo = item as IRepositoryModel;
+                Debug.Assert(repo.Owner != null, "Repository owner cannot be null, did something get changed in the RepositoryModel ctor?");
+                var name = repo.Owner.Login;
                 RepositoryGroup group;
 
                 if (!owner.groups.TryGetValue(name, out group))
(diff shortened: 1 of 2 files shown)


grokys:
This will obviously need a pass by @donokuda too because it's fugly as sin ;)

donokuda:
@grokys I can polish this up today!

Currently it looks like:

image

I did a quick mockup and thinking of moving towards something closer to this:

clone repo new

I'm going to start on this, but I am also open to feedback in the meantime.

grokys:
Looks great Don!

shana:
I like it! ✨

grokys:
Merged @donokuda polish and think this is now ready for review :gem: :octocat:

src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml.cs
```diff
@@ -60,24 +63,60 @@ public RepositoryCloneControl()
             };
         }
 
-        static ListCollectionView CreateRepositoryListCollectionView(IEnumerable<IRepositoryModel> repositories)
+        ListCollectionView CreateRepositoryListCollectionView(IEnumerable<IRepositoryModel> repositories)
```
Haacked: This could be a `static` method, right?

src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml.cs
```diff
@@ -60,24 +63,60 @@ public RepositoryCloneControl()
...
+
+            public RepositoryGroupDescription(RepositoryCloneControl owner)
+            {
+                this.owner = owner;
+            }
+
             public override object GroupNameFromItem(object item, int level, System.Globalization.CultureInfo culture)
             {
-                return ((IRepositoryModel)item).Owner.Login;
+                var name = ((IRepositoryModel)item).Owner.Login;
```
Haacked: We're sure `Owner` can never be null? If so, it might be worth asserting that here for future reviewers benefit. :smile:

Haacked:
Looks good! I just had some really minor nitpicky feedback. Probably need to update this branch from master (or rebase it if you're feeling funky).

Haacked:
:tea:

src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml.cs
grokys: I assumed it couldn't as the previous implementation of this didn't check for null, but sure I can add an assertion.

src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml.cs
grokys: It was static before but with these changes, but now `RepositoryGroupDescription` needs a reference to the `RepositoryCloneControl` (i.e. `this`) in order to collapse the existing group when a second group is added.

Haacked: Ah! How did I not see `this`? :laughing: 
Ok, carry on.

src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml.cs
Haacked: Ah, good point. As you'll learn, I'm a big fan of assertions when it comes to situations like this because in reviewing code, it's hard to tell the difference from a potential null ref or something we're aware of as being never null.

shana:
I fixed some nitpicks regarding `private` and `readonly` and formatting xaml things on the way to merging this to master. Great job! ✨

PR #293, opened 2016-05-05
Polish expander styles

donokuda:
Destination Branch: `276-collapse-clone-groups` (https://github.com/github/VisualStudio/pull/288)

This pull request give a little UI love to the collapsed clone groups. 

**Before**

image

**After**

image

My initial mock had the icons and avatars aligned with each other:

image

However, I liked how the indented icons are consistent with the repo list in the Team Explorer pane:

image

That said, I can align them up if we think it'll look better 😄 

In https://github.com/github/VisualStudio/pull/288, @grokys mentioned:

> Only problem with this is that you can see an initial flash of expanded
> items when loading with multiple groups, however I felt this was
> preferable to everything being collapsed until the whole list is
> loaded...

We about using an animated transition when collapsing and expanding each organization to help with the initial flash of expanded items. That's not included in this PR but would be my next focus!

/cc @shana @simurai @grokys for 👀  and 💭

### DIFF
--- a/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml
+++ b/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCloneControl.xaml
@@ -46,7 +46,68 @@
         <Setter Property="Margin"
                 Value="0" />
       </Style>
-    </DockPanel.Resources>
+
+            <Style x:Key="ExpanderDownHeaderStyle" TargetType="{x:Type ToggleButton}">
+                <Setter Property="Template">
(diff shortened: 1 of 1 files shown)


shana:
I'm not overly worried about the initial flash, since it's not a dialog you see all that often... how bad is this flash? It's probably animated gif time!

shana:
I can barely see the flash the first time it loads, and not at all on subsequent times, tbh.

grokys:
Looks great @donokuda. I'm not worried about the initial flash if nobody else is!

/nogif

PR #294, opened 2016-05-06
WIP: Finishing off gist support.

grokys:
This PR will track the work I"m doing to finish off the gist support added by #194:

- [x] Add error reporting to the "Publish Gist" dialog
- [ ] Check that the user has permission to publish gists and if not prompt/help them to log out and back in
- [ ] Work out why menu sometimes doesn't appear until another GitHub dialog is shown
  * This is because we need a git repo before the plugin is loaded. Investigate loading earlier to get gist support even without a repo.
- [ ] Polish UI
- [ ] Finalize error messages
- [ ] Merge to master


grokys:
@shana was wondering you you could give this some :eyes: to see if I'm on the right track here with detecting the scopes.

src/GitHub.App/Services/GistPublishService.cs
```diff
@@ -0,0 +1,32 @@
...
+using Octokit;
+
+namespace GitHub.Services
+{
+    [Export(typeof(IGistPublishService))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class GistPublishService : IGistPublishService
+    {
+        [ImportingConstructor]
+        public GistPublishService()
```
shana: FYI, with MEF, you don't need the `ImportingConstructor` attribute if there's an empty constructor (or none at all). You only need it if all constructors require parameters or if you don't want the empty constructor to be the one that gets called when MEF instantiates an instance.

src/GitHub.App/Api/ApiClient.cs
```diff
@@ -52,11 +55,38 @@ public IObservable<Repository> CreateRepository(NewRepository repository, string
...
         {
             return gitHubClient.User.Current();
         }
 
+        public IObservable<string[]> GetScopes()
+        {
+            return GetScopesInternal().ToObservable();
+        }
+
+        private async Task<string[]> GetScopesInternal()
```
shana: Private methods don't get marked `private`, by convention.

src/GitHub.App/Services/GistPublishService.cs
grokys: Ok, yep - this was just copied from another service without considering that. Will remove.

shana:
@grokys Yup, looks sane! :+1:

PR #298, opened 2016-05-10
Replace hardcoded fontsizes with proportional text styles

donokuda:
Destination branch: `feature/pr/views`

This PR replaces hardcoded fontsizes with text styles that allow proportional scaling.

See "Formatting (scaling/bolding) reference" in https://msdn.microsoft.com/en-us/library/mt162316.aspx for more info.

Couple of related changes include:

- Remove hardcoded heights
- Remove horizontal scrolling in list
- Truncate long commit messages and timestamps
- Stub tooltips for files changed tab section

--------------

Here's what the views look like (they should still look the same):

pr list

commits

popup

And here's what happens when you change your environment size to 12 and :sparkles: CoMiC SaNs :sparkles: !

comic-pr-list

comic-commits

### DIFF
--- a/src/GitHub.UI/Assets/Controls/FilterTextBox.xaml
+++ b/src/GitHub.UI/Assets/Controls/FilterTextBox.xaml
@@ -13,7 +13,6 @@
         <Setter Property="AllowDrop" Value="true"/>
         <Setter Property="ScrollViewer.PanningMode" Value="VerticalFirst"/>
         <Setter Property="Stylus.IsFlicksEnabled" Value="False"/>
-        <Setter Property="Height" Value="24" />
         <Setter Property="Padding" Value="3,3,18,3" />
         <Setter Property="ContextMenu" Value="{DynamicResource DefaultContextMenu}" />
         <Setter Property="Template">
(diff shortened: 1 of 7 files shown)


grokys:
Excellent, I was wondering how this should be done. This approach should be used instead of the hardcoded size in c754e07762a8d55f77e7765db3eb0df541a9e3a1 too right?

donokuda:
@grokys Yup! I think

```
FontFamily="Segoe UI"
FontSize="12"
```

could be changed to

```xaml
FontFamily="{DynamicResource VsFont.EnvironmentFontFamily}"
FontSize="{DynamicResource VsFont.EnvironmentFontSize}"
```

Although I wonder if this is the equivalent to removing those two lines in the first place. What do you think?

grokys:
Yeah, you might be right. Would be worth trying removing them and changing the environment font size.

donokuda:
@grokys Tried applying those changes and it feels like the equivalent of reverting your commit (c754e07). Considering that commit is primarily for the login modal, I don't think there is anything we need to change unless we want to make those modals responsive to user settings.

shana:
This looks great! Well, you know, besides the fact that it's Comic Sans o_O 

The WindowController needs hardcoded sizes because that UI is not meant to follow VS environment fonts and the window won't resize properly if font sizes change.

shana:
I'm so sorry for the delay in reviewing this! I'll get better at this, promise!

Haacked:
> This looks great! Well, you know, besides the fact that it's Comic Sans o_O

Comic Sans is underrated!

PR #299, opened 2016-05-13
Finished off gist support

grokys:
This PR will track the work I"m doing to finish off the gist support added by #194:

- [x] Add error reporting to the "Publish Gist" dialog
- [x] Check that the user has permission to publish gists and if not prompt/help them to log out and back in
- [x] Work out why menu sometimes doesn't appear until another GitHub dialog is shown
  * This is because we need a git repo before the plugin is loaded. Investigate loading earlier to get gist support even without a repo.
- [x] Polish UI
- [ ] Finalize error messages
- [x] Open gist in browser
- [x] Merge with master


src/GitHub.App/Controllers/UIController.cs
```diff
@@ -254,7 +262,8 @@ public void Start([AllowNull] IConnection connection)
                             .PermitIf(Trigger.Clone, UIViewType.Login, () => !host.IsLoggedIn)
                             .PermitIf(Trigger.Publish, UIViewType.Publish, () => host.IsLoggedIn)
                             .PermitIf(Trigger.Publish, UIViewType.Login, () => !host.IsLoggedIn)
-                            .PermitIf(Trigger.Gist, UIViewType.Gist, () => host.IsLoggedIn)
+                            .PermitIf(Trigger.Gist, UIViewType.Gist, () => host.IsLoggedIn && host.SupportsGist)
+                            .PermitIf(Trigger.Gist, UIViewType.LogoutRequired, () => host.IsLoggedIn && !host.SupportsGist)
```
shana: You don't need this to be evaluated after the first time it runs, right? If you do, there's `PermitDynamic` and `PermitDynamicIf`, which allows you to chose the UIViewType to load every time it gets called (just fyi)

src/GitHub.App/ViewModels/LogoutRequiredViewModel.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            if (repositoryHosts.GitHubHost != null)
+            {
+                await repositoryHosts.GitHubHost.LogOut();
+            }
+
+            if (repositoryHosts.EnterpriseHost != null)
+            {
+                await repositoryHosts.EnterpriseHost.LogOut();
+            }
```
shana: We should check which of the hosts has the wrong scope and only log out that one.

shana:
Looking good!

src/GitHub.App/Controllers/UIController.cs
grokys: This does already seem to be called every time "Create a Gist" is clicked.

src/GitHub.App/ViewModels/LogoutRequiredViewModel.cs
grokys: :+1:

src/GitHub.App/Api/ApiClient.cs
```diff
@@ -62,6 +67,60 @@ public IObservable<User> GetUser()
...
+            return GetScopesInternal().ToObservable();
+        }
+
+        async Task<string[]> GetScopesInternal()
+        {
+            // If auth type is Basic we might be able to read /api/authorizations to get the 
+            // current scopes. However this request sometimes gets mysteriously converted to an
+            // OAuth request so if that doesn't work try reading from / and checking for the 
+            // X-OAuth-Scopes header.
+            var connection = gitHubClient.Connection;
```
shana: There's no point in doing this, we never have access to the actual user/password once login is completed. The password gets swapped with a token at that point.

shana:
Ok, I've updated this and the gist-support to track the new code from the PR feature branch (which is now in master, btw). So now this is up to date, only the changes that are relevant are here plus the new UIController code, and we're 👍 on finishing this work!

shana:
@donokuda Can you work your magic on the "Logout required" UI? It currently looks like this:

screen shot 2016-05-24 at 10 04 57 pm

shana:
@grokys One thing I noticed is that when I click on the create button to create a gist, the dialog closes and there's no feedback. We should probably copy the link to the clipboard and open a webbrowser with the link, is that something that's supposed to be happening already (and I broke the UIController code somehow so it's not...)?

grokys:
Nope, AFAIK that never happened. Will add that to TODO.

donokuda:
> @donokuda Can you work your magic on the "Logout required" UI?

Yup! Working on that right now.

donokuda:
PR for polishing the dialog UI: https://github.com/github/VisualStudio/pull/332

shana:
Is this still a WIP?

grokys:
No, I think this is ready for review!

shana:
Ok, I'll do another pass. Don't forget to remote the WIP label and title and assign it to reviewers (and we now have a Ready for Review label, that helps too)!

grokys:
Oh! I remember why it was still WIP: #336

paladique:
:+1:

shana:
woot!

PR #312, opened 2016-05-19
Allow clicking password box when it is in error

### LINKED ISSUE #305: Cannot click on password box when it is displaying error
Steps to reproduce:

- Login to GitHub in VS
- Enter an invalid password, wait for the `(!)` icon to show up
- Click on the password box - doesn't set focus

I assume this is because the error adorner is handling the click.

grokys:
Fixes #305.


shana:
Awesome one-liner is awesome ^_^

PR #313, opened 2016-05-19
Only display error adorner after user interaction.

### LINKED ISSUE #307: Username error icon displays by default
Steps to reproduce: 
1. Open visual studio
2. Go to connect page
3. Click on Github -> Connect

Expected result: Username error icon only displays after user tabs away from field or enters an invalid value. 
Actual result: Error displays immediately when it opens. Tabbing away from the field clears the icon. 

screen shot 2016-05-18 at 3 24 48 pm 1

grokys:
Fixes #307.

### DIFF
--- a/src/GitHub.UI.Reactive/Controls/Validation/ValidationMessage.cs
+++ b/src/GitHub.UI.Reactive/Controls/Validation/ValidationMessage.cs
@@ -17,6 +17,7 @@ namespace GitHub.UI
     public class ValidationMessage : UserControl
     {
         const double defaultTextChangeThrottle = 0.2;
+        bool userHasInteracted;
 
         public ValidationMessage()
         {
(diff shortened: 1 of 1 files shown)


src/GitHub.UI.Reactive/Controls/Validation/ValidationMessage.cs
```diff
@@ -33,7 +34,7 @@ public ValidationMessage()
                 .Do(CreateBinding)
                 .Select(control =>
                     Observable.Merge(
-                        this.WhenAnyValue(x => x.ShowError),
+                        this.WhenAnyValue(x => x.ShowError).Where(x => userHasInteracted),
```
shana: So I'm curious, why is ShowError signaled here has having been changed if on line 28 it's only set when the ValidationResult is not null? Is `WhenAnyValue` always triggered immediately on setup, and is there a way of having it not triggered until something actually changes in the property?

grokys: Yes, `WhenAnyValue` is triggered immediately. The way of not making it triggered is to use `.WhenAny(...).Skip(1)` but that didn't work here because you get a load of different values pushed, which start off `false` and then become `true`. 

So rather than deal with this, and because the intention is to only show user errors after user interaction, i decided that the best way to do it was explicitly by setting a flag on user interaction.

shana:
✨

PR #314, opened 2016-05-19
Don't call cancel.OnCompleted after disposal.

### LINKED ISSUE #311: Cancelling login dialog after failed login causes ObjectDisposedException
This is a bug that was already fixed on the gists branch caused by `cancel.OnCompleted` being called after the `cancel` subject has already been disposed. We should fix it on master too.

grokys:
`cancel.OnNext` causes `cancel` to get disposed, so don't try calling
`OnCompleted` on it afterwards. Fixes #311.


shana:
Yup. Sigh....

PR #316, opened 2016-05-20
Use Merge instead of Amb on login control.

### LINKED ISSUE #309: Successful enterprise login doesn't close signup modal if error on other tab
Steps to reproduce: 
1. Go to GHfVS signin screen
2. On the Github tab, produce an incorrect password
3. On Enterprise tab, sign in 

Expected result: Signin modal closes after any successful login. 
Actual result: Signin modal stays open if there's an error on github tab.

grokys:
Fixes #309. `Observable.Amb` "emits all of the items from only the firstof these Observables to emit an item or notification". This means that if the user tries to login first to .com with an invalid login, then
tries to login to enterprise with a valid login the login from enterprise will be ignored because the .com observable fired first.Instead use `Observable.Merge` which will emit authentication results from both in any order.

### DIFF
--- a/src/GitHub.App/ViewModels/LoginControlViewModel.cs
+++ b/src/GitHub.App/ViewModels/LoginControlViewModel.cs
@@ -45,7 +45,7 @@ namespace GitHub.ViewModels
 
                 }).ToProperty(this, x => x.LoginMode);
 
-            AuthenticationResults = Observable.Amb(
+            AuthenticationResults = Observable.Merge(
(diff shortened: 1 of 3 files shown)


shana:
Can we put together a test for this somehow?

grokys:
Hmm yeah, we should I guess! Will work on it.

grokys:
Added a unit test.

shana:
:heart_eyes_cat:

PR #320, opened 2016-05-23
Release 1.0.18.4

shana:
To be merged when it's out.


shana:
No merging this one, master already has a version higher than this. This is now the old stable release branch, and if we need more bugfixes on the 1.0 series, they'll go here.

PR #329, opened 2016-05-26
Added tooltips to GitHub pane toolbar buttons.

### LINKED ISSUE #322: Toobar icons in PR view have blank tooltips
image

grokys:
Fixes #322 .


shana:
So, haha, I'm wondering how we're going to localize these...

shana:
Oh wow, we actually need to have separate vsct files per language (https://msdn.microsoft.com/en-us/library/ee943168.aspx). Right, hmmm, let's skip that for now 😝

Issue #303, opened 2016-05-17
SharedDictionaryManager in GitHub.VisualStudio.Helpers should be removed


grokys:
It is now a link to `..\common\SharedDictionaryManager.cs` so  the version in `GitHub.VisualStudio\Helpers` should be removed.

PR #330, opened 2016-05-26
Remove SharedDictionaryManager

paladique:
For #303  This has been moved to `src/common/` folder, but we'll want to keep the reference to the file in the `Github.VisualStudio` project.


shana:
✨ 🚀

PR #331, opened 2016-05-26
Read scopes at the same time as User.

grokys:
Also remove what looked like incorrect logic in that if login failed, the login was read from cache.

### DIFF
--- a/src/GitHub.App/Api/ApiClient.cs
+++ b/src/GitHub.App/Api/ApiClient.cs
@@ -23,11 +23,10 @@ namespace GitHub.Api
 {
     public partial class ApiClient : IApiClient
     {
-        const string scopesHeader = "X-OAuth-Scopes";
+        const string ScopesHeader = "X-OAuth-Scopes";
+        const string ProductName = Info.ApplicationInfo.ApplicationDescription;
         static readonly Logger log = LogManager.GetCurrentClassLogger();
         static readonly Uri userEndpoint = new Uri("user", UriKind.Relative);
-
-        const string ProductName = Info.ApplicationInfo.ApplicationDescription;
 
         readonly IObservableGitHubClient gitHubClient;
         // There are two sets of authorization scopes, old and new:
@@ -75,16 +74,16 @@ namespace GitHub.Api
                 userEndpoint, null, null).ConfigureAwait(false);
             var scopes = default(string[]);
 
-            if (response.HttpResponse.Headers.ContainsKey(scopesHeader))
+            if (response.HttpResponse.Headers.ContainsKey(ScopesHeader))
             {
-                scopes = response.HttpResponse.Headers[scopesHeader]
+                scopes = response.HttpResponse.Headers[ScopesHeader]
                     .Split(',')
                     .Select(x => x.Trim())
                     .ToArray();
             }
             else
             {
-                log.Error($"Error reading scopes: /user succeeded but {scopesHeader} was not present.");
+                log.Error($"Error reading scopes: /user succeeded but {ScopesHeader} was not present.");
             }
 
             return new UserAndScopes(response.Body, scopes);
--- a/src/GitHub.App/Api/ApiClient.cs
+++ b/src/GitHub.App/Api/ApiClient.cs
@@ -17,6 +17,7 @@ using System.Threading.Tasks;
 using System.Reactive.Threading.Tasks;
 using Octokit.Internal;
 using System.Collections.Generic;
+using GitHub.Models;
 
 namespace GitHub.Api
 {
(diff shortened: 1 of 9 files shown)


src/GitHub.App/Services/ModelService.cs
```diff
@@ -97,7 +97,7 @@ IObservable<IEnumerable<string>> GetOrderedGitIgnoreTemplatesFromApi()
         IObservable<IEnumerable<AccountCacheItem>> GetUser()
         {
             return hostCache.GetAndRefreshObject("user",
-                () => apiClient.GetUser().Select(AccountCacheItem.Create), TimeSpan.FromMinutes(5), TimeSpan.FromDays(7))
+                () => apiClient.GetUser().Select(x => new AccountCacheItem(x.User)), TimeSpan.FromMinutes(5), TimeSpan.FromDays(7))
```
shana: Just to keep with the convention of using `Create` methods on these calls, add an overload to `AccountCacheItem.Create` that takes a `UserAndScopes` object instead.

src/GitHub.App/Models/RepositoryHost.cs
```diff
@@ -262,35 +253,21 @@ IObservable<AuthenticationResult> LoginWithApiUser(AccountCacheItem user, string
                     if (result.IsSuccess())
                     {
                         IsLoggedIn = true;
-                        SupportsGist = scopes?.Contains("gist") ?? false;
+                        SupportsGist = userAndScopes.Scopes?.Contains("gist") ?? true;
                     }
 
                     log.Info("Log in from cache for login '{0}' to host '{1}' {2}",
-                        user != null ? user.Login : "(null)",
+                        userAndScopes.User != null ? userAndScopes.User.Login : "(null)",
```
shana: Null checks here, `userAndScopes` can be null

src/GitHub.App/Api/ApiClient.cs
```diff
@@ -62,44 +63,30 @@ public IObservable<Gist> CreateGist(NewGist newGist)
...
+        async Task<UserAndScopes> GetUserInternal()
         {
-            return GetScopesInternal().ToObservable();
-        }
-
-        async Task<string[]> GetScopesInternal()
-        {
-            var connection = gitHubClient.Connection;
+            var response = await gitHubClient.Connection.Get<User>(
+                new Uri("user", UriKind.Relative), null, null);
```
shana: This Uri should be a static readonly field (`userEndpoint`). Also, call `ConfigureAwait(false)` on this, it doesn't have to return to the caller thread (same pattern that the original Octokit call has)

src/UnitTests/GitHub.App/Models/RepositoryHostTests.cs
```diff
@@ -161,6 +161,66 @@ public class TheLoginMethod : TestBaseClass
...
+            Assert.False(host.SupportsGist);
+        }
+
+        [Fact]
+        public async Task SupportsGistIsTrueWhenScopesAreNull()
+        {
+            // TODO: Check assumptions here. From my conversation with @shana it seems that the first login
+            // will be done with basic auth and from then on a token will be used. So if it's the first login,
+            // it's from this version and so gists will be supported. However I've been unable to repro this
+            // behavior.
```
shana: Is this comment still valid?

shana:
I debugged this some and yeah, logic looks about right. The whole login thing should definitely be cleaned up, it's waaay too complicated, but no point doing that here.

src/UnitTests/GitHub.App/Models/RepositoryHostTests.cs
grokys: Yeah, I added this comment with this commit because I wasn't able to repro the case where a response comes back without the OAuth headers, though I'm pretty certain I've seen it in the past, because that's why I introduced the complexity a while back where I checked whether we're using basic auth.

src/GitHub.App/Api/ApiClient.cs
```diff
@@ -25,6 +25,7 @@ public partial class ApiClient : IApiClient
     {
         const string scopesHeader = "X-OAuth-Scopes";
         static readonly Logger log = LogManager.GetCurrentClassLogger();
+        static readonly Uri userEndpoint = new Uri("user", UriKind.Relative);
 
         const string ProductName = Info.ApplicationInfo.ApplicationDescription;
```
shana: Could you reorder these so that const string and static readonly definitions are grouped together?

Nitpick Nancy on the prowl
nitpick_nancy

src/UnitTests/GitHub.App/Models/RepositoryHostTests.cs
shana: Ok, we'll do more testing and figure this one out, but it's not a blocker to merge this.

shana:
bb8_thumbs_up

PR #332, opened 2016-05-26
Polish sign out required dialog

donokuda:
Destination Branch: `grokys/feature/gist-support`

Before:

BEFOREEEEEEEEE

After:

screen shot 2016-05-26 at 11 30 21 am

cc @niik and @simurai for 👀

### DIFF
--- a/src/GitHub.App/Controllers/UIController.cs
+++ b/src/GitHub.App/Controllers/UIController.cs
@@ -259,12 +259,12 @@ namespace GitHub.Controllers
 
         public void Jump(ViewWithData where)
         {
-            Debug.Assert(where.Flow == mainFlow, "Jump called for flow " + where.Flow + " but this is " + mainFlow);
-            if (where.Flow != mainFlow)
+            Debug.Assert(where.ActiveFlow == mainFlow, "Jump called for flow " + where.ActiveFlow + " but this is " + mainFlow);
+            if (where.ActiveFlow != mainFlow)
                 return;
 
             requestedTarget = where;
-            if (activeFlow == where.Flow)
+            if (activeFlow == where.ActiveFlow)
                 Fire(Trigger.Next, where);
         }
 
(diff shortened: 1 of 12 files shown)


shana:
This looks great! At this point, I'm only wondering if the message should be tweaked or not. Is it confusing, not clear enough, will they be scared of doing this?

@Haacked @pmn @paladique thoughts?

Haacked:
What does "this feature" refer to?

shana:
@Haacked It's the "creating a gist" feature. Sounds like we should probably be explicit here and say "To create a Gist, you need to sign in again"?

Haacked:
I think being explicit is good here. I mean, for most users they'll understand  the context. But I could imagine invoking the operation, switching to another task, coming back and not remembering what "this feature" refers to.

paladique:
:+1: to being more explicit, and specifically for wording also agree with replacing "feature" with "creating a gist"

simurai:
>  you need to sign in again.

It might be confusing to say they have to "sign in" but then the button says "Sign out". How about:

`To create a Gist, you need to sign out and then sign in again.`

Although that sounds a bit cumbersome. 🙉

shana:
> It might be confusing to say they have to "sign in" but then the button says "Sign out".

Yeah, I was finding that discrepancy a bit jarring too. Maybe `To create a Gist, you need to sign out and in again.`?

grokys:
Have we come to a conclusion on this? Once this is decided, I think we're ready to merge gists into master.

grokys:
"To create a Gist you need to sign out and back in"?

shana:
I've done a few tweaks to this so that the message and the icon can be set dynamically depending on which feature actually ends up in this dialog. If we ever need to add more scopes or force log outs for some reason, we can just reuse this.

shana:
How it looks now:

screen shot 2016-05-27 at 2 48 49 pm

PR #334, opened 2016-05-27
Set proper visibility for non GitHub Repos

### LINKED ISSUE #301: GitHub context menus show up when in non-github repos
Reported by @acangialosi

When working on a project hosted on VSTS, GitHub Create/Copy link context menus are showing up, and they shouldn't be.

paladique:
Fixes #301 

Seems that the problem points to how we're checking the url in `IsGitHubRepo()` 
I'm thinking we should add something to check that the clone url has "github.com" in it

### DIFF
--- a/src/GitHub.Exports/Services/IMenuHandler.cs
+++ b/src/GitHub.Exports/Services/IMenuHandler.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Threading.Tasks;
 
 namespace GitHub.VisualStudio
 {
--- a/src/GitHub.Exports/Services/IMenuHandler.cs
+++ b/src/GitHub.Exports/Services/IMenuHandler.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Threading.Tasks;
 
 namespace GitHub.VisualStudio
 {
--- a/src/GitHub.Extensions/UriExtensions.cs
+++ b/src/GitHub.Extensions/UriExtensions.cs
@@ -25,6 +25,11 @@ namespace GitHub.Extensions
             return uri.Scheme == "http" || uri.Scheme == "https";
         }
 
+        public static bool IsGithubUri(this Uri uri)
+        {
+            return uri.OriginalString.Contains("github.com");
+        }
+
         public static bool IsSameHost(this Uri uri, Uri compareUri)
         {
             return uri.Host.Equals(compareUri.Host, StringComparison.OrdinalIgnoreCase);
--- a/src/GitHub.Extensions/UriExtensions.cs
+++ b/src/GitHub.Extensions/UriExtensions.cs
@@ -25,11 +25,6 @@ namespace GitHub.Extensions
             return uri.Scheme == "http" || uri.Scheme == "https";
         }
 
-        public static bool IsGithubUri(this Uri uri)
-        {
-            return uri.OriginalString.Contains("github.com");
-        }
-
         public static bool IsSameHost(this Uri uri, Uri compareUri)
         {
             return uri.Host.Equals(compareUri.Host, StringComparison.OrdinalIgnoreCase);
--- a/src/GitHub.VisualStudio/Base/MenuBase.cs
+++ b/src/GitHub.VisualStudio/Base/MenuBase.cs
@@ -85,7 +85,6 @@ namespace GitHub.VisualStudio
         {
             RefreshRepo();
             return ActiveRepo?.CloneUrl?.RepositoryName != null && ActiveRepo?.CloneUrl?.IsGithubUri == true;
-
         }
     }
 }
--- a/src/GitHub.VisualStudio/Base/MenuBase.cs
+++ b/src/GitHub.VisualStudio/Base/MenuBase.cs
@@ -8,9 +8,8 @@ using GitHub.Primitives;
 using GitHub.Services;
 using System.Threading.Tasks;
 using GitHub.Api;
-using GitHub.Primitives;
 using NullGuard;
-using System.Diagnostics;
+using GitHub.UI;
 
 namespace GitHub.VisualStudio
 {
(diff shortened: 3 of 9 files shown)


src/GitHub.Extensions/UriExtensions.cs
```diff
@@ -25,6 +25,11 @@ public static bool IsHypertextTransferProtocol(this Uri uri)
             return uri.Scheme == "http" || uri.Scheme == "https";
         }
 
+        public static bool IsGithubUri(this Uri uri)
+        {
+            return uri.OriginalString.Contains("github.com");
```
paladique: `String.Contains` is case and culture insensitive, should I should change this` .ToLower()`?
Also I wonder if using `String.StartsWith(https://www.github.com)` is a better alternative?

shana:
Oh yes, you're right, this is a bug, darn it.

There is a method in `TeamExplorerItemBase` called `IsAGitHubRepo()`, which does all the required checks to ensure that the repo is, in fact, a github repo. Those tests need to be done by the menus as well.

src/GitHub.VisualStudio/Menus/CopyLink.cs
```diff
@@ -45,7 +48,11 @@ public void Activate([AllowNull]object data = null)
 
         public bool CanShow()
         {
-            return IsGitHubRepo();
+            return System.Threading.Tasks.Task.Run(() =>
+            {
+                return IsGitHubRepo(); 
+            }).Result;
```
shana: This will block the UI for as much time as it takes the task to run, which means it will block the context menu from showing up for that amount of time. You only want to call `Result` on a task when you know the task is done, otherwise there's a risk of deadlocks and visible blocking. You also really don't need to run a task here, `IsGitHubRepo` is already a task.

The best way to do this is to call `Wait` on the task returned by `IsGitHubRepo`, with a time of around 250 milliseconds, which is a delay low enough to not annoy users when they right click, but high enough that if there's no need for internet access, it will complete under that time (for github.com repos).

If `Wait` returns true, then the task is done and you can return the `Result` safely, otherwise it's not done yet and the 250 milliseconds have already elapsed so we don't know if it's one of our repos or not. In this second case we just return false (in case it's not one of our repos), and the next time we evaluate it, `SimpleApi` will have completed and cached the results, so there won't be any delay.

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -4,6 +4,7 @@
 using System.IO;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using SysTask = System.Threading.Tasks;
```
shana: Is this needed?

src/GitHub.VisualStudio/Base/MenuBase.cs
```diff
@@ -3,23 +3,47 @@
...
         readonly IServiceProvider serviceProvider;
+        readonly ISimpleApiClientFactory apiFactory;
+
         protected IServiceProvider ServiceProvider { get { return serviceProvider; } }
 
         protected ISimpleRepositoryModel ActiveRepo { get; private set; }
 
+        ISimpleApiClient simpleApiClient;
+        [AllowNull]
+        public ISimpleApiClient SimpleApiClient
```
shana: Make this `protected`, it's not public API.

src/GitHub.Exports/Services/IMenuHandler.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Threading.Tasks;
```
shana: This isn't needed.

src/GitHub.VisualStudio/Menus/CopyLink.cs
```diff
@@ -45,7 +48,8 @@ public void Activate([AllowNull]object data = null)
 
         public bool CanShow()
         {
-            return IsGitHubRepo();
+            return IsGitHubRepo().Wait(250);
```
shana: Ah, I see what you're trying to do there, but that's not going to return quite the result you expect if it's not a github repo... Consider:

- User has repo with remote set to `http://a-server-that-isn-t-github/user/repo`
- First time `IsGitHubRepo` is called and detects it's not `github.com`, so it checks if this server has been probed already. It hasn't, so it checks if the server returns a github header and caches the result, which is `false`
- 250ms have elapsed before the check returns, so `Wait` returns false and the menu is hidden
- Second time `IsGitHubRepo` is called and detects it's not `github.com`, so it checks if this server has been probed already. It has, so it returns the probe result, which is `false`.
- 250ms have not elapsed, so `Wait` returns `true` and the menu is shown. Ooops!

paladique: Just looked up the return value `Wait()`, definitely not what I was looking for!

paladique:
Since `Wait(250)` returns a bool on whether the task is complete or not, I put this into a conditional operator to condense this into one line. 

So, `IsGitHubRepo` returns false if the task is complete, or the task is complete but is not a GitHub Repo. This can only return true if the task is complete and is a GitHub repo.

src/GitHub.VisualStudio/Menus/CopyLink.cs
```diff
@@ -45,7 +48,9 @@ public void Activate([AllowNull]object data = null)
 
         public bool CanShow()
         {
-            return IsGitHubRepo();
+            Task<bool> githubRepoCheckTask = IsGitHubRepo();
+            return githubRepoCheckTask.Wait(250) ? githubRepoCheckTask.Result : false;
```
shana: 😍

src/GitHub.VisualStudio/Menus/CopyLink.cs
```diff
@@ -45,7 +48,9 @@ public void Activate([AllowNull]object data = null)
 
         public bool CanShow()
         {
-            return IsGitHubRepo();
+            Task<bool> githubRepoCheckTask = IsGitHubRepo();
```
shana: You can make this a `var`, the variable name makes it clear it's a task and the method name makes it clear it's a bool.

shana:
This is great! I just had a little comment, otherwise we're good to go here ^_^

paladique:
:raised_hands: gonna deal with the conflicts now

shana:
Purrfect! I'm going to leave the branch open because we'll want to cherry-pick these into the backport release branch, easier to do it from the branch.

rejoice

PR #337, opened 2016-05-30
Removed DefaultRepositoryName

### LINKED ISSUE #279: Publish button disabled until user clicks repository name
Steps to reproduce: 
`1.` Create new local repository
`2.` Go to publish (double click repository, sync, publish to github) 

screen shot 2016-05-02 at 4 54 14 pm

`3.` Click repository name field, then away from it (notice button is now enabled)

Expected result: Publish button available by default. 
Actual result: Publish button is disabled until user clicks repository name field.

grokys:
From `IRepositoryPublishViewModel`. Seems to not be needed and was causing #279.

The default repository name was being placed directly into `RepositoryPublishControl.nameText` in the control which meant:

- it was only written to `RepositoryName` when the user clicked on the `nameText` control
- which meant that `RepositoryName` was up until that point null
- which meant that the validation for it was failing
- which meant the "Publish" button was disabled.

This commit just sets the `RepositoryName` to the default in the VM ctor meaning that that name gets immediately validated. `DefaultRepositoryName` didn't seem to be used anywhere else, so I removed it.

cc: @Haacked for a sanity check because it looks like you wrote this.

Fixes #279

### DIFF
--- a/src/GitHub.App/SampleData/SampleViewModels.cs
+++ b/src/GitHub.App/SampleData/SampleViewModels.cs
@@ -232,14 +232,6 @@ namespace GitHub.SampleData
             SelectedConnection = Connections[0];
         }
 
-        public string DefaultRepositoryName
-        {
-            get
-            {
-                return "whatever";
-            }
-        }
-
         public bool IsHostComboBoxVisible
         {
             get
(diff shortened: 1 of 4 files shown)


shana:
Yup, this looks very sane! :shipit:

PR #338, opened 2016-05-30
Enable "Cancel" button in PR creation view

### LINKED ISSUE #325: Cancel button in "Create new PR" view does nothing
In addition, if you click the "Back" button to exit it, and then click "Create New" again, nothing happens here either.

grokys:
And allow this view to be invoked >1 time. Fixes #325.

### DIFF
--- a/src/GitHub.App/SampleData/SampleViewModels.cs
+++ b/src/GitHub.App/SampleData/SampleViewModels.cs
@@ -22,19 +22,7 @@ using System.Reactive.Linq;
 namespace GitHub.SampleData
 {
     [ExcludeFromCodeCoverage]
-    public class BaseViewModelDesigner : ReactiveObject, IViewModel
-    {
-        public ICommand Cancel { get; set; }
-        public ReactiveCommand<object> CancelCommand { get; protected set; }
-        public bool IsShowing { get; set; }
-        public string Title { get; set; }
-
-        public void Initialize(ViewWithData data)
-        { }
-    }
-
-    [ExcludeFromCodeCoverage]
-    public class RepositoryCreationViewModelDesigner : BaseViewModelDesigner, IRepositoryCreationViewModel
+    public class RepositoryCreationViewModelDesigner : BaseViewModel, IRepositoryCreationViewModel
     {
         public RepositoryCreationViewModelDesigner()
         {
@@ -369,7 +357,7 @@ namespace GitHub.SampleData
         public void Refresh() { }
     }
 
-    public class RepositoryCloneViewModelDesigner : BaseViewModelDesigner, IRepositoryCloneViewModel
+    public class RepositoryCloneViewModelDesigner : BaseViewModel, IRepositoryCloneViewModel
     {
         public RepositoryCloneViewModelDesigner()
         {
--- a/src/GitHub.App/SampleData/SampleViewModels.cs
+++ b/src/GitHub.App/SampleData/SampleViewModels.cs
@@ -25,6 +25,7 @@ namespace GitHub.SampleData
     public class BaseViewModelDesigner : ReactiveObject, IViewModel
     {
         public ICommand Cancel { get; set; }
+        public ReactiveCommand<object> CancelCommand { get; protected set; }
         public bool IsShowing { get; set; }
         public string Title { get; set; }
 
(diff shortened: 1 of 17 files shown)


src/GitHub.App/SampleData/SampleViewModels.cs
```diff
@@ -25,6 +25,7 @@ namespace GitHub.SampleData
     public class BaseViewModelDesigner : ReactiveObject, IViewModel
     {
         public ICommand Cancel { get; set; }
+        public ReactiveCommand<object> CancelCommand { get; protected set; }
```
shana: Let's kill everything in this designer and inherit it from `BaseViewModel` instead, which already has everything defined.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -7,12 +7,17 @@
...
 
 namespace GitHub.ViewModels
 {
     [ExportViewModel(ViewType = UIViewType.PRCreation)]
     [PartCreationPolicy(CreationPolicy.NonShared)]
     class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationViewModel
     {
+        public PullRequestCreationViewModel()
+        {
+            CancelCommand = ReactiveCommand.Create();
```
shana: Let's move this to the `BaseViewModel` ctor and dedupe some code along the way (and avoid this problem in the future)

src/GitHub.Exports.Reactive/ViewModels/IPullRequestCreationViewModel.cs
```diff
@@ -1,7 +1,10 @@
-﻿namespace GitHub.ViewModels
+﻿using ReactiveUI;
+
+namespace GitHub.ViewModels
 {
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1040:AvoidEmptyInterfaces")]
     public interface IPullRequestCreationViewModel : IViewModel
     {
+        ReactiveCommand<object> CancelCommand { get; }
```
shana: Seeing as this is going to be more useful in the future, how about we put this in its own interface (something like `IReactiveViewModel` that inherits from `IViewModel`) and make `BaseViewModel` implement that? We can add an accessor for it in `SimpleViewUserControl`.

src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml.cs
```diff
@@ -20,6 +21,8 @@ public PullRequestCreationView()
 
             this.WhenActivated(d =>
             {
+                d(this.BindCommand(ViewModel, vm => vm.Cancel, v => v.cancelButton));
```
shana: We don't need to do this here, just add a `Command="{Binding Cancel}"` in the xaml and it'll automagically work.

src/GitHub.App/SampleData/SampleViewModels.cs
grokys: So is `BaseViewModelDesigner` not needed any more?

src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml.cs
grokys: Ok, so are we Officially going to prefer XAML bindings over ReactiveUI bindings from here on? This was like this just because it's what the other views do.

src/GitHub.App/SampleData/SampleViewModels.cs
shana: Nope, not needed. Probably best to just do a quick search&replace on the thing.

src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml.cs
shana: Yeah, I'm trying to clear away the rxui bindings and use xaml. They were needed before because the view model was being set incorrectly, but not anymore. With xaml bindings, don can set up all the data himself when he mocks things up and we don't have to do extra steps to hook it up. :smile:

shana:
I went ahead and cleared out all the `CancelCommand = ReactiveCommand.Create();` instances that were in the other view models, since we don't need that anymore.

Issue #333, opened 2016-05-26
Issue submission guidelines


paladique:
We have a contributor guide for PRs, but not for Issues. This is in anticipation of issues that might be created in relation to the new features we're currently working on. 

I'm thinking we'd add this on to `contributing.md`, to include how to make a log file, and how screenshots, in addition to those log files, will be helpful in diagnosing problems.

Making Issue templates might work too!
Anything else that would be helpful? cc: @shana @Haacked @grokys 

- [x] Add Issue submission info in `contributing.md` 
- [ ] Create Issue templates

shana:
This is a great idea, +10! We should probably document the areas that the extension is responsible for, to avoid more "git is doing X" issues. Having info on how to make a logfile and screenshots will definitely come in handy, I can never remember all the steps by heart :P

We can have an issue template with a message saying where to find this information, in case people are very eager to hit the "New issue" button without reading.

Haacked:
:100: We should also have a pull request template too. :smile:

PR #341, opened 2016-05-31
Information for Issue Contributions

paladique:
Per #333, this is indented to sync up with issue templates that we make.


shana:
This is great! ✨

PR #348, opened 2016-06-07
Always cancel the whole flow if a subflow is cancelled

### LINKED ISSUE #336: Cancelling logout dialog proceeds to Create Gist
@shana this is because cancelling the `UIViewType.LogoutRequired` calls `Go(Trigger.Cancel)`:

https://github.com/github/VisualStudio/blob/grokys/feature/gist-support/src/GitHub.App/Controllers/UIController.cs#L365

Which ends up at `End(false)`:

https://github.com/github/VisualStudio/blob/grokys/feature/gist-support/src/GitHub.App/Controllers/UIController.cs#L369

`End(false)` calls `Fire(Trigger.Next)`:

https://github.com/github/VisualStudio/blob/grokys/feature/gist-support/src/GitHub.App/Controllers/UIController.cs#L607

Which moves the flow to the first UI of the main flow:

https://github.com/github/VisualStudio/blob/grokys/feature/gist-support/src/GitHub.App/Controllers/UIController.cs#L408

I'm not entirely sure how this all should work so I'm hesitant to change anything here, but I will continue poking around. I suspect that `End(false)` shouldn't call `Fire(Trigger.Next)`?

shana:
Fixes #336


chulito1:
Unsubscribe me please  and cancel

shana:
@chulito1 If you're receiving notifications from this repository, it's because you signed up to them by watching the repository. You need to unsubscribe from it yourself.

PR #349, opened 2016-06-09
Add metrics

grokys:
Adds metrics to the extension.


src/GitHub.App/Services/UsageTracker.cs
```diff
@@ -0,0 +1,378 @@
...
+using System.Reactive.Linq;
+using Akavache;
+using GitHub.Caches;
+using GitHub.Extensions.Reactive;
+using GitHub.Models;
+using GitHub.Settings;
+using ReactiveUI;
+using Rothko;
+
+#pragma warning disable CS0649
```
shana: Do you still need this?

grokys: Nope! Well spotted :eyeglasses:! Will remove.

src/GitHub.App/Services/UsageTracker.cs
```diff
@@ -0,0 +1,378 @@
...
+        public UsageTracker(
+            Lazy<ISharedCache> cache,
+            Lazy<IRepositoryHosts> repositoryHosts,
+            IPackageSettings userSettings,
+            IUIProvider serviceProvider)
+        {
+            this.cache = cache;
+            this.repositoryHosts = repositoryHosts;
+            this.userSettings = userSettings;
+            this.client = (IMetricsService)serviceProvider.GetService(typeof(IMetricsService));
```
shana: This is going to blow up miserably if the service is not found. You need to use `TryGetService<IMetricsService>` here.

src/GitHub.App/Services/UsageTracker.cs
```diff
@@ -0,0 +1,378 @@
...
+#pragma warning disable CS0649
+
+namespace GitHub.Services
+{
+    [Export(typeof(IUsageTracker))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class UsageTracker : IUsageTracker
+    {
+        // Whenever you add a counter make sure it gets added to _both_
+        // BuildUsageModel and ClearCounters
```
shana: This is a sign that we need to refactor the code so that we don't accidentally forget things. I opened #359 to fix this, comments welcome!

src/GitHub.App/Services/UsageTracker.cs
```diff
@@ -0,0 +1,376 @@
...
+        {
+            return Observable.Defer(() =>
+            {
+                try
+                {
+                    // Don't even store data locally if the user opts out.
+                    if (!userSettings.CollectMetrics)
+                    {
+                        return Observable.Empty<int>();
+                    }
```
shana: Nitpick Nancy on the prowl: the codebase doesn't use {} when there's only one line.

src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs
```diff
@@ -126,9 +130,12 @@ IObservable<Unit> OnCloneRepository(object state)
                     notificationService.ShowError(Resources.RepositoryCloneFailedNoSelectedRepo);
                     return Observable.Return(Unit.Default);
                 }
+                
                 // The following is a noop if the directory already exists.
                 operatingSystem.Directory.CreateDirectory(BaseRepositoryPath);
-                return cloneService.CloneRepository(repository.CloneUrl, repository.Name, BaseRepositoryPath);
+
+                return cloneService.CloneRepository(repository.CloneUrl, repository.Name, BaseRepositoryPath)
+                    .Do(_ => this.usageTracker.IncrementCloneCount());
```
shana: This one is a bit tricky... VS does the cloning, not us, so we fire off the operation and exit, which means we don't yet know if cloning was successful here. Do we care?

src/GitHub.App/Services/UsageTracker.cs
grokys: Yeah, this came from the GHfW codebase - the changes in your PR looks good to me.

src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs
grokys: Hmm, not sure. I'm tempted to say "the user decided to use the extension to do a clone, so record it - if it didn't work that's a bug rather than a metric"

shana: True... do we validate if the target clone path already exists? That's the only error that I can see us preventing.

EDIT: Nope, we don't, so users can potentially select a base path to clone into and VS fails because there's already a directory with the same name in there. Feels like something we should fix 😜

src/GitHub.App/Services/UsageTracker.cs
shana: Ah, yes, I forgot this is mostly GHfW code. Cool, feel free to merge it in then!

src/GitHub.App/Services/UsageTracker.cs
```diff
@@ -0,0 +1,374 @@
...
+                {
+                    log.Info("User has enabled sending anonymized usage statistics to GitHub");
+
+                    return client.SendOptIn();
+                });
+            }
+            else
+            {
+                return Observable.Return(Unit.Default);
+            }
```
shana: There's those pesky braces again...

src/GitHub.App/Services/UsageTracker.cs
```diff
@@ -0,0 +1,374 @@
...
+                    log.Info("User has disabled sending anonymized usage statistics to GitHub");
+
+                    return ClearCounters(true, true)
+                        .ContinueAfter(() => client.SendOptOut());
+                });
+            }
+            else
+            {
+                return Observable.Return(Unit.Default);
+            }
```
shana: {} 🔫

shana:
Make sure to test this without the script submodule to make sure we're not accidentally relying on internal things and breaking external contributors.

src/GitHub.App/Services/UsageTracker.cs
grokys: Again, these come from GHfW - do you want me to change them or should we be avoiding unnecessary differences with GHfW when the other code is identical?

shana: GHfW is not going to have any changes to this code, so it doesn't matter if it's identical or not (and it's not anymore anyway with the other PR coming in). I'd rather have consistency in the code base  and make this ours.

To be very honest, I'd rather have 

```
            return client != null
                ? Observable.Defer(() =>
                {
                    log.Info("User has enabled sending anonymized usage statistics to GitHub");
                    return client.SendOptIn();
                })
                : Observable.Return(Unit.Default);
```

because that's the form that we use the most in logic like this around the codebase, but that doesn't get on my nerves as much as those floating braces :wink:

shana:
Also, if there's no `MetricsService`, then all the `client.` accesses will go 💥 , so need to check all those nulls ftw.

grokys:
Just checked this with `scripts` de-inited. `TryGetService<T>` needed a `[return: AllowNull]` but now that is fixed, seems to work fine.

shana:
This is looking fine and I'm :+1: for merging, but we can get a couple of tests for the clear counters and usage model building there, to make sure we're clearing everything that matters and that the built model has all the relevant keys? There might be tests already in the Desktop repo that we can reuse.

src/GitHub.Exports/Services/UsageTracker.cs
```diff
@@ -0,0 +1,251 @@
...
+        {
+            ++usage.Model.NumberOfClones;
+            SaveUsage(usage);
+        }
+
+        UsageStore LoadUsage()
+        {
+            var result = File.Exists(storePath) ?
+                SimpleJson.DeserializeObject<UsageStore>(File.ReadAllText(storePath)) :
+                new UsageStore { Model = new UsageModel() };
```
shana: This whole thing is going to need exception handling, there's multiple things that can go wrong here. I'd like to follow the same pattern that https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio/Services/ConnectionManager.cs has - delegates for IO methods that can be mocked later on by our unit tests, a separate method that ensures the cache file exists, and splitting up reading/writing from serializing/deserializing, since one can fail because of the OS but the latter can fail because we break the format (and we can catch that in unit tests)

Note: `ConnectionManager` is not checking for IO exceptions at the moment. It's never failed because that would require two VS instances running at the same time and accessing `ConnectionManager` at exactly the same time, which is unlikely to happen (most cache accesses are user-initiated when logging in, so a user would have to login at the same time in two VS instances, unlikely). However, metrics are timer based which means the likelihood of two VS instances racing for the same file access could potentially be higher, so we need to watch out for this. Any crashes here will crash VS.

shana:
Merging this in. We need to do a build with this, deploy the central metrics branch and test a few metrics, so we can have some data to graph.

Issue #351, opened 2016-06-09
Login error messages not getting fully propagated to the UI


shana:
When a user hasn't verified their email, octokit returns

```
{"message":"Validation Failed","errors":[{"resource":"OauthAccess","code":"custom","field":"user","message":"user must have a verified email address in order to authenticate via OAuth"}],"documentation_url":"https://developer.github.com/v3/oauth_authorizations/#get-or-create-an-authorization-for-a-specific-app"}    at Octokit.Connection.HandleErrors(IResponse response) in Y:\jenkins\workspace\GitHub for Visual Studio speakeasy jenkinskeep\submodules\octokit.net\Octokit\Http\Connection.cs:line 563
```

but our login UI just says `"Could not connect to github.com"`. We need to propagate these messages properly so users don't get confused about why things are failing.

grokys:
:+1: Yes, this bit me just the other day.

grokys:
Fixed by #352

PR #352, opened 2016-06-09
Show login failure message.

grokys:
Initial fix for #351. Shows the error message when connection fails for .com and enterprise.

Needs a design pass! Also not sure if binding to the containing `ErrorMessageDisplay.Message` property is elegant or a hack.

image

image


donokuda:
> Needs a design pass!

I'll open a separate PR with a design pass 🎨

shana:
There's already code that handles showing detailed error messages being used elsewhere, I honestly don't know why we're not doing that (slipped under the radar, I'm sure)

The repo creation viewmodel uses UserErrors.Throw to throw exceptions that then get caught by the `<uirx:UserErrorMessages />` xaml object. 

Throwing:
https://github.com/github/VisualStudio/blob/master/src/GitHub.App/ViewModels/RepositoryCreationViewModel.cs#L296

Showing:
https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio/UI/Views/Controls/RepositoryCreationControl.xaml#L243

(UserErrorMessage registers a handler for the Throw method in https://github.com/github/VisualStudio/blob/master/src/GitHub.UI.Reactive/Controls/Validation/UserErrorMessages.cs#L111-L111, fyi)

So technically what we want to do is have a `<uirx:UserErrorMessages>` block and then use `UserError.Throw` to show the messages (optionally changing the wording on them if we want)

donokuda:
@shana Should I hold off on polish until your suggestions are brought in?

grokys:
@shana ah! Yes I'd forgotten about that control. Will change to use that.

@donokuda yep, hold off until I make the changes @shana suggested.

shana:
@donokuda It might not need polish at all, we should already have all the right stuff put in place for this.

donokuda:
aww yeahhh! Feel free to 🔔 me if you do need any polish! ✨

grokys:
Now updated to use `UserError.Throw` throughout the login viewmodel. The required adding support for recovery options in `UserErrorMessage`. Here's what it looks like now:

Bad username/pass .com:

image

Bad username/pass enterprise:

image

Other error dotcom:

image

src/GitHub.UI.Reactive/Controls/Validation/UserErrorMessages.cs
```diff
@@ -46,7 +46,7 @@ public UserErrorMessages()
             };
         }
 
-        public static readonly DependencyProperty IconMarginProperty = DependencyProperty.Register("IconMargin", typeof(Thickness), typeof(UserErrorMessages), new PropertyMetadata(new Thickness(0,10,7,0)));
+        public static readonly DependencyProperty IconMarginProperty = DependencyProperty.Register("IconMargin", typeof(Thickness), typeof(UserErrorMessages), new PropertyMetadata(new Thickness(0,0,8,0)));
```
shana: This is going to affect all the other dialogs that don't set this property explicitly. Are we sure it's not going to break existing error messages UI?

src/GitHub.UI.Reactive/Controls/Validation/UserErrorMessages.cs
```diff
@@ -106,18 +106,14 @@ public UserError UserError
             Justification = "We're registering a handler for a type so this is appropriate.")]
         public IDisposable RegisterHandler<TUserError>(IObservable<bool> clearWhen) where TUserError : UserError
         {
-            if (IsVisible)
+            return UserError.RegisterHandler<TUserError>(userError =>
```
shana: What's the reason for this change?

src/GitHub.VisualStudio/UI/Views/Controls/LoginControl.xaml.cs
```diff
@@ -22,6 +24,8 @@ public class GenericLoginControl : SimpleViewUserControl<ILoginControlViewModel,
     [PartCreationPolicy(CreationPolicy.NonShared)]
     public partial class LoginControl : GenericLoginControl
     {
+        private IDisposable errorHandler;
```
shana: `private` :gun:

src/GitHub.VisualStudio/UI/Views/Controls/LoginControl.xaml.cs
```diff
@@ -39,7 +43,10 @@ public LoginControl()
                         NotifyDone();
                     }
                 });
+
+                d(Disposable.Create(() => errorHandler.Dispose()));
```
shana: Can `errorHandler` be null here? It's only set in the selection changed event below, and you check for nullness there, is that event guaranteed to be invoked always?

src/GitHub.VisualStudio/UI/Views/Controls/LoginControl.xaml.cs
```diff
@@ -112,5 +107,25 @@ void SetupSelectedAndVisibleTabBindings(Action<IDisposable> d)
                 .Where(x => x == true)
                 .BindTo(this, v => v.enterpriseTab.IsSelected));
         }
+
+        private void hostTabControl_SelectionChanged(object sender, System.Windows.Controls.SelectionChangedEventArgs e)
```
shana: `private` :gun:

src/GitHub.App/ViewModels/LoginTabViewModel.cs
```diff
@@ -18,6 +19,7 @@
 
 namespace GitHub.ViewModels
 {
+    [SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable")]
```
shana: Is this still needed?

shana:
Looking good, left a few comments sprinkled about.  🏈 ➡️

src/GitHub.UI.Reactive/Controls/Validation/UserErrorMessages.cs
grokys: The handler in the login dialog wasn't getting registered because the `UserErrorMessages` control wasn't visible at the time of the call. I saw that check and it looked wrong: if someone wants to not register a handler when a control isn't visible they should do it explicitly IMO. Doing nothing in a situation like this just seems confusing to me...

src/GitHub.VisualStudio/UI/Views/Controls/LoginControl.xaml.cs
grokys: I don't think so - a tab will be selected when the UI is built, causing the selection changed event to fire. Looking at the WPF call stack it looks like this will always happen.

src/GitHub.App/ViewModels/LoginTabViewModel.cs
grokys: Yeah, because it creates a `RecoveryCommand` (which implements `IDisposable`) and assigns it to `NavigateForgotPassword`. I don't think `RecoveryCommand` actually needs to be disposed though - it appears that the only purpose of it is to set its `CanExecute` state.

Should I dispose the `RecoveryCommand` anyway? Or just add a justification?

src/GitHub.UI.Reactive/Controls/Validation/UserErrorMessages.cs
grokys: I managed to get an error message to show in ``RepositoryCreationControl` and it looks fine I think. TBH not sure why it wasn't off-center there.

grokys: Just tried `RepositoryCreationControl` on master and the error doesn't even seem to appear, which is probably why nobody's noticed it's off-center :)

src/GitHub.UI.Reactive/Controls/Validation/UserErrorMessages.cs
shana: heh, yeah, makes sense (or rather, it doesn't...) 👍

src/GitHub.UI.Reactive/Controls/Validation/UserErrorMessages.cs
shana: Hmmm, I wonder if the error there is meant to appear... we do have some errors that show up inline.

shana: Ah, it's the creation repo failure message. I guess we never showed that then? :P

src/GitHub.App/ViewModels/LoginTabViewModel.cs
shana: `RecoveryCommand` leaks a disposable for the `canExecute`, but it should be disposed automatically when the viewmodel gets disposed of (which should happen when the windows is closed), so it should be fine.

PR #358, opened 2016-06-14
Make "Sign in"/"Create an account" links in GitHub pane work.

grokys:


src/GitHub.App/Models/RepositoryHosts.cs
```diff
@@ -157,11 +157,15 @@ public IRepositoryHost LookupHost([AllowNull] HostAddress address)
                     );
                     if (successful)
                     {
+                        // We need to add the connection before assigning to GitHubHost or
+                        // EnterpriseHost so that anything listening to those will find the
+                        // associated connection.
+                        connectionManager.AddConnection(address, usernameOrEmail);
```
shana: But what happens if listening code is expecting to have `GitHubHost/EnterpriseHost` set when they get this notification? Sounds like we have a race condition here.

shana: We'll probably have to set the host fields (without raising notifications), then call `AddConnection`, and then raise the field set events. We don't _really_ need `RaiseAndSetIfChanged` on the `GitHubHost/EnterpriseHost`, it's not something that gets set a lot, so we can probably just do `field = value; this.RaisePropertyChange()` on those setters.

src/GitHub.App/ViewModels/LoggedOutViewModel.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Reactive.Linq;
+using GitHub.Services;
+using GitHub.UI;
+using GitHub.ViewModels;
+using ReactiveUI;
+using System.Reactive;
+using System.Reactive.Subjects;
+using GitHub.Info;
+
+namespace GitHub.VisualStudio.UI.Views
```
shana: Namespace needs updating.

shana:
Looks nice! ✨

src/GitHub.App/Models/RepositoryHosts.cs
grokys: Well as far as I could see, getting the notification from the connection manager requires listening to `Connections.CollectionChanged` and nothing I could see relied on the `GitHubHost`/`EnterpriseHost` fields being set there. I may have missed something though.

grokys: Having said that though - you're right - we can have it both ways. Will fix.

shana: It's more of a "if it can race, I'm sure we'll make code that races" thing. It's the law!

src/GitHub.App/ViewModels/LoggedOutViewModel.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// <inheritdoc/>
+        public IReactiveCommand<object> SignIn { get; }
+
+        /// <inheritdoc/>
+        public IReactiveCommand<object> Register { get; }
+
+        /// <summary>
+        /// Called when the <see cref="SignIn"/> command is executed.
+        /// </summary>
+        private void OnSignIn()
```
shana: Kill the `private`

src/GitHub.App/ViewModels/LoggedOutViewModel.cs
```diff
@@ -0,0 +1,62 @@
...
+            // GitHubPaneViewModel will listen to RepositoryHosts.IsLoggedInToAnyHost and close
+            // this view when the user logs in.
+            uiProvider.SetupUI(UIControllerFlow.Authentication, null);
+            uiProvider.RunUI();
+        }
+
+        /// <summary>
+        /// Called when the <see cref="Register"/> command is executed.
+        /// </summary>
+        private void OnRegister()
```
shana: `private` :gun:

grokys: Arrgh force of habit!

grokys:
Closing as this needs to target master.

Issue #360, opened 2016-06-14
Validate if directory exists when cloning/creating


shana:
When cloning or creating repos, validate that the target directory where VS is going to create the git repo doesn't exist, otherwise when VS tries to clone it will fail and the user has to start the process over.

PR #366, opened 2016-06-15
Fix create button enabled state in create dialog.

### LINKED ISSUE #346: Repository creation button activation not enabled when Name text box filled
Per the test checklist, this is supposed to be enabled once the text box is filled
Only seems to be enabled if you change focus from the text box.

### LINKED ISSUE #347: Repository creation button activation enabled when Name text box empty
1. Clicked on create link
2. Filled out name text box
3. Changed focus to enable button (via #346) 
4. Cleared out name text box
5. Clicked Create button
6. Repo with name that was deleted from textbox was created

grokys:
XAML bindings need to write to the associated VM property on every keypress rather than when focus is lost in order for VM validators to work. Fixes #346, #347 and half of #360.


shana:
let_it_be_done

PR #369, opened 2016-06-16
Check if the clone target directory already exists.

### LINKED ISSUE #360: Validate if directory exists when cloning/creating

grokys:
And show a validation error if it doesn't. Fixes #360.


src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs
```diff
@@ -80,7 +83,17 @@ public class RepositoryCloneViewModel : BaseViewModel, IRepositoryCloneViewModel
...
+                x => x.BaseRepositoryPath,
+                x => x.SelectedRepository,
+                (x, y) => x.Value);
+
+            BaseRepositoryPathValidator = ReactivePropertyValidator.ForObservable(baseRepositoryPath)
+                .IfNullOrEmpty("Please enter a repository path")
+                .IfTrue(x => x.Length > 200, "Path too long")
+                .IfContainsInvalidPathChars("Path contains invalid characters")
+                .IfPathNotRooted("Please enter a valid path")
+                .IfTrue(IsAlreadyRepoAtPath, Resources.RepositoryNameValidatorAlreadyExists);
```
shana: Instead of copying this here, maybe it's best to pass an observable to `CreateBaseRepositoryPathValidator`. The repository creation viewmodel also does these exact same checks, it would be nice not to duplicate code.

grokys: I looked into that, but it was just different enough to cause problems so I gave up and did it like this. I can look into it again though.

shana: The only two differences are that this one takes an observable and the other one makes its own observable, so we can just pass in one instead of letting it do that; and that there's an extra `IsTrue` directive on this one, which can be added to the extension method since both create and clone should be doing the same validations. Everything else is the same, am I missing something?

grokys: The issue was that in the create view model the repository name is validated separately from the base path, whereas in the clone dialog we don't need to validate the name as it's already coming from GitHub. So the options were either merge the two validators in the create view model somehow or have two validators for a single textbox in clone. If that makes sense?

shana:
:sparkles:

PR #370, opened 2016-06-16
Stop the UIController properly in the GitHubPane

shana:
Make sure existing UI is stopped whenever the following conditions change:

- [x] The active repository changes
- [x] The user login status changes

Also make sure `Reload` is reentrant, all those await calls means it can get called multiple times.

### DIFF
--- a/src/GitHub.VisualStudio/UI/Views/GitHubPaneViewModel.cs
+++ b/src/GitHub.VisualStudio/UI/Views/GitHubPaneViewModel.cs
@@ -131,7 +131,7 @@ namespace GitHub.VisualStudio.UI.Views
                 return;
             }
 
-            if (uiController == null || (data != null && data.ActiveFlow != uiController.SelectedFlow))
+            if (uiController == null || (data != null && data.ActiveFlow != uiController.CurrentFlow))
                 StartFlow(data?.ActiveFlow ?? UIControllerFlow.PullRequests, connection, data);
             else if (data != null || currentNavItem >= 0)
                 uiController.Jump(data ?? navStack[currentNavItem]);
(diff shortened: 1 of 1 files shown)


shana:
Still have an annoying problem with the list sometimes not refreshing, even though it clearly changes. But otherwise, looking nice

pr-list-view2

src/GitHub.VisualStudio/UI/Views/GitHubPaneViewModel.cs
```diff
@@ -19,13 +19,16 @@
 using ReactiveUI;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using System.Diagnostics;
```
paladique: Looks like this isn't being used.

src/GitHub.VisualStudio/UI/Views/GitHubPaneViewModel.cs
```diff
@@ -105,53 +108,103 @@ protected override void RepoChanged(bool changed)
...
+                var isLoggedIn = await connection.IsLoggedIn(hosts);
+                if (reloadCallId != latestReloadCallId)
+                    return;
+
+                IsLoggedIn = isLoggedIn;
             }
+
+            if (!IsGitHubRepo.Value)
+            {
+                //LoadView(UIViewType.NotAGitHubRepo);
```
paladique: Doesn't look like this if statement is doing anything anymore, I uncommented to investigate and it looks like it's broken, do we intend to add `NotAGitHubRepo` to `UIViewType` at some point in the future?

shana: Yeah, it's a placeholder for @grokys to stuff things into 😄

paladique:
This looks great :+1:

Issue #368, opened 2016-06-15
Hotfix release 1.0.18.5


shana:
Tracking issues that we want to backport the fixes for to the release branch, for release 1.0.18.5:

- [x] #367 
- [x] #360 
- [x] #301 

@paladique @grokys Anything else we should be including?

paladique:
Looking through closed issues/PRs, I think #311, #292, and #360 are ones that might be nice to include. I've personally come across #360 once or twice, which I can see as being annoying to users.

shana:
#311 and #289 (the issue for #292) were included with the last release (https://github.com/github/VisualStudio/releases/tag/v1.0.18.4)

#360 looks good, I've added it to the list.

shana:
@paladique Could you start prepping this?

First step is to bump the version of branch `release/1.0.18` with the `script/Bump-Version.ps1` ps script to 1.0.18.5. 

Then do a new branch from it for the fix to #367.

For the fix to #360, also branch off from `release/1.0.18` and then cherry-pick the commits for that fix into that.

Do pull requests for both these branches to `release/1.0.18`, and then we can cut a release to test!

paladique:
@shana picking the commits and merging changes to `RepositoryCreationControl.xaml` is really huge, in comparison to what was fixed in #360. Looks like there were some binding issues addressed/resolved in the [last release](https://github.com/github/VisualStudio/tree/v1.0.18.4) that didn't make it into the `release/1.0.18` branch, am I missing this somewhere? Do I need to pick these?

https://github.com/github/VisualStudio/commit/8e14765a964adf572b67ae5d1b64007561df4f4d

shana:
@paladique Oooof, yeah, I don't think we should backport the big changes, they require a bunch of changes. See if just manually applying the little xaml change from https://github.com/github/VisualStudio/pull/366/files makes things work, otherwise we might skip it. It's a hotfix release, so it shouldn't introduce any big changes and only fix easy or critical things.

shana:
Oh! We should also throw in the fix for #301, that was reported by MS too.

PR #373, opened 2016-06-17
[Backport] Add directory validation

paladique:
For the upcoming hot-fix release listed in #368, it was decided that the resolution to #360 should be in this release. Picked those commits.


paladique:
I should probably hold off on picking the rest of this until the second part is complete.

paladique:
So `release/1.0.18` doesn't include the binding fixes from

paladique:
Somehow after picking all commits from #369 it looks like there was still a reference to two files that didn't exist. I just removed them and tested to make sure everything was :+1:

shana:
This looks good! 👍

PR #374, opened 2016-06-20
Display a message when project isn't a git/github repository.

grokys:
We still need to decide on the wording for when the current Team Explorer context isn't a git repository.

### DIFF
--- a/src/GitHub.VisualStudio.UI/RepositoryOrigin.cs
+++ b/src/GitHub.VisualStudio.UI/RepositoryOrigin.cs
@@ -1,0 +1,10 @@
+﻿namespace GitHub.VisualStudio.UI
+{
+    public enum RepositoryOrigin
+    {
+        DotCom,
+        Enterprise,
+        Other,
+        NonGitRepository,
+    }
+}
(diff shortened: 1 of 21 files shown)


src/GitHub.VisualStudio.UI/RepositoryOrigin.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace GitHub.VisualStudio.UI
+{
+    public enum RepositoryOrigin
+    {
+        DotCom,
+        Enterprise,
+        Other,
+        NonGitRepository,
+    }
```
shana: Move the `Other` value up to the first position, or add a `None` value. enums always have a default value, and distinguishing between `DotCom` and `haven't probed yet / this is a default` is hard if they're the same value.

shana:
The current messages look fine for now. We can refine them during the beta phase.
All these strings should be in the resx file in `GitHub.VisualStudio.UI`. Check for existing strings in there that can be used, I'm sure the Getting Started phrase is in there already.

shana:
⚽ back to you!

src/GitHub.VisualStudio.UI/RepositoryOrigin.cs
grokys: Well strictly speaking, the `Unknown` state is `RepositoryOrigin == null` as the `RepositoryOrigin` property is currently nullable, because it came from the nullable `IsGitHubRepo` property. However I agree that it would be better the have that `Unknown` value as an enum value rather than the null state: will change.

shana:
thumbs-up

PR #375, opened 2016-06-20
Catch File I/O exception in GitHub.Exports

### LINKED ISSUE #367: Crash in GitHub.Exports
Hi all,
Sorry for the broad posting. I work on the Visual Studio team at Microsoft as a Program Manager and we have some reported Watson crashes with this extension for Updates 1 and 2 (only 1 crash so far for Update 3, so perhaps the issue has been fixed. I would love to confirm) You can contact me here or at v-kefans@microsoft.com.

paladique:
Fixes #367. Wrapping in a try/catch. Since this is happening in `Logger.cs`, we'll add the exception message to the cache if we continue to get reports about problems here.

### DIFF
--- a/src/GitHub.Exports/Services/Logger.cs
+++ b/src/GitHub.Exports/Services/Logger.cs
@@ -77,7 +77,13 @@ namespace GitHub.VisualStudio
             await System.Threading.Tasks.Task.Delay(500);
             lock(fileLock)
             {
-                File.AppendAllText(defaultLogPath, msg, Encoding.UTF8);
+                try
+                {
+                    File.AppendAllText(defaultLogPath, msg, Encoding.UTF8);
+                }
+                catch(Exception)
+                {
+                }
             }
         }
     }


shana:
Could you squash this into one commit? It's small enough to be a self-contained commit.

src/GitHub.Exports/Services/Logger.cs
```diff
@@ -74,11 +74,19 @@ public static void WriteLine(string message)
...
+                {
+                    File.AppendAllText(defaultLogPath, msg, Encoding.UTF8);
+                }
+            }
+            catch (Exception)
             {
-                File.AppendAllText(defaultLogPath, msg, Encoding.UTF8);
+                throw;
             }
+
```
shana: The `try/catch` should be inside the lock, and only around the `File.AppendAllText`. Also don't throw the exception, if we do that it'll crash Visual Studio.

shana:
✨

Issue #356, opened 2016-06-14
GitHub pane appears blank when current repository is not a GitHub repo.


grokys:
There are 3 situations we need to handle here:

1. There is no project loaded and the Team Explorer has source control context
2. There is a project loaded but it's not a git repository
3. There is a project loaded but it's not a GitHub repository

shana:
Ayup.

grokys:
@donokuda any idea what this should look like?

donokuda:
We could go with the same approach as the [logged out view](https://github.com/github/VisualStudio/pull/266) in terms of layout.

I think we should use this as an opportunity to publish their repository to GitHub so they can start using what our extension offers immediately such as creating a pull request.

Here are some mocks to illustrate what I was thinking:

<img width="628" alt="screen shot 2016-06-15 at 10 25 12 am" src="https://cloud.githubusercontent.com/assets/1174461/16090234/78df430c-32e3-11e6-99cc-f85849c4c735.png">

Here's an alternative version using the primary button styles because yey options!

<img width="620" alt="screen shot 2016-06-15 at 10 22 25 am" src="https://cloud.githubusercontent.com/assets/1174461/16090127/14753f3e-32e3-11e6-9e9e-7e6f20ddf5c2.png">

That said, I'm curious if there's a way where we can ship this message without publishing as a first pass (because having something explaining why the pane is empty is better than nothing.) We could link to the "Create Repo" page, but I'm not sure how to bring the flow back to Visual Studio without making it feel "disconnected."

donokuda:
PS. If there's a view already set up for this, I can stub something out for you 😄

grokys:
Looks good to me! There's no view currently set up - will do that now.

grokys:
@donokuda ok, there's a blank view on the `fixes/356-not-a-github-repository` branch at:

https://github.com/github/VisualStudio/blob/fixes/356-not-a-github-repository/src/GitHub.VisualStudio/UI/Views/NotAGitHubRepositoryView.xaml

Work your magic! :space_invader:

donokuda:
✨ 🎩 ✨ 

I added some content in 64ef6c6ab95fe2c3fa0d3b8fc45959eaf23f9712.

❓ Is the publish button the only thing needed for this flow? After looking at our create a repo modal, I think we'll want to ask for a name, description, and which account to add the repo to:

[...]

PR #376, opened 2016-06-20
Update styles for primary action button

donokuda:
Ref: https://github.com/github/VisualStudio/issues/356#issuecomment-226739689

> On a different note, I've been doing the PR creation and I'm finding the green button pretty hard to read. The white on flat green just disappears. It's also almost impossible to tell when it's enabled/disabled, too :/

This PR updates the color for the primary action button to match the green specified in [primer](http://primercss.io/colors/).

The text is also bolded for better readability / contrast.

EDIT: This only updates the color for the light and blue themes. The dark theme is unaffected aside from bolding the text.

screen shot 2016-06-20 at 11 48 28 am

cc: @shana


shana:
Aah, awesome! Do we also have a style for disabled buttons?

donokuda:
>  Do we also have a style for disabled buttons?

I thought we did but it turns out we didn't! (or at least it wasn't getting set)

Added in ac0edec

image

shana:
The disabled style doesn't make it obvious at all that it's a disabled button... If you've never seen the button in its enabled state, it just looks like just a faded style. This kinda works in the website because the website 1) always has the buttons enabled from the start 2) they have a faux-3d style that becomes flat when they get disabled (which kinda works as a hint... kinda)

I feel like green should only be there to signal that everything is green and ready to go. How about a monochromatic style for disabled buttons?

donokuda:
> Looking at the Commit button in the Changes pane, where there are no other buttons in the area and the Commit button starts disabled - the Light and the Dark theme make the background of the button the same as the background of the window, so only the outline and the gray text is showing. The blue theme turns the button background gray but the window background is white.

That makes sense. What if we treated our primary button the same: appear like a normal disabled button? I also took note of how the blue theme looks like a faded style and tried to make our disabled button's background closer to the window background color:

artboard 20

(I purposefully removed the "Cancel" button so we have a better sense if it works as a disabled button)

And then when enabled, we give it the full green button style:

button-enabled

> BTW, all the themes highlight enabled buttons on mouse hover (but that's learned behaviour, albeit pretty standard on windows). We should probably make use of that as an extra hint of state.

💯 Having a hover state will help out with understanding if a button is disabled or not:

primary-button-disabled

primary-button

By the way, these changes are only on the blue theme. If they look good, then we can go ahead and work the changes to the other themes.

shana:
I've added the missing color definitions on the dark and light themes so that we can pull this in now. The colors are likely not correct and require validation, let's open another PR for that.

shana:
This fixes #356

PR #379, opened 2016-06-21
Handling File I/O crash in GitHub.Exports in master

paladique:
via hotfix #367, for `master`.


shana:
Looks like your Visual Studio configuration might be adding spaces before the parenthesis in method calls and expressions - `lock (...)` instead of `lock(...)` and `catch (...)` instead of `catch(...)`. Could you fix those up please, here and on the backport.

shana:
thumbs-up

PR #389, opened 2016-07-04
Adds Appveyor support.

grokys:


shana:
WOOT! ✨

appveyor.yml
```diff
@@ -0,0 +1,18 @@
+# scripts that run after cloning repository
+install:
+  - if "%APPVEYOR_REPO_BRANCH%" EQU "master" if not defined APPVEYOR_PULL_REQUEST_NUMBER set MASTER_BUILD=1
```
shana: What mainly worries me about this check is whether this can be triggered by anyone creating a PR on a fork. Randomly running anyone else's code with private submodules is a risk if this trigger activates like this.

grokys: Ah yes, indeed that may be a problem. I'll take a look at how to detect if it's running on a fork.

PR #392, opened 2016-07-05
Allow null param to CreateListenerCollection

### LINKED ISSUE #391: NullReferenceException when using `TrackingCollection.CreateListenerCollection`
The default value for the `stickieItemsOnTop` in `TrackingCollection.CreateListenerCollection` is null, but if one tries to use a null here this throws an exception:

    var col = new ObservableCollection<T>(stickieItemsOnTop);

As you cannot create an `ObservableCollection` with a null initializer collection.

grokys:
Fixes #391


shana:
sigh, we need to add tests for these helper methods.

shana:
thumbs-up

PR #395, opened 2016-07-05
Correctly check for existing repo in clone dialog.

### LINKED ISSUE #109: Ctrl-clicking a selection in the list doesn't remove the selection and disable the clone button
**Steps to reproduce:**
1. Click on Clone action link
2. Select a repo in the list
3. ctrl-click the selection to unselect it 

**Expected Results:** The selection is un-selected and the clone button is disabled

**Actual Results:** Nothing. I can un-select a repo by clicking outside the list (but inside the box). I am on a Mac so I tried control + clicking and cmd/command+ clicking as well. Hoping @shana can just take a :eyes: for me. 
:point_down:  is just to show that I can click outside the list but inside the box to unselect. :grin: 
select

### LINKED ISSUE #394: Clone dialog doesn't always detect existing repo.
When using the clone dialog, if I have the following repositories in the destination directory:

image

And I select an existing repository, followed by _another_ existing repository, the Clone button is enabled when it shouldn't be:

image

grokys:
The clone dialog was validating for the presence of an existing repo at the destination path, but the validation was dependent upon the _previous_ validation state. Make sure the current path is always validated. Fixes #394. Fixes #109.

### DIFF
--- a/src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs
+++ b/src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs
@@ -171,7 +171,7 @@ namespace GitHub.ViewModels
 
         bool IsAlreadyRepoAtPath(string path)
         {
-            Debug.Assert(path != null, "RepositoryCloneViewModel.IsAlreadyRepoAtPath cannot be passed null as a path paramter.");
+            Debug.Assert(path != null, "RepositoryCloneViewModel.IsAlreadyRepoAtPath cannot be passed null as a path parameter.");
 
             bool isAlreadyRepoAtPath = false;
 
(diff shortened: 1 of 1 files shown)


src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs
```diff
@@ -175,12 +175,8 @@ bool IsAlreadyRepoAtPath(string path)
```
shana: Let's add a `Debug.Assert(path != null)` here, to protect against accidental regressions of the validator logic that ends up calling this.

src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs
```diff
@@ -171,16 +171,14 @@ IObservable<Unit> OnCloneRepository(object state)
 
         bool IsAlreadyRepoAtPath(string path)
         {
+            Debug.Assert(path != null, "RepositoryCloneViewModel.IsAlreadyRepoAtPath cannot be passed null as a path paramter.");
```
shana: aaaaaaaah

grokys: Haha - oops.

PR #398, opened 2016-07-05
Allow manually adding items to TrackingCollection before subscription

shana:
Make sure items added to `TrackingCollection` via `AddItem` are processed once subscription happens and not just ignored.


shana:
@grokys @paladique This is ready for :eyes:

paladique:
:+1: !

PR #421, opened 2016-07-06
Verify source code is not in a git repo in context menu checks

### LINKED ISSUE #419: Null reference exception in IsGitHubRepo()
**Version**: 2.0 Beta 1
**VS Version**: VS 2015 Update 3

Repro: Right-click in a source file that is not part of a git repository (eg. create a simple Console project)

`ActiveRepo` is null

https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio/Base/MenuBase.cs#L116

paladique:
In `IsGitHubRepo()`, `ActiveRepo` is null if source code is not in a git repo, which throws a null reference exception. Modified to return false if this is the case.

Fixes #419

### DIFF
--- a/src/GitHub.VisualStudio/Base/MenuBase.cs
+++ b/src/GitHub.VisualStudio/Base/MenuBase.cs
@@ -113,6 +113,9 @@ namespace GitHub.VisualStudio
         {
             RefreshRepo();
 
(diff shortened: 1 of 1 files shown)


shana:
thumbs-up

PR #424, opened 2016-07-08
Fix url generation with startLine and endLine

gazab:
When selecting text in Visual Studio startLine and endLine depends on which direction you did the selection. So if your click and drag upwards (or use your keyboard) to make your selection startLine will be greater than endLine. This in turn will make the extension generate a link like this: 
https://github.com/foo/bar/blob/123123/src/dir/file1.cs#L2-L1
which will break Github's line highlighting and not highlight those rows at all.

This PR adds tests to detect this and also fixes it by swapping startLine and endLine when needed.

I also noticed that when you select a single row the extension still tries to make a range link like this:
https://github.com/foo/bar/blob/123123/src/dir/file1.cs#L1-L1

Links like these still work but I find them rather ugly so I added tests and fixed that too.


shana:
Thank you so much for this, it looks great! Gets the official @Haacked seal of approval!

PR #425, opened 2016-07-08
Bump octokit to 0.20.0

shana:
This bumps octokit to 0.20, updates the API for breaking changes, and also includes a local octokit fix (not yet upstream) that adds the PR Review comment count field to the PullRequest object (comment counts include commit comments and pr comments, which are stored in two different fields)

The comment count is not actually getting populated right now because the json result of listing PRs doesn't return these fields, but they are returned when we get individual PRs, so this bump is not strictly required right now.

This needs local testing before merging!


paladique:
:+1: as is, gonna make it even with master and merge :sparkles:

PR #427, opened 2016-07-08
Hide comment icon and count if there's no comments

shana:
Also, yeah, there's lots of silliness going on here...

The CountToVisibilityConverter had its logic backwards because it was being used to show textblocks with "XXX required" messages in the publish control if those fields were not filled out... which is why we have PromptTextBox controls to use!

We need to hide the icon because we don't actually have comment counts on the PR body, because the PR list json returned by the API returns everything except this information (and some other status fields like merge status). So we're going to have to get the list and then query all PRs individually. That's going to be around 300+ requests for a decent-size repo for an initial cache fill.

So we're punting on the comment count for now and figuring this one out later.

GRUMBLE!

### DIFF
--- a/src/GitHub.UI/Converters/CountToVisibilityConverter.cs
+++ b/src/GitHub.UI/Converters/CountToVisibilityConverter.cs
@@ -15,7 +15,7 @@ namespace GitHub.UI
     {
         public override object Convert(object value, Type targetType, [AllowNull] object parameter, [AllowNull] CultureInfo culture)
         {
-            return ((int)value == 0) ? Visibility.Visible : Visibility.Collapsed;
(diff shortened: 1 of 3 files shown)


pmn:
> We need to hide the icon because we don't actually have comment counts on the PR body, because the PR list json returned by the API returns everything except this information (and some other status fields like merge status). So we're going to have to get the list and then query all PRs individually. That's going to be around 300+ requests for a decent-size repo for an initial cache fill.

> So we're punting on the comment count for now and figuring this one out later.

FWIW this should be significantly improved in the new and wonderful GraphQL future. :100: on punting on comment count until that's available.

PR #438, opened 2016-07-14
Remember PR list filter settings.

### LINKED ISSUE #386: Save current filter/sort etc in PR list
Per repository.

grokys:
Fixes #386

### DIFF
--- a/src/GitHub.App/ViewModels/PullRequestListViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestListViewModel.cs
@@ -98,6 +98,8 @@ namespace GitHub.ViewModels
         {
             base.Initialize(data);
 
+            pullRequestsLoaded = false;
+
             PullRequests = repositoryHost.ModelService.GetPullRequests(repository, pullRequests);
             pullRequests.Subscribe(pr =>
             {
--- a/src/GitHub.App/ViewModels/PullRequestListViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestListViewModel.cs
@@ -64,15 +64,19 @@ namespace GitHub.ViewModels
                 new PullRequestState { Name = "All" }
             };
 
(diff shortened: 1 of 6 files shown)


shana:
I did a couple of changes to fix the initial loading of the `SelectedState`, which was not correctly updating the filter.

shana:
@grokys Take a look at my changes when you have a chance, so we can pull this in today.

src/GitHub.App/ViewModels/PullRequestListViewModel.cs
```diff
@@ -88,6 +103,21 @@ public override void Initialize([AllowNull] ViewWithData data)
...
+                {
+                    if (listSettings.SelectedAuthor != null)
+                    {
+                        SelectedAuthor = Authors.FirstOrDefault(x => x.Login == listSettings.SelectedAuthor);
+                    }
+
+                    if (listSettings.SelectedAssignee != null)
+                    {
+                        SelectedAssignee = Assignees.FirstOrDefault(x => x.Login == listSettings.SelectedAssignee);
+                    }
```
shana: If both are set, this causes the list to be refiltered twice, which is a potentially costly operation. It would be preferable if we manually called `UpdateFilter` here for these things, but the problem is going to be that the UI won't update unless we set `SelectedAuthor` and `SelectedAssignee`, and that brings us back to the same problem. So let's add a TODO here noting that this needs some rethinking.

src/GitHub.App/ViewModels/PullRequestListViewModel.cs
```diff
@@ -88,6 +104,24 @@ public override void Initialize([AllowNull] ViewWithData data)
...
+ 
+                    pullRequestsLoaded = true;
```
shana: `Initialize` is called in load and when the lost is refreshed, so this code can potentially run multiple times and you'll need to reset this flag when that happens.

PR #446, opened 2016-07-22
Add information pane

### LINKED ISSUE #443: Add error handling to pull request view
The current pull request views (creation and list for now) need to have a way to let the user know if something has failed, whether it be a validation error or a bug of some sort.

This is related to #406, where the validation exception is being caught, but not handled.

The current solution (per discussion) to this is to put an information/status panel at the top of the GitHub panel (similar to VS's current information panel in the solution explorer to inform of updates and tools). This panel will then listen to instances of `INotificationDispatcher` to display the panel and the relevant message.

paladique:
Adding an information pane in the GitHub pane (that currently contains the PR list and creation views and ) for errors or additional information that need to be relayed to the user.

Fixes #443


shana:
This is how it looks now.

errormsg

PR #447, opened 2016-07-22
Localize strings

### LINKED ISSUE #381: Localization pass
There's a few strings that slipped through during the last localization pass (e.g. `RepositoryCloneViewModel`), and some more strings have been added since, mostly in xaml and some in validator messages. All of these need to be moved to the resx files in `GitHub.VisualStudio.UI` (primary) and `GitHub.App`, or existing resource strings used.

paladique:
Fixes #381

### DIFF
--- a/src/GitHub.App/Resources.Designer.cs
+++ b/src/GitHub.App/Resources.Designer.cs
@@ -250,29 +250,38 @@ namespace GitHub.App {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Please enter a repository path.
+        /// </summary>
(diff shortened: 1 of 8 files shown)


shana:
@paladique One of the commits in this set (https://github.com/github/VisualStudio/pull/447/commits/587bf22fb07ce81e1f851665aa12068aed4d1648 - Track login count metric) is already in master with a different commit hash, so this will conflict when it gets merged into the info pane. You need to rebase it out of the branch.

paladique:
Closing this to point to master.

PR #449, opened 2016-07-25
Localize strings

### LINKED ISSUE #381: Localization pass

paladique:
Fixes #381 
Needs :eyes: to make sure they're all caught.

### DIFF
--- a/src/GitHub.App/Resources.Designer.cs
+++ b/src/GitHub.App/Resources.Designer.cs
@@ -250,29 +250,38 @@ namespace GitHub.App {
         }
         
         /// <summary>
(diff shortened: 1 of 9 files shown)


shana:
Looks good! Not sure there's others to be caught in the source. I feel like an extension that scans the source and xaml files for string literals and lists them all would be useful for this...

PR #451, opened 2016-07-26
Disable PR list filters when loading.

### LINKED ISSUE #450: PR filter dropdowns should be disabled until list is loaded
When the PR list finishes loading, the settings for the filter dropdowns are loaded from the previous session. However before this the user can still change these filters, meaning that the users changes will be overridden when the list finishes loading. The filters should be disabled until the list is loaded.

grokys:
Fixes #450.


shana:
:sparkles:

PR #457, opened 2016-07-27
Clear InfoPanel message when closed

### LINKED ISSUE #454: Error message not always displayed in InfoPanel.
Using 2.0.0 RC, an error message is not always displayed for an error.

This is because [UpdateVisibility](https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs#L39) is only called when the error message changes, and if it has previously been called with the same error message and then dismissed, the changed handler won't be called again because the property hasn't changed.

paladique:
Clears the message when the information panel is dismissed.
Fixes #454

### DIFF
--- a/src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
+++ b/src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
@@ -42,7 +42,7 @@ namespace GitHub.VisualStudio.UI.Controls
             control.Visibility = string.IsNullOrEmpty(control.Message) ? Visibility.Collapsed : Visibility.Visible;
         }
 
-        void Button_Click(object sender, RoutedEventArgs e)
+        void Dismiss_Click(object sender, RoutedEventArgs e)
         {
             this.Message = string.Empty;
         }
--- a/src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
+++ b/src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
@@ -45,6 +45,7 @@ namespace GitHub.VisualStudio.UI.Controls
         void Button_Click(object sender, RoutedEventArgs e)
         {
             this.Visibility = Visibility.Collapsed;
+            this.Message = string.Empty;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
```diff
@@ -45,6 +45,7 @@ static void UpdateVisibility(DependencyObject d, DependencyPropertyChangedEventA
         void Button_Click(object sender, RoutedEventArgs e)
         {
             this.Visibility = Visibility.Collapsed;
+            this.Message = string.Empty;
```
shana: You can actually remove the `this.Visibility = Visibility.Collapsed;` line above and just do this one.
As soon as you set the message to empty, the `UpdateVisibility` method triggers and the visibility will get set to collapsed ([on line 42](https://github.com/github/VisualStudio/pull/457/files#diff-9482f34c5d4f74fce8bf57c770dac73bL42))

Side effects ftw! :smile:

src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
```diff
@@ -45,6 +45,7 @@ static void UpdateVisibility(DependencyObject d, DependencyPropertyChangedEventA
         void Button_Click(object sender, RoutedEventArgs e)
```
paladique: We should probably change this to something a little more descriptive? I'm :thought_balloon: `Dismiss_Click`?

shana: Oh yeah, excellent idea, self-descriptive code ftw!

src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
```diff
@@ -42,9 +42,9 @@ static void UpdateVisibility(DependencyObject d, DependencyPropertyChangedEventA
             control.Visibility = string.IsNullOrEmpty(control.Message) ? Visibility.Collapsed : Visibility.Visible;
         }
 
-        void Button_Click(object sender, RoutedEventArgs e)
+        void Dismiss_Click(object sender, RoutedEventArgs e)
```
shana: You need to change this in the xaml as well, otherwise it won't activate.

shana:
:heart_eyes:

PR #459, opened 2016-07-28
Allow "Create PR" comment TextBox to grow as more content is entered

### LINKED ISSUE #458: PR Comment text box is artificially small
When creating a new PR the comments text box is limited to 6 lines: there's no reason for this, we have the whole pane to work with: it should either be just be bigger or automatically resize to the content up to the size of the containing pane.

grokys:
Allow "Create PR" comment TextBox to grow as more content is entered. Wrapped the containing panel in a ScrollViewer so that it can be scrolled if it gets too large for the GitHub pane.

I would've prefered to allow the TextBox to grow only as far as to fill the available space, but there seems to be no panel with the desired behavior in WPF and creating a new panel for this case seems overkill.

Fixes #458

### DIFF
--- a/src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml
+++ b/src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml
@@ -109,43 +109,45 @@
 
     </DockPanel>
 
-    <StackPanel Orientation="Vertical">
-      <ui:PromptTextBox x:Name="titleText"
-                        Margin="10,5"
-                        Text="{Binding PRTitle, UpdateSourceTrigger=PropertyChanged}"
-                        PromptText="{x:Static prop:Resources.TitleRequired}"
-                        Style="{DynamicResource GitHubVsPromptTextBox}" />
+    <ScrollViewer HorizontalScrollBarVisibility="Hidden" VerticalScrollBarVisibility="Auto">
+        <StackPanel Orientation="Vertical">
+            <ui:PromptTextBox x:Name="titleText"
+                    Margin="10,5"
+                    Text="{Binding PRTitle, UpdateSourceTrigger=PropertyChanged}"
+                    PromptText="{x:Static prop:Resources.TitleRequired}"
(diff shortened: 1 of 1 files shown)


src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml
```diff
@@ -109,43 +109,44 @@
...
+                        HorizontalAlignment="Right"
+                        Command="{Binding CreatePullRequest}"
+                        Content="Create pull request"
+                        Style="{StaticResource GitHubVsPrimaryActionButton}" />
+                <Button DockPanel.Dock="Right"
+                        HorizontalAlignment="Right"
+                        Command="{Binding Cancel}"
+                        Content="Cancel"
+                        Style="{StaticResource GitHubVsButton}" />  
+            </DockPanel>
```
shana: The buttons should not be included in a scrollviewer, they should never be out of view. Neither should the title, probably, since it's only one line.

grokys: Yeah, ideally - however as I said, I don't think there's a WPF panel which could give us this behavior.

shana:
I took out the ready to review label because I didn't know if you were done with this or not, put it back in if it's done.

Could you include some screenshots of how it looks now?

grokys:
Sure - here is the dialog without anything entered:

image

And here is the dialog after a PR description has been entered that is too large to display in the pane. The TextBox enlarges pushing down the buttons until the buttons reach the bottom, at which point the text in the TextBox scrolls:

image

shana:
This is conflicting with the branch selection dropdown PR code that was merged to master

grokys:
Merged master and fixed conflicts.

shana:
We should be showing a scrollbar if we have scrollable content, once the textbox can't grow any more.

grokys:
Well we weren't before, but yeah it would probably be best!

shana:
Yeah, sigh, lots of things we should be doing and aren't :cry:

shana:
:sparkles:

PR #464, opened 2016-07-29
Descriptive Pull Request Creation Errors 

### LINKED ISSUE #453: Getting "Validation Failed" error message when trying to create PR
Using the new 2.0 RC I'm getting a "Validation Failed" message when trying to create a PR:

image

paladique:
In this specific instance, we notify the user when they are trying to a duplicate PR on a branch, but with just a "Validation Failed" message. A more informative message lies within the response from the exception. If the type of exception is checked (like `APIValidationException`), we can convert and grab the actual message we want to convey to the user. If it's not that type (it's null), then just display the exception message.

Fixes #453

### DIFF
--- a/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
@@ -79,7 +79,7 @@ namespace GitHub.ViewModels
                     //TODO:Will need a uniform solution to HTTP exception message handling
                     var apiException = ex as ApiValidationException;
                     var error = apiException?.ApiError?.Errors?.FirstOrDefault();
-                    notifications.ShowError(error.Message ?? ex.Message);
+                    notifications.ShowError(error?.Message ?? ex.Message);
                 }
             });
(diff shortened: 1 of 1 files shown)


src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -75,7 +76,10 @@ public class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationV
             {
                 if (!ex.IsCriticalException())
                 {
-                    notifications.ShowError(ex.Message);
+                    //TODO:Will need a uniform solution to HTTP exception message handling
+                    var apiException = ex as ApiValidationException;
+                    var error = apiException?.ApiError?.Errors?.FirstOrDefault();
+                    notifications.ShowError(error.Message ?? ex.Message);
```
shana: `error` can be null here, so it'll throw if you do `error.Message`, you need to null check there as well.

shana:
:rocket: :ship:

PR #473, opened 2016-08-05
WIP: PR checkout button.

grokys:
Adds a PR checkout button that when clicked, checks out the PR to a new branch


src/GitHub.App/Models/PullRequestModel.cs
```diff
@@ -131,6 +134,8 @@ public bool HasNewComments
         public DateTimeOffset CreatedAt { get; set; }
         public DateTimeOffset UpdatedAt { get; set; }
         public IAccount Author { get; set; }
+        [AllowNull] public GitReference Base { [return: AllowNull] get; set; }
+        [AllowNull] public GitReference Head { [return: AllowNull] get; set; }
```
shana: We already have a `IBranch` somewhere which should probably be representing this information. It'll likely need some tweaking, but it already includes the owner repository, so it can give you access to the clone url.

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -74,6 +132,33 @@ public IObservable<string> GetPullRequestTemplate(ILocalRepositoryModel reposito
...
+        /// generates a branch name based on the pull request name and number.
+        /// </remarks>
+        async Task<NewOrExistingBranch> GetPullRequestBranch(IRepository repository, IPullRequestModel pullRequest)
+        {
+            var branch = $"refs/pull/{pullRequest.Number}/head";
+            var existing = await gitClient.GetBranchStartsWith(repository, "pr/" + pullRequest.Number).DefaultIfEmpty();
+
+            if (existing != null)
+            {
+                return new NewOrExistingBranch(existing);
```
shana: `BranchModel` is probably what you want to use here.

PR #482, opened 2016-08-08
Fix crasher when creating a PR with nothing in the body

### LINKED ISSUE #477: Fails when creating a PR with blank description
Trying to create a PR with a blank description fails. With a `[NullGuard]` error in Debug mode and UI just gets confused and fails in Release mode.

shana:
Fixes #477

### DIFF
--- a/src/UnitTests/GitHub.App/ViewModels/PullRequestCreationViewModelTests.cs
+++ b/src/UnitTests/GitHub.App/ViewModels/PullRequestCreationViewModelTests.cs
@@ -1,0 +1,36 @@
+﻿using System.Reactive.Linq;
+using System.Threading.Tasks;
+using NSubstitute;
+using Xunit;
+using UnitTests;
+using GitHub.Models;
+using System;
+using GitHub.Services;
+using GitHub.ViewModels;
+
+public class PullRequestCreationViewModelTests : TempFileBaseClass
+{
+    [Fact]
+    public async Task NullDescriptionBecomesEmptyBody()
+    {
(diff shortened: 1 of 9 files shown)


src/UnitTests/GitHub.App/ViewModels/PullRequestCreationViewModelTests.cs
```diff
@@ -0,0 +1,36 @@
...
+        var repository = new SimpleRepositoryModel("name", new GitHub.Primitives.UriString("http://github.com/github/stuff"));
+        var title = "a title";
+
+        var vm = new PullRequestCreationViewModel(host, repository, service, notifications);
+        vm.SourceBranch = new BranchModel() { Name = "source" };
+        vm.TargetBranch = new BranchModel() { Name = "target" };
+        vm.PRTitle = title;
+
+        await vm.CreatePullRequest.ExecuteAsync();
+        ms.Received().CreatePullRequest(repository, vm.PRTitle, String.Empty, vm.SourceBranch, vm.TargetBranch);
```
grokys: I think the warning here can be suppressed by appending e.g. `var unused =`.

shana: Or maybe just doing `await` there, it might still complain without it

grokys: That won't work - the method is a mock so it won't return anything to await.

PR #495, opened 2016-08-10
Use PR template if it exists.

### LINKED ISSUE #463: Use PR template if it exists when creating PRs
We're currently ignoring PULL_REQUEST_TEMPLATE.md and .github/* files in the PR creation form, and we should use templates to pre-fill the description, if the templates exist.

grokys:
Closes #463.

### DIFF
--- a/src/GitHub.App/Services/PullRequestService.cs
+++ b/src/GitHub.App/Services/PullRequestService.cs
@@ -1,6 +1,7 @@
 using System;
 using System.ComponentModel.Composition;
 using System.IO;
+using System.Linq;
 using GitHub.Models;
 
 namespace GitHub.Services
@@ -10,6 +11,14 @@ namespace GitHub.Services
     [PartCreationPolicy(CreationPolicy.Shared)]
     public class PullRequestService : IPullRequestService
     {
+        static readonly string[] TemplatePaths = new[]
+        {
+            "PULL_REQUEST_TEMPLATE.md",
+            "PULL_REQUEST_TEMPLATE",
+            ".github\\PULL_REQUEST_TEMPLATE.md",
+            ".github\\PULL_REQUEST_TEMPLATE",
+        };
+
         public IObservable<IPullRequestModel> CreatePullRequest(IRepositoryHost host, ISimpleRepositoryModel repository, string title, string body, IBranch source, IBranch target)
         {
             Extensions.Guard.ArgumentNotNull(host, nameof(host));
@@ -26,13 +35,7 @@ namespace GitHub.Services
         {
             Extensions.Guard.ArgumentNotNull(repository, nameof(repository));
 
-            var paths = new[]
-            {
-                Path.Combine(repository.LocalPath, "PULL_REQUEST_TEMPLATE"),
-                Path.Combine(repository.LocalPath, "PULL_REQUEST_TEMPLATE.md"),
-                Path.Combine(repository.LocalPath, ".github", "PULL_REQUEST_TEMPLATE"),
-                Path.Combine(repository.LocalPath, ".github", "PULL_REQUEST_TEMPLATE.md"),
-            };
+            var paths = TemplatePaths.Select(x => Path.Combine(repository.LocalPath, x));
 
             foreach (var path in paths)
             {
--- a/src/GitHub.App/Services/PullRequestService.cs
+++ b/src/GitHub.App/Services/PullRequestService.cs
@@ -1,5 +1,6 @@
 using System;
 using System.ComponentModel.Composition;
+using System.IO;
 using GitHub.Models;
 
 namespace GitHub.Services
@@ -19,6 +20,29 @@ namespace GitHub.Services
             Extensions.Guard.ArgumentNotNull(target, nameof(target));
 
             return host.ModelService.CreatePullRequest(repository, title, body, source, target);
+        }
+
+        public string GetPullRequestTemplate(ISimpleRepositoryModel repository)
+        {
+            Extensions.Guard.ArgumentNotNull(repository, nameof(repository));
+
+            var paths = new[]
+            {
+                Path.Combine(repository.LocalPath, "PULL_REQUEST_TEMPLATE"),
+                Path.Combine(repository.LocalPath, "PULL_REQUEST_TEMPLATE.md"),
+                Path.Combine(repository.LocalPath, ".github", "PULL_REQUEST_TEMPLATE"),
+                Path.Combine(repository.LocalPath, ".github", "PULL_REQUEST_TEMPLATE.md"),
+            };
+
+            foreach (var path in paths)
+            {
+                if (File.Exists(path))
+                {
+                    try { return File.ReadAllText(path); } catch { }
+                }
+            }
+
+            return null;
         }
     }
 }
--- a/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
@@ -30,6 +30,7 @@ namespace GitHub.ViewModels
 
         readonly IRepositoryHost repositoryHost;
         readonly ISimpleRepositoryModel activeRepo;
+        readonly IPullRequestService service;
         readonly Subject<Unit> initializationComplete = new Subject<Unit>();
         bool initialized;
 
@@ -45,6 +46,7 @@ namespace GitHub.ViewModels
         {
             this.repositoryHost = repositoryHost;
             this.activeRepo = activeRepo;
+            this.service = service;
 
             var repo = GitService.GitServiceHelper.GetRepo(activeRepo.LocalPath);
             this.WhenAny(x => x.Branches, x => x.Value)
@@ -99,6 +101,9 @@ namespace GitHub.ViewModels
         {
             initialized = false;
             base.Initialize(data);
+
+            this.Title = null;
+            this.Description = service.GetPullRequestTemplate(activeRepo) ?? string.Empty;
 
             repositoryHost.ModelService.GetBranches(activeRepo)
                             .ToReadOnlyList()
--- a/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
@@ -55,9 +55,7 @@ namespace GitHub.ViewModels
                 {
                     // what do we do if there's no master?
                     TargetBranch = x.FirstOrDefault(b => b.Name == "master");
-
-                    // 'repo.Head' will be null in unit tests.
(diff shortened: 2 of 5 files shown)


src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -53,7 +55,9 @@ public class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationV
                 {
                     // what do we do if there's no master?
                     TargetBranch = x.FirstOrDefault(b => b.Name == "master");
-                    SourceBranch = x.FirstOrDefault(b => b.Name == repo.Head.FriendlyName);
+
+                    // 'repo.Head' will be null in unit tests.
+                    SourceBranch = x.FirstOrDefault(b => b.Name == repo.Head?.FriendlyName);
```
shana: I dislike changing the code to cater to unit test assumptions, it would be better if the unit tests reflect the data that's supposed to be there. If `Head` is null, that means that the repository is not initialized/invalid (and this might be a thing we have to check before we actually get here) so we'd never get to this point at all. Can we change the unit tests to set the repo head properly?

src/UnitTests/GitHub.App/ViewModels/PullRequestCreationViewModelTests.cs
```diff
@@ -33,4 +33,30 @@ public class PullRequestCreationViewModelTests : TempFileBaseClass
...
+
+        var vm = new PullRequestCreationViewModel(host, repository, service, notifications);
+        vm.Initialize(null);
+
+        Assert.Equal("Test PR template", vm.Description);
+    }
+
+    class BranchMock : LibGit2Sharp.Branch
+    {
+        public BranchMock()
```
shana: Is this being used anywhere?

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
grokys: Yeah, I wasn't happy about that either. Problem is that the `repo` comes from `GitService.GitServiceHelper.GetRepo` which is a static method and can't be mocked. I started refactoring to allow this but aborted as I kinda wanted to keep the PR to the point. I can go back to it if you'd like though.

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -20,5 +21,28 @@ public IObservable<IPullRequestModel> CreatePullRequest(IRepositoryHost host, IS
...
+
+            foreach (var path in paths)
+            {
+                if (File.Exists(path))
+                {
+                    try { return File.ReadAllText(path); } catch { }
+                }
+            }
+
+            return null;
```
shana: Maybe return `String.Empty` here instead of null? That way we can always trust the return value to be something.

src/UnitTests/GitHub.App/ViewModels/PullRequestCreationViewModelTests.cs
grokys: Ah no - that was part of the aborted refactoring I mentioned above! Will remove.

src/GitHub.App/Services/PullRequestService.cs
grokys: I return `null` because they are two different outcomes: an empty string may be caused by an empty `PULL_REQUEST_TEMPLATE`, but null means "the pull request template was not found". This may be relevant if we decide we also want auto-population of the message from the last commit as in GHfW.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
shana: Oh, `GitServiceHelper` is static for convenience so we can easily unit test things. It returns a mocked instance from the service provider when called in unit tests, created [right here](https://github.com/github/VisualStudio/blob/master/src/UnitTests/Substitutes.cs#L119)

shana: For clarification: `VisualStudio.Services.DefaultExportProvider` called in https://github.com/github/VisualStudio/blob/master/src/GitHub.Exports/Services/GitService.cs#L106 goes to https://github.com/github/VisualStudio/blob/master/src/GitHub.Exports/Services/Services.cs#L45 which grabs the `ComponentModel` property in https://github.com/github/VisualStudio/blob/master/src/GitHub.Exports/Services/Services.cs#L44 which is looked up in `GetGlobalService` here: https://github.com/github/VisualStudio/blob/master/src/GitHub.Exports/Services/Services.cs#L32-L42.

When running unit tests, `GetGlobalService` uses the mocked `IServiceProvider` created in https://github.com/github/VisualStudio/blob/master/src/UnitTests/Substitutes.cs#L103-L111, which includes mocks of all our services plus the `ComponentModel` object and other infrastructure stuff. Anything that queries a service provider or goes through `VisualStudio.Services` properties ends up in these mocked objects.

grokys: Ah yeah - I see that I do get a mocked `IRepository` there. My bad - I hadn't investigated enough and assumed that this part of the code wasn't written testably.

src/UnitTests/Substitutes.cs
```diff
@@ -31,7 +32,17 @@ internal static class Substitutes
...
-        public static IGitService IGitService { get { return Substitute.For<IGitService>(); } }
+        public static IGitService IGitService
+        {
+            get
+            {
+                var result = Substitute.For<IGitService>();
+                var repo = Substitute.For<IRepository>();
+                repo.Head.Returns(Substitute.For<Branch>());
+                result.GetRepo(Arg.Any<string>()).Returns(repo);
+                return result;
```
shana: This shouldn't be done here, these are assumptions that are going to be different depending on what we're testing (what if we want to test the behaviour when there's no Head? what if we want to have different things returned when we call `GetRepo`?). The test should be setting all of this up by itself.

The reason we have these mocked service properties is so that they can be singletons to pass to constructors and return via the service provider. They're services that always exist but are empty of any specific data, and tests customize the return values of their properties and methods depending on the situation.

All test classes derive from TestBaseClass, which has Entry and Exit virtual methods that are called by every test method at the beginning and end. If you want to do common setup for every test in a class, you can override them and set things up there.

src/UnitTests/Substitutes.cs
```diff
@@ -206,5 +217,14 @@ public static IPullRequestService GetPullRequestsService(this IServiceProvider p
         {
             return provider.GetService(typeof(IPullRequestService)) as IPullRequestService;
         }
+
+        // LibGit2Sharp's Branch class has a protected constructor, so override it here so we
+        // can use a branch in our mocks.
+        class BranchMock : Branch
+        {
+            public BranchMock()
```
shana: :eyes:

shana:
🏈  back to you!

src/UnitTests/GitHub.App/ViewModels/PullRequestCreationViewModelTests.cs
```diff
@@ -33,4 +34,31 @@ public class PullRequestCreationViewModelTests : TempFileBaseClass
         var unused = ms.Received().CreatePullRequest(repository, vm.PRTitle, String.Empty, vm.SourceBranch, vm.TargetBranch);
     }
 
+    [Fact]
+    public void TemplateIsUsedIfPresent()
+    {
+        var serviceProvider = Substitutes.ServiceProvider;
+        var service = Substitute.For<IPullRequestService>();
```
shana: BTW, if you do this instead of `service.GetPullRequestService`, the instance that's returned by `service.GetService(typeof(IPullRequestService))` is going to be different than the instance you have here. It doesn't matter for this case since there's no codepath that ends up querying the service provider, but if there was, things would break, just fyi.

grokys: Ah yeah - I guess I should fix that!

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -20,5 +21,28 @@ public IObservable<IPullRequestModel> CreatePullRequest(IRepositoryHost host, IS
...
+        public string GetPullRequestTemplate(ISimpleRepositoryModel repository)
+        {
+            Extensions.Guard.ArgumentNotNull(repository, nameof(repository));
+
+            var paths = new[]
+            {
+                Path.Combine(repository.LocalPath, "PULL_REQUEST_TEMPLATE"),
+                Path.Combine(repository.LocalPath, "PULL_REQUEST_TEMPLATE.md"),
+                Path.Combine(repository.LocalPath, ".github", "PULL_REQUEST_TEMPLATE"),
+                Path.Combine(repository.LocalPath, ".github", "PULL_REQUEST_TEMPLATE.md"),
```
shana: We could just keep the strings in an static array (without the path), and only do the Path.Combine when we're actually checking for it down below as we need it, and order the strings for likelihood (we recommend a `PULL_REQUEST_TEMPLATE.md` so we can check for that first). That way we only allocate once and we're likely to only do a single Path.Combine operation (not that it's very costly, but kinda wasteful as-is)

grokys: Sure, if you think that'd be better. However think I prefer to have everything in one place so it can be read more easily. Let me know if you have strong feelings either way.

I can definitely reorder though.

shana: It's less about readibility and more about this being a really wasteful way to do it, every time this method gets called a new array gets allocated and strings get allocated and concatenated, when we can just do it once for the names of the files and one Combine when we check, and if we would want to add more filenames to check we would only change the data (the array) and not the logic (the method). Also, this service is 30 lines, it's not going to be that hard to see the filenames if they're outside the method...

grokys: Well we still have to concatenate the strings, and either create a new array to hold those strings, or use LINQ which would involve even more allocations and a far more complex code-path ;)

But it sounds like you do have strong feelings on this, so will change.

shana:
Woot!

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -9,6 +11,14 @@ namespace GitHub.Services
     [PartCreationPolicy(CreationPolicy.Shared)]
     public class PullRequestService : IPullRequestService
     {
+        static readonly string[] TemplatePaths = new[]
+        {
+            "PULL_REQUEST_TEMPLATE.md",
```
Haacked: I believe that github.com supports `PULL_REQUEST_TEMPLATE.txt` as well. I don't think it checks the file extension.

shana: You're right. I shall steal your other code ^_^

PR #498, opened 2016-08-10
Clear PR Creation Form on cancel and successful creation

### LINKED ISSUE #479: PR Creation: Cancel doesn't clear form
If the PR creation is cancelled, the next time the form is opened it still contains the data filled in before the user cancelled.

paladique:
Fixes #479

### DIFF
--- a/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
@@ -98,13 +98,13 @@ namespace GitHub.ViewModels
                     }));
 
             CancelCommand.Subscribe(_ => ClearForm());
-           
         }
 
-        private void ClearForm()
+        void ClearForm()
         {
             this.PRTitle = string.Empty;
             this.Description = string.Empty;
+            TargetBranch = Branches.FirstOrDefault(b => b.Name == "master");
         }
 
 
@@ -112,7 +112,6 @@ namespace GitHub.ViewModels
         {
             initialized = false;
             base.Initialize(data);
-
             this.Title = null;
             this.Description = service.GetPullRequestTemplate(activeRepo) ?? string.Empty;
 
--- a/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
@@ -108,7 +108,6 @@ namespace GitHub.ViewModels
             TargetBranch = Branches.FirstOrDefault(b => b.Name == "master");
         }
 
-
         public override void Initialize([AllowNull] ViewWithData data)
         {
             initialized = false;
--- a/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
@@ -98,6 +98,7 @@ namespace GitHub.ViewModels
                     }));
 
             CancelCommand.Subscribe(_ => ClearForm());
+            createPullRequest.Subscribe(_ => ClearForm());
         }
 
         void ClearForm()
--- a/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
@@ -96,8 +96,18 @@ namespace GitHub.ViewModels
                         notifications.ShowError(error?.Message ?? ex.Message);
                         return Observable.Empty<IPullRequestModel>();
                     }));
+
+            CancelCommand.Subscribe(_ => ClearForm());
+           
         }
-        
+
+        private void ClearForm()
+        {
+            this.PRTitle = string.Empty;
+            this.Description = string.Empty;
+        }
+
+
         public override void Initialize([AllowNull] ViewWithData data)
         {
             initialized = false;
--- a/src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml.cs
+++ b/src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml.cs
@@ -24,5 +24,11 @@ namespace GitHub.VisualStudio.UI.Views
                 d(ViewModel.CreatePullRequest.Subscribe(_ => NotifyDone()));
             });
         }
+
+        private void Cancel_Click(object sender, System.Windows.RoutedEventArgs e)
+        {
+            titleText.Text = string.Empty;
+            descriptionText.Text = string.Empty;
+        }
     }
 }
--- a/src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml.cs
+++ b/src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml.cs
@@ -22,8 +22,6 @@ namespace GitHub.VisualStudio.UI.Views
             {
                 d(ViewModel.CancelCommand.Subscribe(_ => NotifyCancel()));
                 d(ViewModel.CreatePullRequest.Subscribe(_ => NotifyDone()));
-                //d(Done.Subscribe(_ => ClearForm()));
-
             });
 
             IsVisibleChanged += (s, e) =>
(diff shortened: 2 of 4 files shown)


src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml.cs
```diff
@@ -22,8 +22,6 @@ public PullRequestCreationView()
             {
                 d(ViewModel.CancelCommand.Subscribe(_ => NotifyCancel()));
                 d(ViewModel.CreatePullRequest.Subscribe(_ => NotifyDone()));
-                //d(Done.Subscribe(_ => ClearForm()));
```
paladique: Tried subscribing to the `SimpleViewUserControl`'s Done observable initially to clear the form when a PR is made, which worked but didn't seem like the right approach here. The initial idea was to set `ClearForm()` as the `OnCompleted()` parameter for `CreatePullRequest.Subscribe` but it never reached `ClearForm()`

src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml.cs
```diff
@@ -23,6 +23,18 @@ public PullRequestCreationView()
...
+            {
+                if(!IsVisible)
+                    ClearForm();
+            };
+        }
+
+        private void ClearForm()
+        {
+            titleText.Text = string.Empty;
+            descriptionText.Text = string.Empty;
```
grokys: This should all probably be done in the ViewModel: the View should only contain logic for "view" related stuff but this is all related to data. Putting stuff at the view model level also means that it can be tested with unit tests.

In `PullRequestCreationViewModel` there is `CancelCommand`: you can subscribe to this to detect when the user clicked the "Cancel" button and clear the title and description there. That way you can also create a unit test to make sure that the functionality doesn't break in future; views are impossible to unit test because you need to run the UI...

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -96,8 +96,18 @@ public class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationV
                         notifications.ShowError(error?.Message ?? ex.Message);
                         return Observable.Empty<IPullRequestModel>();
                     }));
+
+            CancelCommand.Subscribe(_ => ClearForm());
+           
+        }
+
+        private void ClearForm()
```
shana: Can you take out the `private` marker here? We don't use it in our codebase.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -96,8 +96,18 @@ public class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationV
...
+        {
+            this.PRTitle = string.Empty;
+            this.Description = string.Empty;
         }
-        
+
+
```
shana: Unnecessary whitespace is unnecessary!

src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml.cs
```diff
@@ -23,6 +23,8 @@ public PullRequestCreationView()
                 d(ViewModel.CancelCommand.Subscribe(_ => NotifyCancel()));
                 d(ViewModel.CreatePullRequest.Subscribe(_ => NotifyDone()));
             });
+
         }
+
```
shana: No need for these whitespace changes.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -96,8 +96,18 @@ public class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationV
...
+            CancelCommand.Subscribe(_ => ClearForm());
+           
+        }
+
+        private void ClearForm()
+        {
+            this.PRTitle = string.Empty;
+            this.Description = string.Empty;
         }
-
```
shana: The other thing that needs to be cleared is the TargetBranch. Setting it to `TargetBranch = Branches.FirstOrDefault(b => b.Name == "master");` should do it.

src/UnitTests/GitHub.App/ViewModels/PullRequestCreationViewModelTests.cs
```diff
@@ -61,4 +61,72 @@ public void TemplateIsUsedIfPresent()
...
+        master.Name.Returns("master");
+        ms.GetBranches(Arg.Any<ISimpleRepositoryModel>()).Returns(Observable.Return(master));
+        host.ModelService.Returns(ms);
+
+        var repository = new SimpleRepositoryModel("name", new GitHub.Primitives.UriString("http://github.com/github/stuff"));
+
+        var vm = new PullRequestCreationViewModel(host, repository, service, notifications);
+        vm.PRTitle = "PR title";
+        vm.Description = "PR Desc";
+        vm.Initialize(null);
```
shana: You probably want to set the TargetBranch to something different here, so you can test whether it actually resets or not. That means returning two IBranch objects in the observable in the `GetBranches` call above.

Also, `PRTitle` and `Description` should be set after the `Initialize()` call (which doesn't need a null parameter, it has a default value). Users can only interact with the form after the `Initialize()` call occurs, so setting those properties after is closer to real world usage.

src/UnitTests/GitHub.App/ViewModels/PullRequestCreationViewModelTests.cs
```diff
@@ -61,4 +61,72 @@ public void TemplateIsUsedIfPresent()
...
```
shana: Same thing here as above.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -96,13 +96,23 @@ public class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationV
...
+
+        void ClearForm()
+        {
+            this.PRTitle = string.Empty;
+            this.Description = string.Empty;
+            TargetBranch = Branches.FirstOrDefault(b => b.Name == "master");
         }
-        
+
+
```
shana: Sneaky whitespace sneaking in again!

src/UnitTests/GitHub.App/ViewModels/PullRequestCreationViewModelTests.cs
```diff
@@ -61,4 +61,72 @@ public void TemplateIsUsedIfPresent()
...
+        gitService.GetRepo(Arg.Any<string>()).Returns(repo);
+
+        var host = serviceProvider.GetRepositoryHosts().GitHubHost;
+        var ms = Substitute.For<IModelService>();
+        var master = Substitute.For<IBranch>();
+        master.Name.Returns("master");
+        ms.GetBranches(Arg.Any<ISimpleRepositoryModel>()).Returns(Observable.Return(master));
+        host.ModelService.Returns(ms);
+
+        var repository = new SimpleRepositoryModel("name", new GitHub.Primitives.UriString("http://github.com/github/stuff"));
```
shana: You can mock this one too, if it makes things easier

shana:
The PR creation fixes have landed on master and there's been so many changes that it's all conflicting (sorry!). It'll probably be easier to redo your changes based on master, it's going to be painful if you try to rebase and resolve all the conflicts.

master also now includes more tests which mock all the things required to run the PR creation viewmodel, so you can copy one of them to test this.

Also, right now the form is being disposed every time we leave it, which means you won't ever see the effects of your code. That's just a temporary solution until we do things properly, so having a method to reset the form is still valuable. Ideally, we'd probably have a `Reset` method in every viewmodel that brings it to a clean condition by setting everything back to the defaults.

So, TL;DR, if you remove [this line](https://github.com/github/VisualStudio/blob/master/src/GitHub.App/Controllers/UIController.cs#L343), the form won't be disposed. Rename your `ClearForm` method to `Reset`, have it do what it does now. Take one of the simpler tests like https://github.com/github/VisualStudio/blob/master/src/UnitTests/GitHub.App/ViewModels/PullRequestCreationViewModelTests.cs#L115 or https://github.com/github/VisualStudio/blob/master/src/UnitTests/GitHub.App/ViewModels/PullRequestCreationViewModelTests.cs#L181 and use that as a base for your tests.

PR #500, opened 2016-08-10
Reduce font-size for titles in PR list

donokuda:
The font size for titles in the pull request list were large, which tends to draw the eye away from the main area of Visual Studio. This PR reduces the font size back to the default size and adds text trimming for long PR titles.

**Before**

image

**After**

image

### DIFF
--- a/src/GitHub.UI/Controls/Buttons/GitHubActionLink.cs
+++ b/src/GitHub.UI/Controls/Buttons/GitHubActionLink.cs
@@ -8,10 +8,19 @@ namespace GitHub.UI
         public static readonly DependencyProperty HasDropDownProperty = DependencyProperty.Register(
             "HasDropDown", typeof(bool), typeof(GitHubActionLink));
 
+        public static readonly DependencyProperty TextTrimmingProperty = DependencyProperty.Register(
+            "TextTrimming", typeof(TextTrimming), typeof(GitHubActionLink));
+
         public bool HasDropDown
         {
             get { return (bool)GetValue(HasDropDownProperty); }
             set { SetValue(HasDropDownProperty, value); }
+        }
+
+        public TextTrimming TextTrimming
+        {
+            get { return (TextTrimming)GetValue(TextTrimmingProperty); }
+            set { SetValue(TextTrimmingProperty, value); }
         }
 
         public GitHubActionLink()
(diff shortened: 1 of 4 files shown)


src/GitHub.VisualStudio.UI/Styles/ActionLinkButton.xaml
```diff
@@ -10,7 +10,7 @@
     <Setter Property="Template">
       <Setter.Value>
         <ControlTemplate TargetType="Button">
-          <TextBlock>
+          <TextBlock TextTrimming="CharacterEllipsis">
```
donokuda: Should I be aware of instances where we don't want this to be set? Is there a better way to approach this?

shana: The better way is to have it as a templated setter (like the `Background`, `Focusable` and `Foreground` properties above). The problem there is that the `Button` object doesn't have a `TextTrimming` property (because buttons can have non-text content), so we need to switch from using a button to using a `GitHubActionLink` and add a custom DependencyProperty to it that we can then use to set the trimming.

shana:
#### I added a touch of SemiBold to the title:
font-size-trimming-weight

#### But then I removed it because it probably doesn't need it, so this is how it looks now:
font-size-trimming-weight-normal

src/GitHub.UI/Controls/Buttons/GitHubActionLink.cs
```diff
@@ -8,12 +8,21 @@ public partial class GitHubActionLink : Button
         public static readonly DependencyProperty HasDropDownProperty = DependencyProperty.Register(
             "HasDropDown", typeof(bool), typeof(GitHubActionLink));
 
+        public static readonly DependencyProperty TextTrimmingProperty = DependencyProperty.Register(
+            "TextTrimming", typeof(TextTrimming), typeof(GitHubActionLink));
```
grokys: Small thing, but you can register this property by adding an owner to the `TextBlock.TextTrimmingProperty` like so:

```
public static readonly DependencyProperty TextTrimmingProperty =
    TextBlock.TextTrimmingProperty.AddOwner(typeof(GitHubActionLink));
```

It probably won't make much difference to how the code runs but it tells the reader (at least it tells me) that this is an existing property that is being used, and that I can look up `TextBlock.TextTrimming` for more info.

shana: Good point!

donokuda:
@shana @grokys 😍 Thanks for the insights around my question! This implementation is much, much cleaner.

> I added a touch of SemiBold to the title
But then I removed it because it probably doesn't need it, so this is how it looks now

Yeah, it's hard to notice that it's bolded slightly at that font size but it's fine without it being bold.

Here's where we're at after merging master into the branch:

image

I think this is good to go, what do you all think?

grokys:
LGTM!

Issue #401, opened 2016-07-05
Description left out in PR Creation


paladique:
Made a few test PRs and noticed that the description does not show up in the PR.

![2016-07-05](https://cloud.githubusercontent.com/assets/1414307/16593611/16617f16-42b4-11e6-906a-9625b0685498.png)

![2016-07-05 1](https://cloud.githubusercontent.com/assets/1414307/16593640/3ad0fc78-42b4-11e6-9b3d-84991b2f3420.png)

grokys:
Yep, can confirm.

Issue #481, opened 2016-08-08
Closing a PR on the website does not show it as closed on the list


shana:
I closed a PR on the website so I could create a new PR from the extension for the same branch. The new one shows up, but so does the old one, both in the "Open" filter. The date on the new PR is updating every time I refresh, so for some reason it seems the status is not updating on the old PR, so the list thinks it's open when it's not.

grokys:
This also appears to be happening with new PRs being opened on the site. However after a certain amount of time (not sure, about 5 mins?) it updates. Related #407.

grokys:
Fixed by 009d24e47f51fc8526f2994e93790ddfc61cda83

PR #502, opened 2016-08-11
PR Creation: Push local branches and handle forks properly

### LINKED ISSUE #501: PR Creation: If branch hasn't been pushed, push it
If the current branch hasn't been pushed yet and the user goes to the PR creation form, we should push the branch and then create the PR.

shana:
Fixes #501

When creating PRs, we should do the following

- [x] Push local commits
- [x] Set local branch to track remote branch if it's not
- [x] If local branch is tracking a differently-named remote branch, use that remote branch name as the source for the PR
- [x] If the repo is a fork, list all branches from both local and upstream and set upstream master as the default PR target branch


shana:
Ok, this is ready for a first pass at review. There's no progress bars yet while creating a PR, but the form is disabled while it's happening. If a PR is created from a fork to a branch upstream, the list won't show any PRs, which will certainly be confusing, so we need to show a messsage indicating the link to the newly created PR, which is not done yet.

Expected behaviour:

- The list of branches should show upstream branches (prefixed with owner of repo) and list of fork branches, if the repo is a fork. If it isn't, it should only show the repo branches, no prefix
- When creating a PR from a fork, it should select the upstream default branch.
- When creating a PR from a non-fork, it should select the repo's default branch
- When creating a PR targeting the same repo, the PR should show up on the list afterwards
- When creating a PR targeting an upstream branch, a message should show up linking to the new PR on the website (it's not listed) - **this isn't happening yet**
- While loading data, form should be disabled (might be too fast to see)
- While processing a PR creation, form should be disabled

src/GitHub.App/Services/RepositoryPublishService.cs
```diff
@@ -37,13 +37,12 @@ public string LocalRepositoryName
...
         {
-            return Observable.Defer(() => Observable.Return(activeRepository))
-                .SelectMany(r => apiClient.CreateRepository(newRepository, account.Login, account.IsUser)
-                    .Select(gitHubRepo => Tuple.Create(gitHubRepo, r)))
-                    .SelectMany(repo => gitClient.SetRemote(repo.Item2, "origin", new Uri(repo.Item1.CloneUrl)).Select(_ => repo))
-                    .SelectMany(repo => gitClient.Push(repo.Item2, "master", "origin").Select(_ => repo))
-                    .SelectMany(repo => gitClient.Fetch(repo.Item2, "origin").Select(_ => repo))
-                    .SelectMany(repo => gitClient.SetTrackingBranch(repo.Item2, "master", "origin").Select(_ => repo.Item1));
+            return Observable.Defer(() => apiClient.CreateRepository(newRepository, account.Login, account.IsUser)
+                         .Select(remoteRepo => new { RemoteRepo = remoteRepo, LocalRepo = activeRepository }))
```
shana: Ooops, I forgot I included this. I've been meaning to clean up this tuple for ages and always end up not doing it. Oh well, now done.

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -19,33 +26,75 @@ public class PullRequestService : IPullRequestService
...
-                    try { return File.ReadAllText(path); } catch { }
+                    if (os.File.Exists(path))
+                    {
+                        try { ret = os.File.ReadAllText(path, Encoding.UTF8); } catch { }
+                        if (ret != null)
+                            break;
+                    }
                 }
-            }
+                return Observable.Return(ret);
```
grokys: This will return an observable that contains `null` if a template is not found, but it should probably return an empty sequence. It was returning `null` before to say "template file not found" but now we're using observables an empty sequence is a more natural fit for that,

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -19,33 +26,75 @@ public class PullRequestService : IPullRequestService
...
+            var repo = await Task.Run(() => gitService.GetRepository(sourceRepository.LocalPath));
+            var remote = await gitClient.GetHttpRemote(repo, "origin");
+            await gitClient.Push(repo, sourceBranch.Name, remote.Name);
 
-            return null;
+            if (!repo.Branches[sourceBranch.Name].IsTracking)
+                await gitClient.SetTrackingBranch(repo, sourceBranch.Name, remote.Name);
+
+            // delay things a bit to avoid a race between pushing a new branch and creating a PR on it
+            if (Splat.ModeDetector.Current.InUnitTestRunner().GetValueOrDefault())
```
grokys: Should this be delaying if we're *not* in the unit test runner?

shana: Bleh, yes.

src/GitHub.App/Services/RepositoryPublishService.cs
```diff
@@ -37,13 +37,12 @@ public string LocalRepositoryName
...
+                    .SelectMany(repo => gitClient.SetRemote(repo.LocalRepo, "origin", new Uri(repo.RemoteRepo.CloneUrl)).Select(_ => repo))
+                    .SelectMany(repo => gitClient.Push(repo.LocalRepo, "master", "origin").Select(_ => repo))
+                    .SelectMany(repo => gitClient.Fetch(repo.LocalRepo, "origin").Select(_ => repo))
+                    .SelectMany(repo => gitClient.SetTrackingBranch(repo.LocalRepo, "master", "origin").Select(_ => repo.RemoteRepo));
```
grokys: Would the intention here would be clear as:

```
            return Observable.Defer(() => apiClient.CreateRepository(newRepository, account.Login, account.IsUser)
                         .Select(remoteRepo => new { RemoteRepo = remoteRepo, LocalRepo = activeRepository }))
                    .Do(repo => gitClient.SetRemote(repo.LocalRepo, "origin", new Uri(repo.RemoteRepo.CloneUrl)))
                    .Do(repo => gitClient.Push(repo.LocalRepo, "master", "origin"))
                    .Do(repo => gitClient.Fetch(repo.LocalRepo, "origin"))
                    .Do(repo => gitClient.SetTrackingBranch(repo.LocalRepo, "master", "origin"))
                    .Select(x => x.RemoteRepo);
```

?

The `SelectMany` statements aren't _really_ "selecting" - they're used for side-effects, so `Do` seems more relevant.

grokys: Actually no, ingore this: `Do` won't wait for the previous operation to finish.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -11,63 +11,85 @@
...
-                x => x.TargetBranch,
-                x => x.SourceBranch,
-                (target, source) => new { Source = source.Value, Target = target.Value })
-                .Where(_ => initialized)
-                .Merge(initializationComplete.Select(_ => new { Source = SourceBranch, Target = TargetBranch }));
+                    x => x.TargetBranch,
+                    x => x.SourceBranch,
+                    (target, source) => new { Source = source.Value, Target = target.Value })
+                .Where(_ => Initialized)
+                .Merge(this.WhenAnyValue(x => x.Initialized).Where(x => x).Select(_ => new { Source = SourceBranch, Target = TargetBranch }));
```
grokys: Good change to remove the `initializationComplete`! However I think this could be further simplified:

```
            var branchObs = this.WhenAnyValue(
                    x => x.Initialized,
                    x => x.TargetBranch,
                    x => x.SourceBranch,
                    (init, target, source) => new { Initialized = init, Source = source, Target = target })
                .Where(x => x.Initialized);
```

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -96,68 +119,93 @@ public class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationV
                         notifications.ShowError(error?.Message ?? ex.Message);
                         return Observable.Empty<IPullRequestModel>();
                     }));
+            isExecuting = CreatePullRequest.IsExecuting.ToProperty(this, x => x.IsExecuting);
```
grokys: Are `IsExecuting` and `IsBusy` not the same thing? Are they related?

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -96,68 +119,93 @@ public class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationV
...
+            get { return branchesList; }
+            set
+            {
+                branchesList = value;
+                this.RaisePropertyChanged();
+                this.RaisePropertyChanged(nameof(Branches));
+            }
         }
 
+        public IReadOnlyList<IBranch> Branches => branchesList;
```
grokys: Why are both `Branches` and `BranchesList` needed? The branches are never mutated after they're set so can't there be just `Branches`?

```
        IReadOnlyList<IBranch> branches;
        public IReadOnlyList<IBranch> Branches
        {
            get { return branches; }
            private set
            {
                branches = value;
                this.RaisePropertyChanged();
            }
        }
```

src/GitHub.Exports/Extensions/SimpleRepositoryModelExtensions.cs
```diff
@@ -10,7 +10,7 @@ public static class SimpleRepositoryModelExtensions
     {
         public static bool HasCommits(this ISimpleRepositoryModel repository)
         {
-            var repo = GitService.GitServiceHelper.GetRepo(repository.LocalPath);
+            var repo = GitService.GitServiceHelper.GetRepository(repository.LocalPath);
```
grokys: :+1: to not using abbreviations! :)

src/GitHub.Exports/Models/IBranch.cs
```diff
@@ -1,7 +1,15 @@
...
 {
-    public interface IBranch
+    public interface IBranch : ICopyable<IBranch>,
+        IEquatable<IBranch>, IComparable<IBranch>
     {
+        string Id { get; }
         string Name { get; }
+        ISimpleRepositoryModel Repository { get; set; }
+        bool IsTracking { get; set; }
+        string DisplayName { get; set; }
```
grokys: Could some of these properties be made readonly? It kinda worries me to have them all read/write as it can be difficult to tell how they can change. In what circumstances can they change?

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
shana: `IsBusy` reflects a combination of multiple states that signify the vm is doing something and so data should not be externally changed until that something is finished (when loading branches, initializing, executing the creation, etc)

I really don't like having to have this `IsExecuting` property just to listen to the changes in the observable of the button, but right now it's the best way to combine it with the rest of the property changes. I really want to find a cleaner way of doing this next week.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
shana: They were mutated at one point in this rewrite, but they're not anymore. I'll fixicate!

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
grokys: :+1: i guess this will become clear when the `IsBusy` state is actually updated.

grokys: Oh, and then looking 3 lines below I see it is. I missed that previously, it would've answered my question.

src/GitHub.Exports/Models/IBranch.cs
shana: Remnants of prototyping. `DisplayName` is the only one that's currently modified externally based on what the viewmodel wants the branch to look like in lists.

grokys: Ok - yeah it makes sense for `DisplayName` to be mutable to me.

src/GitHub.Exports/Models/SimpleRepositoryModel.cs
```diff
@@ -133,39 +144,50 @@ public string HeadSha
...
-            return other != null && String.Equals(Name, other.Name) && String.Equals(CloneUrl, other.CloneUrl) && String.Equals(LocalPath?.TrimEnd('\\'), other.LocalPath?.TrimEnd('\\'), StringComparison.CurrentCultureIgnoreCase);
+            return other != null && String.Equals(Name, other.Name) && String.Equals(Owner, other.Owner) && String.Equals(CloneUrl, other.CloneUrl) && String.Equals(LocalPath?.TrimEnd('\\'), other.LocalPath?.TrimEnd('\\'), StringComparison.CurrentCultureIgnoreCase);
         }
 
         bool IEquatable<SimpleRepositoryModel>.Equals(SimpleRepositoryModel other)
         {
             if (ReferenceEquals(this, other))
                 return true;
-            return other != null && String.Equals(Name, other.Name) && String.Equals(CloneUrl, other.CloneUrl) && String.Equals(LocalPath?.TrimEnd('\\'), other.LocalPath?.TrimEnd('\\'), StringComparison.CurrentCultureIgnoreCase);
+            return other != null && String.Equals(Name, other.Name) && String.Equals(Owner, other.Owner) && String.Equals(CloneUrl, other.CloneUrl) && String.Equals(LocalPath?.TrimEnd('\\'), other.LocalPath?.TrimEnd('\\'), StringComparison.CurrentCultureIgnoreCase);
```
grokys: Kinda worries me to see this long line copy/pasted in 2 places? Could you make `Equals(SimpleRepositoryModel)` a normal method (not an explicit interface method) and call `Equals(SimpleRepositoryModel)` from `Equals(object)`?

shana: Yes, definitely should not have this in two places. Good catch!

src/GitHub.Exports/Services/GitService.cs
```diff
@@ -69,37 +49,23 @@ public static UriString GetUriFromPath(string path)
...
-        }
-
-        /// <summary>
         /// Returns a <see cref="UriString"/> representing the uri of the "origin" remote with no modifications.
         /// </summary>
         /// <param name="repo"></param>
+        /// <param name="remote">The name of the remote to look for</param>
         /// <returns></returns>
-        public static UriString GetOriginUri(IRepository repo)
+        public UriString GetOriginUri(IRepository repo, string remote = "origin")
```
grokys: This probably should now be called "GetRemoteUri" and shouldn't have "origin" in its doc comments any more.

shana:
Ok, we're now showing notifications when a PR is created, with a link to the created PR on the website. If the PR is created upstream, it won't show up on the PR list but the user will be able to go to it so hopefully it won't be too confusing.

This is how it works (the url opened is wrong in this gif but that's already fixed)
creating-a-pr-upstream

src/GitHub.Exports/ViewModels/IInfoPanel.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace GitHub.ViewModels
+{
+    public interface IInfoPanel
+    {
+        string Message { get; set; }
+        MessageType MessageType { get; set; }
+    }
```
shana: Interface probably not needed, but added it anyway to have a place to keep the `MessageType` enum and be explicit about things (maybe unit test it later, too)

shana:
This also fixes #401 and #481

Issue #503, opened 2016-08-14
Cannot create repository,vs crash


TracyDa:
I cannot use it to create repository ,it will make my visual studio crash.
Really disappointed！

shana:
Can you post screenshots of the steps you do up to the crash, and send us the logs as described in https://github.com/github/VisualStudio/blob/master/CONTRIBUTING.md#submitting-an-issue? Thanks!

TracyDa:
OK,I have sent you email with the screenshot and log file.Hope you can fix it.

shana:
@TracyDa Thanks, this is helpful!

Looks like we have a bug handling the dropdown list in certain locales:
```
System.NullReferenceException: δ??????????????ʵ????
   ? GitHub.UI.IsLastItemInContainerConverter.Convert(Object value, Type targetType, Object parameter, CultureInfo culture) λ? Y:\jenkins\workspace\GitHub for Visual Studio speakeasy jenkinskeep\src\GitHub.UI\Converters\IsLastItemInContainerConverter.cs:???17
   ? System.Windows.Data.BindingExpression.TransferValue(Object newValue, Boolean isASubPropertyChange)
   ? System.Windows.Data.BindingExpression.HandlePropertyInvalidation(DependencyObject d, DependencyPropertyChangedEventArgs args)
   ? System.Windows.Data.BindingExpressionBase.OnPropertyInvalidation(DependencyObject d, DependencyPropertyChangedEventArgs args)
   ? System.Windows.Data.BindingExpression.OnPropertyInvalidation(DependencyObject d, DependencyPropertyChangedEventArgs args)
   ? System.Windows.DependentList.InvalidateDependents(DependencyObject source, DependencyPropertyChangedEventArgs sourceArgs)
   ? System.Windows.DependencyObject.NotifyPropertyChange(DependencyPropertyChangedEventArgs args)
   ? System.Windows.DependencyObject.UpdateEffectiveValue(EntryIndex entryIndex, DependencyProperty dp, PropertyMetadata metadata, EffectiveValueEntry oldEntry, EffectiveValueEntry& newEntry, Boolean coerceWithDeferredReference, Boolean coerceWithCurrentValue, OperationType operationType)
[...]

PR #506, opened 2016-08-17
Added null check in IsLastItemInContainerConverter.

grokys:
#503 is a crash that seems to occur here. Can't reproduce locally but there's only so much that can cause an NRE in this method so try to avoid it in future.

### DIFF
--- a/src/GitHub.UI/Converters/IsLastItemInContainerConverter.cs
+++ b/src/GitHub.UI/Converters/IsLastItemInContainerConverter.cs
@@ -6,27 +6,20 @@ using System.Windows.Data;
 
 namespace GitHub.UI
 {
-	// http://stackoverflow.com/questions/12125764/change-style-of-last-item-in-listbox
-	public class IsLastItemInContainerConverter : IValueConverter
-	{
-		public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
-		{
-			DependencyObject item = (DependencyObject)value;
-			ItemsControl ic = ItemsControl.ItemsControlFromItemContainer(item);
+    // http://stackoverflow.com/questions/12125764/change-style-of-last-item-in-listbox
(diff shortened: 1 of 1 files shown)


src/GitHub.UI/Converters/IsLastItemInContainerConverter.cs
```diff
@@ -14,8 +14,15 @@ public object Convert(object value, Type targetType, object parameter, CultureIn
...
-		}
+            if (ic != null && ic.Items != null)
+            {
+                return ic.ItemContainerGenerator.IndexFromContainer(item) == ic.Items.Count - 1;
+            }
+            else
+            {
+                return false;
+            }
+        }
```
shana: Indentation seems heavily screwed up here.

Also, this is probably enough:

```
return ic != null ? ic.ItemContainerGenerator.IndexFromContainer(item) == ic.Items.Count - 1 : false;
```

If `ic` is not null than `Items` won't be null either (I peeked, it's created in the getter).

src/GitHub.UI/Converters/IsLastItemInContainerConverter.cs
```diff
@@ -6,27 +6,20 @@
...
 
-            if (ic != null && ic.Items != null)
-            {
-                return ic.ItemContainerGenerator.IndexFromContainer(item) == ic.Items.Count - 1;
-            }
-            else
-            {
-                return false;
-            }
+            return ic?.ItemContainerGenerator.IndexFromContainer(item) == ic.Items.Count - 1;
```
shana: Won't it still crash on the right-hand side in `ic.Items.Count`?

grokys: D'oh!

shana:
:rocket:

PR #510, opened 2016-08-19
[WIP] Generate Copy/Open link based on the latest commit

### LINKED ISSUE #343: Link generation for Open/Link to GitHub should use the latest commit of the current file
The "Open on GitHub" feature always seems to try to open the file on master - if I'm on a branch that file might not exist or be completely different. It would be good to open the file on the correct branch on the site.

paladique:
Fixes #343

### DIFF
--- a/src/GitHub.Exports/Models/ISimpleRepositoryModel.cs
+++ b/src/GitHub.Exports/Models/ISimpleRepositoryModel.cs
@@ -1,6 +1,7 @@
 ﻿using GitHub.Primitives;
 using GitHub.UI;
 using System.ComponentModel;
+using System.Threading.Tasks;
 
 namespace GitHub.Models
 {
@@ -20,6 +21,6 @@ namespace GitHub.Models
         /// Updates the url information based on the local path
         /// </summary>
         void Refresh();
-        UriString GenerateUrl(string path = null, int startLine = -1, int endLine = -1);
+        Task<UriString> GenerateUrl(string path = null, int startLine = -1, int endLine = -1);
     }
 }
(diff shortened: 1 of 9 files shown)


shana:
Probably worth checking whether the branch is tracking a remote branch and whether the commit hashes match there, and then maybe checking if any commits are reachable on that tracking branch, as an optimization, before going on enumerating all the remotes branches and all the local commits. Very likely the branch has been pushed at least once so that might hit a large chunk of use cases.

paladique:
@shana so I want to test this new method `GetLatestPushedSha` (GitServiceTests seemed fitting), and I think I need to mock a directory like in `SimpleRepositoryModelTests`, but the base classes between the two are different so I can't create a local directory with a `TempBaseClass`.

Didn't push this yet, but if I swap `TempBaseClass`  for `TempFileBaseClass` in `GitServiceTests` it didn't seem to break anything so I'm wondering if to get access to methods like `Directory.CreateSubdirectory` is a fair reason to switch the base class.

paladique:
I realize I might need to make a new PR with the merged refactor in master, but wanna get through this test before I do that

shana:
> Didn't push this yet, but if I swap TempBaseClass for TempFileBaseClass in GitServiceTests it didn't seem to break anything so I'm wondering if switching the base class is fair reason to get access to methods like Directory.CreateSubdirectory

The TempFileBaseClass has some... issues. It's kinda fundamentally flawed,  https://github.com/github/VisualStudio/pull/544 takes a different approach on that. It's probably best to close this PR and leave it be as-is, create a new branch from master, port your changes over to it, to start with. Then it'll be easier to write tests using temp dirs. If you port over this PR to master within the next three hours, I'll still be around to review it!

PR #514, opened 2016-08-22
Record Visual Studio version properly

### LINKED ISSUE #496: Record a more precise VS version in usage stats.
When we send usage stats for the extension, the VS version is sent as e.g. "14.0". We need to  find a way to discover the specific update version that is shown in VS the about page.

shana:
Fixes #496


paladique:
:sparkles:

PR #520, opened 2016-08-26
Add a loader for external packages

### LINKED ISSUE #522: Start Page support for VS "15" Preview 4
VS "15" Preview 4 ships a revamped Start Page: https://www.visualstudio.com/en-us/news/releasenotes/vs15-relnotes#vside

Add support for it.

shana:
Fixes #522 

This PR adds a way to load external packages that are included in the vsix and expose the `IUIProvider` as a global service, so those packages can make use of our services.

The way it works is that we now ship two packages - one that only loads when the git ui context loads (as before), which tracks app startups and loads menus and things, and one that loads as soon as VS loads and registers the `IUIProvider` service as a global service. This service loader package is an async package, so it loads in the background, and doesn't create the `IUIProvider` instance until someone requests it via a `GetService` call.

This PR also includes a submodule for the [StartPage package](https://github.com/editor-tools/StartPage), includes it in the vsix if it is available, and only loads it at runtime if we're running on VS "15" Preview 4 (and only then).

I've also added a little change to make sure we only record a clone metric if the clone method returns successfully. It doesn't necessarily mean that the clone is completed successfully, but we can't track that so this is the best we can do right now.

@haacked I could really use your 👀 on this one.

### DIFF
--- a/src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs
+++ b/src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs
@@ -22,6 +22,7 @@ using Rothko;
 using System.Collections.ObjectModel;
 using GitHub.Collections;
 using GitHub.UI;
+using GitHub.Extensions.Reactive;
 
 namespace GitHub.ViewModels
 {
@@ -152,12 +153,16 @@ namespace GitHub.ViewModels
                     notificationService.ShowError(Resources.RepositoryCloneFailedNoSelectedRepo);
                     return Observable.Return(Unit.Default);
                 }
-                
+
                 // The following is a noop if the directory already exists.
                 operatingSystem.Directory.CreateDirectory(BaseRepositoryPath);
 
                 return cloneService.CloneRepository(repository.CloneUrl, repository.Name, BaseRepositoryPath)
-                    .Do(_ => this.usageTracker.IncrementCloneCount());
+                    .ContinueAfter(() =>
+                    {
+                        usageTracker.IncrementCloneCount();
+                        return Observable.Return(Unit.Default);
+                    });
             })
             .SelectMany(_ => _)
             .Catch<Unit, Exception>(e =>
--- a/src/GitHub.Exports/Helpers/AssemblyResolver.cs
+++ b/src/GitHub.Exports/Helpers/AssemblyResolver.cs
@@ -52,7 +52,7 @@ namespace GitHub.Helpers
             {
                 var requestedName = e.Name.TrimSuffix(".dll", StringComparison.OrdinalIgnoreCase);
                 var name = new AssemblyName(requestedName).Name;
-                if (!ourAssemblies.Contains(name, new LambdaComparer<string>((lhs, rhs) => String.Compare(lhs, rhs, StringComparison.OrdinalIgnoreCase))))
+                if (!ourAssemblies.Contains(name, StringComparer.OrdinalIgnoreCase))
                     return null;
                 var path = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
                 var filename = Path.Combine(path, name + ".dll");
--- a/src/GitHub.Exports/Helpers/AssemblyResolver.cs
+++ b/src/GitHub.Exports/Helpers/AssemblyResolver.cs
@@ -7,6 +7,8 @@ using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.VisualStudio.Shell;
+using System.Diagnostics;
 
 namespace GitHub.Helpers
 {
@@ -47,13 +49,15 @@ namespace GitHub.Helpers
         {
             try
             {
-                var name = new AssemblyName(e.Name);
-                if (!ourAssemblies.Contains(name.Name))
+                var requestedName = e.Name.TrimSuffix(".dll", StringComparison.OrdinalIgnoreCase);
+                var name = new AssemblyName(requestedName).Name;
+                if (!ourAssemblies.Contains(name))
                     return null;
                 var path = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
-                var filename = Path.Combine(path, name.Name + ".dll");
+                var filename = Path.Combine(path, name + ".dll");
                 if (!File.Exists(filename))
                     return null;
+
                 return Assembly.LoadFrom(filename);
             }
             catch (Exception ex)
--- a/src/GitHub.Exports/Helpers/AssemblyResolver.cs
+++ b/src/GitHub.Exports/Helpers/AssemblyResolver.cs
@@ -9,6 +9,7 @@ using System.Text;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.Shell;
 using System.Diagnostics;
+using GitHub.Collections;
 
 namespace GitHub.Helpers
 {
@@ -51,7 +52,7 @@ namespace GitHub.Helpers
             {
                 var requestedName = e.Name.TrimSuffix(".dll", StringComparison.OrdinalIgnoreCase);
                 var name = new AssemblyName(requestedName).Name;
-                if (!ourAssemblies.Contains(name))
+                if (!ourAssemblies.Contains(name, new LambdaComparer<string>((lhs, rhs) => String.Compare(lhs, rhs, StringComparison.OrdinalIgnoreCase))))
                     return null;
                 var path = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
                 var filename = Path.Combine(path, name + ".dll");
--- a/src/GitHub.VisualStudio/GitHubPackage.cs
+++ b/src/GitHub.VisualStudio/GitHubPackage.cs
@@ -82,9 +82,9 @@ namespace GitHub.VisualStudio
     [Guid(StartPagePackageId)]
     public sealed class ServiceProviderPackage : AsyncPackage
     {
-        static Version vsversion;
-        public static Version VSVersion
         const string StartPagePackageId = "D5CE1488-DEDE-426D-9E5B-BFCCFBE33E53";
+        Version vsversion;
+        Version VSVersion
         {
             get
             {
--- a/src/GitHub.VisualStudio/GitHubPackage.cs
+++ b/src/GitHub.VisualStudio/GitHubPackage.cs
@@ -79,11 +79,12 @@ namespace GitHub.VisualStudio
     [ProvideService(typeof(IUIProvider), IsAsyncQueryable = true)]
     [ProvideAutoLoad(UIContextGuids.NoSolution)]
     [ProvideAutoLoad(UIContextGuids.SolutionExists)]
-    [Guid("D5CE1488-DEDE-426D-9E5B-BFCCFBE33E53")]
+    [Guid(StartPagePackageId)]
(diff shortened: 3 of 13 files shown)


Haacked:
My :eyes: are on the prize!

src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs
```diff
@@ -152,12 +153,16 @@ IObservable<Unit> OnCloneRepository(object state)
                     notificationService.ShowError(Resources.RepositoryCloneFailedNoSelectedRepo);
                     return Observable.Return(Unit.Default);
                 }
-                
+
                 // The following is a noop if the directory already exists.
                 operatingSystem.Directory.CreateDirectory(BaseRepositoryPath);
 
                 return cloneService.CloneRepository(repository.CloneUrl, repository.Name, BaseRepositoryPath)
-                    .Do(_ => this.usageTracker.IncrementCloneCount());
```
Haacked: Oh good call. Does `CloneRepository` send multiple signals?

Haacked:
Weird. This is failing on Janky but fine on appveyor.

src/GitHub.Exports/Helpers/AssemblyResolver.cs
```diff
@@ -47,13 +49,15 @@ static Assembly LoadAssemblyFromRunDir(object sender, ResolveEventArgs e)
         {
             try
             {
-                var name = new AssemblyName(e.Name);
-                if (!ourAssemblies.Contains(name.Name))
+                var requestedName = e.Name.TrimSuffix(".dll", StringComparison.OrdinalIgnoreCase);
```
Haacked: Today I learned about `TrimSuffix` included in `PackageUtilities`. :smile:

src/GitHub.Exports/Helpers/AssemblyResolver.cs
```diff
@@ -47,13 +49,15 @@ static Assembly LoadAssemblyFromRunDir(object sender, ResolveEventArgs e)
...
+                var name = new AssemblyName(requestedName).Name;
+                if (!ourAssemblies.Contains(name))
```
Haacked: To be on the safe side, should we make this comparison case insensitive?

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -80,4 +74,60 @@ public GHClient(IProgram program)
...
+                    {
+                        // this will return Microsoft.VisualStudio.Shell.Immutable.14.0 in VS15
+                        // but Microsoft.VisualStudio.Shell.Framework in Dev15
+                        var vinfo = FileVersionInfo.GetVersionInfo(asm.Location);
+                        vsversion = new Version(vinfo.FileMajorPart, vinfo.FileMinorPart, vinfo.FileBuildPart, vinfo.FilePrivatePart);
+                    }
+                    catch
+                    {
+                        // something wrong, fallback to assembly version
+                        vsversion = asm.GetName().Version;
```
Haacked: Do we ever expect this to go wrong? Should we log it?

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -80,4 +74,60 @@ public GHClient(IProgram program)
...
+                return vsversion;
+            }
+        }
+
+        protected override async tasks.Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
+        {
+            AddService(typeof(IUIProvider), CreateService, true);
+
+            // Load the start page package only for Dev15 Preview 4
+            if (VSVersion.Major == 15 && VSVersion.Build == 25618)
```
Haacked: What happens post Preview 4? Should this be `VSVersion.Build >= 25618` for example?

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -80,4 +74,60 @@ public GHClient(IProgram program)
...
+            {
+                var shell = await GetServiceAsync(typeof(SVsShell)) as IVsShell;
+                IVsPackage startPagePackage;
+                if (ErrorHandler.Failed(shell?.LoadPackage(new Guid("3b764d23-faf7-486f-94c7-b3accc44a70d"), out startPagePackage) ?? -1))
+                {
+                    // ¯\_(ツ)_/¯
```
Haacked: Can we display a full screen vector art version of this ascii drawing in this case and play the sad trombone music?

:trollface:

src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs
shana: Not right now, but one of my alternate implementations used a custom clone service using libgit2sharp directly instead of the TeamExplorer service, overriding the default service just for the start page cloning, and that one signaled multiple times :-P

src/GitHub.Exports/Helpers/AssemblyResolver.cs
shana: oh yes, indeed!

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -80,4 +74,60 @@ public GHClient(IProgram program)
...
+    [ProvideAutoLoad(UIContextGuids.SolutionExists)]
+    [Guid("D5CE1488-DEDE-426D-9E5B-BFCCFBE33E53")]
+    public sealed class ServiceProviderPackage : AsyncPackage
+    {
+        static Version vsversion;
+        public static Version VSVersion
+        {
+            get
+            {
+                if (vsversion == null)
```
Haacked: Given that `InitializeAsync` is an async method, are we sure that access to this `VSVersion` property is synchronized? I'd be worried about a race condition here, in which case I think making use of `Lazy<Version>` might be warranted.

src/GitHub.VisualStudio/GitHubPackage.cs
shana: Yeah, but without good exception telemetry, we're not going to see the exceptions unless a user explicitly sends the logs, and why would they, it never fails :-P

Need to send things to haystack, is the conclusion.

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -80,4 +74,60 @@ public GHClient(IProgram program)
...
+        protected override async tasks.Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
+        {
+            AddService(typeof(IUIProvider), CreateService, true);
+
+            // Load the start page package only for Dev15 Preview 4
+            if (VSVersion.Major == 15 && VSVersion.Build == 25618)
+            {
+                var shell = await GetServiceAsync(typeof(SVsShell)) as IVsShell;
+                IVsPackage startPagePackage;
+                if (ErrorHandler.Failed(shell?.LoadPackage(new Guid("3b764d23-faf7-486f-94c7-b3accc44a70d"), out startPagePackage) ?? -1))
```
Haacked: Should this GUID be a constant like some of the others? It's not clear to me what that guid represents. I assume it's the package id for the Start Page package.

src/GitHub.VisualStudio/GitHubPackage.cs
shana: Post preview 4 the interfaces are going to have breaking changes. At that time we probably update the package to those new types, bump the version check, and stop loading this on preview 4.

Haacked:
Done with my review. This is some deep magic going on here. :stuck_out_tongue: Just a few comments and concerns to address, but as far as I can tell, looks like a good approach.

src/GitHub.VisualStudio/GitHubPackage.cs
shana: Yeah. Needs to be clearer, and we should probably change it every time it has incompatible changes, just in case something is cached somewhere.

src/GitHub.VisualStudio/GitHubPackage.cs
shana: hmmmmmmmm! tempting....

src/GitHub.VisualStudio/GitHubPackage.cs
shana: `InitializeAsync` only runs once (guaranteed), it's just a way of loading packages in a background thread, so reentrancy shouldn't be a problem right now, given that this property is only used for the package check. It might be an issue if other code uses it though, so maybe it shouldn't be static nor public.

shana:
> Weird. This is failing on Janky but fine on appveyor.

That's because the bot that appveyor uses to checkout private repos has permissions on the repo where this stuff is coming from, but janky's doesn't yet, so it blows up on checkout.

src/GitHub.Exports/Helpers/AssemblyResolver.cs
```diff
@@ -51,7 +52,7 @@ static Assembly LoadAssemblyFromRunDir(object sender, ResolveEventArgs e)
             {
                 var requestedName = e.Name.TrimSuffix(".dll", StringComparison.OrdinalIgnoreCase);
                 var name = new AssemblyName(requestedName).Name;
-                if (!ourAssemblies.Contains(name))
+                if (!ourAssemblies.Contains(name, new LambdaComparer<string>((lhs, rhs) => String.Compare(lhs, rhs, StringComparison.OrdinalIgnoreCase))))
```
Haacked: I think you could just pass in `StringComparer.OrdinalIgnoreCase` here rather than the `LambdaComparer`. :smile:

src/GitHub.VisualStudio/Services/UIProvider.cs
```diff
@@ -105,7 +134,11 @@ public object TryGetService(Type serviceType)
...
                 return instance;
 
-            instance = serviceProvider.GetService(serviceType);
+            instance = serviceProvider?.GetService(serviceType);
+            if (instance != null)
+                return instance;
+
+            instance = asyncServiceProvider?.GetServiceAsync(serviceType);
             if (instance != null)
                 return instance;
```
Haacked: One thought, perhaps collapse the above few lines to

```csharp
var instance = AddToDisposables(tempContainer.GetExportedValueOrDefault<object>(contract))
    ?? AddToDisposables(ExportProvider.GetExportedValues<object>(contract).FirstOrDefault(x => contract.StartsWith("github.", StringComparison.OrdinalIgnoreCase) ? x.GetType().Assembly.GetName().Version == currentVersion : true))
    ?? serviceProvider?.GetService(serviceType)
    ?? asyncServiceProvider?.GetServiceAsync(serviceType);

if (instance != null)
    return instance;
```

Haacked:
:turtle:

src/GitHub.Exports/Helpers/AssemblyResolver.cs
shana: Oh duh. My brain today...

andreia
github.com/shana
@sh4na <http://twitter.com/sh4na>

src/GitHub.VisualStudio/Services/UIProvider.cs
shana: I want to be clear about the ordering of these loaders, and I need to add some documentation for it, because the ordering is important (and we may use it to override the default MEF exports altogether), so I'm going to clarity and spacing here.

PR #523, opened 2016-08-26
Clear Pull Request Form when canceled or completed.  

### LINKED ISSUE #479: PR Creation: Cancel doesn't clear form

paladique:
Fixes #479 

Instead subscribing to the cancel and completion actions of the PullRequestCreation view model, `ResetViewModel()` was added to `IViewModel`  that will execute the VMs overridden `BaseViewModel`'s `Reset()`,

### DIFF
--- a/src/GitHub.App/Controllers/UIController.cs
+++ b/src/GitHub.App/Controllers/UIController.cs
@@ -339,7 +339,8 @@ namespace GitHub.Controllers
                 .OnEntry(tr => RunView(UIViewType.PRCreation, CalculateDirection(tr)))
                 .PermitDynamic(Trigger.Next, () => Go(Trigger.Next))
                 .PermitDynamic(Trigger.Cancel, () => Go(Trigger.Cancel))
-                .PermitDynamic(Trigger.Finish, () => Go(Trigger.Finish));
+                .PermitDynamic(Trigger.Finish, () => Go(Trigger.Finish))
+                .OnExit(() => Reset());
 
             uiStateMachine.Configure(UIViewType.Login)
                 .OnEntry(tr => RunView(UIViewType.Login, CalculateDirection(tr)))
--- a/src/GitHub.App/Controllers/UIController.cs
+++ b/src/GitHub.App/Controllers/UIController.cs
@@ -339,8 +339,7 @@ namespace GitHub.Controllers
                 .OnEntry(tr => RunView(UIViewType.PRCreation, CalculateDirection(tr)))
                 .PermitDynamic(Trigger.Next, () => Go(Trigger.Next))
                 .PermitDynamic(Trigger.Cancel, () => Go(Trigger.Cancel))
-                .PermitDynamic(Trigger.Finish, () => Go(Trigger.Finish))
-                .OnExit(() => DisposeView(activeFlow, UIViewType.PRCreation));
+                .PermitDynamic(Trigger.Finish, () => Go(Trigger.Finish));
 
             uiStateMachine.Configure(UIViewType.Login)
                 .OnEntry(tr => RunView(UIViewType.Login, CalculateDirection(tr)))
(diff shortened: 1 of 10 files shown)


src/GitHub.App/SampleData/SampleViewModels.cs
```diff
@@ -122,12 +122,6 @@ public ReactivePropertyValidator<string> RepositoryNameValidator
             private set;
         }
 
-        public ICommand Reset
```
paladique: I had to the option to remove or hide this with `new`, I decided to remove it since this was not in use anywhere,

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -186,6 +186,13 @@ public void Dispose()
             GC.SuppressFinalize(this);
         }
 
+       public override void Reset()
+        {
```
shana: Indentation is weird here.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -186,6 +186,13 @@ public void Dispose()
...
+            this.PRTitle = string.Empty;
+            this.Description = string.Empty;
```
shana: You don't need the `this.` qualifiers here, that is only required in constructors when you need to distinguish between a constructor argument and a field or property that have the same name.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -186,6 +186,13 @@ public void Dispose()
...
+            TargetBranch = Branches.FirstOrDefault(b => b.Name == "master");
```
shana: The `TargetBranch` default value is not hardcoded to master anymore, it's whatever it gets set to in https://github.com/github/VisualStudio/blob/35d1a987f7750d3360e2a6718a823fbf635f66c6/src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs#L69 from the value in the `GitHubRepository` property.

shana:
So to avoid adding the `Reset` method to the `GitHubPaneViewModel`, which doesn't need it, an more interesting implementation can be:

- Add a new interface called `IResettable` and move the `Reset` method definition in it
- Add the new interface to the list of interfaces that the `PullRequestCreationViewModel` implements
- In the `UIController` reset method, check whether the view model is an `IResettable` and only call the method if it is.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
paladique: At this point, I'm assuming I'll just need the `FirstOrDefault` `IBranch` and not have to search by name?

`Branches.FirstOrDefault(b => b == TargetBranch)`

src/GitHub.App/Controllers/UIController.cs
```diff
@@ -421,8 +421,7 @@ void ResetViewModel(UIViewType viewType)
             if (flowList.TryGetValue(viewType, out pair))
             {
                 var vm = pair.ViewModel as IResettable;
-                //TODO:Can I even do this?
-                //vm?.Reset();
+                vm?.Reset();
```
paladique: This seems like a handy C# 6 thing! What I'm doing here is using the null conditional operator to  speculate that the VM could be a member of `IResettable` and to access `Reset()` if it is, so if `vm` is null, and the operator avoids a null reference exception

shana: Yup, this is a very handy thing indeed!

shana:
@paladique I forget, did we figure out why these tests fail in CI? Do they pass in Visual Studio?

paladique:
@shana ran the tests in VS and it's failing on the new test added for the `Reset()` method, `ResetIsCalled()`. Clearing the form is successful when doing it in the UI, but in the test `vm?.Reset()` is skipped because the `PullRequestCreationProxy` is not being recognized as a member of `IResettable` 

Should I be mocking IResettable in some way? That seems unnecessary because `PullRequestCreation` is already a member. This must be missing something else.

paladique:
This was mostly an exploratory PR, so closing for now!

PR #524, opened 2016-08-26
Use ProjectItem.Name for file name consistency 

### LINKED ISSUE #472: Open on GitHub sometimes 404s when path contains upper case
- GitHub Extension version: 1.0.18.6

Right click on code windows and select 'GitHub > Open on GitHub'. If the source file or directory contains an upper case char, the GitHub page sometimes opens with a 404.

It appears there is an issue with Visual Studio. If a source file was opened automatically when a solution loads, `Document.Name` will return an all lower case path. If the user explicitly opens the document after the solution loads, the path will contain upper and lower chase chars es expected.

You can see what's happening here. I've executed the `go` method using TestDriven.Net's `Test With > In-Proc` command:

document name

The highlighted `netcoretestrunnerlibrarytests.cs` was opened with the solution, but `Scratch.cs` was opened (or created) after the solution was opened. You can see `ProjectItem.Name` always returns a path with the expected casing, but with `Document.Name` it varies.

I'm guessing you could fix it by changing the following line:
https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio/Menus/LinkMenuBase.cs#L21

to...

`return repo.GenerateUrl(activeDocument.ProjectItem.Name, activeDocument.StartLine, activeDocument.EndLine);`

I hope that's useful. :smile: 

You can find discussion about it on Twitter here:
https://twitter.com/jcansdale/status/760547405926395904

paladique:
Fixes #472 per suggestions.

### DIFF
--- a/src/GitHub.VisualStudio/Helpers/ActiveDocumentSnapshot.cs
+++ b/src/GitHub.VisualStudio/Helpers/ActiveDocumentSnapshot.cs
@@ -19,7 +19,7 @@ namespace GitHub.VisualStudio
         public ActiveDocumentSnapshot([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider)
         {
             StartLine = EndLine = -1;
-            Name = Services.Dte2?.ActiveDocument?.FullName;
+            Name = Services.Dte2?.ActiveDocument?.ProjectItem.Name;
 
             var textManager = serviceProvider.GetService(typeof(SVsTextManager)) as IVsTextManager;
             Debug.Assert(textManager != null, "No SVsTextManager service available");
--- a/src/UnitTests/GitHub.Exports/SimpleRepositoryModelTests.cs
+++ b/src/UnitTests/GitHub.Exports/SimpleRepositoryModelTests.cs
@@ -44,6 +44,8 @@ public class SimpleRepositoryModelTests : TempFileBaseClass
     [InlineData(17, true, "https://github.com/foo/bar", "", @"src\dir\file1.cs", -1, 2, "https://github.com/foo/bar")]
     [InlineData(18, true, null, "123123", @"src\dir\file1.cs", 1, 2, null)]
     [InlineData(19, false, "git@github.com/foo/bar", "123123", @"src\dir\file1.cs", -1, -1, "https://github.com/foo/bar/blob/123123/src/dir/file1.cs")]
+    [InlineData(20, false, "git@github.com/foo/bar", "123123", @"src\dir\File1.cs", -1, -1, "https://github.com/foo/bar/blob/123123/src/dir/File1.cs")]
+    [InlineData(21, false, "git@github.com/foo/bar", "123123", @"src\dir\ThisIsFile1.cs", -1, -1, "https://github.com/foo/bar/blob/123123/src/dir/ThisIsFile1.cs")]
     public void GenerateUrl(int testid, bool createRootedPath, string baseUrl, string sha, string path, int startLine, int endLine, string expected)
     {
         SetupRepository(sha);


src/GitHub.VisualStudio/Helpers/ActiveDocumentSnapshot.cs
```diff
@@ -19,7 +19,7 @@ public class ActiveDocumentSnapshot : IActiveDocumentSnapshot
         public ActiveDocumentSnapshot([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider)
         {
             StartLine = EndLine = -1;
-            Name = Services.Dte2?.ActiveDocument?.FullName;
+            Name = Services.Dte2?.ActiveDocument?.ProjectItem.Name;
```
shana: What happens when there's no project loaded? Will `ProjectItem` still exist, or will it be null? This needs to be tested by opening a file on a repo when there's no solution opened.

If `ActiveDocument` is not returning the correct file case and if `ProjectItem` can be null, then perhaps a better way of fixing this is to get the full name of the file from the filesystem to get the correct case.

paladique: I gave this a try by making a new solution without any projects and added a .txt file and `ProjectItem` still exists. VS might be considering a solution a project in the case when no project exists, or still considers it a project if it does.

shana: You need to be in a git repo context without a solution open. You can either create an empty git repo and switch to it via the local git repositories list, or if you have a solution open, close it. Then you open or drag a file into the editor and see what if the property is still there.

paladique: `ProjectItem.ContainingProject` references the project where the file is, and turns out to be null when testing it out. However, debugging `Services.Dte2.Solution.Projects` contains 1 item, which I assume is the solution.

It's possible also to find a file through `Services.Dte2.Solution.FindProjectItem("somefile")` which might mean that any file, whether it be in solution or project, is considered a `ProjectItem`

paladique: When you open a file in VS without a solution open, VS creates a temporary solution file so  `ProjectItem` still exists, but the link that's generated points to the latest commit of the git repo that's currently in the context. Looks like a bug there, it probably should generate a link if the solution isn't open, I'll put that in another issue.

shana: Yeah, I got that behaviour too. We should probably be checking the repo where the file is and not the current repo to determine the link to go to. That's likely a bigger change though :P

shana: Oh wait, you mean the file is in the repo but it generates the wrong url? Sounds like a bug with the path ir something?

paladique: The file wasn't in the repo. I just had a git repo context with no solution open, and opened a random file on my PC.

paladique:
@shana so the end result of this was that:

- In all known cases/scenarios tested, ProjectItem is never null, (solutions are treated like projects, a temporary solution file is created in solution explorer when there is no solution file)

- A [bug](https://github.com/github/VisualStudio/issues/530) was discovered where using the open/copy link context menu items can generate URLs when opening a file that may not be part of the current git repo context

shana:
✨ 

I'll 👀 this first thing in the morning with a fresh brain 😄

shana:
@paladique Update the test manifest with test cases that check for this issue.

PR #528, opened 2016-08-30
Load the package asynchronously

shana:
Load everything asynchronously so it doesn't delay VS itself loading. This requires moving some stuff around, because during the package loading process there can't be any calls to `GetService` or `GetGlobalService`, and those happen in a bunch of places.

### DIFF
--- a/src/GitHub.VisualStudio/GitHubPackage.cs
+++ b/src/GitHub.VisualStudio/GitHubPackage.cs
@@ -14,6 +14,7 @@ using System.ComponentModel.Design;
 using System.Diagnostics;
 using System.Threading;
 using tasks = System.Threading.Tasks;
+using Microsoft.VisualStudio.ComponentModelHost;
 
 namespace GitHub.VisualStudio
 {
@@ -75,6 +76,7 @@ namespace GitHub.VisualStudio
         }
     }
 
+    [NullGuard.NullGuard(NullGuard.ValidationFlags.None)]
     [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [ProvideService(typeof(IUIProvider), IsAsyncQueryable = true)]
     [ProvideAutoLoad(UIContextGuids.NoSolution)]
@@ -126,11 +128,15 @@ namespace GitHub.VisualStudio
             }
         }
 
-        tasks.Task<object> CreateService(IAsyncServiceContainer container, CancellationToken cancellationToken, Type serviceType)
+        async tasks.Task<object> CreateService(IAsyncServiceContainer container, CancellationToken cancellationToken, Type serviceType)
         {
-            AssemblyResolver.InitializeAssemblyResolver();
-            var ret = Services.ComponentModel.DefaultExportProvider.GetExportedValueOrDefault<IUIProvider>();
-            return tasks.Task.FromResult((object)ret);
+            if (serviceType == null)
+                return null;
+            string contract = AttributedModelServices.GetContractName(serviceType);
+            var cm = await GetServiceAsync(typeof(SComponentModel)) as IComponentModel;
+            if (cm == null)
+                return null;
+            return await tasks.Task.Run(() => cm.DefaultExportProvider.GetExportedValueOrDefault<object>(contract));
         }
     }
 }
--- a/src/GitHub.VisualStudio/GitHubPackage.cs
+++ b/src/GitHub.VisualStudio/GitHubPackage.cs
@@ -26,7 +26,7 @@ namespace GitHub.VisualStudio
     [ProvideAutoLoad("11B8E6D7-C08B-4385-B321-321078CDD1F8")]
     [ProvideToolWindow(typeof(GitHubPane), Orientation = ToolWindowOrientation.Right, Style = VsDockStyle.Tabbed, Window = EnvDTE.Constants.vsWindowKindSolutionExplorer)]
     [ProvideOptionPage(typeof(OptionsPage), "GitHub for Visual Studio", "General", 0, 0, supportsAutomation: true)]
-    public class GitHubPackage : Package
+    public class GitHubPackage : AsyncPackage
     {
 
         readonly IServiceProvider serviceProvider;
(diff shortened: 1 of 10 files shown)


src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -25,7 +26,7 @@ namespace GitHub.VisualStudio
     [ProvideAutoLoad("11B8E6D7-C08B-4385-B321-321078CDD1F8")]
```
BertanAygun: To utilize async loading during auto load, you also need AllowsBackgroundLoad parameter set to true here

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -14,6 +14,7 @@
 using System.Diagnostics;
 using System.Threading;
 using tasks = System.Threading.Tasks;
```
AArnott: You can avoid this articial `tasks` alias that you have to use everywhere by adding these two usings instead:

```csharp
using System.Threading.Tasks;
using Task = System.Threading.Tasks.Task; // disambiguate with MS.VS.Shell.Task
```

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -45,24 +46,23 @@ public GitHubPackage(IServiceProvider serviceProvider)
             this.serviceProvider = serviceProvider;
         }
 
-        protected override void Initialize()
+        protected override async tasks.Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
         {
-            base.Initialize();
-            IncrementLaunchCount();
+            var usageTracker = await GetServiceAsync(typeof(IUsageTracker)) as IUsageTracker;
+            await tasks.Task.Run(() => usageTracker.IncrementLaunchCount());
```
AArnott: What's the point of the Task.Run when `AsyncPackage.InitializeAsync` already starts on a background thread? (is that true, @BertanAygun?)

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -45,24 +46,23 @@ public GitHubPackage(IServiceProvider serviceProvider)
...
 
             foreach (var menu in menus.DynamicMenus)
                 serviceProvider.AddCommandHandler(menu.Guid, menu.CmdId, menu.CanShow, () => menu.Activate());
-        }
 
-        void IncrementLaunchCount()
-        {
-            var usageTracker = serviceProvider.GetExportedValue<IUsageTracker>();
-            usageTracker.IncrementLaunchCount();
+            await base.InitializeAsync(cancellationToken, progress);
```
AArnott: I think the recommended pattern is for you to call the base class method before your own implementation.

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -25,7 +26,7 @@ namespace GitHub.VisualStudio
     [ProvideAutoLoad("11B8E6D7-C08B-4385-B321-321078CDD1F8")]
     [ProvideToolWindow(typeof(GitHubPane), Orientation = ToolWindowOrientation.Right, Style = VsDockStyle.Tabbed, Window = EnvDTE.Constants.vsWindowKindSolutionExplorer)]
     [ProvideOptionPage(typeof(OptionsPage), "GitHub for Visual Studio", "General", 0, 0, supportsAutomation: true)]
-    public class GitHubPackage : Package
+    public class GitHubPackage : AsyncPackage
```
AArnott: Update your `PackageRegistration` attribute to include `AllowsBackgroundLoading = true`

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -126,11 +131,15 @@ Version VSVersion
...
-            AssemblyResolver.InitializeAssemblyResolver();
-            var ret = Services.ComponentModel.DefaultExportProvider.GetExportedValueOrDefault<IUIProvider>();
-            return tasks.Task.FromResult((object)ret);
+            if (serviceType == null)
+                return null;
+            string contract = AttributedModelServices.GetContractName(serviceType);
+            var cm = await GetServiceAsync(typeof(SComponentModel)) as IComponentModel;
+            if (cm == null)
+                return null;
+            return await tasks.Task.Run(() => cm.DefaultExportProvider.GetExportedValueOrDefault<object>(contract));
```
AArnott: A more convenient way to switch to a background thread is:

```csharp
using Microsoft.VisualStudio.Threading; // at the top of your file.

await TaskScheduler.Default; // switches to a threadpool thread, if not already on one.
```

src/GitHub.VisualStudio/Services/UsageTracker.cs
```diff
@@ -184,9 +163,24 @@ void UpdateTimer(bool initialCall)
             // Subsequent timer ticks should occur every 8 hours.
             timer.Interval = TimeSpan.FromHours(8);
 
+            if (userSettings == null)
+            {
+                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
AArnott: I'm glad to see you're taking care to switch to the main thread when necessary, and using this mechanism in particular.

src/GitHub.VisualStudio/GitHubPackage.cs
BertanAygun: InitializeAsync would always be called in thread pool if package load is asynchronous, so call to usageTracker can be made directly as in both cases it will be executed in thread pool.

src/GitHub.VisualStudio/Services/UsageTracker.cs
```diff
@@ -156,25 +148,12 @@ void SaveUsage(UsageStore store)
...
-            timer.Stop();
-            timer.Interval = TimeSpan.FromMinutes(1);
-
-            if (userSettings.CollectMetrics && client != null)
-                timer.Start();
+            timer.Interval = TimeSpan.FromMinutes(3);
+            timer.Start();
         }
 
         async void TimerTick(object sender, EventArgs e)
```
AArnott: `async void` methods will crash the process if they ever throw an exception. Please wrap your entire method in a try/catch block and swallow the exception (after reporting the exception some other way like telemetry or an error dialog) to prevent this.

src/GitHub.VisualStudio/GitHubPackage.cs
BertanAygun: Similar to comment above, CreateService should also be called from thread pool threads in async services case so GetExportedValueOrDefault can be called directly.

src/GitHub.VisualStudio/GitHubPackage.cs
BertanAygun: Correct, I just left a similar comment both for async services and packages, the initialize methods should always be called in background threads.

src/GitHub.VisualStudio/GitHubPackage.cs
shana: That is so handy to know! :+1:

src/GitHub.VisualStudio/GitHubPackage.cs
BertanAygun: For async packages, Init is always called form background thread.

shana: @AArnott @BertanAygun I was just testing and got a deadlock on the `InitializeAsync` method doing a `GetServiceAsync` call, and the `InitializeAsync` method is definitely getting called on the main thread (see screenshot). The package is to `AllowsBackgroundLoading = true`.

screen shot 2016-11-02 at 1 20 43 pm

shana: The deadlock is probably unrelated. In any case, the `InitializeAsync` method is definitely getting called on the main thread in most cases here.

PR #537, opened 2016-09-06
Small refactor of repository models.

grokys:
Renamed the repository model classes:

- SimpleRepositoryModel -> LocalRepositoryModel
- RepositoryModel -> RemoteRepositoryModel

It makes sense to review this PR as two separate commits.

The first commit refactors the repository model hierarchy to make `IRepositoryModel` and `ISimpleRepositoryModel` inherit from a common base interface `IRepositoryModelBase` with an implementation in `RepositoryModelBase`. This means that information that is not relevant to a remote repository (local path etc) is no longer present on `RemoteRepositoryModel`.

The second commit renames the classes as described above and renames `IRepositoryModelBase` to `IRepositoryModel` as that name is no longer taken.


src/GitHub.Exports/Models/RepositoryModelBase.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// Initializes a new instance of the <see cref="RepositoryModelBase"/> class.
+        /// </summary>
+        /// <param name="name">The repository name.</param>
+        /// <param name="cloneUrl">The repository's clone URL.</param>
+        protected RepositoryModelBase(
+            string name,
+            UriString cloneUrl)
+        {
+            Name = name;
+            this.cloneUrl = cloneUrl;
```
shana: Should set the property, not the field

src/GitHub.Exports/Models/RepositoryModelBase.cs
```diff
@@ -0,0 +1,86 @@
...
+                    cloneUrl = value;
+                    RaisePropertyChanged();
+                }
+            }
+        }
+
+        /// <summary>
+        /// Gets the name of the owner of the repository, taken from the clone URL.
+        /// </summary>
+        public string Owner => CloneUrl.Owner;
```
shana: CloneUrl can actually be null because we aren't doing null checks anywhere. Technically it should never be null right now because we earlier filter out repos that don't have remotes set, but in the future this assumption may change.
Might be a good idea to add nullchecks to the constructor for now, to document assumptions... or make sure this doesn't blow up if we don't have null checks there.

src/GitHub.Exports/Models/LocalRepositoryModel.cs
```diff
@@ -5,63 +5,40 @@
...
-        {
-            Name = cloneUrl.RepositoryName;
-            Owner = cloneUrl.Owner;
-            CloneUrl = cloneUrl;
-            Icon = Octicon.repo;
-        }
-
-        public SimpleRepositoryModel(string path)
+        public LocalRepositoryModel(string path)
+            : base(ExtractRepositoryName(path), ExtractCloneUrl(path))
```
shana: This will cause IO and git repo loading calls to happen twice every time we create one of these, and these calls happen a lot, so it would be good to avoid it. These calls cost more than the side effect of calling `PropertyChanged` when there's no subscribers.

grokys: Yeah, I didn't like doing this, but there's no way to evaluate once and pass to 2 different base ctor arguments. I will make the base ctor accept a path.

src/GitHub.Exports/Models/LocalRepositoryModel.cs
```diff
@@ -5,63 +5,40 @@
...
 {
+    /// <summary>
+    /// A local repository.
+    /// </summary>
     [DebuggerDisplay("{DebuggerDisplay,nq}")]
-    public class SimpleRepositoryModel : NotificationAwareObject, ISimpleRepositoryModel, IEquatable<SimpleRepositoryModel>
+    public class LocalRepositoryModel : RepositoryModelBase, ILocalRepositoryModel, IEquatable<LocalRepositoryModel>
     {
-        public SimpleRepositoryModel(string name, UriString cloneUrl, string localPath = null)
+        public LocalRepositoryModel(string name, UriString cloneUrl, string localPath = null)
```
grokys: @shana I'm guessing we should disallow a null local path here?

src/GitHub.Exports/Models/RepositoryModelBase.cs
```diff
@@ -0,0 +1,111 @@
...
+        protected RepositoryModelBase(string path)
+        {
+            Guard.ArgumentNotNull(path, nameof(path));
+
+            var dir = new DirectoryInfo(path);
+            if (!dir.Exists)
+                throw new ArgumentException("Path does not exist", nameof(path));
+            var uri = GitService.GitServiceHelper.GetUri(path);
+            Name = uri?.RepositoryName ?? dir.Name;
+            CloneUrl = GitService.GitServiceHelper.GetUri(path);
```
shana: If `CloneUrl` is null here (and it can be), then the `Owner` getter will blow up below. We probably need to be aware/handle the case of loading a git repository that doesn't have a remote set by returning null on the `Owner` below in this case (probably good enough?)

shana: Actually, it was `String.Empty` before, so that's what we need to do in the `Owner` property

PR #542, opened 2016-09-08
Fix refresh of PR list switching to "create" form.

### LINKED ISSUE #480: Refreshing the PR list sometimes switches to the PR creation form
In certain conditions, when going to the PR creation form and back to the list and then refreshing the list causes the PR creation form to be shown again.

refresh-goes-to-creation

OlsonAndrewD:
Fixes #480. Issue was happening because of an extra PRCreation
item in the nav stack when Refresh was clicked 2nd time.

It was there because when navigating forward for the 2nd Create,
GoForward method wasn't removing everything after the current item.


OlsonAndrewD:
I went to go write tests for this fix but then I did a fetch and noticed that @grokys started a refactor of this code a few weeks ago, including tests. Went ahead and submitted this now in case you'd like to use it. (It looks like the new class might have the same issue in it.) Would you like me to go ahead and add tests for this, or let @grokys take over since he's refactoring anyway?

OlsonAndrewD:
(Hi, BTW!) 😊

grokys:
Hi @OlsonAndrewD, thanks for this! My code wasn't a refactor - it was more just trying out ideas for how we might do this stuff differently and won't be merged to master, so please feel free to add tests!

src/GitHub.VisualStudio/UI/Views/GitHubPaneViewModel.cs
```diff
@@ -279,8 +279,8 @@ void StartFlow(UIControllerFlow controllerFlow, [AllowNull]IConnection conn, Vie
         void GoForward(ViewWithData data)
         {
             currentNavItem++;
-            if (currentNavItem < navStack.Count - 1)
-                navStack.RemoveRange(currentNavItem, navStack.Count - 1 - currentNavItem);
+            if (currentNavItem < navStack.Count)
+                navStack.RemoveRange(currentNavItem, navStack.Count - currentNavItem);
             navStack.Add(data);
```
capone3691: ACCESS

OlsonAndrewD:
I reverted my change, created a single failing test for the specific scenario illustrated in the issue, then re-applied my change, and the test went green. :green_heart: I could make more tests if you like, but I thought I'd go ahead and send this up for your consideration/feedback. It took kind of a lot of code to get everything mocked, so do let me know if there's a simpler way that could be done.

grokys:
Thanks @OlsonAndrewD! I will take a look at this this afternoon and see if there's any way to simplify the wiring up of the service provider - I think ``Substitutes.GetFullyMockedServiceProvider` should provide a good start. We (well I at least) definitely want to try and simplify this stuff...

grokys:
(Also @shana `GitHubPaneViewModel` is still in the views folder - this needs moving right?)

shana:
> (Also @shana GitHubPaneViewModel is still in the views folder - this needs moving right?)

Yeah, sigh, it needs to go into `GitHub.App/ViewModels` for sure. Probably best to do that when there's no PRs touching it so it's a conflict-free operation, what do you think?

shana:
> We (well I at least) definitely want to try and simplify this stuff...

Oh yes, it is currently a mess in there, I agree. Let me just add refactoring the service provider story to the list of refactors we need to do... /me unrolls two meters of parchment and scribbles a note

OlsonAndrewD:
> Yeah, sigh, it needs to go into GitHub.App/ViewModels for sure. Probably best to do that when there's no PRs touching it so it's a conflict-free operation, what do you think?

FWIW, based on my limited experience with the project, I agree it might make sense to merge this fix (if you're ok with it) before doing that move. VS definitely shows lots of red squiggles in that file if I move it, and I don't know how far the ripple effect would go. I could maybe take a closer look later if you like though, if I have a chance.

In any case, sounds like maybe the move should be a separate issue on the two meters of parchment? :smiley:

src/GitHub.VisualStudio/Properties/AssemblyInfo.cs
```diff
@@ -1,6 +1,8 @@
 ﻿using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
 [assembly: AssemblyTitle("GitHub.VisualStudio")]
 [assembly: AssemblyDescription("GitHub for Visual Studio VSPackage")]
-[assembly: Guid("fad77eaa-3fe1-4c4b-88dc-3753b6263cd7")]
\ No newline at end of file
+[assembly: Guid("fad77eaa-3fe1-4c4b-88dc-3753b6263cd7")]
+[assembly: InternalsVisibleTo("UnitTests")]
```
grokys: Unfortunately, this causes problems here internally because we sign our official releases of the extension and so `InternalsVisibleToFails`. I think the best thing to do here is just make the class you need access to (``PkgCmdIDList`) public.

src/UnitTests/GitHub.VisualStudio/UI/Views/GitHubPaneViewModelTests.cs
```diff
@@ -0,0 +1,139 @@
...
+        #region Start tests already logged in with an active repo.
+
+        var repositoryHosts = Substitutes.RepositoryHosts;
+        repositoryHosts.IsLoggedInToAnyHost.Returns(true);
+
+        var teamExplorerServiceHolder = Substitute.For<ITeamExplorerServiceHolder>();
+        var activeRepo = Substitute.For<ILocalRepositoryModel>();
+        activeRepo.CloneUrl.Returns(new UriString("https://github.com/foo/foo"));
+        teamExplorerServiceHolder
+            .When(_ => _.Subscribe(Arg.Any<object>(), Arg.Any<Action<ILocalRepositoryModel>>()))
```
grokys: The `_` symbol as a lamda parameter usually means ["this parameter is ignored"](http://stackoverflow.com/questions/2778362/what-does-this-mean-in-lambda-expressions), but here you use it. `x` is our usual first choice for simple lambda params.

src/UnitTests/GitHub.VisualStudio/UI/Views/GitHubPaneViewModelTests.cs
```diff
@@ -0,0 +1,139 @@
...
+{
+    private readonly IServiceProvider serviceProvider;
+    private readonly IUIController uiController;
+    private readonly FakeMenuCommandService menuCommandService;
+    private readonly GitHubPaneViewModel viewModel;
+    private UIViewType lastUiControllerJump;
+
+    public GitHubPaneViewModelTests()
+    {
+        #region Start tests already logged in with an active repo.
```
grokys: This is a nit, but I have unreasonable hatred of `#region`s. Could you remove these? It would make me happy :)

src/UnitTests/GitHub.VisualStudio/UI/Views/GitHubPaneViewModelTests.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        #endregion
+
+        #region Wire up service/UI provider.
+
+        serviceProvider = Substitutes.GetFullyMockedServiceProvider();
+        menuCommandService = new FakeMenuCommandService();
+        serviceProvider.GetService(typeof(IMenuCommandService)).Returns(menuCommandService);
+
+        // TODO: Maybe could avoid cast by making a new interface that inherits IServiceProvider and IUIProvider
```
grokys: Not sure what you mean here - `IUIProvider` implements `IServiceProvider`. I do know however that our service locator stuff is confusing :(

src/UnitTests/GitHub.VisualStudio/UI/Views/GitHubPaneViewModelTests.cs
```diff
@@ -0,0 +1,139 @@
...
+using System.ComponentModel.Composition;
+using System.ComponentModel.Design;
+using System.Linq;
+using System.Reactive.Linq;
+using UnitTests;
+using Xunit;
+
+public class GitHubPaneViewModelTests : TestBaseClass
+{
+    private readonly IServiceProvider serviceProvider;
```
grokys: Convention in the extension is to not explicitly specify `private`.

grokys:
@OlsonAndrewD sorry I didn't manage to get the review done yesterday. In general, great PR! I'm impressed you managed to get all that wiring done! :D

src/GitHub.VisualStudio/Properties/AssemblyInfo.cs
OlsonAndrewD: Done.

src/UnitTests/GitHub.VisualStudio/UI/Views/GitHubPaneViewModelTests.cs
OlsonAndrewD: Removed.

src/UnitTests/GitHub.VisualStudio/UI/Views/GitHubPaneViewModelTests.cs
OlsonAndrewD: Sure, no problem.

src/UnitTests/GitHub.VisualStudio/UI/Views/GitHubPaneViewModelTests.cs
OlsonAndrewD: Yeah, after re-reading that comment, it definitely didn't make sense; removed it.

It felt weird to cast the mock IServiceProvider to IUIProvider. It does work because of the way the mock is set up, but the consumer wouldn't know that based solely on the return type of GetFullyMockedServiceProvider. Maybe Substitutes could have a method that returns IUIProvider?

OlsonAndrewD:
Thanks @grokys! This was a fun challenge. :-)

grokys:
Merged! 🎉 Thanks @OlsonAndrewD!

OlsonAndrewD:
Awesome, you're welcome!

PR #544, opened 2016-09-09
Removed TempFileBaseClass.

### LINKED ISSUE #543: Tests that use TempFileBaseClass cannot be run in parallel.
[TempFileBaseClass](
https://github.com/github/VisualStudio/blob/master/src/UnitTests/Helpers/TestBaseClass.cs#L67) handily creates a temporary directory per method (so every test gets their own little playground), but then it stores that in a `Directory` property which is of course shared for the class, and then deletes that directory when the method exits. This means that tests running in parallel in the same class will run into each other

grokys:
Instead tests should construct a new instance of
`TestBaseClass.TempDirectory` and use that. Fixes #543.

### DIFF
--- a/src/UnitTests/GitHub.App/Models/RepositoryModelTests.cs
+++ b/src/UnitTests/GitHub.App/Models/RepositoryModelTests.cs
@@ -54,31 +54,37 @@ public class RepositoryModelTests
     }
 
     [Collection("PackageServiceProvider global data tests")]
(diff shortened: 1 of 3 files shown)


shana:
Take out https://github.com/github/VisualStudio/blob/master/src/UnitTests/Properties/AssemblyInfo.cs#L39 and see how CI likes it

shana:
Yay! ✨

PR #546, opened 2016-09-12
Remove duplicate guard class.

grokys:
And fix potential NullReferenceException in ArgumentNotEmptyString.


shana:
Yay! ✨ 

Do we have unit tests for these somewhere?

PR #548, opened 2016-09-12
Add more detailed logs when login fails

shana:
Some more detailed logs for login failures.


shana:
Not merging this one right now, we'll need to figure out logging and our login problems in a more holistic way.

PR #553, opened 2016-09-14
Generate Copy/Open link based on the latest commit

### LINKED ISSUE #343: Link generation for Open/Link to GitHub should use the latest commit of the current file

paladique:
Fixes #343 
Continuation of #510

### DIFF
--- a/src/GitHub.Exports/Models/LocalRepositoryModel.cs
+++ b/src/GitHub.Exports/Models/LocalRepositoryModel.cs
@@ -66,7 +66,7 @@ namespace GitHub.Models
             if (CloneUrl == null)
                 return null;
 
-            var sha = await GitService.GitServiceHelper.GetLatestPushedSha(path);
+            var sha = path != null ? await GitService.GitServiceHelper.GetLatestPushedSha(path) : HeadSha;
             // this also incidentally checks whether the repo has a valid LocalPath
             if (String.IsNullOrEmpty(sha))
                 return CloneUrl.ToRepositoryUrl().AbsoluteUri;
--- a/src/GitHub.Exports/Models/LocalRepositoryModel.cs
+++ b/src/GitHub.Exports/Models/LocalRepositoryModel.cs
@@ -7,6 +7,7 @@ using GitHub.Primitives;
 using GitHub.UI;
 using GitHub.Services;
 using GitHub.Extensions;
+using System.Threading.Tasks;
 
 namespace GitHub.Models
 {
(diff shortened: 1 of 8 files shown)


src/GitHub.Exports/Models/LocalRepositoryModel.cs
```diff
@@ -60,12 +61,12 @@ public void Refresh()
...
         /// <param name="endLine">The end of a line range on the specified file.</param>
         /// <returns>An UriString with the generated url, or null if the repository has no remote server configured or if it can't be found locally</returns>
-        public UriString GenerateUrl(string path = null, int startLine = -1, int endLine = -1)
+        public async Task<UriString> GenerateUrl(string path = null, int startLine = -1, int endLine = -1)
         {
             if (CloneUrl == null)
                 return null;
 
-            var sha = HeadSha;
+            var sha = await GitService.GitServiceHelper.GetLatestPushedSha(path);
```
shana: Is this the path where the repository is? The `GenerateUrl` method takes a `path` argument that represents the file to generate the url for, not the path to the repository where the file is.

src/GitHub.Exports/Services/GitService.cs
```diff
@@ -70,5 +71,29 @@ public UriString GetRemoteUri(IRepository repo, string remote = "origin")
         }
 
         public static IGitService GitServiceHelper => VisualStudio.Services.DefaultExportProvider.GetExportedValueOrDefault<IGitService>() ?? new GitService();
+
+        public Task<string> GetLatestPushedSha(string path)
```
shana: Is this argument allowed to be null? If yes, it needs to be documented as such, and code that relies on it needs null checks. If no, it needs a null check at the start of the method.

src/GitHub.Exports/Services/GitService.cs
```diff
@@ -70,5 +72,36 @@ public UriString GetRemoteUri(IRepository repo, string remote = "origin")
         }
 
         public static IGitService GitServiceHelper => VisualStudio.Services.DefaultExportProvider.GetExportedValueOrDefault<IGitService>() ?? new GitService();
+
+        /// <summary>
+        /// Finds the latest pushed commit of a file and returns the sha of that commit. Returns null when no commits have 
+        /// been found in any remote branches or the current local branch. 
+        /// </summary>
+        /// <param name="path">The local path of the file. This cannot be null.</param>
```
shana: I think `The local path of a repository or a file inside a repository. This cannot be null.` will be clearer.

src/GitHub.Exports/Services/GitService.cs
```diff
@@ -70,5 +72,36 @@ public UriString GetRemoteUri(IRepository repo, string remote = "origin")
...
+        /// <returns></returns>
+        public Task<string> GetLatestPushedSha(string path)
+        {
+            Guard.ArgumentNotNull(path, nameof(path));
+            var repo = GetRepository(path);
```
shana: If the `path` points to a file or a directory that's not inside a repository, this will return null, so you need to check that before using the variable, and likely return early.

src/GitHub.Exports/Services/IGitService.cs
```diff
@@ -44,5 +45,7 @@ public interface IGitService
         /// <param name="repo"></param>
         /// <returns></returns>
         UriString GetRemoteUri(IRepository repo, string remote = "origin");
+
+        Task<string> GetLatestPushedSha(string path);
```
shana: You'll need to copy the documentation added to the method to here. Classes and interfaces don't share docs, unfortunately.

shana:
:sparkles:

PR #556, opened 2016-09-15
Allow canonical refs in SetTrackingBranch.

grokys:
Allow canonical or friendly refs to be passed to `GitClient.SetTrackingBranch`. Also fixed up what appeared to be incorrect doc comments in `IGitClient`.

### DIFF
--- a/src/GitHub.App/Services/GitClient.cs
+++ b/src/GitHub.App/Services/GitClient.cs
@@ -97,7 +97,7 @@ namespace GitHub.Services
                 });
         }
 
-        private static bool IsCanonical(string s)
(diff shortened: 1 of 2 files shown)


src/GitHub.App/Services/GitClient.cs
```diff
@@ -96,5 +96,10 @@ public IObservable<Remote> GetHttpRemote(IRepository repo, string remote)
                     return ret;
                 });
         }
+
+        private static bool IsCanonical(string s)
```
shana: :eyes: `private` :eyes:

PR #557, opened 2016-09-15
Added Fetch method that accepts refspecs.

grokys:
Adds a `Fetch` method that accepts refspecs to `GitClient`.

### DIFF
--- a/src/GitHub.App/Services/GitClient.cs
+++ b/src/GitHub.App/Services/GitClient.cs
@@ -51,6 +51,18 @@ namespace GitHub.Services
             });
         }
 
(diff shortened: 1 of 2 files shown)


shana:
I got a little comment but it really doesn't much matter for this PR, it's fine as-is.

src/GitHub.App/Services/GitClient.cs
```diff
@@ -51,6 +51,18 @@ public IObservable<Unit> Fetch(IRepository repository, string remoteName)
             });
         }
 
+        public IObservable<Unit> Fetch(IRepository repository, string remoteName, params string[] refspecs)
```
grokys: There doesn't seem to be from a quick google, no...

Issue #403, opened 2016-07-06
User icon missing in Create Gist window


grokys:
The user icon is sometimes missing in the Create Gist window:

![image](https://cloud.githubusercontent.com/assets/1775141/16612830/5a226296-4369-11e6-8601-383cb26f5923.png)

paladique:
Do you remember what you happened to be doing before coming to this window? I'm not getting the same behavior. 

![2016-07-06 2](https://cloud.githubusercontent.com/assets/1414307/16619676/21ee3028-435d-11e6-8933-55aae3e70e06.png)

shana:
It's likely the same issue as #408

grokys:
Yeah, it seems to work most of the time, so likely the same as #408 as @shana said...

StanleyGoldman:
I sent in a pull request for #408 and I was hoping it would fix this issue as well.
I was unable to understand the cause, but the avatar will appear correctly the first time, but subsequent views don't.

StanleyGoldman:
Resolved by #584

Issue #408, opened 2016-07-06
Clicking refresh in PR list sometimes makes PR icon disappear


grokys:
This one is a little difficult to reproduce, but sometimes when clicking the refresh button in the PR list, the icon disappears. Sometimes clicking it more makes it reappear, sometimes not.

shana:
Which icon? Can we haz screenshot on this?

grokys:
The user icon next to the PR in the list. I wasn't able to capture a screenshot because it stopped happening - as I said it can be difficult to reproduce. It was happening fairly regularly and then it stopped :(

grokys:
Caught it!

![406b76d01c9d84c4780eba99a9ad1021](https://cloud.githubusercontent.com/assets/1775141/16614477/ce11489a-4371-11e6-9b44-8c74c9d9cac4.gif)

shana:
Yeah, got that too when loading PRs for the first time.

![screen shot 2016-07-06 at 11 41 13 am](https://cloud.githubusercontent.com/assets/310137/16615331/3add8d0e-4376-11e6-9848-003912d711cc.png)

This is the avatar provider loading things from cache in one of the akavache threads. Any exceptions in the log file?

grokys:
This is the only thing that looks like it could be relevant:

> 2016-06-15 00:29:52.8413|ERROR|thread: 1|ImageCache|Could not vacuum image cache
> System.ObjectDisposedException: The cache 'SqlitePersistentBlobCache' was disposed.

shana:
I have some of those too, but that's in June, not today

grokys:
I suspect this issue is caused by this line:

https://github.com/github/VisualStudio/blob/master/src/GitHub.App/Models/Account.cs#L35

I think that should be setting the property rather than the backing field.

StanleyGoldman:
I'll check that out a bit later today.

femisimon:
This was reproduced.

Cc @grokys @shana 

![snip20161108_2](https://cloud.githubusercontent.com/assets/17054747/20109988/6433ecd8-a597-11e6-9758-2f68101261ed.png)

StanleyGoldman:
I'm unable to reproduce this issue anymore

paladique:
Can't repro this either, closing!

grokys:
Yeah, I think this was fixed!

PR #561, opened 2016-09-25
Removing Avatar from CopyFrom in Account

StanleyGoldman:
Fix for #408 
I do not believe that `Avatar` should be copied in `Account.CopyFrom()`.

### DIFF
--- a/src/GitHub.App/Models/Account.cs
+++ b/src/GitHub.App/Models/Account.cs
@@ -15,6 +15,8 @@ namespace GitHub.Models
     {
         BitmapSource avatar;
 
+        IObservable<BitmapSource> BitmapSource;
+        private IDisposable _bitmapSourceSubscription;
         public Account(
             string login,
             bool isUser,
@@ -30,9 +32,11 @@ namespace GitHub.Models
             PrivateReposInPlan = privateRepositoryInPlanCount;
             IsOnFreePlan = privateRepositoryInPlanCount == 0;
             HasMaximumPrivateRepositories = OwnedPrivateRepos >= PrivateReposInPlan;
+            BitmapSource = bitmapSource;
 
-            bitmapSource.ObserveOn(RxApp.MainThreadScheduler)
-                .Subscribe(x => avatar = x);
+            _bitmapSourceSubscription = bitmapSource
+                .ObserveOn(RxApp.MainThreadScheduler)
+                .Subscribe(x => Avatar = x);
         }
 
         public Account(Octokit.Account account)
@@ -80,6 +84,16 @@ namespace GitHub.Models
             IsOnFreePlan = other.IsOnFreePlan;
             HasMaximumPrivateRepositories = other.HasMaximumPrivateRepositories;
             Avatar = other.Avatar;
+
+            var otherAccount = other as Account;
+            if (otherAccount != null)
+            {
+                _bitmapSourceSubscription.Dispose();
+
+                _bitmapSourceSubscription = otherAccount.BitmapSource
+                    .ObserveOn(RxApp.MainThreadScheduler)
+                    .Subscribe(x => Avatar = x);
+            }
         }
 
         public override bool Equals([AllowNull]object obj)
--- a/src/UnitTests/GitHub.App/Models/AccountModelTests.cs
+++ b/src/UnitTests/GitHub.App/Models/AccountModelTests.cs
@@ -1,0 +1,94 @@
+﻿using System;
+using System.Reactive.Linq;
+using System.Threading;
+using System.Windows.Media.Imaging;
+using GitHub.Collections;
+using GitHub.Models;
+using GitHub.Services;
+using ReactiveUI;
+using Xunit;
+
+namespace UnitTests.GitHub.App.Models
+{
+    public class AccountModelTests : TestBaseClass
+    {
+        [Fact]
+        public void CopyFromDoesNotLoseAvatar()
+        {
+            //A function that will return this image in an observable after X seconds
+            var image = AvatarProvider.CreateBitmapImage("pack://application:,,,/GitHub.App;component/Images/default_user_avatar.png");
+            Func<int, IObservable<BitmapImage>> generateObservable = seconds =>
+            {
+                return Observable.Generate(
+                    initialState: 0, 
+                    condition: i => i < 1, 
+                    iterate: state => state + 1,
+                    resultSelector: i => image, 
+                    timeSelector: i => TimeSpan.FromSeconds(seconds));
+            };
+
+            var evt = new ManualResetEvent(false);
+
+            //Creating an initial account with an observable that returns immediately
+            const string login = "foo";
+            const int initialOwnedPrivateRepositoryCount = 1;
+
+            var createdAt = DateTime.Now;
+            var initialAccount = new Account(login, true, false, initialOwnedPrivateRepositoryCount, 0, generateObservable(0));
+
+            //Creating the test collection
+            var col = new TrackingCollection<IAccount>(Observable.Empty<IAccount>(), OrderedComparer<IAccount>.OrderByDescending(x => x.Login).Compare);
+            col.Subscribe(account =>
+            {
+                evt.Set();
+            }, () => { });
+
+            //Adding that account to the TrackingCollection
+            col.AddItem(initialAccount);
+
+            //Waiting for the collection add the item
+            evt.WaitOne();
+            evt.Reset();
+
+            //Checking some initial properties
+            Assert.Equal(login, col[0].Login);
+            Assert.Equal(initialOwnedPrivateRepositoryCount, col[0].OwnedPrivateRepos);
+
+            //Demonstrating that the avatar is present
+            Assert.NotNull(col[0].Avatar);
+
+            //Creating an observable that will return in one second
+            var updatedBitmapSourceObservable = generateObservable(1);
+
+            //Creating an account update with an observable
+            const int updatedOwnedPrivateRepositoryCount = 2;
+
+            var updatedAccount = new Account(login, true, false, updatedOwnedPrivateRepositoryCount, 0, updatedBitmapSourceObservable);
+
+            //Updating the accout in the collection
+            col.AddItem(updatedAccount);
+
+            //Waiting for the collection to process the update
+            evt.WaitOne();
+            evt.Reset();
+
+            updatedBitmapSourceObservable.Subscribe(bitmapImage =>
+            {
+                evt.Set();
+            });
+
+            //Waiting for the delayed bitmap image observable
+            evt.WaitOne();
+            evt.Reset();
+
+            //Login is the id, so that should be the same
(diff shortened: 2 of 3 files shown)


grokys:
Hi @StanleyGoldman.

Could you explain how this fixes the issue? I can see that maybe the problem is the subscription to the `bitmapSource` isn't being copied as well, but simply not copying the avatar doesn't look correct on first glance to me. Surely if the account is different when `CopyTo` is called then the avatar needs to be updated? I think we'd need unit tests here to be sure...

Disclaimer: I did not write this code, so take what I say with a pinch of salt ;)

StanleyGoldman:
Hey @grokys.

After looking at it a bit more, you are definitely right, removing that line from the `CopyFrom` can't be the right fix. I'm trying to better understand why a null is being copied over, that part doesn't make sense.

The `CopyFrom` is called by `TrackingCollection` when `SortedNone`, but in other situations the Avatar is not null. I started reading unit tests for `TrackingCollection` last night, I should be able to reproduce the issue in a unit test when I understand it better.

StanleyGoldman:
Hey @grokys 

I understand the problem now, unfortunately that didn't make the solution pop out at me. Hopefully someone with more project experience can make a recommendation.

The problem is with the `trackingAuthors` and `trackingAssignees` collections in `PullRequestViewModel`. During `Initialize()` as the pullRequests Observable is being consumed Authors and Assignees are being added to those collections with the `AddItem()` method. If a user previously exists in that collection, as is the case in a pull request refresh, `CopyFrom()` is called to update the original Account object.

If `Account.CopyFrom()` is called before the `bitmapSource` observable in the "replacement object" is complete, `CopyFrom()` will copy a null avatar value to the "destination object". By the time `bitmapSource` is complete, the "replacement object" has conceptually been discarded.

I believe I have successfully demonstrated this in the unit test, by intentionally delaying the `bitmapSource` observable in the "replacement object".

What do you think?

StanleyGoldman:
So I was thinking about how to solve this and I had an idea, but it didn't work out as I had hoped.

I remembered that the `TrackingCollection` has a `NewerComparer` which is documented as...

        /// <summary>
        /// Set a comparer that determines whether the item being processed is newer than the same
        /// item seen before. This is to prevent stale items from overriding newer items when data
        /// is coming simultaneously from cache and from live data. Use a timestamp-like comparison
        /// for best results
        /// </summary>

I was hoping that If a `CreatedAt` property were added to `Account` all the instances in a `PullRequest` list would have the same value as they came from the same API response; that in turn would prevent `CopyFrom` from being called.

It turns out that the `TrackingCollection` does not skip elements that have the same value and `CopyFrom` is still called. Is that correct?

Rebased changes to latest master.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -540,7 +540,7 @@ ActionData ProcessItem(ActionData data, List<T> list)
                 if (newer != null)
                 {
                     // the object is "older" than the one we have, ignore it
-                    if (newer(item, old) > 0)
+                    if (newer(item, old) >= 0)
```
StanleyGoldman: So this little edge case is the culprit.

StanleyGoldman:
Of course that fix caused a few other unit tests to fail, but I believe there are errors in the tests themselves.

src/TrackingCollectionTests/TrackingCollectionTests.cs
```diff
@@ -333,7 +333,7 @@ public void FilterTitleRun2()
         var total = 1000;
 
         var list1 = new List<Thing>(Enumerable.Range(1, total).Select(i => GetThing(i, i, i, "Run 1")).ToList());
-        var list2 = new List<Thing>(Enumerable.Range(1, total).Select(i => GetThing(i, i, i, "Run 2")).ToList());
+        var list2 = new List<Thing>(Enumerable.Range(1, total).Select(i => GetThing(i, i, i + 1, "Run 2")).ToList());
```
StanleyGoldman: So in this unit test, we are expecting the collection to reflect a list or Things with the Title "Run 2".
But if the updated times are the same, that should not be the case, so I am making a change to ensure the "Run 2" things have a newer `UpdatedAt` and now the unit test passes.

src/TrackingCollectionTests/TrackingCollectionTests.cs
```diff
@@ -429,8 +429,8 @@ public void OrderByMatchesOriginalOrder()
         var count = 0;
         var total = 1000;
 
-        var list1 = new List<Thing>(Enumerable.Range(1, total).Select(i => GetThing(i, i, total - i, "Run 1")).ToList());
-        var list2 = new List<Thing>(Enumerable.Range(1, total).Select(i => GetThing(i, i, total - i, "Run 2")).ToList());
+        var list1 = new List<Thing>(Enumerable.Range(1, total).Select(i => GetThing(i, i, i, "Run 1")).ToList());
+        var list2 = new List<Thing>(Enumerable.Range(1, total).Select(i => GetThing(i, i, i + 1, "Run 2")).ToList());
```
StanleyGoldman: Again, making sure we have the correct `UpdatedAt` times.

src/TrackingCollectionTests/TrackingCollectionTests.cs
```diff
@@ -464,7 +464,7 @@ public void OrderByMatchesOriginalOrder()
 
         Assert.AreEqual(total, count);
         Assert.AreEqual(total, col.Count);
-        CollectionAssert.AreEqual(col, list2);
+        CollectionAssert.AreEqual(col, list2.Reverse<Thing>());
```
StanleyGoldman: This list is being sorted descending so list2 should be reversed.

src/GitHub.App/Services/ModelService.cs
```diff
@@ -164,13 +165,13 @@ public IObservable<AccountCacheItem> GetUserFromCache()
...
                             // this could blow up due to the collection being disposed somewhere else
-                            try { collection.RemoveItem(Create(item)); }
+                            try { collection.RemoveItem(Create(item, DateTime.Now)); }
                             catch (ObjectDisposedException) { }
                         },
                         TimeSpan.Zero,
                         TimeSpan.FromDays(7))
                 )
-                .Select(Create)
+                .Select(prCacheItem => Create(prCacheItem, createdAt))
```
StanleyGoldman: Setting the create date of a set of pull requests

src/GitHub.App/ViewModels/PullRequestListViewModel.cs
```diff
@@ -68,6 +68,10 @@ public class PullRequestListViewModel : BaseViewModel, IPullRequestListViewModel
                 OrderedComparer<IAccount>.OrderByDescending(x => x.Login).Compare);
             trackingAssignees = new TrackingCollection<IAccount>(Observable.Empty<IAccount>(),
                 OrderedComparer<IAccount>.OrderByDescending(x => x.Login).Compare);
+
+            trackingAuthors.NewerComparer = OrderedComparer<IAccount>.OrderByDescending(x => x.CreatedAt).Compare;
+            trackingAssignees.NewerComparer = OrderedComparer<IAccount>.OrderByDescending(x => x.CreatedAt).Compare;
+
```
StanleyGoldman: Adding this `NewerComparer` will prevent `CopyFrom()` being called multiple times if a user is the author/assignee of multiple pull requests in the set.

src/GitHub.App/Models/Account.cs
```diff
@@ -75,11 +85,22 @@ public void CopyFrom(IAccount other)
...
+
+            var otherAccount = other as Account;
+            if (otherAccount != null)
+            {
+                _bitmapSourceSubscription.Dispose();
+
+                _bitmapSourceSubscription = otherAccount.BitmapSource
+                    .ObserveOn(RxApp.MainThreadScheduler)
+                    .Subscribe(x => Avatar = x);
+            }
```
StanleyGoldman: If we are copying this `Account` from another `Account` instance. We should cancel the subscription to the previous `BitmapSource` and start a new subscription against the new `BitmapSource`.

src/GitHub.App/Models/Account.cs
```diff
@@ -30,9 +35,12 @@ public class Account : ReactiveObject, IAccount
             PrivateReposInPlan = privateRepositoryInPlanCount;
             IsOnFreePlan = privateRepositoryInPlanCount == 0;
             HasMaximumPrivateRepositories = OwnedPrivateRepos >= PrivateReposInPlan;
+            CreatedAt = createdAt ?? DateTime.Now;
+            BitmapSource = bitmapSource;
```
StanleyGoldman: Keep track of the BitmapSource in case we are copied to another instance.

src/GitHub.App/Models/Account.cs
```diff
@@ -30,9 +35,12 @@ public class Account : ReactiveObject, IAccount
...
 
-            bitmapSource.ObserveOn(RxApp.MainThreadScheduler)
-                .Subscribe(x => avatar = x);
+            _bitmapSourceSubscription = bitmapSource
+                .ObserveOn(RxApp.MainThreadScheduler)
+                .Subscribe(x => Avatar = x);
```
StanleyGoldman: Keep track of the `BitmapSource` `Subscribe()` disposable so we can stop the subscription if we are copied into.

StanleyGoldman:
I'm able to refresh on the pull request screen now.
Refresh

StanleyGoldman:
This also resolves #403

grokys:
Hi Stanley, I've tried to review this PR but I don't quite understand. There seem to be two parts to the fix: the avatar `CopyFrom` fix and changes to `TrackingCollection`'s comparer. I've read your comment above that mentions trying to prevent `CopyFrom` from being called by using a comparer, but I'm not sure how the two parts go together because if you've fixed `CopyFrom` then the `TrackingCollection` changes shouldn't be needed. What am I missing?

StanleyGoldman:
Hey @grokys I think you understand perfectly ;)

You are right, I can definitely (and probably should) separate this pull request into two.

src/UnitTests/GitHub.App/Models/AccountModelTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            //A function that will return this image in an observable after X seconds
+            var image = AvatarProvider.CreateBitmapImage("pack://application:,,,/GitHub.App;component/Images/default_user_avatar.png");
+            Func<int, IObservable<BitmapImage>> generateObservable = seconds =>
+            {
+                return Observable.Generate(
+                    initialState: 0, 
+                    condition: i => i < 1, 
+                    iterate: state => state + 1,
+                    resultSelector: i => image, 
+                    timeSelector: i => TimeSpan.FromSeconds(seconds));
```
StanleyGoldman: Yea, that's probably a better idea. I'll work those changes in.

PR #565, opened 2016-09-27
Sprinkle some UI polish for description section

donokuda:
Couple of small tweaks to the PR description:

- Shrink avatar to 20px
- Minor margin/padding adjustments
- Shrink images if they are included in the markdown file.

body-changes

/cc @grokys since this targets your branch

### DIFF
--- a/src/GitHub.App/SampleData/PullRequestDetailViewModelDesigner.cs
+++ b/src/GitHub.App/SampleData/PullRequestDetailViewModelDesigner.cs
@@ -23,7 +23,9 @@ namespace GitHub.SampleData
 
 ViewModels are responsible for handling the UI on the view they control, but they shouldn't be handling UI for things outside of the view. In this case, we're showing errors in VS outside the view, and that should be handled by the section that is hosting the view.
 
-This requires that errors be propagated from the viewmodel to the view and from there to the host via the IView interface, since hosts don't usually know what they're hosting.";
+This requires that errors be propagated from the viewmodel to the view and from there to the host via the IView interface, since hosts don't usually know what they're hosting.
+
+![An image](https://cloud.githubusercontent.com/assets/1174461/18882991/5dd35648-8496-11e6-8735-82c3a182e8b4.png)";
 
             var gitHubDir = new PullRequestDirectoryViewModel("GitHub");
             var modelsDir = new PullRequestDirectoryViewModel("Models");
--- a/src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
+++ b/src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
@@ -163,14 +163,14 @@
                 </Grid.RowDefinitions>
 
                 <!-- Author avatar -->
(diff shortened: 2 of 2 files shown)


src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
```diff
@@ -41,6 +41,14 @@
                 <Setter Property="FontSize" Value="12"/>
                 <Setter Property="TextAlignment" Value="Left"/>
                 <Setter Property="PagePadding" Value="0"/>
+
+                <Style.Resources>
+                    <Style TargetType="Image">
+                        <Setter Property="Stretch" Value="UniformToFill" />
+                        <Setter Property="StretchDirection" Value="DownOnly" />
+                        <Setter Property="MaxWidth" Value="{Binding ActualWidth, ElementName=bodyMarkdown}" />
```
donokuda: Is there a better way to bind this value? Coupling it specifically to `ElementName` doesn't seem very smart or reusable. 🔬 🐶

src/GitHub.App/SampleData/PullRequestDetailViewModelDesigner.cs
```diff
@@ -23,7 +23,9 @@ public PullRequestDetailViewModelDesigner()
 
 ViewModels are responsible for handling the UI on the view they control, but they shouldn't be handling UI for things outside of the view. In this case, we're showing errors in VS outside the view, and that should be handled by the section that is hosting the view.
 
-This requires that errors be propagated from the viewmodel to the view and from there to the host via the IView interface, since hosts don't usually know what they're hosting.";
+This requires that errors be propagated from the viewmodel to the view and from there to the host via the IView interface, since hosts don't usually know what they're hosting.
+
+![An image](https://cloud.githubusercontent.com/assets/1174461/18882991/5dd35648-8496-11e6-8735-82c3a182e8b4.png)";
```
donokuda: Are there better ways to test images rendering?

src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
grokys: I suspect here that using a `RelativeSource AncestorType` binding might be the best way to go: http://stackoverflow.com/questions/84278/how-do-i-use-wpf-bindings-with-relativesource

src/GitHub.App/SampleData/PullRequestDetailViewModelDesigner.cs
grokys: Not sure what you mean... Do you mean instead of using a URL?

grokys:
Merging and will tidy up the markup relating to the 2 questions myself (if it works ;) )

src/GitHub.App/SampleData/PullRequestDetailViewModelDesigner.cs
grokys: Actually if that is what you mean, I think a URL is best. We could probably reference a file in our resources, but that would mean that the file was distributed with the extension unnecessarily.

PR #567, opened 2016-09-28
Maintainer workflow

grokys:
This adds a PR details view to the GitHub pane, and allows checking out PR branches both from the same repository and from forks.

Known issues:
- Navigation in the GitHub pane is broken - we know this but didn't want to make this PR any bigger than it already is
- Some of the messages are not very useful - this will be addressed in a separate UI polish PR
- No open file/diff file support yet - that will also come in a separate PR

cc @Haacked to notify you of the known issues above


src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
```diff
@@ -1,23 +1,508 @@
...
         {
-            System.Windows.MessageBox.Show(String.Format(CultureInfo.InvariantCulture, "{0}", data.Data));
+            var prNumber = (int)data.Data;
+
+            IsBusy = true;
+
+            // TODO: Catch errors.
+            Observable.CombineLatest(
+                    repositoryHost.ApiClient.GetPullRequest(repository.Owner, repository.CloneUrl.RepositoryName, prNumber),
+                    repositoryHost.ApiClient.GetPullRequestFiles(repository.Owner, repository.CloneUrl.RepositoryName, prNumber).ToList(),
```
grokys: Yes, I did originally start writing it using `ModelService` but then I remembered how we ended up essentially bypassing the caching for the PR list, so I went straight to the API. The question I wanted to discuss with you was do our reasons for bypassing the cache for the PR list apply here?

src/GitHub.App/ViewModels/PullRequestFileViewModel.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace GitHub.ViewModels
+{
+    public class PullRequestFileViewModel : IPullRequestFileViewModel
```
shana: Shouldn't this be a model and not a viewmodel?

src/GitHub.App/ViewModels/PullRequestDirectoryViewModel.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace GitHub.ViewModels
+{
+    public class PullRequestDirectoryViewModel : IPullRequestDirectoryViewModel
```
shana: Shouldn't this be a model?

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
grokys: Ok, that makes sense. I did assume that I would need to use it in the end, but for development it was easier just to bypass and not have to maintain yet another class and interface. Will make it use the model service.

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
```diff
@@ -1,23 +1,508 @@
...
+            model = pullRequest;
+            State = CreatePullRequestState(pullRequest);
+            SourceBranchDisplayName = GetBranchDisplayName(pullRequest.Head.Label);
+            TargetBranchDisplayName = GetBranchDisplayName(pullRequest.Base.Label);
+            CommitCount = pullRequest.Commits;
+            Title = pullRequest.Title;
+            Number = pullRequest.Number;
+            Author = new Models.Account(pullRequest.User, avatarProvider.GetAvatar(new AccountCacheItem(pullRequest.User)));
+            CreatedAt = pullRequest.CreatedAt;
+            Body = !string.IsNullOrWhiteSpace(pullRequest.Body) ? pullRequest.Body : "*No description provided.*";
```
shana: Instead of copying all of this to the view model, how about using the `IPullRequestModel` which is already set up for most of this information (adding any missing bits) and exposing a `PullRequest` property that the view can get the information from?

src/GitHub.App/ViewModels/PullRequestFileViewModel.cs
grokys: It's specifically tailored to the needs of the view, so that makes it a view model to me. In particular it wouldn't be a tree if it weren't for the fact that we can display as a tree in the view and later we might want to add things like a custom icon depending on file type etc, which makes it a VM in my mind.

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
grokys: Yes, I did consider this but then we'd lose out on change notifications, as models shouldn't be implementing `INotifyPropertyChanged`. I could however use the `IPullRequestModel` fields as the backing instead of having them as fields in the VM itself.

src/GitHub.App/ViewModels/PullRequestFileViewModel.cs
shana: I don't see anything that's specific to this view in particular. It's just a representation of a file change in a pull request. It has no logic, only data, usable by any view or viewmodel to do stuff. Screams of model to me 😄

grokys: Well to me the model is the model you get from the backend (`PullRequestFile` currently). This is transformed into a tree at the viewmodel layer for view purposes, making it a view model :)

src/GitHub.App/ViewModels/PullRequestDirectoryViewModel.cs
shana: Actually, this is only used internally in the `PullRequestDetailViewModel`, so it probably shouldn't be exposed publicly, it's an implementation detail. We can just have this class in the `PullRequestDetailViewModel` as a helper class. I would definitely still not call it a ViewModel - the only thing it does is represent data, it has no logic for a view to interact with data, it's just a handy model (or in this case, a handy helper class)

src/GitHub.App/Services/ModelService.cs
```diff
@@ -427,39 +451,98 @@ public AccountCacheItem Owner
...
+                : this(pr, new PullRequestFile[0])
+            {
+            }
+
+            public PullRequestCacheItem(PullRequest pr, IList<PullRequestFile> files)
             {
                 Title = pr.Title;
                 Number = pr.Number;
+                Base = new GitReferenceModel { Label = pr.Base.Label, Ref = pr.Base.Ref, RepositoryCloneUrl = pr.Base.Repository.CloneUrl };
+                Head = new GitReferenceModel { Label = pr.Head.Label, Ref = pr.Head.Ref, RepositoryCloneUrl = pr.Head.Repository.CloneUrl };
```
Haacked: It's possible for the `pr.Head.Repository` property to be null causing a `NullReferenceException` here. For example, check out this PR: https://github.com/github/Scientist.net/pull/20 After it was merged, the user deleted the repository.

src/GitHub.App/ViewModels/PullRequestFileNode.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public PullRequestFileNode(string fullPath, PullRequestFileStatus status)
+        {
+            FileName = System.IO.Path.GetFileName(fullPath);
+            Path = System.IO.Path.GetDirectoryName(fullPath);
+            Status = status;
+        }
+
+        public string FileName { get; }
+        public string Path { get; }
```
Haacked: I tend to find property names like `Path` confusing. Path to what? How about `DirectoryName` or `DirectoryPath`? Bonus, that'll let you use `Path.GetFileName` rather than the fully qualified calls. :smile:

src/GitHub.App/Models/PullRequestModel.cs
```diff
@@ -107,30 +106,51 @@ public string Title
...
+            get { return body; }
+            set { body = value; this.RaisePropertyChange(); }
+        }
+
+        public GitReferenceModel Base { get; set; }
+        public GitReferenceModel Head { get; set; }
         public DateTimeOffset CreatedAt { get; set; }
         public DateTimeOffset UpdatedAt { get; set; }
         public IAccount Author { get; set; }
+        public IList<IPullRequestFileModel> ChangedFiles { get; set; } = new IPullRequestFileModel[0];
```
grokys: Yes, again a very good point - I always forget about ~~Dre~~ `IReadOnlyCollection`.

src/GitHub.App/Services/ModelService.cs
grokys: Very good catch - I didn't think of that! Added code to deal with that case and a unit test. In addition, I've made `GitReferenceModel` immutable and made it check for non-null values in its fields as an additional sanity check.

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -74,6 +84,156 @@ public IObservable<string> GetPullRequestTemplate(ILocalRepositoryModel reposito
...
+            {
+                var repo = gitService.GetRepository(repository.LocalPath);
+                var result = GetLocalBranchesInternal(repository, repo, pullRequest).Select(x => new BranchModel(x, repository));
+                return result.ToObservable();
+            });
+        }
+
+        public bool IsPullRequestFromFork(ILocalRepositoryModel repository, IPullRequestModel pullRequest)
+        {
+            return pullRequest.Head.RepositoryCloneUrl.ToRepositoryUrl() != repository.CloneUrl.ToRepositoryUrl();
```
grokys: Ah, I didn't think of looking there. So if I check that do I still need to compare the clone URLs?

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -74,6 +84,156 @@ public IObservable<string> GetPullRequestTemplate(ILocalRepositoryModel reposito
...
+        }
+
+        public IObservable<Unit> SwitchToBranch(ILocalRepositoryModel repository, IPullRequestModel pullRequest)
+        {
+            return Observable.Defer(async () =>
+            {
+                var repo = gitService.GetRepository(repository.LocalPath);
+                var branchName = GetLocalBranchesInternal(repository, repo, pullRequest).First();
```
grokys: 👍

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -95,5 +255,33 @@ public IObservable<string> GetPullRequestTemplate(ILocalRepositoryModel reposito
...
+            var origin = repo.Network.Remotes["origin"];
+            var existing = origin.FetchRefSpecs.FirstOrDefault(x => x.Specification == spec);
+
+            if (existing == null)
+            {
+                repo.Network.Remotes.Update(origin, x => x.FetchRefSpecs.Add(spec));
+            }
+        }
+
+        static string GetSafeBranchName(string name)
```
grokys: The branch name that is created here is based on the PR title, so I think that carrying that through into the branch name wouldn't be desired. For example, if I create a PR called "Fix foo/bar errors" then it doesn't make much sense for the PR to be called `pr/123-fix foo/bar errors` especially as Visual Studio will display it in the "Manage Branches" window as:

- pr
  - 123-fix foo
    - bar errors

src/GitHub.App/ViewModels/PullRequestFileNode.cs
grokys: Thing is that this is either a path to a file *or* a directory, as the property comes from the `IPullRequestChangeNode` which is implemented by `PullRequestDirectoryNode` as well as `PullRequestFileNode`. As such, I'm struggling to come up with a better name.

src/GitHub.App/Services/PullRequestService.cs
Haacked: > Ah, I didn't think of looking there. So if I check that do I still need to compare the clone URLs?

Yeah, I suppose you still do because it's possible that the PR target is the fork and not upstream.

src/GitHub.App/Services/PullRequestService.cs
Haacked: Ah, I see. Ok, carry on. Is it possible for a PR title to only contain invalid characters? Then what would be the branch name? Just a single `-`?

src/GitHub.App/ViewModels/PullRequestFileNode.cs
Haacked: Ah! I didn't catch that. Carry on.

src/GitHub.App/Services/PullRequestService.cs
grokys: Hmm I guess someone *could* create a PR using only punctuation and no letters or digits, though I'm not sure how likely it would be! And yeah, in that case the branch name would be `PR/123--`.

grokys: Thinking about this a little over the weekend, I think the current implementation could cause problems with non-Latin character sets. So I guess that if the title has no Latin characters we should fall back to simply `pr/123`.

PR #571, opened 2016-09-30
Fix build instructions

eogas:


eogas:
Or should the .gitmodules file use the public URL for this repo?

shana:
@eogas Ooops, apologies about breaking the public builds, the repo was private for a bit there. Yes, it should definitely use the public URL.

eogas:
Cool, sorry for the bad PR.  I figured it was one or the other, but I'm not really familiar with the whole git modules thing.

shana:
@eogas No worries, it was a useful PR! We would have missed the problem otherwise. It's fixed in master now, so you should be able to update the submodules now.

eogas:
Sounds good.  Thanks!

Issue #196, opened 2016-01-14
Notify the user once a new Gist has been created


georgebearden:
To further refine the `Gist` creation work being done in #194, it would be nice to automatically copy the URL returned from the GitHub API to the clipboard, and also generate a VS notification to notify the user using one of the `VSServices` methods [here](https://github.com/github/VisualStudio/blob/master/src/GitHub.Exports/Services/VSServices.cs#L29-L30).

To get the URL needed to copy to the clipboard one can subscribe to the `IReactiveCommand` [here](https://github.com/github/VisualStudio/blob/feature/gist-support/src/GitHub.Exports.Reactive/ViewModels/IGistCreationViewModel.cs#L12).

shana:
I'm adding an API for showing notifications in the status bar, Team Explorer notifications are way too intrusive imo (and requires Team Explorer to be visible, which might or might not be true)

georgebearden:
Now that #201 has been merged into master, would now be a good time to implement this feature, targeting the feature/gist-support branch?

shana:
@georgebearden Indeed! It's already implemented by the `Copy link` feature, so should be an easy addition.

paladique:
Fixed in #572

PR #572, opened 2016-10-01
Adding "Gist Created" Notification

StanleyGoldman:
Adding a simple "Gist Created" notification as requested in #196

### DIFF
--- a/src/GitHub.App/ViewModels/GistCreationViewModel.cs
+++ b/src/GitHub.App/ViewModels/GistCreationViewModel.cs
@@ -8,6 +8,7 @@ using GitHub.Exports;
 using GitHub.Extensions;
 using GitHub.Models;
 using GitHub.Services;
+using Microsoft.VisualStudio.Shell;
 using NLog;
 using NullGuard;
 using Octokit;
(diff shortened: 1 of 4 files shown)


src/GitHub.App/ViewModels/GistCreationViewModel.cs
```diff
@@ -8,6 +8,7 @@
 using GitHub.Extensions;
 using GitHub.Models;
 using GitHub.Services;
+using Microsoft.VisualStudio.Shell;
```
shana: No need for this change :)

shana:
Thank you for the PR! Apologies on the radio silence, we've been traveling and it's disrupted our flow a bit, but it's not forgotten 😄

src/GitHub.App/ViewModels/GistCreationViewModel.cs
StanleyGoldman: How did that get there...
Removed.

StanleyGoldman:
Fixed and rebased.

grokys:
LGTM, but it's out of date with master again :( Do you want to enable us pushing to the branch so we can do the merging with master?

StanleyGoldman:
I'll rebase it in a little bit

StanleyGoldman:
Rebased

grokys:
Thanks Stanley! 🎉 It may take us a while, but we do merge eventually ;)

Issue #483, opened 2016-08-09
Switch to serilog for logging


shana:
We need a logging framework that is small, doesn't block, doesn't blow up, is reliable, easy to configure.
[Serilog](https://github.com/serilog/serilog) seems to fit this bill way better than NLog does.

StanleyGoldman:
I've never used Serilog before, I've usually fall back on NLog, but I didn't mind the excuse to check it out.

This extension helps, [Visual Studio Gallery - Serilog Analyzer](https://visualstudiogallery.msdn.microsoft.com/b00c584e-71c0-4e5f-85ab-a3d2347a053c), it adds a few code analyzers to make sure you are using serilog correctly. [Github - Serilog Analyzer](https://github.com/Suchiman/SerilogAnalyzer#user-content-analyses)

PR #573, opened 2016-10-02
Switch to serilog for logging

StanleyGoldman:
Switch to serilog for logging as requested in #483


StanleyGoldman:
Hmm, I was playing with this and i realized it's not perfect. I tried logging some random things in the UIProvider after it calls `logging.Configure()` and I was not seeing anything in the log. It might have something to do with the logger instances being static members. I'll have to investigate more.

src/GitHub.App/Api/ApiClient.cs
```diff
@@ -84,7 +84,7 @@ public IObservable<UserAndScopes> GetUser()
             }
             else
             {
-                log.Error($"Error reading scopes: /user succeeded but {ScopesHeader} was not present.");
+                log.Error("Error reading scopes: /user succeeded but {ScopesHeader} was not present.", ScopesHeader);
```
shana: This changes the log message from string interpolation to a normal params method, but `{ScopesHeader}` is still in the string so this is probably not going to work. Is there a particular reason this change is needed?

shana: Or maybe `Log.Error` is doing something magical with the parameter name that I'm not familiar with?

shana:
Is it flushing the data to disk immediately when the logging methods are called?

StanleyGoldman:
I didn't get a chance to test it (I'm still stuck at work now) but here is what I think is going on.
There is a NLog.config which I think is being found by NLog and provides the initial logging configuration. Then `logging.Configure()` is being called. Which I think is just replacing the NLog output sinks with an async handler.

With Serilog, the configuration options are limited. You either do it in code, or you do it in an app.config or application.json (for .Net Core). I didn't think I can include an app.config in a Visual Studio extension. So I am performing the configuration in `logging.Configure()` which is unfortunately too late for the loggers that are declared statically on `UIProvider`.

I still have to test this to verify what I'm saying.

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -114,6 +119,16 @@ Version VSVersion
...
+            var environment = new Rothko.Environment();
+            var logPath = Path.Combine(environment.GetLocalGitHubApplicationDataPath(), "extension.log");
+
+            Log.Logger = new LoggerConfiguration()
+                .Enrich.WithThreadId()
+                .WriteTo.File(logPath,
+                    fileSizeLimitBytes: 2L * 1024L * 1024L,
+                    outputTemplate: "{Timestamp:yyyy-MM-dd HH:mm:ss.fff}|{Level}|Thread:{ThreadId}|{SourceContext}|{Message}{NewLine}{Exception}")
+                .CreateLogger();
+
```
StanleyGoldman: This is the earliest entry point I could find that allowed me to configure serilog.
It has to be done before `UIProvider` is statically constructed.

I'm not sure this is where this code should stay permanently.

shana: Yeah, entry points are tricky. There are other areas of the code, particularly the Team Explorer UI, that may be initialized before the package is, they are almost completely separate entities.

Since there are multiple initialization points and all this code needs logging (and currently not everything is using NLog because reasons), what we likely need is a separate class that run this initialization code once and everyone uses it as a gateway to the logging system. If you could move this to such a class and have every caller use that, then we don't have to worry as much about entry points.

StanleyGoldman: What other entry points do you know of?

tocsoft:
For a way to ensure that configure is all ways called before any logging occurs you could do the following;

 rename ``LoggingConfiguration`` to ``LogManager`` 

Add a static field of type ``Lazy<ILogger>`` which just calls your current ``InitializeLogging()`` method and returns the configured logger.

Now add a ``static ILogger ForContext<TSource>();`` method which internally calls into ``LazyLogger.Value.ForContext<TSource>()``.

Then everywhere your calling ``Log.ForContext<T>()`` you can now call ``LogManager.ForContext<T>()`` and you can be sure that serilog is properly configured not matter where you try and call it from at any point without having to hope certain classes are called first.

StanleyGoldman:
@tocsoft That's a good way around it.

src/GitHub.App/Infrastructure/LogManager.cs
```diff
@@ -0,0 +1,35 @@
...
+                .Enrich.WithThreadId()
+                .WriteTo.File(logPath,
+                    fileSizeLimitBytes: fileSizeLimitBytes,
+                    outputTemplate: outputTemplate)
+                .CreateLogger();
+        }
+
+        static Lazy<Logger> Logger { get; } = new Lazy<Logger>(CreateLogger);
+
+        public static ILogger ForContext<T>() => Logger.Value.ForContext<T>();
```
StanleyGoldman: So a function `FooBar<T>()` that does not return something with type `T` is apparently a code analysis error.
It can't disabled with a `#pragma`. It would have to be modified out of the rule set.

https://msdn.microsoft.com/library/ms182150.aspx

src/GitHub.App/Infrastructure/LogManager.cs
```diff
@@ -0,0 +1,36 @@
...
+                    fileSizeLimitBytes: fileSizeLimitBytes,
+                    outputTemplate: outputTemplate)
+                .CreateLogger();
+        }
+
+        static Lazy<Logger> Logger { get; } = new Lazy<Logger>(CreateLogger);
+
+        //Violates CA1004
+        //public static ILogger ForContext<T>() => Logger.Value.ForContext<T>();
+
```
StanleyGoldman: So a function `FooBar<T>()` that does not return something with type `T` is apparently a code analysis error.
It can't disabled with a `#pragma`. It would have to be modified out of the rule set.

[CA1004: Generic methods should provide type parameter](https://msdn.microsoft.com/library/ms182150.aspx)

shana: @StanleyGoldman Yeah, that's fine, we already have that rule suppressed somewhere else for the same reason.

shana:
Is this PR finished, or are you still working on it?

src/GitHub.App/Infrastructure/LogManager.cs
StanleyGoldman: Cool. So then I'll roll back the last change, and make a change to the rule set so it will pass code analysis again.

StanleyGoldman:
I'd like to handle that change above so it can use `ForContext<T>()`.
But other than that, I believe it is complete.

src/GitHub.App/Infrastructure/LogManager.cs
shana: You can add the exception directly in the code instead of in the suppressions file, that way it's easy to see

shana:
@StanleyGoldman Awesome! You might want to rebase too, there's conflicts going on with some other stuff that's already landed on master. We are currently putting out some fires in some other areas, which is why things have been so delayed around here, but we'll definitely get to it. Thank you for your patience!

src/GitHub.App/Infrastructure/LogManager.cs
```diff
@@ -28,7 +28,10 @@ static Logger CreateLogger()
 
         static Lazy<Logger> Logger { get; } = new Lazy<Logger>(CreateLogger);
 
+        //Violates CA1004 - Generic methods should provide type parameter
+        #pragma warning disable CA1004
         public static ILogger ForContext<T>() => Logger.Value.ForContext<T>();
```
StanleyGoldman: So this threw me for a loop originally, but CA1004 is an error by default and we can only pragma disable warnings.

StanleyGoldman:
Well that's strange, I didn't expect to still fail the build. This change makes the difference for me building locally.

src/GitHub.App/ViewModels/PullRequestListViewModel.cs
```diff
@@ -110,7 +114,7 @@ public override void Initialize([AllowNull] ViewWithData data)
                 .ObserveOn(RxApp.MainThreadScheduler)
                 .Catch<System.Reactive.Unit, Octokit.AuthorizationException>(ex =>
                 {
-                    // TODO: Do some decent logging here
+                    log.Error(ex, "Error Processing PullRequests");
```
tocsoft: This is explicitly an AuthorizationException not a general one, probably should have a better message i.e.
"Authorization error listing pull requests"

StanleyGoldman:
Rebased.

StanleyGoldman:
- Added a static `Log.Information(..)`, `Log.Debug(..)` etc.
- Changing every call to `Debug.Assert` in `GitHub.App` and `GitHub.VisualStudio`

grokys:
As mentioned in https://github.com/github/VisualStudio/pull/1172#discussion_r134268444 - it would be nice to have logging available to all our assemblies. At the moment, it's not available in e.g. `GitHub.Extensions`.

StanleyGoldman:
🎉

grokys:
Moved to #1223

PR #574, opened 2016-10-03
Moar details polish

donokuda:
Cleans up a little more of the details pane:

- Update directory icon
   - Update color of folder to yellow (blue for dark theme)
- Use "alert" icon for status that requires action
- Remove "alert" icon for errors and use red instead.

details-changes

/cc @grokys since this targets your branch


shana:
@grokys when you merge a PR, you should also delete the branch if it's on the same repo, so we don't accumulate merged branches.

grokys:
Yeah, I was hovering over the delete button but then I figured it was don's
branch and I didn't want to delete it from under him... Will do in future.

PR #584, opened 2016-10-08
Create Git Avatar Fix

StanleyGoldman:
The bug mentioned in #403 is fixed by making sure `RaisePropertyChanged()` gets called.
This fix is also contained in #561


grokys:
Thanks Stanley! Apologies for the delay, the team have been abroad and we're only just now catching up!

shana:
thumbs-up

PR #594, opened 2016-10-11
Fixing Account's CopyFrom

StanleyGoldman:
Fix for #408.
Edited from #561 as per @grokys request.

### DIFF
--- a/src/GitHub.App/Models/Account.cs
+++ b/src/GitHub.App/Models/Account.cs
@@ -15,6 +15,8 @@ namespace GitHub.Models
     {
         BitmapSource avatar;
 
+        IObservable<BitmapSource> BitmapSource;
+        private IDisposable _bitmapSourceSubscription;
         public Account(
             string login,
             bool isUser,
@@ -30,8 +32,10 @@ namespace GitHub.Models
             PrivateReposInPlan = privateRepositoryInPlanCount;
             IsOnFreePlan = privateRepositoryInPlanCount == 0;
             HasMaximumPrivateRepositories = OwnedPrivateRepos >= PrivateReposInPlan;
+            BitmapSource = bitmapSource;
 
-            bitmapSource.ObserveOn(RxApp.MainThreadScheduler)
+            _bitmapSourceSubscription = bitmapSource
+                .ObserveOn(RxApp.MainThreadScheduler)
                 .Subscribe(x => Avatar = x);
         }
 
@@ -80,6 +84,16 @@ namespace GitHub.Models
             IsOnFreePlan = other.IsOnFreePlan;
             HasMaximumPrivateRepositories = other.HasMaximumPrivateRepositories;
             Avatar = other.Avatar;
+
+            var otherAccount = other as Account;
+            if (otherAccount != null)
+            {
+                _bitmapSourceSubscription.Dispose();
+
+                _bitmapSourceSubscription = otherAccount.BitmapSource
+                    .ObserveOn(RxApp.MainThreadScheduler)
+                    .Subscribe(x => Avatar = x);
+            }
         }
 
         public override bool Equals([AllowNull]object obj)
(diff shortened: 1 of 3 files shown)


src/GitHub.App/Models/Account.cs
```diff
@@ -15,6 +15,8 @@ public class Account : ReactiveObject, IAccount
     {
         BitmapSource avatar;
 
+        IObservable<BitmapSource> BitmapSource;
+        private IDisposable _bitmapSourceSubscription;
```
shana: Private fields should be lowercase with no underscore prefix. Feel free to use `this` during constructor assignment if the constructor argument name matches the field it's being assigned to.

We also don't use the `private` keyword in the codebase (it is not by any means a consensual choice, but we're following it for consistency)

src/GitHub.App/Models/Account.cs
```diff
@@ -30,9 +32,11 @@ public class Account : ReactiveObject, IAccount
             PrivateReposInPlan = privateRepositoryInPlanCount;
             IsOnFreePlan = privateRepositoryInPlanCount == 0;
             HasMaximumPrivateRepositories = OwnedPrivateRepos >= PrivateReposInPlan;
+            BitmapSource = bitmapSource;
 
-            bitmapSource.ObserveOn(RxApp.MainThreadScheduler)
-                .Subscribe(x => avatar = x);
+            _bitmapSourceSubscription = bitmapSource
+                .ObserveOn(RxApp.MainThreadScheduler)
+                .Subscribe(x => Avatar = x);
```
shana: The change that fixed the property setting is already on master so your branch is conflicting on it, need to resolve that before we can merge.

src/UnitTests/GitHub.App/Models/AccountModelTests.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            Func<int, BitmapImage, IObservable<BitmapImage>> generateObservable = (seconds, image) =>
+            {
+                return Observable.Generate(
+                    initialState: 0, 
+                    condition: i => i < 1, 
+                    iterate: state => state + 1,
+                    resultSelector: i => image, 
+                    timeSelector: i => TimeSpan.FromSeconds(seconds));
+            };
```
shana: Like @grokys mentioned in the other PR, timings are iffy, it's safer to use a `Subject` or perhaps a `ReplaySubject` and do `OnNext` on it whenever you want to push something. Saves on waits, too :)

shana:
So I was wondering the consequences of disposing and resubscribing to the bitmap source so I went looking at the image cache that's serving this observable and oh boy oh boy is it a mess down there. Note to self, need to clean up the ImageCache code, it's very silly.

src/GitHub.App/Models/Account.cs
StanleyGoldman: Rebased

src/GitHub.App/Models/Account.cs
StanleyGoldman: Gotcha

src/UnitTests/GitHub.App/Models/AccountModelTests.cs
StanleyGoldman: Good call. I was able to write a better unit test without it.

src/UnitTests/GitHub.App/Models/AccountModelTests.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            //Updating the account in the collection
+            col.AddItem(updatedAccount);
+
+            //Waiting for the collection to process the update
+            collectionEvent.WaitOne();
+            collectionEvent.Reset();
+
+            //Providing the second avatar
+            bitmapImageObservable.OnNext(orgImage);
```
StanleyGoldman: Updated

StanleyGoldman:
Rebased

shana:
Woot! Thank you for this fix, apologies (again) for the delays, we're really shorthanded around here.

PR #595, opened 2016-10-12
Fixes/tracking collection newer comparer

StanleyGoldman:
Edited from #561 as per @grokys request.

While debugging #408 I discovered that `CopyFrom()` is called a quite often for objects that I thought were equivalent.

After some digging I noticed that with the `NewerComparer` property set, objects that are older/newer were being correctly ignored/copied, but objects that were created at the same time were being incorrectly copied.

### DIFF
--- a/src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
+++ b/src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
@@ -539,7 +539,7 @@ namespace GitHub.Collections
                 var old = list[idx];
                 if (newer != null)
                 {
-                    // the object is "older" than the one we have, ignore it
+                    // the object is not "newer" than the one we have, ignore it
                     if (newer(item, old) >= 0)
                         return new ActionData(TheAction.Ignore, list, item, null, idx, idx);
                 }
--- a/src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
+++ b/src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
@@ -540,7 +540,7 @@ namespace GitHub.Collections
                 if (newer != null)
                 {
                     // the object is "older" than the one we have, ignore it
-                    if (newer(item, old) > 0)
+                    if (newer(item, old) >= 0)
                         return new ActionData(TheAction.Ignore, list, item, null, idx, idx);
                 }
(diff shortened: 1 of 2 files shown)


src/TrackingCollectionTests/TrackingCollectionTests.cs
```diff
@@ -429,8 +429,8 @@ public void OrderByMatchesOriginalOrder()
         var count = 0;
         var total = 1000;
 
-        var list1 = new List<Thing>(Enumerable.Range(1, total).Select(i => GetThing(i, i, total - i, "Run 1")).ToList());
-        var list2 = new List<Thing>(Enumerable.Range(1, total).Select(i => GetThing(i, i, total - i, "Run 2")).ToList());
+        var list1 = new List<Thing>(Enumerable.Range(1, total).Select(i => GetThing(i, i, i, "Run 1")).ToList());
```
shana: This test is testing whether `OrderBy` will correctly retain the order of the collection if all items added to it are in the same order as the `OrderBy` comparer. You're changing the order here, which changes the conditions of the test (by forcing the collection to actually reorder). Not that it's going to fail the test, but that's not what this test is supposed to do, this test is exercising the code path where no re-ordering is required.

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
```diff
@@ -540,7 +540,7 @@ ActionData ProcessItem(ActionData data, List<T> list)
                 if (newer != null)
                 {
                     // the object is "older" than the one we have, ignore it
-                    if (newer(item, old) > 0)
```
shana: Should probably update the comment as well, since it no longer matches the logic 😄

shana:
Nice catch, thanks for the patch! I just have some comments on a few of your changes, nothing major 😄

src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
StanleyGoldman: :)

StanleyGoldman:
I believe i have corrected this unit tests code path, let me know if you agree.

shana:
Looks great! ✨ 
thank-you

PR #596, opened 2016-10-12
Backport to 2.0.15.1: Async Package load

shana:
This is #528 applied to 2.0.15.1. We may need to do a release for it before master is ready, and I don't want to lose the conversation on #528.


src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -112,31 +154,36 @@ Version VSVersion
...
-                var shell = await GetServiceAsync(typeof(SVsShell)) as IVsShell;
-                IVsPackage startPagePackage;
-                if (ErrorHandler.Failed(shell?.LoadPackage(new Guid(StartPagePreview4PackageId), out startPagePackage) ?? -1))
-                {
-                    // ¯\_(ツ)_/¯
-                }
-            }
+            AddService(typeof(IUsageTracker), CreateService, true);
+            AddService(typeof(IMenuProvider), CreateService, true);
+            return Task.FromResult<object>(null);
```
tocsoft: I feel ``Task.FromResult(true)`` is slightly easier to read than  `Task.FromResult<object>(null)`. Less brackets to track while mentally parsing.

tocsoft:
@shana you will probably want to back port #621 (for version numbers) into this as well, and possibly #651 (for broken icon) if you are pushing out a release not based on master.

src/GitHub.VisualStudio/GitHubPackage.cs
shana: We have `Task.CompletedTask` for this purpose, let's use that.

shana:
@tocsoft Yes, there are several other fixes we'll need, we're still trying to figure out if those go in this release or separately.

tocsoft:
👍

PR #598, opened 2016-10-13
Correct path casing in open/copy links

### LINKED ISSUE #472: Open on GitHub sometimes 404s when path contains upper case

paladique:
Fixes #472 

Compares Document.FullName to ProjectItem.FileName[1] to find proper casing.

<s>per discussion of iterating through `FileNames` to find the proper casing.</s>

<s>While examining the values of the indexes in FileNames, it seems as if they're never null and/or always the same.  Instead of going through all 32,767 values, a small sample was selected. 

<s>Compared the two paths to find where they don't match and changed the FileName to lowercase to compare against the Name to see if they're the same, which would then mean that that `FileName` has the proper casing so exit the loop.</s>


src/GitHub.VisualStudio/Helpers/ActiveDocumentSnapshot.cs
```diff
@@ -19,8 +19,14 @@ public class ActiveDocumentSnapshot : IActiveDocumentSnapshot
         public ActiveDocumentSnapshot([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider)
         {
             StartLine = EndLine = -1;
-            Name = Services.Dte2?.ActiveDocument?.ProjectItem.Name;
+            Name = Services.Dte2?.ActiveDocument?.FullName;
+            var projectItem = Services.Dte2?.ActiveDocument?.ProjectItem;
 
+            if ((String.Compare(Name, projectItem.FileNames[1], StringComparison.Ordinal) != 0) && (String.Compare(Name, projectItem.FileNames[1], StringComparison.OrdinalIgnoreCase) == 0))
```
shana: Any particular reason to do two comparisons here?

paladique: If the `FileName` and `Name` are the same in the first comparison then it's safe to assume we have the proper casing. But I guess that check is done in the second comparison as well. This could probably be condensed with a one line conditional operator for `Name` instead of possibly changing the value of `Name` in the `if` statement.

shana: Right, you could technically only grab the name from `FileNames` if the casing is wrong in `ActiveDocument.FullName`, but in practice, `FileNames` always has the correct case while `ActiveDocument.FullName` sometimes has the wrong case, which means we might as well trust `FileNames` always and only do one comparison to confirm that the `FileNames` entry is the correct one.

shana:
This looks good! What test cases should we have for QA to make sure this works and doesn't regress?

paladique:
I would suggest:
- Testing copy/open on a file with a long URL/directory (about 3/4 directories deep) to confirm proper casing

ie: something like https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml

- Updating the Setting VS that will load the last solution that was open, and trying open/copy on any file, and confirming the casing

This should probably go into the manifest? @shana

shana:
@femisimon This is going in and @paladique has noted some tests that should be done to validate this above. Let us know if you need more info on how this works for your test cases.

femisimon:
@shana I will sync with @paladique on this to get more information and add more test cases

PR #600, opened 2016-10-14
Make ScrollViewer hack reusable.

grokys:
WPF's ScrollViewer is broken, so we need to hack around it. We already had this hack in GitHubConnectContent, so moved the hack to a central place that can be used wherever it's needed.


shana:
thumbs-up-scroll

PR #601, opened 2016-10-14
Update Styles.xaml

kessanro:
#3da4f4 bit diffrent at line i forgot

### DIFF
--- a/src/GitHub.UI/Assets/Styles.xaml
+++ b/src/GitHub.UI/Assets/Styles.xaml
@@ -8,7 +8,7 @@
                     mc:Ignorable="PresentationOptions">
 
     <!-- The blue accent color used through-out the app -->
-    <Color x:Key="GitHubAccentColor">#4ea6ea</Color>
+    <Color x:Key="GitHubAccentColor">#3da4f4</Color>
     <SolidColorBrush x:Key="GitHubAccentBrush" Color="{StaticResource GitHubAccentColor}"
                      PresentationOptions:Freeze="true" />
 


shana:
Hi @kessanro! We appreciate the effort, but don't really see the relevance of this change. Is there something in our blue shade that's bothering you?

kessanro:
kindof it does not match github blue

kessanro:
this is a start im going to try to make it asyer since my laptop was in for repair

shana:
@donokuda Could you double check out shades of blue, see if they match our style in the website and other apps?

donokuda:
I did a search for `GitHubAccentColor` and it looks like we don't use it in our extension at all. This might have been a color imported from GitHub Desktop for Windows.

image

As far as I know, we don't use specify any blue in our extension. If we do, we should strongly consider using the color Visual Studio provides us instead of a custom blue (for controls like a hyperlink.)

shana:
> I did a search for GitHubAccentColor and it looks like we don't use it in our extension at all. This might have been a color imported from GitHub Desktop for Windows.

Oh, haha, fun, dead code! 😄 

> As far as I know, we don't use specify any blue in our extension. If we do, we should strongly consider using the color Visual Studio provides us instead of a custom blue (for controls like a hyperlink.)

The dialogs for login and friends use a custom theme that matches the desktop app but not VS (on purpose). The login/clone/create buttons on those dialogs are blue (have a blue border and animate to blue when hovered upon). I wonder what blue we're using there...

donokuda:
> The dialogs for login and friends use a custom theme that matches the desktop app but not VS (on purpose). The login/clone/create buttons on those dialogs are blue (have a blue border and animate to blue when hovered upon). I wonder what blue we're using there...

D'oh! Forgot about those dialogs.

It looks like we're using two different blues: `#4a88c3` for the blue text and `#49a5ed` for the circle/icon buttons.

shana:
@kessanro So it looks like the color that you changed in this PR is not actually being used, so there's no real point in doing this change. If you have suggestions on how to improve our UI, please open an issue for it, preferably with mockups of what you want to do, so we can have a discussion about that!

kessanro:
ok

PR #606, opened 2016-10-18
Remove old package test

### LINKED ISSUE #541: GitHubPackageTests doesn't have [Fact] attribute
The [AddsTopLevelMenuItems](https://github.com/github/VisualStudio/blob/62567638cf02ea43885233147659f5b76665ddb6/src/UnitTests/GitHubPackageTests.cs#L11) test doesn't have a `[Fact]` attribute so isn't recognised as a test.

paladique:
With updates coming to the package loader, this test won't be needed anymore.
Fixes #541


shana:
Woot! 🚀

PR #607, opened 2016-10-18
More polish for pull request details

donokuda:
This pull request reworks the header and tweaks a few things in the PR details view:

- Moved PR status, number, and author information under title
- Move PR branch, commits/files changed count, and branch switcher into the scrollviewer to give us more vertical room.

---

I tried a version with the pull request icon in the title but wasn't sure how I felt about it:

---

❗️ Heads up: the indentation in `PullRequestDetailView.xaml` hasn't been corrected to keep this diff sane and minimize the number of potential conflicts. After this is merged, lines 138-314 and 316-472 need to be indented. I can quickly follow up this PR with one that fixes the indentation.

Closes https://github.com/github/VisualStudio/issues/625


donokuda:
Any idea on how I can change the icon for toggling the menu in the files changed tree view?

image

StanleyGoldman:
Definitely looks better without the icon in the title.

donokuda:
> Definitely looks better without the icon in the title.

Yeah, I'm leaning towards the title without the icon.

----

@shana @grokys Is there something I need to do myself to get janky to pass? I'm at a loss with how to fix my builds.

Other than that, this is ready for review.

tocsoft:
I feel the `open`/`closed`/`merged` labels would look better, and more in keeping with .com, if they were more like tags with the icon in tag with green/red/purple background and white text. Added bonus this should work well across all the VS themes (and any other) without needing to worry about contrast etc.

something like this is what i was thinking:
label
_excuse the quality of the image just a quick crop cut & paste_

donokuda:
@tocsoft Thanks for the suggestion! An earlier iteration of the details view used to treat the status in a similar fashion as your suggestion (see https://github.com/github/VisualStudio/pull/564). I removed the green background since it tended to draw more attention than it needed to. Additionally, having a flat green background made it feel more like a button.

I felt like it was a good compromise to remove the background and use the status color as the text's foreground instead. 😄

tocsoft:
Seems a logical conclusion.

I think my main problem is I personally find the green text a little hard to read, on both the dark and light themes. I feel like the contrast is just isn't quite high enough for me. Saying that the more a look as some of the other parts of the VS UI the blue text they use, when its on white, isn't much better. 

But the more I think about it, it could quite as easily be the rubbish monitors i'm using.

shana:
@donokuda Your animated screenshots reminded me of that conversation we had waaaaay back at the beginning (and I think I had it with simurai too iirc), about docking things in a way that we keep certain elements in view when scrolling content. I feel like maybe the files changed section is one of those that should be always in view regardless of the size of the PR description, what do you think?

I think we're going to leave this open for now, until we get back on track with the maintainer workflow, so there's no rush on this (maybe nice that we have the pause and then can go back and look at what we have with fresher eyes 😄)

shana:
Oh, just noticed your comment about gaining more space by putting things inside the scroll viewer. Yup, that's an excellent point! Maybe we should just keep the actionable items (like the link to github) always visible so users don't have to wonder how to get to them and let everything else be scrollable.

Also, :+1: on title without icon, agree it looks better.

paladique:
Also 👍 without the icon!

donokuda:
Closing this out to address https://github.com/github/VisualStudio/issues/713

PR #614, opened 2016-10-22
Image Cache Refactor

StanleyGoldman:
Based on a comment from @shana on #594

### DIFF
--- a/src/GitHub.App/Caches/ImageCache.cs
+++ b/src/GitHub.App/Caches/ImageCache.cs
@@ -64,18 +64,12 @@ namespace GitHub.Caches
         IObservable<BitmapSource> GetImageImpl(Uri url)
         {
             var maxCacheDuration = TimeSpan.FromDays(30);
+
+            // Between 1 and 2 days.
+            var refreshInterval = TimeSpan.FromSeconds(random.Next(60 * 60 * 24, 60 * 60 * 24 * 2));
+
             return Observable.Defer(() =>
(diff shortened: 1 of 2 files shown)


paladique:
👋 Hey there! How's this coming along? Still a WIP?

StanleyGoldman:
Hey @paladique 👋 ..short answer is yes this is still a WIP.

I was planning on refactoring some of the areas in the ImageCache that looked a little strange. So i just commented all the unit tests to make sure I understood them and I added a few tests around the constructor and the "vacuum" functionality.

The tests I added work, I just need to spend some time actually refactoring the code now.

StanleyGoldman:
The two sections that looked strange to me were the parts around `Vacuum()` and around `GetAndRefresh()` with the large comment block. I haven't looked at it in a while and it makes more sense to me than it did before, so that is good.

The vacuuming code, is mostly a protection to not call Akavache's `Vacuum` method too often. Not sure it is a necessary protection, but it can't hurt.

`GetImageImpl` will generally cache images for 30 days. Akavache's `GetAndRefresh` is a method is rather useful and allows for us to return the cached image meanwhile refreshing the image if it is older than 1 or 2 days. I'm not sure that there is a benefit around randomizing the 1-2 day refresh period, but I believe that `lock` around the `random` object is not necessary as `GetImageImpl()` is only called through a `SerializedObservableProvider`.

Rebased.

PR #615, opened 2016-10-23
add ICanBeBusy and show indeterminate progress bar

### LINKED ISSUE #554: Add IsBusy state to viewmodels that can have it
We need a way to indicate busy state outside of the viewmodel, so that hosting containers (like the GitHubPane) can disable the UI and show a busy state indicator. This should be a thing that's automatically handled if a viewmodel indicates busy state.

## The proposal is:

- Have an interface `ICanBeBusy` or `IHasBusy` (or `IXKCD303` if especially busy :trollface:)
- Have a property `IsBusy`
- Have an implementation with `NotificationAwareObject` that notifies on changes
- Have the GitHubPaneViewModel hook up listeners for viewmodels that it is hosting that support the interface
- When `IsBusy` is signaled true, disable the hosted UI and either:
  - Show an overlay with a busy indicator
  - Show a busy "progress" bar indicating work under way
  - Any other way of indicating we're doing things in the UI

Thoughts, opinions, protest signs welcome

tocsoft:
Wanted to help out and thought this issue looks like it shouldn't be too difficult or would require any sweeping changes to implement. 
### Gotchas

`ICanBeBusy` will only work when the class also implements `INotifyPropertyChanged` (basically if the ViewModel inherits from `NotificationAwareObject` then it'll work as expected)
### Overview of changes:
- Added a new interface `ICanBeBusy`
- Updated `GitHubPaneViewModel` to proxy `IsBusy` from hosted views whose VM implement `ICanBeBusy`. 
- I show an indeterminate progress bar on `GitHubPaneView` and disabled hosted view if `IsBusy` set to true.
- Proxy `IsBusy` from `ICanBeBusy` to the `IsBusy` Observable in `SimpleViewUserControl`. 
  _(not sure is this one wanted doing but it looked like it was there for a reason)_
### Other things of note

This PR doesn't actually enable any of the current viewmodels to use `ICanBeBusy` (yet), would prefer some guidance on which one should use this, didn't just want to be disabling any random views as I assumed they are already internally handling there own busy state UI.

What I could do is just stop disabling the view (let the view handle its own disabling if required) and then I feel pretty confident that I could just slap `ICanBeBusy` on anything that currently setting the `IsBusy` property on `BaseViewModel`. We can always add a second interface that explicitly opts into the view being disabled wholesale an `ICanBeReallyBusy` for example.

Fixes #554

### DIFF
--- a/src/GitHub.Exports/GitHub.Exports.csproj
+++ b/src/GitHub.Exports/GitHub.Exports.csproj
@@ -160,6 +160,7 @@
     <Compile Include="Settings\RepositoryUIState.cs" />
     <Compile Include="Settings\UIState.cs" />
     <Compile Include="SimpleJson.cs" />
+    <Compile Include="ViewModels\ICanBeBusy.cs" />
     <Compile Include="ViewModels\IInfoPanel.cs" />
     <Compile Include="ViewModels\IServiceProviderAware.cs" />
     <Compile Include="UI\IView.cs" />
--- a/src/GitHub.Exports/ViewModels/ICanBeBusy.cs
+++ b/src/GitHub.Exports/ViewModels/ICanBeBusy.cs
@@ -1,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GitHub.ViewModels
+{
+    public interface ICanBeBusy
+    {
+        bool IsBusy { get; }
+    }
+}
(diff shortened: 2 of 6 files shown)


tocsoft:
just been reading over 'Maintainer workflow' and think that this PR is probably superfluous as it looks like some of this stuff has been rolled into that one... i'll leave this here anyway as it looks like the 2 implementations are a bit different but if you want to close this in favour of whats coming in there then that's understandable.

shana:
Thank you so much for this! We've been going back and forth on how to handle busy states and forward them from hosted view/viewmodels out to the container, and what you did here on this PR is one of the possible approaches, we just had no time to actually implement it to try it out.

I'll wait until @grokys has a chance to look this over and comment on it as he's been doing most of the work that needs this.

tocsoft:
That's fine.. feel free to just treat this more as a suggested option that a real request for merging in.

sguthals:
Closing because this is stale.

PR #621, opened 2016-10-24
Fix the about version number to match release number.

### LINKED ISSUE #619: Version number in Help About does not match Extension Manager
The version of the install github extension always says v1.0 in Help About

image

However, in the Extension Manager, it displays the actual version number, e.g. v2.0.15.1

image

tocsoft:
Makes the visual studio about screen show the same version number as the package.

fixes #619 and prevents it form coming out of sync as long as `AssemblyVersionInformation` is updated correctly.


grokys:
Thanks @tocsoft ! 🎈 🎉

tocsoft:
😀 no problem, just small fix but every little helps.

PR #634, opened 2016-10-25
Update TE Invitation Buttons to match VS 

### LINKED ISSUE #599: Fix VsButtonStyleKey to match TE button style
`VsButtonStyleKey` is matching the standard VS button style, but the TE button style is slightly different. Since we're only using this button style in TE sections, we should fix it to match TE.

## TE button, non-hover and hover state, dark theme
not-hover-dark-theme
hover-dark-theme

## TE button, non-hover and hover state, white theme
not-hover-white-theme
hover-white-theme

## Our button, hover state, dark theme
Background and border need fixing
![our-style-hover-dark-theme]
(https://cloud.githubusercontent.com/assets/310137/19354869/89d99b60-9168-11e6-8442-81d093dff51b.png)

## Our button, non-hover and hover state, white theme
Border needs fixing
our-style-non-hover-white-theme

our-style-hover-white-theme

paladique:
Fixes #599 

Paired on this with @donokuda to find out that although `VSButtonStyle` was using the default VS button style, it didn't have any additional info that a custom style could be set. So instead we updated `GitHubVsButton`, replaced `VSButtonStyle` with it, and made small changes to the button states to make it uniform with TE buttons.

Don also noticed that the margins within the button were a little off in comparison to the other TE buttons, should we adjust that?

Another thing of note is that `VSButtonStyle` is still being used on the Publish Repository button, should that be switched out as well?

# Black
black
blackhover

#  Blue
blue
bluehover

# White
whitw
whitehover


grokys:
Does this only apply to VS15, not 2015?

paladique:
@grokys Testing this out was all done in VS2015, but I'll check the effects it has in VS15

shana:
I checked, VS15 is the same.

WRT margins, I like our button, it looks balanced. The TE buttons look heavy at the bottom. I'm fine with it as-is personally, it's going to be nicely balanced on the `Publish` UI too (let's get screenshots of that one to make sure)

paladique:
## Blue
blue
bluehover

## Light

white
whitehover


## Dark
black
blackhover

shana:
> Another thing of note is that VSButtonStyle is still being used on the Publish Repository button, should that be switched out as well?

Yup, let's switch that one as well and get everything uniform.

shana:
Weird, I forgot to hit enter on the previous comment until now... looks good to me! We do look a bit fatter on that button but dunno, @donokuda?

donokuda:
> Weird, I forgot to hit enter on the previous comment until now... looks good to me! We do look a bit fatter on that button but dunno, @donokuda?

We do,  I think it's better to adjust the padding so that we're consistent with the default button styles.

If I recall correctly, if we adjust the [padding on the button](https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio.UI/Styles/Buttons.xaml#L11), then we also need to do the same with the [primary button styles](https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio.UI/Styles/Buttons.xaml#L103) so that they'll look consistent on the create pull request view.

We could follow up the padding in a separate PR if that makes sense. Don't want that blocking this PR since it feels like fairly minor cosmetic nit-pick.

shana:
Sounds good, let's do that!

PR #636, opened 2016-10-25
Fix checking out PR branch from fork.

### LINKED ISSUE #635: Checkout PR Branch from a fork returns error messages
# Description
PR branch from a fork currently returns error messages when repo owner tries to **Checkout Pull request.** 
Screenshot attached.

- GitHub Extension version: v2.0.15.3-alpha1 
- Visual Studio version: Enterprise 2015 Update 3

**Steps to reproduce**
- Fork a repo as a contributor.
- Switch to a branch.
- Make a change, commit and make a PR
- Launch new instance of Visual Studio
- Login as owner
- On Team Explorer pane ==> Home Icon ==> Pull Requests
- Select the PR from contributor
- On PR detail pane click `Checkout pull request`

**Expected** - Switch branch to the PR branch, providing details of the PR(File changed in File Explorer)

**Actual:** Returns Error messages:
- Value cannot be null
- Parameter name: input

snip20161024_89

grokys:
Use the `Model.Title` property which contains the PR title rather than the `BaseViewModel.Title` property which is unused in PullRequestDetailsViewModel.

Fixes #635.

### DIFF
--- a/src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
@@ -399,7 +399,7 @@ namespace GitHub.ViewModels
                     break;
                 case CheckoutMode.Fetch:
                     operation = pullRequestsService
-                        .GetDefaultLocalBranchName(repository, Model.Number, Title)
+                        .GetDefaultLocalBranchName(repository, Model.Number, Model.Title)
                         .SelectMany(x => pullRequestsService.FetchAndCheckout(repository, Model.Number, x));
                     break;
                 case CheckoutMode.Switch:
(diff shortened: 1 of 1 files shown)


tocsoft:
Wouldn't it make more sense to push Model.Title into BaseViewModel.Title so your not left with having to remember that on some view models you use BaseViewModel.Title and on others you use some other property for the title?

grokys:
Yeah, I did use `BaseViewModel.Title` previously, but @shana didn't like the fact I was copying model properties to the view model. To be honest I don't think `Title` belongs on `BaseViewModel` - not all view models will have a title.

tocsoft:
Ah ok... that makes sense, the real fix is removing the `Title` from `BaseViewModel` and re-adding it higher up the chain as required. (Totally outside the scope of this PR.)

shana:
Yeah, the `Title` in the `BaseViewModel` is more the display name of the viewmodel itself rather than the specific title of the PR that we're showing. It probably needs rethinking, it's not really appropriate there.

PR #647, opened 2016-10-25
Fix the remote handling and tweak the threading a bit

shana:
@grokys 

The branch is blowing up on the `VisualStudio` repo because some of the PRs we have have had their original repos deleted, and the `GitReferenceModel` code can't handle that. I've tweaked the `Guard` methods so that they don't actually throw in debug mode outside of unit tests, so we don't crash while we test (they're already asserting, throwing is useless).

We should probably be storing the `IRepositoryModel` itself in `GitReferenceModel`, instead of the clone url, that would make things a lot easier. ~~As is, we probably need to use `UriString? RepositoryCloneUrl` to work around the nullness there.~~ This is on the target branch already.

In the meantime, I got another crash due to my clone using ssh urls, and none of the `GitClient` new methods calling `GetHttpRemote` to make sure they're using http urls that we can handle, so I went ahead and did that for all of them. Since `GetHttpRemote` is an async call, we can switch to a background thread from it for "free" by doing `ConfigureAwait`, so we don't have to manually do `Task.Factory.StartNew`.

And while I was at it, following the comment on https://github.com/github/VisualStudio/pull/528#discussion_r81361057 which I think is a pretty neat way of switching to background threads if you're in the main thread, I added an extension method to do it so we don't have to take a dependency on `Microsoft.VisualStudio.Threading` for it.

This is targeting the diff branch for now, once that's in master I'll retarget it.

### DIFF
--- a/src/GitHub.App/Extensions/AwaitExtensions.cs
+++ b/src/GitHub.App/Extensions/AwaitExtensions.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace GitHub.Extensions
+{
+    [NullGuard.NullGuard(NullGuard.ValidationFlags.None)]
+    public static class AwaitExtensions
+    {
+        public static TaskSchedulerAwaiter GetAwaiter(this TaskScheduler scheduler)
+        {
+            Guard.ArgumentNotNull(scheduler, nameof(scheduler));
+            return new TaskSchedulerAwaiter(scheduler);
+        }
+
+        public struct TaskSchedulerAwaiter : INotifyCompletion
+        {
+            readonly TaskScheduler scheduler;
+
+            public bool IsCompleted
+            {
+                get
+                {
+                    bool isThreadPoolThread = Thread.CurrentThread.IsThreadPoolThread;
+                    return (scheduler == TaskScheduler.Default & isThreadPoolThread) || (scheduler == TaskScheduler.Current && TaskScheduler.Current != TaskScheduler.Default);
+                }
+            }
+
+            public TaskSchedulerAwaiter(TaskScheduler scheduler)
+            {
+                Guard.ArgumentNotNull(scheduler, nameof(scheduler));
+                this.scheduler = scheduler;
+            }
+
+            public void OnCompleted(Action continuation)
+            {
+                Task.Factory.StartNew(continuation, CancellationToken.None, TaskCreationOptions.None, scheduler);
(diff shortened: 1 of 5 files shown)


tocsoft:
ConfigureAwait(false) doesn't guarantee continuation on a thread pool thread. See this for details https://msdn.microsoft.com/en-us/magazine/jj991977.aspx

If the awaited task was complete already then it won't capture its context and immediately continue on the current thread. You might want to force it to a background thread first.

tocsoft:
Sorry it was late and I was on my phone.. there's a section in the article that talks about `ConfigureAwait(false)` in that section it mentions that it might not continue on a background thread and what can cause it.

shana:
@tocsoft Ah yes, of course, that's what I get for pushing code at 1am.

grokys:
>I've tweaked the Guard methods so that they don't actually throw in debug mode outside of unit tests, so we don't crash while we test (they're already asserting, throwing is useless).

Yeah, to be honest I'm not sure that using `Debug.Fail` even in debug mode is a good idea - assertions cause havoc with the XAML designer - you can get into a place where VS appears to have completely hung, until you notice that there an assert dialog box in the background (that sometimes can't be clicked to dismiss and a VS restart is required). For this reason, I'm going off the idea of assertions altogether; I'd prefer to just throw.

> As is, we probably need to use UriString? RepositoryCloneUrl to work around the nullness there.

UriString is a class, so it can already be null. There's already [code for handling a null Head reference](https://github.com/github/VisualStudio/blob/39a27bf958eab1794519f40f132d5f7dab37432b/src/GitHub.App/Services/ModelService.cs#L490) but it sounds like I'm just handling it wrong - it sounds like only `Head.Repository` is null rather than `Head` itself. Could you point me to a PR that causes a problem?

shana:
> Yeah, to be honest I'm not sure that using Debug.Fail even in debug mode is a good idea - assertions cause havoc with the XAML designer

Good point. Throwing will make it harder to run debug builds in VS, but I guess we always do that while attached to a debugger anyway. It's a fairly easy thing to switch back in any case and it only affects us, so I'm fine not doing asserts.

shana:
> UriString is a class, so it can already be null. There's already code for handling a null Head reference but it sounds like I'm just handling it wrong - it sounds like only Head.Repository is null rather than Head itself. Could you point me to a PR that causes a problem?

Nevermind, it's already sorted in the target branch of the PR.

shana:
Leaving this one open until we can get back to the maintainer workflow.

grokys:
This should have been fixed by #724.

Issue #646, opened 2016-10-25
Deleted files not Indicated with strikethrough in PR details


femisimon:
# Description

Deleted files in a branch with open/closed PR should indicate that the particular file(s) were removed or deleted from the repo by been strikethrough. Currently, deleted files are listed under `Changes` is seen as other file changed.
- GitHub Extension version: v2.0.15.3-alpha1 
- Visual Studio version: Enterprise 2015 Update 3.
## **Steps to reproduce**
- Launch Vs and login to GHfVS
- Select a repo with Open PR and switch to the branch
- Delete a file from the branch
- Commit changes made and sync.
- Click Team Explorer ==> Home Icon ==> Pull Requests
- Select the open PR with deleted file(s)
### **Expected**:
- Deleted file(s) has ~~strikethrough~~ and colored red.
### **Actual:**
- Deleted file(s) is displayed as a regular file on the `Changes` section

grokys:
This feature is coming in #637. However, the file isn't colored red as I was trying to match the way that Team Explorer displays files. @donokuda do we want to display deleted files in red, different from Team Explorer?

shana:
This UI is ours, not Team Explorer's, so there's no reason for us to use red here, we should have our own branded style.

@donokuda Do we have a set GitHub style currently for displaying deleted files? If not, my personal preference is not red, red is... not really subtle. Maybe a duller or more faded color? Gray?

donokuda:
> @donokuda Do we have a set GitHub style currently for displaying deleted files? If not, my personal preference is not red, red is... not really subtle. Maybe a duller or more faded color? Gray?

Sorry I missed this earlier. We don't have a style for deleted files yet and 💯 on gray. I think the strike-through w/ gray should be enough to communicate that a file was deleted.

PR #648, opened 2016-10-26
Fixing the display of deleted files in pull requests

StanleyGoldman:
Fixing the issue mentioned in #646

Image

### DIFF
--- a/src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
+++ b/src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
@@ -451,7 +451,8 @@
                                                 <TextBlock.Style>
                                                     <Style TargetType="TextBlock">
                                                         <Style.Triggers>
-                                                            <DataTrigger Binding="{Binding ChangeType}" Value="Removed">
+                                                            <DataTrigger Binding="{Binding Status}" Value="Removed">
+                                                                <Setter Property="Foreground" Value="Red"/>
                                                                 <Setter Property="TextDecorations" Value="Strikethrough"/>
                                                             </DataTrigger>
                                                         </Style.Triggers>
--- a/src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
+++ b/src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
@@ -452,7 +452,7 @@
                                                     <Style TargetType="TextBlock">
                                                         <Style.Triggers>
                                                             <DataTrigger Binding="{Binding Status}" Value="Removed">
-                                                                <Setter Property="Foreground" Value="Red"/>
+                                                                <Setter Property="Foreground" Value="{DynamicResource GitHubDeletedFileBrush}"/>
                                                                 <Setter Property="TextDecorations" Value="Strikethrough"/>
                                                             </DataTrigger>
                                                         </Style.Triggers>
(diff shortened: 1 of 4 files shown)


grokys:
Thanks @StanleyGoldman, only question is if we want to retain the style used elsewhere in Team Explorer where red is not used (which is what the current style is based on). If we decide we want to use a different style, I will merge this.

src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
```diff
@@ -448,7 +448,8 @@
                                                 <TextBlock.Style>
                                                     <Style TargetType="TextBlock">
                                                         <Style.Triggers>
-                                                            <DataTrigger Binding="{Binding ChangeType}" Value="Removed">
+                                                            <DataTrigger Binding="{Binding Status}" Value="Removed">
+                                                                <Setter Property="Foreground" Value="Red"/>
```
shana: Colors shouldn't be hardcoded, as the UI needs to adapt to theme changes. We have different files defining the colors for each theme under the `Styles` folder, any control that needs theme-aware colors should reference resources from there.

StanleyGoldman: Fixed up.

StanleyGoldman:
@donokuda What kind of grey should be used instead?

donokuda:
Hey @StanleyGoldman! I pushed 7063bfc to your branch (success! 🎉) to use the `VsColor.GrayText` brush for a more subdued gray. I chose VisualStudio's gray so we'll have the right color for each theme out of the box.

image

The only thing I couldn't figure out is updating the Octicon's `Foreground` to be the same gray when the file is deleted.

shana:
> The only thing I couldn't figure out is updating the Octicon's Foreground to be the same gray when the file is deleted.

Oh, probably the same way as we're doing in the name, with an inline style with triggers to switch the background color in https://github.com/StanleyGoldman/VisualStudio/blob/7063bfc8a4f1837d686c52aeba3f2375f9a78fd6/src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml#L446

shana:
This looks great! Perhaps we should add the icon color change as well in this PR and have it all nicely wrapped up, what do you think @StanleyGoldman?

StanleyGoldman:
Yep, let's wrap this one up.

src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
```diff
@@ -443,7 +443,18 @@
...
                                                     ContextMenu="{StaticResource FileContextMenu}"
                                                     Tag="{Binding DataContext, RelativeSource={RelativeSource Mode=FindAncestor,AncestorType={x:Type te:SectionControl}}}">
-                                            <ui:OcticonImage Icon="file_code" Foreground="{DynamicResource GitHubVsWindowText}"/>
+                                            <ui:OcticonImage Icon="file_code" Foreground="{DynamicResource GitHubVsWindowText}">
+                                                <ui:OcticonImage.Style>
+                                                    <Style TargetType="ui:OcticonImage" BasedOn="{StaticResource OcticonImage}">
+                                                        <Style.Triggers>
+                                                            <DataTrigger Binding="{Binding Status}" Value="Removed">
+                                                                <Setter Property="Foreground" Value="{DynamicResource GitHubDeletedFileIconBrush}"/>
+                                                                <!--<Setter Property="Foreground" Value="Red"/>-->
```
StanleyGoldman: I'm having a bit of trouble with the binding here. I know I have the trigger right, as I'm able to see the effect of the value `Red` in the Visual Studio designer, but at runtime neither the DynamicResource binding nor the value `Red` seem to have an effect.

StanleyGoldman:
@donokuda I put the grey on the icon, but I don't know if it stands out well enough. Feel free to change the color as you see fit.

image

StanleyGoldman:
Yea, i see that point to the light grey icon doesn't really stand out with all that ~~strikeout~~ text right next to it.
I think it's good now though, the washed out icon has already grown on me. 😄

donokuda:
@StanleyGoldman Awesome! Going to merge this now. Thanks for this! 🙇

PR #651, opened 2016-10-26
Use BitmapImageMoniker instead of BitmapResourceID to reference logo

### LINKED ISSUE #642: Broken GitHub pane icon
- GitHub Extension version: 2.0.15.1
- Visual Studio version: VS2015 update 3

When you have the 'GitHub Pane' docked such that it shows as a tab 
image
but then you resize the sidebar until the tabs change to icons then it appears to have a broken icon image
image

and in blue
image
image
image

tocsoft:
If we use an `ImageMoniker` instead of setting the `BitmapResourceID`  on the `ToolWindowPane` then the icon displays correctly.

https://msdn.microsoft.com/en-us/library/mt628927.aspx was an interesting read on `ImageMoniker`'s which led me to this fix.

image

### Warning
I had to add a dependency on  `Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime` to get access to the `ImageMoniker` class not sure what effect that might have in terms of compatibility with other versions of VS.

fixes #642

### DIFF
--- a/src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
+++ b/src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
@@ -110,6 +110,9 @@
     </Reference>
     <Reference Include="Microsoft.VisualStudio.ComponentModelHost, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
     <Reference Include="Microsoft.VisualStudio.CoreUtility, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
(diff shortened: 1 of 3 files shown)


tocsoft:
**side note:** if you read though the article then, _from how i read it,_ it looks like we can get access to the images/icons that VS has already provided for free. For example we can use their refresh, back and forward icons (as I notice that the once on the GitHub pane don't quite match the ones on the Team Explorer one).

grokys:
Nice find! I need to read through that article again ;)

shana:
Yay finally they documented this stuff! 
Yup, let's switch the icons to what VS uses, it would be nice to be consistent with the icons in Solution Explorer (which are the ones most familiar to most users, I would say) 😄

shana:
Taking the dependency is fine, these interop wrapper assemblies are guaranteed to be available in future versions 👍

shana:
bb8_thumbs_up

PR #654, opened 2016-10-28
added .editorconfig to help prevent wrong indent types Tabs vs Spaces

tocsoft:
Must have editor config plugin installed to have any effect otherwise it has zero effect and doesn't harm anything.

Can be installed from https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+# top-most EditorConfig file
+root = true
+
+[*.cs]
+indent_style = space
+indent_size = 4


shana:
I'm merging this because I don't think any of our dependent submodules have different settings than what this file is set to, but we should probably do this separately for our submodule forks, too, at some point.

tocsoft:
this one only set the configs for *.cs might want to add setting for things like `xml`/`xaml` etc too as i think the repo is currently using 2 spaces for `xaml` (well at least in the one view I last looked at it seemed to be)

shana:
Oh, those things... yeah, they're all over the place, I use Xaml Styler to fix them up as I go but they're not consistent at all. They require a separate cleanup pass after all releases are done.

Issue #11, opened 2015-07-24
Opening github.com files in Visual Studio


RandomlyKnighted:
Per my conversation with @Haacked in #10, I would like to be able to browse code on GitHub from Visual Studio and then open that code file in Visual Studio. 

@Haacked also mentioned that when viewing a file on GitHub there is a button titled ```Open this file in GitHub for Windows```, which will open the code file in your default text editor for that file type. However, this only works if you have GitHub for Windows installed as it's not supported by default in Visual Studio. 

I'm interested in finding out what GitHub for Windows does that enables that to be be viewed in a text editor and then help add that to Visual Studio whether that by making it a part of this extension or by creating another one.

shiftkey:
> I'm interested in finding out what GitHub for Windows does that enables that to be be viewed in a text editor and then help add that to Visual Studio whether that by making it a part of this extension or by creating another one.

I can explain this bit - the URL you get when clicking "Open this file in GitHub for Windows" will contain the relative path to the file you clicked. So after cloning the repository GHfW just shells out to Windows to use whatever has been configured under the [File Associations](https://msdn.microsoft.com/en-us/library/cc144158(VS.85).aspx) on the local machine.

RandomlyKnighted:
@shiftkey So let's say we were able to browse GitHub files in Visual Studio. Would downloading the file to a temporary folder then opening it accomplish the same thing? Specifying of course that it be opened with Visual Studio (basically accomplishing the same task as right clicking on a file and saying Open with -> Visual Studio)

Haacked:
> I would like to be able to browse code on GitHub from Visual Studio and then open that code file in Visual Studio.

What would this look like?

> So let's say we were able to browse GitHub files in Visual Studio. Would downloading the file to a temporary folder then opening it accomplish the same thing?

Sure. If we implement something in Visual Studio, we can do whatever we want. We don't even need a temporary file since there are API calls we can make to get the contents of a file.

RandomlyKnighted:
Well what can we do with the VS API? I'd imagine it would connect to GitHub and show the files in a tree view similar to Solution Explorer.

Haacked:
@RandomlyKnighted That's a possibility. But we need to design the experience for how we get there? How would I launch the ability to browse or search repositories. What would that look like? It's worth thinking through the whole experience and even mock it up using Balsamiq, Powerpoint, or even line drawings on a napkin. :smile:

Naphier:
[...]

PR #659, opened 2016-10-30
add extension method and separate fork name

Drans78:
#11


shana:
@Drans78 Please don't create PRs from random branches you don't own.

PR #662, opened 2016-11-03
Use 'Sign in' rather than 'Login'

### LINKED ISSUE #657: Inconsistent use of language Login/Sign in
- GitHub Extension version: 2.0.15.1
- Visual Studio version: 2015 u3

There are inconsistencies between the usage of the terms **sign in/sign out** and **login** the code base seems to use both interchangeably.

image

image

image

should be normalised to one or the other.. Probably should be **Sign In** every where including the credentials window to be consistent with the site.

tocsoft:
Replaced all instances of 'login' and equivalent with 'sign in'

fixes #657 
image

more in keeping with github.com (as proof)

image

### DIFF
--- a/src/GitHub.App/Resources.Designer.cs
+++ b/src/GitHub.App/Resources.Designer.cs
@@ -151,7 +151,7 @@ namespace GitHub.App {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Make sure to use your password and not a Personal Access token to log in..
+        ///   Looks up a localized string similar to Make sure to use your password and not a Personal Access token to sign in..
         /// </summary>
         internal static string LoginFailedForbiddenMessage {
             get {
@@ -169,7 +169,7 @@ namespace GitHub.App {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Login failed..
+        ///   Looks up a localized string similar to Sign in failed..
         /// </summary>
         internal static string LoginFailedText {
             get {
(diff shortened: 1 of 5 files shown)


paladique:
My only concern is that when we start introducing localization, between the two, login vs sign in, which is an easier translation? This may differ from language to language but I only really know one 😆 but I definitely agree that consistency between the verbiage is the most important part out of all of this.

LGTM 👍 but will let the others take a 👀 too

tocsoft:
I think in terms of localisation it'd be a non-issue. You translate the meaning not the words, it just happens that in English there are 2 ways to represent signing in to a system, in other languages that might not be an issue.

paladique:
You are absolutely right, great point! 👍

shana:
Nice catch, thanks for this!

@donokuda Could you check what's the standard we use for this at GitHub in our native apps and online?

@saraford Could you check what terminology Microsoft uses in VS for these things?

The reason I ask is that I think the inconsistency comes from ours and MS standard being different, and if that's the case, I'd like to make sure we're going the right way on this :)

shana:
celebrate

tocsoft:
😁

saraford:
It seems like Sign In is the terminology VS is using these days. I found 4 prominent places where Sign In is used: 

1. the upper right-hand corner sign in
2. Product Registration
3. Connect to Team Explorer - Switch User
4. several in-browser experiences, e.g. signing into the forums. 

Sign In is also used as a screenshot example in the [Visual Studio UX Guidelines - Text - Style and Tone documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/dn742477(v=vs.85).aspx). You'll need to scroll down to see it, not text searchable.

I found a couple of counter examples from some of the much, much older UI, e.g. Log on to database and Connect as (for a server).   

I :+1: on Sign In :) 

FYI @donokuda

PR #652, opened 2016-10-28
Fixed branch name in PR created succesfully message

jswietek:
Changed target name of successful PR creation message from <owner>/<target_branch_name> (e.g. jswietek/master) to <owner>/<repo_name> (e.g. jswietek/VisualStudio)

PR #665, opened 2016-11-07
Add fork separator in PR Creation Form

### LINKED ISSUE #605: Create PR view uses / as separator for forks
The create PR view separates the fork name and branch name with a `/` which is confusing as branches often contain the `/` character. For example if I fork `github/VisualStudio` to `grokys` and now i have things like `grokys/grokys/maintainer-workflow`. Use a `:` as a fork:branch separator as on GitHub itself.

### LINKED ISSUE #640: Create PR success message isn't correct.
- GitHub Extension version: 2.0.15.1 (i.e. this is present in current stable release)

I create a PR to the github/VisualStudio repository, and once created the success message displayed was this:

image

I think that message is wrong: "github/master" doesn't make sense.

paladique:
- Replaced `\` with `:`  in forked branches in the pull request creation form dropdown using `IMultiConverter` to get full information on the branch owner. This also includes updates to `LinkDropDown.cs`
- Update the wording/style of the PR creation success message to include the source branch

Fixes #605 
Fixes #640


src/GitHub.UI/Controls/ComboBoxes/LinkDropDown.cs
```diff
@@ -87,9 +87,18 @@ private void UpdateLinkText()
                 // will work as long as DisplayMemberPath is just a property name, which is
                 // all we need right now.
                 if (string.IsNullOrWhiteSpace(DisplayMemberPath))
-                {
-                    LinkText = item.ToString();
+                {                   
+                    if (ItemTemplate == null)
```
paladique: If I'm reading this correctly, then yes, we can't have `DisplayMemberPath` and `ItemTemplate` set at the same time. So if  for some reason either of those are not set then we can just display the item.

src/GitHub.UI/Converters/BranchNameConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace GitHub.UI.Converters
+{
+    public class BranchNameConverter : IMultiValueConverter
+    {
+
+        public object Convert(object[] values, Type targetType, object parameter, CultureInfo culture)
+        {
+            var branch = values[0] as IBranch;
+            var activeRepo = values[1] as IRepositoryModel;
```
shana: Before accessing values here directly, you should check to make sure the array actually contains at least two elements and return if you don't

tocsoft: might not something like this be better?
```c#
var branch = values.OfType<IBranch>().FirstOrDefault();
var activeRepo = values.OfType<IRepositoryModel>().FirstOrDefault();
```
will automatically work for when you fail to provide all values and will even work for when you provide the parts in a different order.

paladique: Either way, should I just return `null` if only one or neither of these elements exist?

tocsoft: I would personally return  `values.FirstOrDefault()?.ToString()` then it will at least return something that should be vaguely correct.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -108,10 +108,9 @@ public class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationV
                         notifications.ShowError(error?.Message ?? ex.Message);
                         return Observable.Empty<IPullRequestModel>();
                     }))
-
             .OnExecuteCompleted(pr =>
             {
-                notifications.ShowMessage(String.Format(CultureInfo.CurrentCulture, Resources.PRCreatedUpstream, TargetBranch.Id,
+                notifications.ShowMessage(String.Format(CultureInfo.CurrentCulture, Resources.PRCreatedUpstream, TargetBranch.Repository.Owner + "/" + TargetBranch.Repository.Name,
```
paladique: @shana you mentioned I merge in #652 into here, but remind me again what I was supposed to do here? 😅 The message currently displays the owner and the repo, but not the branch.

shana: Oh, that was because the `Id` that was being displayed before meant that the message would show the branch name only, and now it's been change to show the repo owner and name. Since you're working on fixing the names of branches to match the nomenclature we use on .com, I thought it would be useful to also aggregate this PR and make sure what what we show in this message also makes sense - we might also want to show the PR target branch name here as well.

paladique:
### What this currently looks like:
capture

### Just the source branch
capture2

### Source and target (but will be a `:` instead of a `\` to specify owner)

capture3


### Source and target with `:` separator
capture5


cc @saraford @donokuda per [this conversation](https://github.com/github/VisualStudio/pull/665#pullrequestreview-7923294) we're working with the verbiage here. Should we include target and source branch info here too? Or should we just leave it alone?

donokuda:
> Should we include target and source branch info here too? Or should we just leave it alone?

Maybe something like "Pull request for branch `foo` successfully created at `paladique/Talks`"

I'm guess that the link goes to the pull request itself? Right now it reads to me as if it would go to the repo page.

@saraford Does Team Explorer have verbiage that we could pull some inspiration from?

paladique:
>I'm guessing that the link goes to the pull request itself?

Yup!

@donokuda per our chat convo and suggestions here, this is what that would look like

capture

grokys:
"Created successfully created" :)

tocsoft:
I feel the branch name gets lost in there some what. Is it worth bolding it? (Just to pull it out from the rest of the sentence)

paladique:
@tocsoft I was thinking the same thing too! Glad someone else noticed it 👍

paladique:
Here's what it looks like with the bold branch text.

capture

paladique:
👋 @shana @grokys this is ready for another pass when you have a minute

donokuda:
@paladique One more quick design feedback: Can we either bring back the info icon? 😄 

image

paladique:
@donokuda oh right! It's been disappearing between screenshots! Odd. Anyone have any ideas why?

paladique:
Looks like that icon issue this is a bug, details here: https://github.com/github/VisualStudio/issues/692

grokys:
🎇 Merged!

PR #668, opened 2016-11-08
Lazy load the services needed by UsageTracker.

grokys:
To improve the startup time of the extension.

### DIFF
--- a/src/GitHub.VisualStudio/GitHubPackage.cs
+++ b/src/GitHub.VisualStudio/GitHubPackage.cs
@@ -52,8 +52,9 @@ namespace GitHub.VisualStudio
             await base.InitializeAsync(cancellationToken, progress);
             await EnsurePackageLoaded(new Guid(ServiceProviderPackage.ServiceProviderPackageId));
 
-            //var usageTracker = await GetServiceAsync(typeof(IUsageTracker)) as IUsageTracker;
-            //usageTracker.IncrementLaunchCount();
+            // Activate the usage tracker by forcing an instance to be created.
+            await GetServiceAsync(typeof(IUsageTracker));
+
             InitializeMenus().Forget();
         }
 
@@ -177,6 +178,11 @@ namespace GitHub.VisualStudio
             {
                 var sp = await GetServiceAsync(typeof(IUIProvider)) as IUIProvider;
                 return new MenuProvider(sp);
+            }
+            else if (serviceType == typeof(IUsageTracker))
+            {
+                var uiProvider = await GetServiceAsync(typeof(IUIProvider)) as IUIProvider;
+                return new UsageTracker(uiProvider);
             }
             // go the mef route
             else
--- a/src/GitHub.VisualStudio/GitHubPackage.cs
+++ b/src/GitHub.VisualStudio/GitHubPackage.cs
@@ -53,7 +53,7 @@ namespace GitHub.VisualStudio
             await EnsurePackageLoaded(new Guid(ServiceProviderPackage.ServiceProviderPackageId));
 
             // Activate the usage tracker by forcing an instance to be created.
-            await GetServiceAsync(typeof(IUsageTracker));
+            GetServiceAsync(typeof(IUsageTracker)).Forget();
 
             InitializeMenus().Forget();
         }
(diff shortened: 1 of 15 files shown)


src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -52,8 +52,9 @@ public GitHubPackage(IServiceProvider serviceProvider)
             await base.InitializeAsync(cancellationToken, progress);
             await EnsurePackageLoaded(new Guid(ServiceProviderPackage.ServiceProviderPackageId));
 
-            //var usageTracker = await GetServiceAsync(typeof(IUsageTracker)) as IUsageTracker;
-            //usageTracker.IncrementLaunchCount();
+            // Activate the usage tracker by forcing an instance to be created.
+            await GetServiceAsync(typeof(IUsageTracker));
```
shana: We probably don't have to wait for it to be done, could we just fire and forget like we do with the menus?

src/GitHub.VisualStudio/Services/UsageTracker.cs
```diff
@@ -194,21 +204,23 @@ void TimerTick(object sender, EventArgs e)
...
-            // Subsequent timer ticks should occur every 8 hours.
-            timer.Interval = TimeSpan.FromHours(8);
+            if (firstRun)
+            {
+                IncrementLaunchCount();
+                timer.Interval = TimeSpan.FromHours(8);
+                firstRun = false;
+            }
 
             if (userSettings == null)
```
shana: If you're calling `Initialize` above, then this condition will always be false, right? Do we need any of the code inside this `if` block?

src/GitHub.VisualStudio/GitHubPackage.cs
grokys: Yep, we can.

src/GitHub.VisualStudio/Services/UsageTracker.cs
grokys: The `userSettings` check is unnecessary, you're right. However we still need to stop the timer if `client == null`. I think you added the `SwitchToMainThreadAsync` there - is that still necessary? Not sure why it's needed.

shana: Because this might get called from a background thread (if it was triggered from a background thread) and we were doing `GetExportedValue`, which needs to be called on the main thread. Speaking of which, we need to be very careful about what we call in `Initialize`, there's side effects that can deadlock, I'll comment there.

src/GitHub.VisualStudio/Services/UsageTracker.cs
```diff
@@ -145,8 +135,29 @@ public void IncrementLoginCount()
...
+        void Initialize()
+        {
+            // The services needed by the usage tracker are loaded when they are first needed to
+            // improve the startup time of the extension.
+            if (userSettings == null)
+            {
+                client = uiProvider.GetService<IMetricsService>();
+                connectionManager = uiProvider.GetService<IConnectionManager>();
+                userSettings = uiProvider.GetService<IPackageSettings>();
+                vsservices = uiProvider.GetService<IVSServices>();
```
shana: Any or all of these things can trigger a MEF call via `GetExportedValue`, so we need to make sure this is done from the main thread, not a background thread. We'll need `await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();` before these calls, `GetService` is not thread-safe.

src/GitHub.VisualStudio/Services/UsageTracker.cs
grokys: Well `TimerTick` will only ever get called on the main thread as it's triggered from a `DispatcherTimer`, so it was unnecessary before. But you're right that it's necessary now as `Initialize` may be called on a b/g thread from one of the `IncrementX` methods.

src/GitHub.VisualStudio/Services/UsageTracker.cs
shana: Right, the timer is a `DispatcherTimer` so the tick method will always call this from the main thread so that's fine. I don't think anything else will be calling this from background threads, but in case that happens, probably best to make sure we're on the right thread here.

shana:
✨

PR #669, opened 2016-11-08
WIP: Release 2.1.0.x

shana:
Ongoing work on VS "15" support.


src/GitHub.App/ViewModels/StartPageCloneViewModel.cs
```diff
@@ -0,0 +1,208 @@
...
+using System.Collections.ObjectModel;
+using GitHub.Collections;
+using GitHub.UI;
+using GitHub.Extensions.Reactive;
+
+namespace GitHub.ViewModels
+{
+    [ExportViewModel(ViewType=UIViewType.StartPageClone)]
+    [PartCreationPolicy(CreationPolicy.NonShared)]
+    public class StartPageCloneViewModel : BaseViewModel, IBaseCloneViewModel
```
tocsoft: Would a better name for this stuff be `ReloneViewModel` its not really start page specific it just happens to be currently only be used by the start page stuff but its not really specific to that, I can see times you might want to duplicate a repository for other reasons that could also use this functionality.

src/GitHub.VisualStudio/UI/Views/Controls/StartPageCloneView.xaml.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    public class GenericStartPageCloneView : SimpleViewUserControl<IBaseCloneViewModel, StartPageCloneView>
+    {}
+
+    /// <summary>
+    /// Interaction logic for CloneRepoControl.xaml
+    /// </summary>
+    [ExportView(ViewType=UIViewType.StartPageClone)]
+    [PartCreationPolicy(CreationPolicy.NonShared)]
+    public partial class StartPageCloneView : GenericStartPageCloneView
```
tocsoft: see `StartPageCloneViewModel` comment `RecloneView` might be better

src/GitHub.App/ViewModels/StartPageCloneViewModel.cs
shana: I don't plan on keeping this specific viewmodel around for long but right now there's Reasons(tm) for it.

tocsoft: ah those "reasons"... no point changing stuff around if its not going to live long. 😁

PR #670, opened 2016-11-08
Update the start page clone dialog.

grokys:
- Show just the repository to clone (with a logo) - this is ugly and needs a @donokuda pass.
- Re-enable detection of existing repository in said clone dialog.


shana:
Looks good! Can we get a screenshot of how it looks currently?

@donokuda this branch won't build on 2015 but the xaml designer should be able to load the view. If not, just a mockup will do, we can fix it up ourselves. This is basically the clone dialog, only without the repository list and filter textbox, and instead just showing the name of the repository we're going to clone. Since the size of the window is pretty much fixed (because it has to accommodate the login dialog before), we have a lot of empty space that we need filling.

We should probably have:
- The avatar that owns the account
- The name of the account
- The name of the repository
- The url to it (just to be clear and you know gotta fill that space)
- repo/lock/fork icons maybe?

Since this includes a fix to actually make it work properly, we should probably merge this and do the polish on another PR targeting release/2.1.0. @grokys thoughts?

grokys:
This is how it looks currently in all its monochrome glory:

image

donokuda:
@shana @grokys Can you walk me through how a user gets to this screen?
Does this replace the existing clone from GitHub flow or is this something like a confirmation screen?

shana:
@donokuda In VS15 Preview 5 there's a new API in the Start Page for cloning repositories that you have already cloned in some other machine - it synchronizes the roaming profile information, which includes information on what you've cloned before (only what you've cloned from the Start Page, right now). So this is the same clone dialog that we have, only we already know exactly what repository we want to clone, and we just need the user to set the path to clone to.

Users will get to this dialog by clicking on the url of a repository in the Start Page, which pops up our dialog, optionally goes through login if needed, then shows this UI with the information about the repository they selected and where they want to clone it to.

PR #677, opened 2016-11-09
Handled NotFoundException, when opening repo that was deleted on github

jswietek:
Added catch method for PRs observable. This handles NotFoundExceptions when user tries to open and retrieve Pull Requests for repo, that was deleted through GitHub website.


src/GitHub.App/ViewModels/PullRequestListViewModel.cs
```diff
@@ -113,6 +113,11 @@ public override void Initialize([AllowNull] ViewWithData data)
                     // TODO: Do some decent logging here
                     return repositoryHost.LogOut();
                 })
+                .Catch<System.Reactive.Unit, Octokit.NotFoundException>(ex =>
+                {
+                    //this is caused when repository was deleted on github
+                    return repositoryHost.LogOut();
```
tocsoft: You probably shouldn't log the user out when a repo is not found

this will want to return `Observable.Empty<System.Reactive.Unit>()` instead this will stop it crashing but leave the user signed in.

grokys:
Thanks @jswietek ✨ - apologies for the delay - we're working through the backlog!

PR #678, opened 2016-11-09
Switch to main thread on repository creation.

grokys:
Before checking out created repository, this is needed because `vsGitServices.Clone` ends up calling into `AsyncPackage.GetService` which cannot be called from a background thread.


shana:
:+1:

PR #679, opened 2016-11-09
Don't use RxUI for Repositories binding.

grokys:
As it was sometimes setting the selection on the `ListBox` without user interaction. Instead, use XAML bindings. Unfortunately this required binding to a dependency property on the view itself. Better solutions
moving forward would be:

- Declare the collection view in XAML. This would require the Group classes to be made public, but they have a dependency on state in the view.
- Declate the collection view in the view model.

I didn't implement either of these solutions as I felt the changes would be too impactful on the codebase at this point in time.


grokys:
Fixes #681.

PR #690, opened 2016-11-16
Update the start page clone dialog UI

### LINKED ISSUE #684: Update design for start page clone dialog
Follow up to https://github.com/github/VisualStudio/pull/670

> @donokuda this branch won't build on 2015 but the xaml designer should be able to load the view. If not, just a mockup will do, we can fix it up ourselves. This is basically the clone dialog, only without the repository list and filter textbox, and instead just showing the name of the repository we're going to clone. Since the size of the window is pretty much fixed (because it has to accommodate the login dialog before), we have a lot of empty space that we need filling.
>
> We should probably have:
> - The avatar that owns the account
> - The name of the account
> - The name of the repository
> - The url to it (just to be clear and you know gotta fill that space)
> - repo/lock/fork icons maybe? 

Here's what I'm thinking the updated design could look like

grokys:
As described in #684:

image

Note that this doesn't currently display the Organization icon: this is because a `SimpleRepositoryModel` is being passed to the clone dialog and this class doesn't expose an `IAccount` for the organization. The repository models have been refactored on `master` and should allow us to display this information should we want it for the next minor version (i.e. alongside the maintainer workflow).

Further discussion here: https://github.com/github/VisualStudio/issues/684

# Further notes:

This PR has turned into a bit of a "make stuff work in VS2017" PR so not everything is related to the start page clone UI. Of particular note is the fact that I've moved the start page assembly from a submodule into the main repository.

Fixes #684.


grokys:
@donokuda you may want to tweak the spacing etc on this, I don't know if you can do it on VS2015, but if not let me know and I can make the changes. If you can tweak it yourself please feel free to commit directly to the branch!

grokys:
Waiting for a reply from MS regarding the styling: https://halp.githubapp.com/discussions/431bf3feaba211e681409a36d271ae86

shana:
Feels like the GitHub mark is way too big on this dialog, it's overwhelming it somewhat. Maybe we should be aligning this more closely with the create repository dialog, I left a comment about that in the issue.

shana:
@grokys Could you upload a screenshot of how this looks like now?

grokys:
Unfortunately not as I'm unable to get the start page functionality working on VS2017 currently. I will attach a screenshot when I get that working again.

src/GitHub.App/SampleData/RemoteRepositoryModelDesigner.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1065:DoNotRaiseExceptionsInUnexpectedLocations")]
+        public bool Equals(IRemoteRepositoryModel other)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void SetIcon(bool isPrivate, bool isFork)
+        {
+            throw new NotImplementedException();
```
shana: No need to throw exceptions on this class, this is only for the xaml designer, so it will never run. And if it does, better not to crash the designer, it's awfully finicky as it is.

grokys: Ah yeah good point.

src/GitHub.StartPage/StartPagePackage.cs
```diff
@@ -0,0 +1,176 @@
...
+{
+    [PackageRegistration(UseManagedResourcesOnly = true)]
+    [Guid(guidPackage)]
+    [ProvideCodeContainerProvider("GitHub Container", guidPackage, Guids.ImagesId, 1, "#110", "#111", typeof(GitHubContainerProvider))]
+    public sealed class StartPagePackage : ExtensionPointPackage
+    {
+        static IServiceProvider serviceProvider;
+        internal static IServiceProvider ServiceProvider { get { return serviceProvider; } }
+
+        public const string guidPackage = "3b764d23-faf7-486f-94c7-b3accc44a70e";
```
shana: Since the project is now in the repo, this can go into the Guids class with all the other Guids, and get a proper name like "StartPagePackageId"

src/GitHub.StartPage/StartPagePackage.cs
```diff
@@ -0,0 +1,176 @@
...
+        public StartPagePackage()
+        {
+            serviceProvider = this;
+        }
+    }
+
+    [Guid(CodeContainerProviderId)]
+    public class GitHubContainerProvider : ICodeContainerProvider
+    {
+        public const string CodeContainerProviderId = "6CE146CB-EF57-4F2C-A93F-5BA685317660";
```
shana: Let's move this one to the Guids class too

src/GitHub.StartPage/StartPagePackage.cs
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        CloneRequest ShowCloneDialog(IGitHubServiceProvider gitHubServiceProvider, IGitRepositoriesExt gitRepositories, IRemoteRepositoryModel repository = null)
+        {
+            string basePath = null;
+
+            gitHubServiceProvider.AddService(this, gitRepositories);
+            var uiProvider = gitHubServiceProvider.GetService<IUIProvider>();
+            var controller = uiProvider.Configure(repository == null ? UIControllerFlow.Clone : UIControllerFlow.StartPageClone,
+                null //TODO: set the connection corresponding to the repository if the repository is not null
```
shana: Do we want to fix this one now? It's an edge case in case of the user being logged in to two accounts.

src/GitHub.StartPage/StartPagePackage.cs
grokys: How about `guidPackage` in `StartPagePackage`?

src/GitHub.StartPage/StartPagePackage.cs
grokys: To be honest, I'm a little wary of messing with things now... We can totally do it though if you think it's something that might bite us.

shana: Let's open an issue to track it, it's a nice low hanging bug fix.

PR #693, opened 2016-11-18
Set default info panel icon color and type 

### LINKED ISSUE #692: Information panel information icon missing in some cases
When working on #665, in some of the [screenshots](https://github.com/github/VisualStudio/pull/665#issuecomment-261030796), the information icon in the information panel is missing. 

From that example, the screenshots where the icon is missing are when I successfully created a pull request without any errors. When accidentally creating an already existing pull request for a branch, going to .com to delete it, and making the pull request again within the extension, the information icon is visible. It is also visible after I make more pull requests. 

Closing the info panel and creating a successful pull request right after will also not display the icon. Also, if I make two (or more) pull requests in a row successfully, the icon will still not show up. 

To sum this up, the icon is missing when:
- You close the info panel (if it's currently open/visible)
- The first message is information (successful creation, not an error)

It's not missing when:
- The first message was a warning or error

paladique:
Since `Information` is the default type, set it as the information icon and style in the `InfoPanel`'s constructor.

Fixes #692

### DIFF
--- a/src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
+++ b/src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
@@ -34,7 +34,7 @@ namespace GitHub.VisualStudio.UI.Controls
 
         public MessageType MessageType
         {
-            get { return (MessageType)GetValue(MessageProperty); }
+            get { return (MessageType)GetValue(MessageTypeProperty); }
             set { SetValue(MessageTypeProperty, value); }
         }
 
@@ -80,7 +80,9 @@ namespace GitHub.VisualStudio.UI.Controls
         {
             InitializeComponent();
 
-            this.DataContext = this;
+            DataContext = this;
+            Icon = Octicon.info;
+            IconColor = InfoColorBrush;
 
             LinkCommand = new RelayCommand(x =>
             {


src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
```diff
@@ -34,7 +34,7 @@ public string Message
 
         public MessageType MessageType
         {
-            get { return (MessageType)GetValue(MessageProperty); }
+            get { return (MessageType)GetValue(MessageTypeProperty); }
```
paladique: This was previously set to `MessageProperty`, which didn't seem right, given the property type. Nothing looked like it was breaking because of it, and it doesn't look like anything broke after changing it.

src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
```diff
@@ -80,7 +80,9 @@ public InfoPanel()
         {
             InitializeComponent();
 
-            this.DataContext = this;
+            DataContext = this;
```
paladique: Also removed the `this` keyword here, since we don't need it here (also VS suggested we do it).

src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
shana: Yup that was definitely a typo :P

shana:
thumbs-up-scroll

PR #699, opened 2016-11-23
On Hold: Allow building .vsix packages for VS2017 on VS2015.

grokys:
**On hold: we're going to wait a while to see if the VS team can point us in the right direction to get this done without the hacks in this PR.**

Until VS2017 is out we need the option of building .vsix packages for VS2017 on VS2015. To do this supply the `/p:VsixVersion=15` option to msbuild. This will use a slightly the hacked `lib\Microsoft.VsSDK.15.targets` instead of the standard one.


shana:
Looks a-ok to me! :shipit:

grokys:
Closing this as it seems official support from MS has been released.

PR #701, opened 2016-12-01
Allow wrapping in GitHubActionLink.

grokys:
`GitHubActionLink` already had a `TextTrimming` property - this PR add the associated `TextWrapping` property so that text in the can be word-wrap.


shana:
:sparkles:

PR #711, opened 2016-12-08
Update PR details view design

### LINKED ISSUE #641: Inconsistent UI for open PR branch
- GitHub Extension version:  v2.0.15.3-alpha1

The _'local branch is missing'_ UI is inconsistent between when a pull request is from a fork or from the same origin.

**Local/remote branch**
image

**Fork**
image

I would expect the UI to look them same, not having the same actual contents each would be worded appropriately to the case, but either both to be buttons or both be warning messages with a link to download the pull request. 

My preference would be the warning text with a link.

grokys:
This PR updates the PR details view:

- Tries to decrease wasted whitespace (this still needs some work)
- Use the same link appearance for checking out PR branches from local branches and forks (fixes #641)
- Use push/pull buttons for syncing with tracked branch
- When checking out a PR from a fork, adds a remote to push/pull directly from the fork branch

Here is the current state of the UI:

ui-update

TODO:

- [ ] Final polish of UI
- [ ] Clean up unused remotes
- [ ] Move UI messages to resources

cc: @donokuda


src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
```diff
@@ -426,38 +413,60 @@ string GetBranchDisplayName(string targetBranchLabel)
...
+            return Observable.Defer(async () =>
             {
-                case CheckoutMode.NeedsPull:
-                    operation = pullRequestsService.Pull(repository);
-                    break;
-                case CheckoutMode.Fetch:
-                    operation = pullRequestsService
+                var localBranches = await pullRequestsService.GetLocalBranches(repository, Model).ToList();
+
+                if (localBranches.Count > 0)
```
grokys: Actually scratch that, it's an observable so returning many is no different. I can use `FirstOrDefaultAsync` however.

grokys: Actually, thinking further, I think it's best to leave this as-is in case _somehow_ we end up with >1 local branch created. Using `FirstOrDefaultAsync` would mean that the actual branch could never be found, meaning that another branch will be checked out each time the link is clicked, exacerbating the problem.

shana: :+1:

shana:
Merging this, we'll handle the outstanding TODOs in other PRs.

PR #720, opened 2016-12-09
Add renamed file status

### LINKED ISSUE #719: Add renamed file status
Files that have been moved will have a renamed status. `PullRequestFileStatus` only contains the added, removed, and modified status. When the [ModelService tries to parse the renamed status](https://github.com/github/VisualStudio/blob/master/src/GitHub.App/Services/ModelService.cs#L559), an invalid argument exception is thrown.

grokys:
Handle renamed status in PR details view, and display the old filename/path following Team Explorer's behavior as described in #719.

Fixes #719.

### DIFF
--- a/src/GitHub.App/Services/PullRequestService.cs
+++ b/src/GitHub.App/Services/PullRequestService.cs
@@ -172,6 +172,17 @@ namespace GitHub.Services
             });
         }
 
+        public IObservable<TreeChanges> GetTreeChanges(ILocalRepositoryModel repository, IPullRequestModel pullRequest)
+        {
+            return Observable.Defer(async () =>
+            {
+                var repo = gitService.GetRepository(repository.LocalPath);
+                await gitClient.Fetch(repo, "origin");
+                var changes = await gitClient.Compare(repo, pullRequest.Base.Sha, pullRequest.Head.Sha, detectRenames: true);
+                return Observable.Return(changes);
+            });
+        }
+
         public IObservable<IBranch> GetLocalBranches(ILocalRepositoryModel repository, IPullRequestModel pullRequest)
         {
             return Observable.Defer(() =>
--- a/src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
@@ -272,6 +272,7 @@ namespace GitHub.ViewModels
             IsBusy = true;
 
             modelService.GetPullRequest(repository, prNumber)
+                .TakeLast(1)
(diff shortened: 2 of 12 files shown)


src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -172,6 +172,17 @@ public IObservable<BranchTrackingDetails> CalculateHistoryDivergence(ILocalRepos
             });
         }
 
+        public IObservable<TreeChanges> GetTreeChanges(ILocalRepositoryModel repository, IPullRequestModel pullRequest)
+        {
+            return Observable.Defer(async () =>
+            {
+                var repo = gitService.GetRepository(repository.LocalPath);
+                await gitClient.Fetch(repo, "origin");
+                var changes = await gitClient.Compare(repo, pullRequest.Base.Sha, pullRequest.Head.Sha, detectRenames: true);
```
grokys: After thinking about this for the weekend, I think what I'm doing here is the best thing to do as firstly it reflects the underlying API. Secondly (and less importantly) should we need any other options adding later (there are a number of potential options), by using named parameters the call sites won't need to change.

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
```diff
@@ -411,6 +415,30 @@ string GetBranchDisplayName(string targetBranchLabel)
...
+                    {
+                        return Path.GetDirectoryName(change.OldPath) == Path.GetDirectoryName(change.Path) ?
+                            Path.GetFileName(change.OldPath) : change.OldPath;
+                    }
+                    else
+                    {
+                        return "rename";
+                    }
+                default:
+                    return null;
```
shana: How about returning a `String.Empty` here so we don't have to worry about dealing with nulls in the caller code?

grokys: The null is used in a trigger to hide the `TextBlock` that shows the status display: https://github.com/github/VisualStudio/blob/34a4071fe432e0daf0673c0772ca9b0137be9a5b/src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml#L360

I've run into problems before trying to make triggers work with empty strings, and indeed the suggested solution found here (and elsewhere) isn't pretty: http://stackoverflow.com/a/2582441/6448 . I can't remember the exact problem I've had in the past though - I can give it a try if you like?

grokys: Also note that this property is likely only to be used for binding, so nulls shouldn't be a problem (famous last words I know...)

shana:
Linkback to #733 which depends on this.

Issue #715, opened 2016-12-08
Maintainer Workflow: Move messages to resource dictionary


grokys:
There are a number of messages in the PR details view that are using hardcoded strings. These need to be moved to resources. Might be worth waiting until #713 is finished before doing this.

Issue #713, opened 2016-12-08
Maintainer Workflow: Final pass of PR details UI


grokys:
We need a final pass of the PR details UI before the maintainer workflow feature is ready. The current state of the UI can be seen here: https://github.com/github/VisualStudio/pull/711

Pinging @donokuda to work his magic here.

donokuda:
This should be finished with https://github.com/github/VisualStudio/pull/730 but feel free to reopen if y'all think otherwise.

PR #724, opened 2016-12-09
Handle PRs from deleted repositories.

### LINKED ISSUE #721: Maintainer Workflow: assert failure on loading PR from ghost account
Attempting to load a PR such as https://github.com/aspnet/mvc/pull/2057 which is created by a ghost account causes an assertion failure. We also need to disable attempting to check out such a PR.

grokys:
If a PR was submitted from a fork and that repository subsequently deleted, we previously asserted in the PR list. Handle this case and disable the checkout button in the PR details view.

This PR builds on #720 so that should be merged first.

Fixes #721

### DIFF
--- a/src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
@@ -325,9 +325,17 @@ namespace GitHub.ViewModels
                 var caption = localBranches.Count > 0 ?
                     "Checkout " + localBranches.First().DisplayName :
                     "Checkout to " + (await pullRequestsService.GetDefaultLocalBranchName(repository, Model.Number, Model.Title));
-                var disabled = await pullRequestsService.IsWorkingDirectoryClean(repository) ?
-                    null :
-                    "Cannot checkout as your working directory has uncommitted changes.";
+                var clean = await pullRequestsService.IsWorkingDirectoryClean(repository);
+                string disabled = null;
+
+                if (!pullRequest.Head.RepositoryCloneUrl.IsValidUri)
+                {
+                    disabled = "The source repository is no longer available.";
+                }
+                else if (!clean)
+                {
+                    disabled = "Cannot checkout as your working directory has uncommitted changes.";
+                }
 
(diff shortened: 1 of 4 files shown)


paladique:
Gave this a try and worked fine for me! Got the message "The source repository is no longer available." when I moused over the checkout link. I was questioning whether or not this would be obvious to the user, but I imagine they'll navigate to the conversation in the browser and see what happened.

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
```diff
@@ -325,9 +325,17 @@ public override void Initialize([AllowNull] ViewWithData data)
...
                     "Checkout to " + (await pullRequestsService.GetDefaultLocalBranchName(repository, Model.Number, Model.Title));
-                var disabled = await pullRequestsService.IsWorkingDirectoryClean(repository) ?
-                    null :
-                    "Cannot checkout as your working directory has uncommitted changes.";
+                var clean = await pullRequestsService.IsWorkingDirectoryClean(repository);
+                string disabled = null;
+
+                if (pullRequest.Head == null || !pullRequest.Head.RepositoryCloneUrl.IsValidUri)
+                {
+                    disabled = "The source repository is no longer available.";
```
shana: These should go into the resources file. Are we doing that on a later pass for all strings, or doing it as we go along? I remember we talked about this but not what we decided.

paladique: Yup! There's an issue to do that after #713 is taken care of: https://github.com/github/VisualStudio/issues/715

shana: Aaaah indeed indeed! :+1:

shana:
> Gave this a try and worked fine for me! Got the message "The source repository is no longer available." when I moused over the checkout link. I was questioning whether or not this would be obvious to the user, but I imagine they'll navigate to the conversation in the browser and see what happened.

That's a good point! Let's keep that in mind while we test and see if it's enough or if we need to add more obvious messaging.

shana:
Linkback to #733 which depends on this.

PR #728, opened 2016-12-09
Pane Navigation & UIController

### LINKED ISSUE #628: GitHub pane navigation broken
Now that the maintainer workflow branch has been merged to master, it's clear that navigation in the GitHub pane is broken: switching branches switches the pane back to the PR list and subsequent navigation with the back/forward/PR buttons is broken.

StanleyGoldman:
Attempting to resolve #628

In `UIController` When `Jump()` uses `Fire()` to change states, `Fire()` does not correctly determine if the requested state in `requestedTarget` requires an argument.

### DIFF
--- a/src/GitHub.App/Controllers/UIController.cs
+++ b/src/GitHub.App/Controllers/UIController.cs
@@ -91,6 +91,8 @@ namespace GitHub.Controllers
     [PartCreationPolicy(CreationPolicy.NonShared)]
     public class UIController : IUIController, IDisposable
     {
+        static readonly NLog.Logger log = NLog.LogManager.GetCurrentClassLogger();
+
         internal enum Trigger
         {
             None,
@@ -168,10 +170,21 @@ namespace GitHub.Controllers
             ConfigureLogicStates();
 
             uiStateMachine = new StateMachineType(UIViewType.None);
+            uiStateMachine.OnUnhandledTrigger((viewType, trigger) =>
+            {
+                log.Warn("UIMachine Unhandled {0} from {1}",  trigger,  viewType);
+            });
+            uiStateMachine.OnTransitioned(transition1 =>
+            {
+                var source = transition1.Source;
+                var destination = transition1.Destination;
+                var trigger = transition1.Trigger;
+
+                log.Info(CultureInfo.InvariantCulture, "UIMachine Transitioned from {0} by {1} to {2}", source, trigger, destination);
+            });
+
             triggers = new Dictionary<Trigger, StateMachineType.TriggerWithParameters<ViewWithData>>();
-
             ConfigureUIHandlingStates();
-
         }
 
         public IObservable<LoadData> SelectFlow(UIControllerFlow choice)
@@ -210,7 +223,7 @@ namespace GitHub.Controllers
                     .ObserveOn(RxApp.MainThreadScheduler)
                     .Subscribe(_ => { }, () =>
                     {
-                        Debug.WriteLine("Start ({0})", GetHashCode());
+                        log.Info("Start");
                         Fire(Trigger.Next);
                     });
             }
@@ -250,7 +263,7 @@ namespace GitHub.Controllers
                     .ObserveOn(RxApp.MainThreadScheduler)
                     .Subscribe(_ => { }, () =>
                     {
-                        Debug.WriteLine("Start ({0})", GetHashCode());
+                        log.Info("Start");
                         Fire(Trigger.Next);
                     });
             }
@@ -260,9 +273,9 @@ namespace GitHub.Controllers
         {
             if (where.ActiveFlow != mainFlow)
             {
-                var message = string.Format(CultureInfo.CurrentCulture,
-                    "Jump called for flow {0} but this is {1}", 
-                    where.ActiveFlow, 
+                var message = string.Format(CultureInfo.InvariantCulture,
+                    "Jump called for flow {0} but this is {1}",
+                    where.ActiveFlow,
                     mainFlow);
 
                 throw new ArgumentException(message);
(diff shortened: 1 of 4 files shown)


src/GitHub.VisualStudio/UI/Views/PullRequestListView.xaml.cs
```diff
@@ -28,7 +28,7 @@ public PullRequestListView()
 
             OpenPR = new RelayCommand(x =>
             {
-                NotifyOpen(new ViewWithData { Data = x });
+                NotifyOpen(new ViewWithData(UIControllerFlow.PullRequests) { ViewType = UIViewType.PRDetail, Data = x});
```
StanleyGoldman: Based on the unit tests I believe this is required.

src/GitHub.App/Controllers/UIController.cs
```diff
@@ -264,7 +264,8 @@ public void Jump(ViewWithData where)
 
             requestedTarget = where;
             if (activeFlow == where.ActiveFlow)
-                Fire(Trigger.Next, where);
+                //TODO: Fix this
+                Fire(Trigger.Next, where, true);
```
StanleyGoldman: Temporary fix to identify a `Fire()` call from `Jump()`

src/GitHub.App/Controllers/UIController.cs
```diff
@@ -799,13 +800,28 @@ bool CreateViewAndViewModel(UIViewType viewType, ViewWithData data = null)
...
         {
             Debug.WriteLine("Firing {0} from {1} ({2})", next, uiStateMachine.State, GetHashCode());
-            if (arg != null && triggers.ContainsKey(next))
-                uiStateMachine.Fire(triggers[next], arg);
+            //TODO: Fix this
+            if (fromJump)
+            {
+                if (requestedTarget.ViewType == UIViewType.PRDetail)
+                {
+                    uiStateMachine.Fire(triggers[triggers.Keys.First()], arg);
```
StanleyGoldman: Temporary hard coded fix.

src/GitHub.App/Controllers/UIController.cs
```diff
@@ -802,10 +802,33 @@ bool CreateViewAndViewModel(UIViewType viewType, ViewWithData data = null)
...
+            if (arg != null)
+            {
+                if (next == Trigger.Next)
+                {
+                    switch (requestedTarget?.ViewType)
+                    {
+                        case UIViewType.PRDetail:
+                            viewWithDataTrigger = triggers[Trigger.PRDetail];
+                            break;
+                    }
```
StanleyGoldman: Checking if the `requestedTarget` needs a parameter.

src/GitHub.App/Controllers/UIController.cs
```diff
@@ -258,9 +258,15 @@ public void Start([AllowNull] IConnection conn)
...
             if (where.ActiveFlow != mainFlow)
-                return;
+            {
+                var message = string.Format(CultureInfo.CurrentCulture,
+                    "Jump called for flow {0} but this is {1}", 
+                    where.ActiveFlow, 
+                    mainFlow);
+
+                throw new ArgumentException(message);
+            }
```
StanleyGoldman: I think it's better to throw an error in this situation than `Assert()`.

src/UnitTests/GitHub.App/Controllers/UIControllerTests.cs
```diff
@@ -308,37 +446,71 @@ public void AuthFlowWith2FA()
...
+                    Assert.Equal(testViewCount + 1, count);
+                    count++;
                 });
 
                 uiController.Start(null);
-                Assert.Equal(3, count);
+                Assert.Equal(testViewCount + 2, count);
                 Assert.True(uiController.IsStopped);
+                Assert.True(success.HasValue);
+                Assert.True(success);
```
StanleyGoldman: `Assert.True(success);` bothers me here. If the last action performed was Cancel on the clone flow, shouldn't this value be False?

StanleyGoldman:
All the changes up to this point resolve the "Back and Forth" issue reported in #628, but I'm noticing a bit different behavior on the "Refresh".

StanleyGoldman:
I'm seeing an invalid state transition from PRDetail to PRDetail

StanleyGoldman:
I think the problem is due to the Next Trigger attempting to redirect to PRDetail which is a TriggerWithParam.

My original fix attempted to intercept the Next Trigger in that scenario and issue a PRDetail TriggerWithParam, that works in a few scenarios, but it would require permitting a PRDetail TriggerWithParam everywhere a Next Trigger is specified.

I cleaned up, uniformed  and commented the tests in `UIControllerTests` them while I read them. I added a `Jump` unit test to `PullRequestsFlow` which I believe reproduces the issue at hand.

In `UIController` I changed the `Debug.WriteLine()` calls to `Log.Info()`; and I added additional logs for state machine transitions and unhandled transitions. Reading the log of `UIController` really helped me understand it.

StanleyGoldman:
After not thinking about it for a day, I think one solution could be to make every trigger a TriggerWithParam. Even if a param is not used/required; that would allow the Next operation to work regardless if the "Next" state needs a param or not.

src/UnitTests/GitHub.App/Controllers/UIControllerTests.cs
shana: iirc, `success` is `false` if the flow failed at the end without user intervention. The user cancelling a flow would not be a failure of the flow itself, merely an early exit. And yes, it doesn't feel very logical. I need to revisit some of the thoughts on this one...

shana:
To be honest, I think UIController is taking on too much responsibility and shouldn't be handling all this work, which is manifesting itself in all these problems. It's meant to be a state machine handling logic for ui flows, which really doesn't work well when forced to jump to arbitrary points in the middle of a particular flow.

grokys:
Thanks for the attempt @StanleyGoldman, but we decided to rewrite the navigation controller in #752 so closing this.

Issue #709, opened 2016-12-07
Target/source branch text hard to read in dark theme 


paladique:
Noticed this while testing `grokys/maintainer-workflow-ui-update`

![image](https://cloud.githubusercontent.com/assets/1414307/20979959/b63c3dd6-bc7c-11e6-8467-c82cd5f4c5cf.png)

cc @donokuda @grokys

Issue #625, opened 2016-10-24
Maintainer Workflow: Change the message icon to be less scary


shana:
Currently we're using a warning icon for our messages, which is a bit scary to users (what did I just do??? panic). Making it an informational icon (or a friendly octocat or something) will probably make it less scary and more of a guidance to next steps (which is what most messages are all about).

donokuda:
The fix for this is included with https://github.com/github/VisualStudio/pull/607

![image](https://cloud.githubusercontent.com/assets/1174461/19662606/196f251a-99ed-11e6-830d-1b65f89d14fd.png)

donokuda:
We reworked these messages in https://github.com/github/VisualStudio/pull/730

Issue #302, opened 2016-05-16
Pull requests: details view exploration


donokuda:
Exploring another layout treatment for the pull request details view. The layout is inspired by some of @simurai's work for Atom ✨:

![pull request details](https://cloud.githubusercontent.com/assets/1174461/15298624/16b1bf7a-1b55-11e6-95cc-57d35d40495d.png)

The top part of the details view would have a summary of the pull request. At a quick glance, you should be able to understand the following:
- What is the current state of the PR?
- How large is the branch?
- Is this branch stable?
- Can this branch be merged or are there conflicts that need to be resolved first?

![artboard 1](https://cloud.githubusercontent.com/assets/1174461/15299788/b8a87b66-1b5a-11e6-8478-450e666f6664.png)

If there is something with the branch that needs to be addressed, then we should be able to guide the developer to solving it. For example, if there are builds failing, we could link to a list of build failures for that pull request on GitHub.com. If everything looks good, then we could merge the branch from this view (if that's something we want to support.)

I also thought about how we could support opening and closing a pull request in the UI. One idea was to have the open/close badge be a dropdown. Another idea is to have the "close" button next to the comment button (closer to how we do it on GitHub.com)

![pull-request-state](https://cloud.githubusercontent.com/assets/1174461/15300358/112d0296-1b5d-11e6-9274-5c89c91d2240.png)

Below the branch summary will show more details for the pull request. From left to right are sketches for details, commits, files changed, and inline (code) comments.

![pull request details](https://cloud.githubusercontent.com/assets/1174461/15230658/a592025c-184b-11e6-983d-bce84a5adddb.png)

You could navigate between the different views (details, code comments, commits, files changed) via a dropdown. 

![desktop copy](https://cloud.githubusercontent.com/assets/1174461/15232635/7fa63dae-1857-11e6-97a7-29df5bff49dc.png)

This navigation treatment could have the benefit of allowing us to add more views in the future. For example, if we were to add a "builds" view, it would appear as another choice in the dropdown.

Feedback welcome 🙇 

(/cc @shana @simurai)

simurai:
> One idea was to have the open/close badge be a dropdown.

Heh.. that's a great way to optimize the available space. :sunglasses:

> Another idea is to have the "close" button next to the comment button (closer to how we do it on GitHub.com)

[...]

PR #730, opened 2016-12-09
Polish pull request details view

donokuda:
Fixes https://github.com/github/VisualStudio/issues/709 https://github.com/github/VisualStudio/issues/713 https://github.com/github/VisualStudio/issues/625 https://github.com/github/VisualStudio/issues/302

This pull request reworks the details view. I reworked around the information so that it's quicker to parse than before.

**Before**

screen shot 2016-12-12 at 3 48 31 pm

**After**

screen shot 2016-12-12 at 2 44 50 pm

**Push / Pull branch actions**

screen shot 2016-12-12 at 3 53 59 pm

**Shorter title and checkout branch action**

screen shot 2016-12-12 at 3 54 19 pm

**Themes!**

screen shot 2016-12-12 at 2 44 35 pm
screen shot 2016-12-12 at 2 44 14 pm

### TODO

- [x] Dark theme
- [x] Light theme
- [x] Don't hardcode font sizes
- [x] Double check design for error messages
- [ ] Handle long branch names (wrap or truncate perhaps?)

### DIFF
--- a/src/GitHub.VisualStudio.UI/Styles/ThemeDesignTime.xaml
+++ b/src/GitHub.VisualStudio.UI/Styles/ThemeDesignTime.xaml
@@ -4,7 +4,7 @@
   xmlns:PresentationOptions="http://schemas.microsoft.com/winfx/2006/xaml/presentation/options">
 
   <ResourceDictionary.MergedDictionaries>
-    <ResourceDictionary Source="pack://application:,,,/GitHub.VisualStudio.UI;component/Styles/ThemeLight.xaml" />
+    <ResourceDictionary Source="pack://application:,,,/GitHub.VisualStudio.UI;component/Styles/ThemeBlue.xaml" />
   </ResourceDictionary.MergedDictionaries>
 
   <!-- Design time colors taken from the VS dark theme
--- a/src/GitHub.VisualStudio.UI/Styles/ThemeDesignTime.xaml
+++ b/src/GitHub.VisualStudio.UI/Styles/ThemeDesignTime.xaml
@@ -4,7 +4,7 @@
   xmlns:PresentationOptions="http://schemas.microsoft.com/winfx/2006/xaml/presentation/options">
 
   <ResourceDictionary.MergedDictionaries>
-    <ResourceDictionary Source="pack://application:,,,/GitHub.VisualStudio.UI;component/Styles/ThemeBlue.xaml" />
+    <ResourceDictionary Source="pack://application:,,,/GitHub.VisualStudio.UI;component/Styles/ThemeLight.xaml" />
   </ResourceDictionary.MergedDictionaries>
 
   <!-- Design time colors taken from the VS dark theme
@@ -16,9 +16,11 @@
   <SolidColorBrush x:Key="GitHubVsSearchBoxBackground" Color="#FF333337" />
   <SolidColorBrush x:Key="GitHubVsWindowText" Color="#FFF1F1F1" />
   <SolidColorBrush x:Key="GitHubVsBrandedUIBorder" Color="#FF3F3F46" />
+  <SolidColorBrush x:Key="GitHubBranchNameBackgroundBrush" Color="#FF4B4D50" />
   -->
 
   <!-- Design time colors taken from the VS light theme
+  -->
   <SolidColorBrush x:Key="GitHubVsToolWindowText" Color="#FF1E1E1E" />
   <SolidColorBrush x:Key="GitHubVsToolWindowBackground" Color="#FFF5F5F5" />
   <SolidColorBrush x:Key="GitHubVsGrayText" Color="#FF717171" />
@@ -28,10 +30,10 @@
   <SolidColorBrush x:Key="GitHubVsWindowText" Color="#FF1E1E1E" />
   <SolidColorBrush x:Key="GitHubVsBrandedUIBorder" Color="#FFCCCEBD" />
   <SolidColorBrush x:Key="GitHubVsBrandedUIBackground" Color="#FFEEEEF2" />
-  -->
+  <SolidColorBrush x:Key="GitHubDirectoryIconForeground" Color="#FFF5D367" />
+  <SolidColorBrush x:Key="GitHubBranchNameBackgroundBrush" Color="#FFDAE8F6" />
 
   <!-- Design time colors taken from the VS blue theme
-  -->
   <SolidColorBrush x:Key="GitHubVsToolWindowText" Color="#FF000000" />
   <SolidColorBrush x:Key="GitHubVsToolWindowBackground" Color="#FFFFFFFF" />
   <SolidColorBrush x:Key="GitHubVsGrayText" Color="#FF6d6d6d" />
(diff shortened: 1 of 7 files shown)


saraford:
The 2nd image implies that the name of the local repo will always match the remote repo `commits-behind` (since I would click on `checkout commits-behind` in the previous pane). This got me thinking... is there any way for users to get to this pane where the local branch has a different name than the remote? E.g. is it possible to rename this local branch in TE? If so, what would happen if the user tried to pull or push?

saraford:
another thought... at the top we show the user the PR number and Repo. Would showing the current local branch also be helpful? Or does that become too wordy?

grokys:
> The 2nd image implies that the name of the local repo will always match the remote repo commits-behind (since I would click on checkout commits-behind in the previous pane).

`commits-behind` here is the name of the branch, not the repository.

> is there any way for users to get to this pane where the local branch has a different name than the remote? E.g. is it possible to rename this local branch in TE? If so, what would happen if the user tried to pull or push?

Yes, it is possible to rename a branch in TE. In the case of a PR from the same repository, this is no different from checking out the PR branch normally (i.e. not using the PR pane) and renaming it however. All we're doing in that case is checking out a branch normally. In the case of a branch checked out from a fork, then the push and pull should still push and pull to the correct branch on the fork,

> Would showing the current local branch also be helpful? Or does that become too wordy?

The current branch is always shown in the status bar at the bottom right, so I'm not sure that adding it here would add any extra info.

src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
```diff
@@ -113,108 +120,118 @@
...
+
+                    <!-- Pull/push buttons -->
+                    <StackPanel Orientation="Horizontal" VerticalAlignment="Center"
+                                Visibility="{Binding UpdateState.UpToDate, FallbackValue=Collapsed, Converter={ui:BooleanToInverseVisibilityConverter}}">
+                        <ui:OcticonImage Icon="arrow_down"/>
+                        <TextBlock Text="{Binding UpdateState.CommitsBehind}" VerticalAlignment="Center"/>
+                        <ui:GitHubActionLink Content="Pull"
+                                             Command="{Binding Pull}"
+                                             Margin="4,0"
+                                             ToolTip="{Binding UpdateState.PullDisabledMessage}"
```
donokuda: Would it make sense to have a tooltip that says something like "Pull changes from the remote branch foo-branch to your local branch?" If so what's the best way to do this so that it also works with this disabled message?

src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
```diff
@@ -113,108 +120,118 @@
...
+                    </TextBlock>
+
+                    <!-- Git operation error message -->
+                    <TextBox Foreground="Red" 
+                             Margin="0 4"
+                             Text="{Binding OperationError}"
+                             VerticalAlignment="Center"
+                             TextWrapping="Wrap"
+                             Style="{StaticResource FlatReadOnlyTextBox}"
+                             Visibility="{Binding OperationError, Converter={ui:NullToVisibilityConverter}}"/>
```
donokuda: What's the best way to test this? Should I set the `Visibility` to something like `true`?

grokys: Yeah you can do that, or you can set the `OperationError` property on the sample data.

src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
grokys: Hmm, that will require a few changes. Probably best to merge this without them an add them after?

donokuda: Yup, that's a good call 👍

src/GitHub.VisualStudio.UI/Styles/ThemeDesignTime.xaml
```diff
@@ -4,7 +4,7 @@
   xmlns:PresentationOptions="http://schemas.microsoft.com/winfx/2006/xaml/presentation/options">
 
   <ResourceDictionary.MergedDictionaries>
-    <ResourceDictionary Source="pack://application:,,,/GitHub.VisualStudio.UI;component/Styles/ThemeBlue.xaml" />
+    <ResourceDictionary Source="pack://application:,,,/GitHub.VisualStudio.UI;component/Styles/ThemeLight.xaml" />
```
grokys: Did you mean to commit this change or was this part of you checking other themes?

grokys:
LGTM @donokuda! I commited a small change to remove a comment that didn't seem necessary any more and asked one question. Once you've answered the question, looks good to merge!

src/GitHub.VisualStudio.UI/Styles/ThemeDesignTime.xaml
donokuda: This was me checking out other themes. I'll change this back to make the diff look better by one line 😉

src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
donokuda: Awesome. I'll check this out and then merge if it looks good.

donokuda:
@grokys I reworked the `<StackPanel>` to be a `<Grid>` instead so that error messages don't get cut off. I wasn't able to test it in the wild (not exactly sure how to, but I'll take a pass if someone points me in the right direction.)

Let me know how this looks and if it's good to merge still. 🙌

grokys:
Are we ready for a merge? There's still "Double check design for error messages" left unchecked.

shana:
I'm really late to this party but just wanted to note that I think this design is really nice and readable, looks great! :heart:

donokuda:
@grokys I just doubled checked and looks like I need to tidy up this example of modern art 😄 

image

Throwing back on the WIP label but will 🏓 when it's ready.

grokys:
It's a masterpiece! ;)

I also noticed that the "`foo` into `bar`" message gets cut off if the branch names are long. Don't know if you wanted that to wrap?

image

donokuda:
hah! more like "`foo` into the margin of the tool window!" I'll try a version of it wrapping. If it feel awkward, then I'll try a pass with truncating.

Good looking out 👀

donokuda:
Error messages now look like this:

screen shot 2016-12-16 at 10 33 40 am

I'm also trying to figure out the best way to handle long branch names. I moved everything to a `<Grid>` so that the width of the branch name adjusts based on the available name. This works fine when branch names (specifically the source branch) is long:

screen shot 2016-12-16 at 4 28 38 pm

But it falls apart when the branch name is short 😢:

screen shot 2016-12-16 at 2 28 14 pm

Ideally, what I would like it to do is have the `MaxWidth` of the column with the source branch name to be the equivalent of `Width="*"` and the `Width` to be `Auto` (if that makes any sense.) I'm stumped on how to do this... 😕 

I can haz halp? 🙏

grokys:
Just spoke with @shana and we'd like to get a beta prepared so going to merge this as-is and fix the spacing after.

donokuda:
@grokys Sounds good. Thanks for pushing this one over the finish line! 🙇

Issue #722, opened 2016-12-09
Add filtering and sorting to PR list.


grokys:
When the pull request list is large, it can be hard to find the one you're looking for. It would be useful to have options for filtering (by number and title keyword) and sorting in the PR list.

PR #732, opened 2016-12-12
Pull Request Filtering

StanleyGoldman:
As requested in #722, functionality to filter pull requests by title or number.

Pull Request Filtering


src/GitHub.Exports.Reactive/ViewModels/IPullRequestListViewModel.cs
```diff
@@ -38,5 +38,7 @@ public interface IPullRequestListViewModel : IViewModel
         IAccount SelectedAuthor { get; set; }
         ObservableCollection<IAccount> Assignees { get; }
         IAccount SelectedAssignee { get; set; }
+        string FilterText { get; set; }
+        bool IsLoaded { get; }
```
StanleyGoldman: I get the feeling `IsLoaded` should be here.

shana:
This is pretty neat! 😄

StanleyGoldman:
It looks good. Just needs some 🏋️ from @donokuda

grokys:
This looks great! A few things:

- While the list is loading, the filter text box is enabled but typing into it does nothing. We probably need to fix this to make filtering etc work too while loading (all the PRs for e.g. ([aspnet/Mvc](https://github.com/aspnet/Mvc) take a while to load) but for the moment we should probably disable the textbox while loading like we do with the sort dropdowns
- I tried to search for a PR by number by typing "#3" but PR 3 was not found - I'd expect that I'd search for PRs by number by using the "#" prefix
- Should probably [put a delay on the search text box binding](http://www.jonathanantoine.com/2011/09/21/wpf-4-5-part-4-the-new-bindings-delay-property/), as currently typing lags when a project has a lot of PRs (again as in aspnet/Mvc)

StanleyGoldman:
Ah, testing against a repo with a lot of pull requests is something I need to start doing then.

StanleyGoldman:
Updated the gif above.

Let me know if you find the "#" functionality odd..
It wont assume you are performing a pull request number search until and while you follow the "#" with digits. So the initial "#" starts to search for pull requests with "#" in the title.

src/GitHub.VisualStudio/UI/Views/PullRequestListView.xaml
```diff
@@ -141,7 +141,7 @@
     </Grid>
 
     <ui:FilterTextBox x:Name="filterText"
-            IsEnabled="{Binding IsLoaded}"
+            IsEnabled="{Binding IsLoaded, Delay=500}"
```
grokys: This delay should be on the `Text` property, not `IsEnabled` ;)

grokys:
No, I think the "#" works as you've implemented it. I just wonder whether it should be a full string search on the PR number instead of a substring search. For example in `aspnet/Mvc` if I search "#2" I'm probably looking for PR #2, but instead I get _all_ of the PRs which start with the number 2 (which is a lot in this case!)

src/GitHub.VisualStudio/UI/Views/PullRequestListView.xaml
StanleyGoldman: 😨

StanleyGoldman:
Fixed up

StanleyGoldman:
Yea, I tried it with exact number matching, I can push a branch if you wanna try it out.
It just felt kind of... empty... until you got to the number you were typing, lol.

grokys:
Hmm yeah, I see what you're saying. One thing I tried (to see if it'd work) was adding a space after the number to see if that forced exact matching - I'm not sure how many people would think to try that though?

tocsoft:
could you not use both searching methods?
order exact matches first then partial matches below?

grokys:
That's a really good idea @tocsoft !

StanleyGoldman:
I think that's actually a sort not a filter. We would have to land the sorting pull request to see how to integrate the logic.

I'm willing to check it out, but I'm afraid a user wouldn't be able to ascertain what it's doing.

donokuda:
I'm assuming that if someone is searching for a PR by its number, then they would know exactly which number they're looking for. If that's the case, then I'm not convinced of the value spent trying to refine filtering for PR numbers beyond how it currently works.

I also can't remember the last time when I went "Oh, I need to search for PR #732." Rather, I'm probably going to be searching for keywords like "pull request filtering" instead. That said, I would love to hear your experiences when searching for a specific pull request number is more useful and quicker (perhaps in tandem with an issue tracker 😄) 

However, if matching worked on something like branch names (which could be considered out of scope for this pull request), then searching for a number would be more useful when I know a branch is [fixing a specific issue](https://github.com/github/VisualStudio/pull/523) (`fixes/479-clear-pr-creation-form`).

grokys:
> I also can't remember the last time when I went "Oh, I need to search for PR #732." Rather, I'm probably going to be searching for keywords like "pull request filtering" instead. 

I do this all the time: for example if someone mentions a PR by number in slack. In addition if I'm looking at a PR on github.com and want to find that PR in Visual Studio I'll likely want to search by number,

StanleyGoldman:
So I think we are agreeing?
Summary: Partial number matching of any kind is misleading and exact number matching is best.

grokys:
Yeah, I've just tried out the PR with the exact matching, and while I can see your problem with it, I think the behavior is more useful. So personally I think exact number matching is best, though I can't speak for anyone else.

StanleyGoldman:
Agreed. Exact matching is best.

shana:
Let's go with that and see how it goes 😄

jcansdale:
This seems to work well. 😄 

For consistency, it would be good to use the magnifying glass icon:

image

It looks like standard icons can be accessed via `IVsImageService2`. See:
https://github.com/madskristensen/ExtensibilityTools/blob/master/src/Misc/Commands/ImageMonikerDialog.xaml.cs

grokys:
@jcansdale yes that would be nice, to properly fit in with other VS search boxes I think we need a completely different control (as the one we're using is styled for our dialog which uses a different color scheme etc). This also has the downside that the search box is white on a dark theme. I think that would be another PR tbh...

grokys:
I've updated the PR to use the built-in VS search functionality, which works around the styling issues. Now looks like this:

image

grokys:
Moved to #1312.

PR #733, opened 2016-12-12
Handle diff/open of non-checked out files.

### LINKED ISSUE #725: Maintainer Workflow: opening/diffing files in PRs from forks doesn't work.
When viewing a PR from a fork in the PR details view, if the fork has not been fetched locally then "Compare File" and "Open File" fail silently. This is because to get the files to compare, we look up the PR HEAD commit and get the file from the returned tree.

We do a fetch before looking up the PR HEAD but if the fork is a repository that does not yet have a remote then we won't have fetched the commit.

I think we need to send an API request for [Get Contents](https://developer.github.com/v3/repos/contents/#get-contents) and then use the returned `download_url` to fetch the raw contents.

grokys:
When viewing a PR from a fork in the PR details view, if the fork has not been fetched locally then "Compare File" and "Open File" were
failing silently. This is because to get the files to compare, we looked up the PR HEAD commit and get the file from the returned tree. We did a
fetch before looking up the PR HEAD but if the fork is a repository that does not yet have a remote then we won't have fetched the commit.

This adds a fallback to using the API and storing a cached version of the file contents.

#720 and #724 should be merged before this (which should resolve the conflicts)

Fixes #725

### DIFF
--- a/src/GitHub.App/Services/PullRequestService.cs
+++ b/src/GitHub.App/Services/PullRequestService.cs
@@ -247,27 +247,57 @@ namespace GitHub.Services
             });
         }
 
-        public IObservable<string> ExtractFile(ILocalRepositoryModel repository, string commitSha, string fileName)
+        public IObservable<string> ExtractFile(
+            ILocalRepositoryModel repository,
+            IModelService modelService,
+            string commitSha,
+            string fileName)
         {
             return Observable.Defer(async () =>
             {
                 var repo = gitService.GetRepository(repository.LocalPath);
                 await gitClient.Fetch(repo, "origin");
-                var result = await gitClient.ExtractFile(repo, commitSha, fileName);
+                var result = await ExtractFile(repository, repo, modelService, commitSha, fileName);
                 return Observable.Return(result);
             });
         }
 
-        public IObservable<Tuple<string, string>> ExtractDiffFiles(ILocalRepositoryModel repository, IPullRequestModel pullRequest, string fileName)
+        public IObservable<Tuple<string, string>> ExtractDiffFiles(
+            ILocalRepositoryModel repository,
+            IModelService modelService,
+            IPullRequestModel pullRequest,
+            string fileName)
         {
             return Observable.Defer(async () =>
             {
                 var repo = gitService.GetRepository(repository.LocalPath);
                 await gitClient.Fetch(repo, "origin");
-                var left = await gitClient.ExtractFile(repo, pullRequest.Base.Sha, fileName);
-                var right = await gitClient.ExtractFile(repo, pullRequest.Head.Sha, fileName);
+                var left = await ExtractFile(repository, repo, modelService, pullRequest.Base.Sha, fileName);
+                var right = await ExtractFile(repository, repo, modelService, pullRequest.Head.Sha, fileName);
                 return Observable.Return(Tuple.Create(left, right));
             });
+        }
+
+        private async Task<string> ExtractFile(
+            ILocalRepositoryModel repository,
+            IRepository repo,
+            IModelService modelService,
+            string commitSha,
+            string fileName)
+        {
+            var result = await gitClient.ExtractFile(repo, commitSha, fileName);
+
+            if (result == null)
+            {
+                var contents = await modelService.GetFileContents(repository, commitSha, fileName);
+                var tempDir = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+                var tempFileName = $"{Path.GetFileNameWithoutExtension(fileName)}@{commitSha}{Path.GetExtension(fileName)}";
+                result = Path.Combine(tempDir, tempFileName);
+                Directory.CreateDirectory(tempDir);
+                File.WriteAllBytes(result, contents);
+            }
+
+            return result;
         }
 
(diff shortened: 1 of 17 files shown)


src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -248,29 +248,58 @@ public IObservable<Unit> UnmarkLocalBranch(ILocalRepositoryModel repository)
...
+
+                // The right file - if it comes from a fork - may not be fetched so fall back to
+                // getting the file contents from the model service.
+                var right = await gitClient.ExtractFile(repo, pullRequest.Head.Sha, fileName) ??
+                            await modelService.GetFileContents(repository, pullRequest.Head.Sha, fileName, fileSha);
                 return Observable.Return(Tuple.Create(left, right));
             });
         }
 
+        private async Task<string> ExtractFile(
```
shana: /me stares at the `private`

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -248,29 +248,58 @@ public IObservable<Unit> UnmarkLocalBranch(ILocalRepositoryModel repository)
...
+            string fileName,
+            string fileSha)
         {
             return Observable.Defer(async () =>
             {
                 var repo = gitService.GetRepository(repository.LocalPath);
                 await gitClient.Fetch(repo, "origin");
-                var result = await gitClient.ExtractFile(repo, commitSha, fileName);
+                var result = await gitClient.ExtractFile(repo, commitSha, fileName) ??
+                             await modelService.GetFileContents(repository, commitSha, fileName, fileSha);
```
shana: Given that these are IO calls, there's a ton of different ways this can fail. We need to handle exceptions from these calls somewhere and fail sanely.

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -248,29 +248,58 @@ public IObservable<Unit> UnmarkLocalBranch(ILocalRepositoryModel repository)
...
+            ILocalRepositoryModel repository,
+            IRepository repo,
+            IModelService modelService,
+            string commitSha,
+            string fileName,
+            string fileSha)
+        {
+            return await gitClient.ExtractFile(repo, commitSha, fileName) ??
+                   await modelService.GetFileContents(repository, commitSha, fileName, fileSha);
+        }
```
shana: Is this method used somewhere? This seems to be a duplicate of the one above, did you mean to call this one instead?

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -248,29 +248,58 @@ public IObservable<Unit> UnmarkLocalBranch(ILocalRepositoryModel repository)
...
                 await gitClient.Fetch(repo, "origin");
+
+                // The left file is the target of the PR so this should already be fetched.
                 var left = await gitClient.ExtractFile(repo, pullRequest.Base.Sha, fileName);
-                var right = await gitClient.ExtractFile(repo, pullRequest.Head.Sha, fileName);
+
+                // The right file - if it comes from a fork - may not be fetched so fall back to
+                // getting the file contents from the model service.
+                var right = await gitClient.ExtractFile(repo, pullRequest.Head.Sha, fileName) ??
+                            await modelService.GetFileContents(repository, pullRequest.Head.Sha, fileName, fileSha);
```
shana: Lots of duplicated calls going on here, probably best to call the Task method below? Maybe handle exceptions down there as well.

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
```diff
@@ -346,7 +354,7 @@ public override void Initialize([AllowNull] ViewWithData data)
         public Task<string> ExtractFile(IPullRequestFileNode file)
         {
             var path = Path.Combine(file.DirectoryPath, file.FileName);
-            return pullRequestsService.ExtractFile(repository, model.Head.Sha, path).ToTask();
+            return pullRequestsService.ExtractFile(repository, modelService, model.Head.Sha, path, file.Sha).ToTask();
         }
```
shana: You're totally right, we should be having methods that return `Task` when there's only one result, they are way better suited to convey the meaning of what's going on. If you want to change the signature of the `PullRequestService.ExtractFile` to return Task and use that everywhere, 👍

src/GitHub.App/Services/PullRequestService.cs
grokys: If an exception is thrown inside an `Observable.Defer` then it will be converted into an `OnError` handler, so we're safe from crashing at least. The question is what do we want to do with any errors encountered? Show in the status bar? In the PR details pane?

src/GitHub.App/Services/PullRequestService.cs
grokys: Yep, I wrote that method and then forgot to use it!

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
grokys: Ok, if I do this we *will* need to handle errors at the call sites ;)

PR #734, opened 2016-12-13
Pull Request Sorting

StanleyGoldman:
As requested in #722, functionality to sort pull requests.

Pull Request Sort


grokys:
Thanks @StanleyGoldman! Might also be nice to also be able to sort by PR number?

StanleyGoldman:
Hey @grokys.
I thought about that, but wouldn't that would be the same as sorting by "Created At".
I could put PR Number in the menu if it's clearer.

grokys:
Yeah, you're right of course! It was early ;) But yeah perhaps calling it "PR Number" might be clearer? Not sure.

grokys:
I guess GitHub.com is the best place to look for inspiration, here it uses:

image

StanleyGoldman:
Lol, you are right. Looking at that made me laugh, because I was trying to figure out how to handle ascending/descending sort..

StanleyGoldman:
I updated the gif above.
I feel like "Least recently updated" takes up a lot of space.

grokys:
> I feel like "Least recently updated" takes up a lot of space.

Yeah, you're right. We might be best diverging from .com behavior here. @donokuda @shana @paladique any ideas?

paladique:
Maybe "Newest updates/Oldest updates"?

donokuda:
I'm not too worried about "Least Recently Updated" taking up space here. If we're worried about the text taking up too much space, we could consider truncating it instead.

paladique:
Another thing comes to mind: should these dropdown options go into the resource dictionary?

grokys:
Yeah, they should - there's already plenty of strings in the PR detail view and an issue to clean that up, so we can do it all together assuming we merge this for release at the same time.

shana:
While I'm poking at #737, a few comments on the design:

### Largest sorting selection with default layout and pane sizes
image

The sorting text is too long for the default pane width (which is the default width of the solution explorer since we're docked to the same area as it). This is a minor issue, we can tweak the text.

The bigger issue is that by adding sorting to the dropdown list, we're now conflating two different user actions without differentiation: filtering and sorting. Given what the other dropdowns do, it's easy to conclude that "Newest" means there are PRs that are "older" that are not being shown (since all the other dropdowns actively remove information and there's no indication that this one is any different).

### In .com
image

Note that all filters have the field that they will be filtering on, while "Sort" represents a completely different action. While I'm not a big fan of this UX either, it's clearer at a glance to understand that "Sort" is an action, not a field (unless, of course, you're danish and your brain translates Sort to Type and you get all confused for a second :P).

Sooo, I think we may need to move the sorting to its own line - we don't have space on this line and the UX is not very helpful as-is.

@donokuda @saraford thoughts?

grokys:
How about moving sorting to a dropdown next to the filter textbox? Something like:

image

(I'm not sure the icon is particularly clear but we can use whatever)

grokys:
BTW, I don't think conflating filtering and sorting is such a bad idea - it's a pattern used pretty much everywhere. However the lack of horizontal space for another dropdown is definitely an issue.

shana:
@StanleyGoldman I see your PR and raise you a PR! https://github.com/StanleyGoldman/VisualStudio/pull/1

> How about moving sorting to a dropdown next to the filter textbox?

That might work indeed! Will require a bit of custom UI though.

shana:
For this PR, I'd favour just moving the sorting to a separate line below the filtering. We can figure out later if we want to do something different, like a sorting icon or some other custom UI for it.

StanleyGoldman:
Rebased

src/GitHub.App/ViewModels/PullRequestListViewModel.cs
```diff
@@ -61,39 +103,38 @@ public class PullRequestListViewModel : BaseViewModel, IPullRequestListViewModel
...
+                .Where(x => pullRequests != null && x != EmptyUser && IsLoaded)
                 .Subscribe(a => UpdateFilter(SelectedState, a, SelectedAuthor));
 
             this.WhenAny(x => x.SelectedAuthor, x => x.Value)
-                .Where(x => PullRequests != null && x != EmptyUser && IsLoaded)
+                .Where(x => pullRequests != null && x != EmptyUser && IsLoaded)
                 .Subscribe(a => UpdateFilter(SelectedState, SelectedAssignee, a));
 
+            this.WhenAny(x => x.SelectedSortOrder, x => x.Value)
+                .Where(x => pullRequests != null)
```
shana: @StanleyGoldman Ooops, I missed a `x != null` check here, and it's breaking the tests. Do you want to add it or shall I?

shana:
@StanleyGoldman Sorry, my bad, I broke a couple of things. Pushed fixes to those to your branch so it's not blowing up horribly.

StanleyGoldman:
No worries, reviewing now.

StanleyGoldman:
There were quite a few things I didn't know about. Thanks for the help. 👍

shana:
Ok, so this looks great and I'd like to merge it, but the dropdown needs to move to the next line first, currently it breaks the UI and we can't get it in as-is.

src/GitHub.App/ViewModels/PullRequestListViewModel.cs
```diff
@@ -61,39 +103,38 @@ public class PullRequestListViewModel : BaseViewModel, IPullRequestListViewModel
...
             Authors = trackingAuthors.CreateListenerCollection(EmptyUser, this.WhenAnyValue(x => x.SelectedAuthor));
             Assignees = trackingAssignees.CreateListenerCollection(EmptyUser, this.WhenAnyValue(x => x.SelectedAssignee));
 
             PullRequests = new TrackingCollection<IPullRequestModel>();
-            pullRequests.Comparer = OrderedComparer<IPullRequestModel>.OrderByDescending(x => x.UpdatedAt).Compare;
             pullRequests.NewerComparer = OrderedComparer<IPullRequestModel>.OrderByDescending(x => x.UpdatedAt).Compare;
 
             this.WhenAny(x => x.SelectedState, x => x.Value)
-                .Where(x => PullRequests != null)
+                .Where(x => pullRequests != null)
```
grokys: Out of interest @shana why did you change these to use the underlying field rather than the property?

src/GitHub.Exports.Reactive/ViewModels/ViewModelItemContainer.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace GitHub.ViewModels
+{
+    /// <summary>
+    /// Simple container for complex objects that we stuff
+    /// into dropdowns and the likes
+    /// </summary>
+    public abstract class ViewModelItemContainer
```
grokys: Hmm, this seems to be a rather general class. and the name doesn't really describe it very well. How about something like `NamedItem`?

src/GitHub.Exports.Reactive/ViewModels/ViewModelItemContainer.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        public string Name;
```
grokys: @shana If we're refactoring this, we might as well change this to be a read-only property right? I'm assuming that classes derived from would expect this value to be immutable judging from the lack of INPC? If so a ctor taking the name would be preferable.

Public fields are euurgh.

StanleyGoldman:
Moved the sort dropdown to the next line. Updated gif above.

src/GitHub.Exports.Reactive/ViewModels/ViewModelItemContainer.cs
Haacked: 👍 Agreed. Should be a read-only property.

Haacked:
What's left to do in this PR besides resolving conflicts? :smile:

StanleyGoldman:
This needs some ❤️ from @donokuda. The position of the sort dropdown is not ideal.

StanleyGoldman:
I'll address the requested code changes & inquiries

src/GitHub.Exports.Reactive/ViewModels/IPullRequestListViewModel.cs
```diff
@@ -4,39 +4,33 @@
...
-        public override string ToString()
-        {
-            return Name;
-        }
+    }
+
+    public class PullRequestSortOrder : ViewModelItemContainer
+    {
+        public IComparer<IPullRequestModel> Comparer;
+        public SortOrder SortOrder;
```
Haacked: These should probably be properties. Most likely they should be `readonly` properties. I know that makes creating instances of these not look as nice as using property initializers when constructing an instance, but if these properties should never be mutated after construction, then it guides future consumers to the right behavior.

However, if you're two-way binding these properties to a UI, then ignore the `readonly` part and make them properties. But in that case they should probably be reactive properties (or implement INPC).

src/GitHub.App/ViewModels/PullRequestListViewModel.cs
```diff
@@ -138,10 +179,8 @@ public override void Initialize([AllowNull] ViewWithData data)
                 });
         }
 
-        void UpdateFilter(PullRequestState state, [AllowNull]IAccount ass, [AllowNull]IAccount aut)
+        void UpdateFilter(PullRequestState state, IAccount ass, IAccount aut)
```
Haacked: While I have no problems with naming a variable `ass` I think we should probably avoid swear words in case Microsoft runs their Policheck tool against this code. 😛

src/GitHub.App/ViewModels/PullRequestListViewModel.cs
```diff
@@ -23,6 +24,40 @@ namespace GitHub.ViewModels
...
+                Name = Resources.SortOrderCreatedAscending,
+                Comparer = OrderedComparer<IPullRequestModel>.OrderBy(x => x.CreatedAt)
+            }
+        };
+
+        // a helper lookup table for easy access to the sorting comparers
+        static Dictionary<SortOrder, PullRequestSortOrder> sortOrderIndex;
+
+        // TODO: Move strings to resources
+        IReadOnlyList<PullRequestState> states = new List<PullRequestState> {
```
Haacked: So this is just casting a mutable list to an `IReadOnlyList` which can be cast back to `List`. I know we probably won't do that, but I'm thinking of the next person to come along. The safe thing to do is to wrap `List` with a `ReadOnlyList`. We have an extension method `ToReadOnlyList` for observables. Wouldn't be hard to write one for `IEnumerable`.

StanleyGoldman: So I discovered this `Array.AsReadOnly(...)` method...

src/GitHub.App/ViewModels/PullRequestListViewModel.cs
StanleyGoldman: 👍

src/GitHub.Exports.Reactive/ViewModels/IPullRequestListViewModel.cs
StanleyGoldman: 👍

src/GitHub.Exports.Reactive/ViewModels/ViewModelItemContainer.cs
StanleyGoldman: 👍

src/GitHub.App/ViewModels/PullRequestListViewModel.cs
StanleyGoldman: paging: @shana

sguthals:
Closing because stale.

PR #664, opened 2016-11-04
Fixed #627: Project recently published to remote pre-fills local name for new project

jswietek:
2 main changes:
1) RepositoryPublishService now has a reference to VSGitService. This is used to call GetActiveRepo() everytime when needed instead of getting that value in constructor. This is needed as active repo may change and the reference was never updated in the RepositoryPublishService

2) GitHubPublishSection.cs now re-initialize the section view when active repo changes.

Fixes #627

PR #736, opened 2016-12-14
Project recently published to remote pre-fills local name for new project

grokys:
This PR is a modification of #664 which fixes the problem where `RepositoryPublishService.PublishRepository` was being called on a background thread, and ends up invoking AsyncPackage.GetService which must be called from the main thread.


src/GitHub.App/Services/RepositoryPublishService.cs
```diff
@@ -49,5 +52,11 @@ public string LocalRepositoryName
                                  return repo.RemoteRepo;
                              });
         }
+
+        async Task<IRepository> GetActiveRepo()
+        {
+            await ThreadingHelper.SwitchToMainThreadAsync();
+            return vsGitServices.GetActiveRepo();
```
shana: I looked at the docs and the code and `GetService` should be totally safe to call now. Sooo we really need to test this again and see if we still get deadlocks.

shana:
thumbs-up-scroll

rubenscholfield:
Can u look for my server, and if its in good mood,
Thank you, staff.
RUBEN SCHOLFIELD

Am 16.12.2016 12:12 nachm. schrieb "Steven Kirk" <notifications@github.com>:

PR #739, opened 2016-12-14
Handle non-latin PR titles.

### LINKED ISSUE #618: Handle non-ascii issue titles in PR checkouts
When we check out a PR branch from a fork we use a branch name that is taken from the PR title by replacing non alphanumeric characters with `-` characters. It's possible that a PR title could be made up of only these characters in e.g. a non-Latin language. In this case we should probably fall back to a simple `pr/ID` branch name.

grokys:
Previously a non-latin PR title resulted in a PR branch called `pr/[Number]-`. Use just `pr/[Number]` in this case.

Fixes #618.


shana:
:heart:

PR #740, opened 2016-12-14
Pull Request Filtering w/ Exact PR Number Match

StanleyGoldman:
A variation of #732 that matches the number exactly.


StanleyGoldman:
@grokys check this version out when you get a chance.

Issue #738, opened 2016-12-14
PR list causes VS to hang when checking out PR in project with lots of PRs


grokys:
Steps to reproduce (note you may need to try this a few times - it doesn't always happen):

1. Check out `aspnet/Mvc`
2. Run with "break on CLR exceptions" disabled
3. Click on a PR
4. Click on the checkout link for the PR
5. Watch VS hang for a good amount of time

If you look at the output window, about 2250 `ObjectDisposedException`s are being thrown by `TrackingCollection.RemoveItem`. These exceptions are being caught in `ModelService.GetPullRequests` but I assume the sheer amount of exceptions being thrown and caught is causing a hang.

StanleyGoldman:
I almost got a unit test where I can reproduce the scenario.

StanleyGoldman:
You can also reproduce the scenario by performing any activity in the git repo with the view open.
For instance by checking in a simple file using the command line.

grokys:
Fixed by #742

Issue #597, opened 2016-10-13
Occasional exception in TrackingCollection.


grokys:
I get very occasional exceptions in TrackingCollection - usually when the PR list is updating. Here's the details:

```
At TrackingCollection.cs line 476

System.ObjectDisposedException occurred
  HResult=-2146232798
  Message=Cannot access a disposed object.
Object name: 'TrackingCollection'.
  ObjectName=TrackingCollection
  Source=GitHub.Exports.Reactive
  StackTrace:
       at GitHub.Collections.TrackingCollection`1.RemoveItem(T item) in D:\projects\VisualStudio\src\GitHub.Exports.Reactive\Collections\TrackingCollection.cs:line 476
  InnerException: 
```

As you can see, the stack trace shows only the current method: `RemoveItem`.

This crashes VS.

shana:
Feels like a race condition between disposal of the current listeners in preparation for processing the new results and the current list still processing.

StanleyGoldman:
I tried my best to reproduce this bug or see how it could happen and I checked the master and maintainer-workflow-diff branches. The only usages of `RemoveItem()` I could find are wrapped in `try/catch` blocks that swallow exceptions of type `ObjectDisposedException`. I'm not sure how you are getting that exception.

grokys:
Just got this again with the following stack trace:

```
GitHub.Exports.Reactive.dll!GitHub.Collections.TrackingCollection<GitHub.Models.IPullRequestModel>.RemoveItem(GitHub.Models.IPullRequestModel item) Line 476	C#
GitHub.App.dll!GitHub.Services.ModelService.GetPullRequests.AnonymousMethod__4(GitHub.Services.ModelService.PullRequestCacheItem item) Line 168	C#
GitHub.App.dll!GitHub.Extensions.AkavacheExtensions.GetAndFetchLatestFromIndex.AnonymousMethod__13() Line 238	C#
mscorlib.dll!System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.InvokeMoveNext(object stateMachine)	Unknown
mscorlib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx)	Unknown
[...]

PR #742, opened 2016-12-14
ModelService prevents ObjectDisposedException

StanleyGoldman:
A fix for #738 and #597
I realized I wouldn't be able to make a unit test for an exception that was being discarded.
And that's when I realized the fix might just be to "take the hint".
If the collection throws an `ObjectDisposedException`, simply stop trying to access it.

### DIFF
--- a/src/GitHub.App/Services/ModelService.cs
+++ b/src/GitHub.App/Services/ModelService.cs
@@ -157,33 +157,25 @@ namespace GitHub.Services
             var keyobs = GetUserFromCache()
                 .Select(user => string.Format(CultureInfo.InvariantCulture, "{0}|{1}:{2}", CacheIndex.PRPrefix, user.Login, repo.Name));
 
-            var source = Observable.Defer(() =>
-            {
-                var collectionIsDisposed = false;
-
-                return keyobs
-                    .SelectMany(key => hostCache.GetAndFetchLatestFromIndex(key, () =>
+            var source = Observable.Defer(() => keyobs
+                .SelectMany(key =>
+                    hostCache.GetAndFetchLatestFromIndex(key, () =>
                         apiClient.GetPullRequestsForRepository(repo.CloneUrl.Owner, repo.CloneUrl.RepositoryName)
-                            .Select(PullRequestCacheItem.Create), item =>
-                            {
-                                try
-                                {
-                                    if (!collectionIsDisposed)
-                                    {
-                                        // this could blow up due to the collection being disposed somewhere else
-                                        collection.RemoveItem(Create(item));
-                                    }
-                                }
-                                catch (ObjectDisposedException)
-                                {
-                                    // take the hint and stop trying
-                                    collectionIsDisposed = true;
-                                }
-                            }, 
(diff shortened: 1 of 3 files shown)


src/GitHub.App/Services/ModelService.cs
```diff
@@ -157,22 +157,33 @@ public IObservable<AccountCacheItem> GetUserFromCache()
...
-                            // this could blow up due to the collection being disposed somewhere else
-                            try { collection.RemoveItem(Create(item)); }
-                            catch (ObjectDisposedException) { }
-                        },
-                        TimeSpan.Zero,
-                        TimeSpan.FromDays(7))
-                )
-                .Select(Create)
-            );
+                            .Select(PullRequestCacheItem.Create), item =>
```
shana: Could you move item to the next line and unindent it, as well as the lambda block for it? This indentation makes it look like the lambda block is scoped at the same level as the `.Select` statement, which is misleading.

src/GitHub.App/Services/ModelService.cs
```diff
@@ -157,22 +157,33 @@ public IObservable<AccountCacheItem> GetUserFromCache()
...
+                            {
+                                try
```
shana: I am totally fine with adding a `public bool Disposed => disposed;` property in `TrackingCollection` so we can check here if it's disposed instead of waiting for the exception. We still need the try catch block (because race conditions), but in the majority of cases we should be able to avoid exceptions being thrown completely, which would be nice.

StanleyGoldman: That code tricked my eyes 👓

src/GitHub.App/Services/ModelService.cs
StanleyGoldman: That code tricked my eyes 👓

src/GitHub.App/Services/ModelService.cs
```diff
@@ -164,6 +164,9 @@ public IObservable<AccountCacheItem> GetUserFromCache()
                                  .Select(PullRequestCacheItem.Create),
                         item =>
                         {
+                            if(collection.Disposed)
```
shana: Missing a space here

src/GitHub.App/Services/ModelService.cs
```diff
@@ -164,6 +164,9 @@ public IObservable<AccountCacheItem> GetUserFromCache()
...
+                            { return; }
```
shana: You can drop the `{ }` here and just have the return, it's fine.

src/GitHub.App/Services/ModelService.cs
```diff
@@ -206,6 +209,9 @@ public ITrackingCollection<IRemoteRepositoryModel> GetRepositories(ITrackingColl
                                  .Select(RepositoryCacheItem.Create),
                         item =>
                         {
+                            if (collection.Disposed)
+                            { return; }
```
shana: These `{ }` can go too :)

StanleyGoldman:
Cleaned up

PR #744, opened 2016-12-15
Clear the sort cache when changing the sorter

### LINKED ISSUE #737: Sorted PR list crashes VS when GitHub pane closed and re-opened
While testing #734 I came upon what I think is a bug in `TrackingCollection`. To repro:

1. Check out #734 
2. Open the GitHub pane and sort PRs by "Least recently updated"
3. Close the GitHub pane
4. Click Pull Requests in Team Explorer to re-open the GitHub pane
5. Exception:

```
Exception thrown: 'System.ArgumentException' in GitHub.App.dll

Additional information: Instance to copy from doesn't match this instance. this:(id:22 title:Fix CI for ultimate happiness created:2015-07-30 21:59:17Z updated:2015-07-30 22:00:49Z) other:(id:732 title:Pull Request Filtering created:2016-12-12 15:53:29Z updated:2016-12-14 09:58:08Z)

   at GitHub.Models.PullRequestModel.CopyFrom(IPullRequestModel other) in D:\projects\VisualStudio\src\GitHub.App\Models\PullRequestModel.cs:line 30
```

shana:
Fixes #737

Also make sure we scan the whole range of items that have potentially been affected by an item move due to updated sorting fields.
Clean up some tests and add more tests.

### DIFF
--- a/src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
+++ b/src/GitHub.Exports.Reactive/Collections/TrackingCollection.cs
@@ -492,6 +492,7 @@ namespace GitHub.Collections
 
         void RecalculateSort(List<T> list, int start, int end)
         {
+            sortedIndexCache.Clear();
             list.Sort(start, end, new LambdaComparer<T>(comparer));
         }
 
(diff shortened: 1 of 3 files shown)


shana:
> , the sortedIndexCache being out of date was what the problem looked like to me.

Yup, that was mostly the problem, it didn't get nuked when the comparer got changed. I ended up running into another sorting bug in the process, which is where the second change comes in - on a filtered and sorted list, when an item gets updated with data that makes it move to another position on the list (usually because the UpdatedAt date is changed), other items might fall off the filter because of that. The `start++` code that was there was skipping the first item on the range of items to reevaluate, because usually that first item is the one that got moved, but in some cases it isn't, and in those cases the filter wasn't applied properly. The new tests I added and tweaked were hitting that codepath, and were previously passing because the `sortedIndexCache` wasn't getting nuked properly. So hopefully two less bugs!

PR #745, opened 2016-12-16
Fix typo in RemoveUnusedRemotes.

grokys:
It was called RemoteUnusedRemotes, which makes no sense ;)


shana:
My inner grammar nazi approves this PR.

PR #749, opened 2016-12-20
Fix PR branch display layout

grokys:
Finishing off the work in #730, this PR attempts to fix the handling of long branch names in the PR details view. The current results are:

Short branch names:

image

Long branch names (screenshot taken in designer for this one):

image

Mixture of long and short branch names:

image

As you can see this is almost right, except with a mixture of short and long branch names both the short and long branch name take up max 50% of the free space, where the long branch name could take up more. To fix this, I think we'd need to move to a custom `Panel`. Thoughts @donokuda @shana ?

### DIFF
--- a/src/GitHub.App/SampleData/PullRequestDetailViewModelDesigner.cs
+++ b/src/GitHub.App/SampleData/PullRequestDetailViewModelDesigner.cs
@@ -39,8 +39,8 @@ namespace GitHub.SampleData
                 CommitCount = 9,
             };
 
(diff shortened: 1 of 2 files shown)


donokuda:
I also snuck in a small change were we'll truncate long branch names in the `Checkout <branch-name>` text:

**Before**

screen shot 2016-12-20 at 9 14 38 am

**After**

screen shot 2016-12-20 at 9 19 19 am

donokuda:
giphy 13

PR #755, opened 2017-01-05
Implemented SectionControl.

grokys:
We were previously using the one supplied by TeamFoundation but we shouldn't be taking a dependency on TeamFoundation controls, as they're not part of the API.

### DIFF
--- a/src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
+++ b/src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
@@ -10,7 +10,6 @@
                                     xmlns:ui="clr-namespace:GitHub.UI;assembly=GitHub.UI"
                                     xmlns:uir="clr-namespace:GitHub.UI;assembly=GitHub.UI.Reactive"
                                     xmlns:vm="clr-namespace:GitHub.ViewModels;assembly=GitHub.App"
-                                    xmlns:te="clr-namespace:Microsoft.TeamFoundation.Controls.WPF.TeamExplorer.Framework;assembly=Microsoft.TeamFoundation.Controls"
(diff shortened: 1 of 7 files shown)


src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml
```diff
@@ -293,50 +293,61 @@
...
+                                                        <Setter Property="Header" Value="{x:Static prop:Resources.SwitchToListView}"/>
+                                                    </DataTrigger>
+                                                    <DataTrigger Binding="{Binding ChangedFilesViewType}" Value="ListView">
+                                                        <Setter Property="Header" Value="{x:Static prop:Resources.SwitchToTreeView}"/>
+                                                    </DataTrigger>
+                                                </Style.Triggers>
+                                            </Style>
+                                        </MenuItem.Style>
+                                        <MenuItem.Icon>
+                                            <ui:OcticonPath Icon="list_unordered" Fill="Black" Height="16"/>
```
shana: Is the fill always black here, is this going to work on theme changes?

grokys: Ah yes good point!

grokys: Oh no, false alarm: this icon is in a popup menu, and they don't change color according to theme.

shana: Ah cool! I don't think I saw any other thingies that should be themeable or defined somewhere else, so all good!

shana:
bb8_thumbs_up

PR #758, opened 2017-01-06
Fix repository list refresh in Connect page

shana:
Things broke on the connect page when https://github.com/github/VisualStudio/commit/1be202d9 got merged and the order in which `Refresh` gets called changed slightly.

Refresh gets called twice for each connect section when loading, once from the constructor and once from initialize. We can only call `RefreshRepositories` when `Initialize` get called (that's when the Team Explorer IServiceProvider is available and we can make sure the git services we rely upon are registered on that IServiceProvider), and we only want to refresh repositories once, since that command refreshes all repositories for all connections.

This needs to be tested with two accounts logged in (github.com and enterprise) so it creates two connect sections and we make sure it doesn't double-refresh or otherwise break.

### DIFF
--- a/src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
+++ b/src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
@@ -181,7 +181,7 @@ namespace GitHub.VisualStudio.TeamExplorer.Connect
             // watch for new repos added to the local repo list
             var section = GetSection(TeamExplorerConnectionsSectionId);
             if (section != null)
-                sectionTracker = new SectionStateTracker(section, () => RefreshRepositories(true));
+                sectionTracker = new SectionStateTracker(section, () => RefreshRepositories(sectionIndex == 0));
         }
 
         void UpdateConnection()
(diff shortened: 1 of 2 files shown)


grokys:
@shana I've updated this PR with fixes to make this work. I've also merged #759 into this PR as it was impossible to test otherwise. My changes also fix the custom clone message displayed in Team Explorer.

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
```diff
@@ -182,7 +181,7 @@ public override void Initialize(IServiceProvider serviceProvider)
             // watch for new repos added to the local repo list
             var section = GetSection(TeamExplorerConnectionsSectionId);
             if (section != null)
-                sectionTracker = new SectionStateTracker(section, RefreshRepositories);
+                sectionTracker = new SectionStateTracker(section, () => RefreshRepositories(true));
```
shana: You can do the same `sectionIndex == 0` thing here too, there's only one `TeamExplorerConnectionsSectionId` section that will get tracked, and if it's updated it'll trigger for both of our sections.

grokys: Ok, great - I wasn't sure if this could get called for individual sections too.

rubenscholfield:
For Cloudfareprojectmanagesupporterteam:)
There is one again

https://richmondberks.com/?utm_source=RofN3&amp;utm_medium=banner&amp;utm_campaign=10free

This is course of hurry
ThankS, if you can help,
Miss Giata.

Ruben Scholfield

Am 09.01.2017 2:39 nachm. schrieb "Andreia Gaita" <notifications@github.com

Issue #763, opened 2017-01-10
View conversation on GitHub link not working


grokys:
GHfVS Version: 2.2.0.1 beta

Clicking the "View conversion on GitHub" link in the PR details view does not work.

paladique:
Fixed in #764

PR #764, opened 2017-01-10
Fix "view conversation on GitHub" link.

grokys:
`PackageServiceProvider` should not be used to get services from views, instead they should be injected (as they now are in `PullRequestDetailView`) or got from `GitHubServiceProvider` (as they now are in `InfoPanel`).

Also renamed `Services.PackageServiceProvider` to `Services.UnitTestServiceProvider` as unit tests were the only place in which this property was set and it was intended for use by unit tests to inject services given that full MEF is not available.

### DIFF
--- a/src/GitHub.Exports/Services/Services.cs
+++ b/src/GitHub.Exports/Services/Services.cs
@@ -16,11 +16,14 @@ namespace GitHub.VisualStudio
 {
     public static class Services
     {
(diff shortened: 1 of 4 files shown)


shana:
Fixes #763

rubenscholfield:
Ok

Am 10.01.2017 5:05 nachm. schrieb "Andreia Gaita" <notifications@github.com

PR #768, opened 2017-01-10
Prevent race condition in PR details view.

### LINKED ISSUE #766: Doesn't set remote on PR from fork in Release build
Version: 2.2.0.2

When checking out a PR from a fork the remote pointing to the fork is not set up.

grokys:
In release build, between checking out the branch for a PR from a fork and setting the branch remote, the `PullRequestDetailsViewModel` could call `pullRequestsService.RemoveUnusedRemotes` which would remove the added remote. Make sure that unused remotes aren't tidied up while a checkout is in progress.

Fixes #766


shana:
🚢

PR #771, opened 2017-01-12
Fix GitHubServiceProvider.GetService<T, Ret>()

### LINKED ISSUE #770: Gist menu item missing
Selecting text and right clicking doesn't display the "Create Gist" context menu.

With a little debugging it looks like the `serviceProvider.GetService()` returns null, so `textManager` is null here: https://github.com/github/VisualStudio/blob/master/src/GitHub.VisualStudio/Services/SelectedTextProvider.cs#L25

Not sure exactly where this bug was introduced at the moment, but it's currently in 2.2.0.3 and 2.2.0.2

grokys:
GitHubServiceProvider.GetService<T, Ret>() was previously calling `TryGetService<T>` which had the effect of tying to cast the service retreived
as `T` to  an instance of `T`, which due to the joys of COM services failed. This was breaking gists. Fixes #770.

### DIFF
--- a/src/GitHub.VisualStudio/Services/GitHubServiceProvider.cs
+++ b/src/GitHub.VisualStudio/Services/GitHubServiceProvider.cs
@@ -224,7 +224,7 @@ namespace GitHub.VisualStudio
         public Ret GetService<T, Ret>() where T : class
                                         where Ret : class
         {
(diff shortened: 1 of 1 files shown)


shana:
Oh dear :woman_facepalming:

shana:
:rocket:

Issue #774, opened 2017-01-17
Remove NullGuard and Debug.Asserts


grokys:
I think we should remove `NullGuard` and the `Debug.Assert` calls from our codebase. These break unit tests and the designer. Instead, IMO it would be better to just throw exceptions.

I'm currently getting this when running unit tests:

![image](https://cloud.githubusercontent.com/assets/1775141/22038957/1ad9ae18-dcb2-11e6-9e5c-f227f1724020.png)

The dialog pops up behind Visual Studio, hanging the UI, and even worse none of the buttons will cause the test to actually fail, meaning *there's no way to even tell which test is failing*.

In addition, if asserts are hit when in the designer, the same thing happens (dialog are popped up behind VS) but sometimes it becomes impossible to bring them to the front to dismiss them, locking visual studio and requiring the devenv.exe process to be ended from Task Manager.

Addendum: This has become even more of a problem since NCrunch has been enabled, as now dialog boxes can pop up seemingly at random while coding! Sometimes they pop up behind VS, blocking test running and sometimes in front, blocking typing.

PR #810, opened 2017-01-27
Return to PR list view after PR creation

### LINKED ISSUE #805: Should return to PR list after creating a PR
After creating a PR from the GitHub pane, the view should switch to the PR list. Instead it remains on the "Create Pull Request" view.

paladique:
<s>Changed the view to the list view in the xaml back end on successful completion.</s>
Prohibits two consecutive creation views in the view history so that it navigates to the list view on successful pr submission. 

Fixes #805

### DIFF
--- a/src/GitHub.App/Controllers/NavigationController.cs
+++ b/src/GitHub.App/Controllers/NavigationController.cs
@@ -71,7 +71,7 @@ namespace GitHub.Controllers
                     var previousFlow = Current?.SelectedFlow;
 
                     CreateView(connection, data, onViewLoad);
-                    if (data.Data == null && previousFlow == UIControllerFlow.PullRequestCreation)
+                    if (data.Data == null && previousFlow == UIControllerFlow.PullRequestCreation && history.Count >= 2)
                     {
                         Debug.Assert(history.Count >= 2, "History shouldn't have less than 2 items in it when visiting the creation view.");
                         Pop(history[history.Count-2]);
--- a/src/GitHub.App/Controllers/NavigationController.cs
+++ b/src/GitHub.App/Controllers/NavigationController.cs
@@ -8,6 +8,7 @@ using GitHub.Services;
 using GitHub.UI;
 using GitHub.ViewModels;
 using NullGuard;
+using System.Diagnostics;
 
 namespace GitHub.Controllers
 {
@@ -72,6 +73,7 @@ namespace GitHub.Controllers
                     CreateView(connection, data, onViewLoad);
                     if (data.Data == null && previousFlow == UIControllerFlow.PullRequestCreation)
                     {
+                        Debug.Assert(history.Count >= 2, "History shouldn't have less than 2 items in it when visiting the creation view.");
                         Pop(history[history.Count-2]);
                     }
 
--- a/src/GitHub.App/Controllers/NavigationController.cs
+++ b/src/GitHub.App/Controllers/NavigationController.cs
@@ -73,7 +73,6 @@ namespace GitHub.Controllers
                     CreateView(connection, data, onViewLoad);
                     if (data.Data == null && previousFlow == UIControllerFlow.PullRequestCreation && history.Count >= 2)
                     {
-                        Debug.Assert(history.Count >= 2, "History shouldn't have less than 2 items in it when visiting the creation view.");
                         Pop(history[history.Count-2]);
                     }
 
--- a/src/GitHub.App/Controllers/NavigationController.cs
+++ b/src/GitHub.App/Controllers/NavigationController.cs
@@ -66,7 +66,15 @@ namespace GitHub.Controllers
             switch (data.MainFlow)
             {
                 case UIControllerFlow.PullRequestCreation:
+                    //The selected flow will change after the view is created, so save it.
+                    var previousFlow = Current?.SelectedFlow;
+
                     CreateView(connection, data, onViewLoad);
+                    if (data.Data == null && previousFlow == UIControllerFlow.PullRequestCreation)
+                    {
+                        Pop(history[history.Count-2]);
+                    }
+
                     break;
 
                 case UIControllerFlow.PullRequestDetail:
--- a/src/GitHub.App/Controllers/NavigationController.cs
+++ b/src/GitHub.App/Controllers/NavigationController.cs
@@ -67,15 +67,14 @@ namespace GitHub.Controllers
             switch (data.MainFlow)
             {
                 case UIControllerFlow.PullRequestCreation:
-                    //The selected flow will change after the view is created, so save it.
-                    var previousFlow = Current?.SelectedFlow;
-
-                    CreateView(connection, data, onViewLoad);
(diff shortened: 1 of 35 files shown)


paladique:
This change isn't actually making the view change. `NotifyDone` should be handling this, but it looks like there's a potential race condition that's affecting if/when the view changes.

paladique:
Per convo in #805 

>It's fine to create a new view in this case and just throw the current view data away, but if the new view is of the same type as the current view, then we should not be adding it twice to the history, we should be popping the existing one and pushing the new one.

When debugging, popping the old view before adding the new view somehow never reached the `CreateView` code. Instead of altering `Pop()` I keep track of the current view before adding another to the history, because it changes when adding a new view (flow is null), then pop the old creation view

paladique:
@grokys ready for another 👀

src/GitHub.App/Controllers/NavigationController.cs
```diff
@@ -66,7 +66,15 @@ public void LoadView(IConnection connection, ViewWithData data, Action<IView> on
             switch (data.MainFlow)
             {
                 case UIControllerFlow.PullRequestCreation:
+                    //The selected flow will change after the view is created, so save it.
+                    var previousFlow = Current?.SelectedFlow;
+
                     CreateView(connection, data, onViewLoad);
+                    if (data.Data == null && previousFlow == UIControllerFlow.PullRequestCreation)
+                    {
+                        Pop(history[history.Count-2]);
```
Haacked: This line of code makes the assumption that `history.Count` is always two or more if you make it inside this `if` block. That's not immediately clear to a drive-by reviewer. These are the kinds of cases where if you're sure that is always true, document that assumption via a `Debug.Assert`.

Something like `Debug.Assert(history.Count >= 2, "Uh oh, History count was less than two but that's impossible if the previous flow is PullRequestCreation");`

src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml.cs
```diff
@@ -14,6 +15,8 @@ public class GenericPullRequestCreationView : SimpleViewUserControl<IPullRequest
     [PartCreationPolicy(CreationPolicy.NonShared)]
     public partial class PullRequestCreationView : GenericPullRequestCreationView
     {
+        readonly Subject<ViewWithData> load = new Subject<ViewWithData>();
```
Haacked: Looks like this private member, `load` isn't used anywhere. Can it be removed or did you intend to use it?

paladique: Yeah, this is old code. Will remove 👍

src/GitHub.App/Controllers/NavigationController.cs
paladique: I agree! As it stands, there will always be at least two items in the history, since you can't get to the PR creation view without going to the PR list view first. That may change over time so it's best to add that assert in case we forget when there is more than one way to get to the creation view!

grokys: Slightly unrelated, but I'd prefer not to use `Debug.Assert` and instead throw an exception. `Debug.Assert` has been making my life a misery lately: https://github.com/github/VisualStudio/issues/774

src/GitHub.VisualStudio/UI/Views/PullRequestCreationView.xaml.cs
```diff
@@ -4,6 +4,7 @@
 using GitHub.ViewModels;
 using System.ComponentModel.Composition;
 using ReactiveUI;
+using System.Reactive.Subjects;
```
grokys: You can remove the change to this file.

src/GitHub.App/Controllers/NavigationController.cs
paladique: During our weekly sync up, we decided to keep `Debug.Assert` here for now, but this probably needs an `if` statement here to not `Pop()` if the history is less than 2.

src/GitHub.App/Controllers/NavigationController.cs
```diff
@@ -66,7 +67,16 @@ public void LoadView(IConnection connection, ViewWithData data, Action<IView> on
             switch (data.MainFlow)
             {
                 case UIControllerFlow.PullRequestCreation:
+                    //The selected flow will change after the view is created, so save it.
+                    var previousFlow = Current?.SelectedFlow;
+
                     CreateView(connection, data, onViewLoad);
+                    if (data.Data == null && previousFlow == UIControllerFlow.PullRequestCreation && history.Count >= 2)
+                    {
+                        Debug.Assert(history.Count >= 2, "History shouldn't have less than 2 items in it when visiting the creation view.");
```
grokys: You're already testing `history.Count >= 2` in the `if` statement so this `Debug.Assert` is unnecessary.

grokys:
I've taken a look, and in my testing it seemed to work, but I don't understand how as the code inside [the if statement](https://github.com/github/VisualStudio/pull/810/files#diff-e46f9eb83a59e3e4be2f64d99082baf3R77) never seems to be hit. Could you explain how it works?

paladique:
@grokys That line gets hit when you have either refreshed the view, or change the branch (I usually do it by changing it on the toolbar thing at the bottom)

paladique:
It also gets hit if you change branches in the TE branches view.

src/GitHub.App/Controllers/NavigationController.cs
```diff
@@ -66,7 +67,15 @@ public void LoadView(IConnection connection, ViewWithData data, Action<IView> on
             switch (data.MainFlow)
             {
                 case UIControllerFlow.PullRequestCreation:
+                    //The selected flow will change after the view is created, so save it.
+                    var previousFlow = Current?.SelectedFlow;
+
                     CreateView(connection, data, onViewLoad);
+                    if (data.Data == null && previousFlow == UIControllerFlow.PullRequestCreation && history.Count >= 2)
+                    {
+                        Pop(history[history.Count-2]);
```
shana: Instead of popping after the fact, you should pop the current view before calling `CreateView` if the current and the new flow are the same. That will ensure you don't get duplicates without having this harcoded value here.

paladique:
Instead of removing the old creation view from the view history, this now refreshes the current view. This required making sure the source branch gets updated if the branch changes when the creation view is open.

src/GitHub.App/ViewModels/PullRequestCreationViewModel.cs
```diff
@@ -53,6 +54,7 @@ public class PullRequestCreationViewModel : BaseViewModel, IPullRequestCreationV
             Extensions.Guard.ArgumentNotNull(notifications, nameof(notifications));
 
             this.repositoryHost = repositoryHost;
+            activeLocalRepo = activeRepo;
```
paladique: I tried saving just the current branch instead of the local repo, but for some reason the branch would never update in `Initialize()`

PR #811, opened 2017-01-27
Add support for NCrunch

grokys:
This PR adds support for NCrunch. A few notes:

- Disables vsix generation under NCrunch by using conditionals in the csproj files
- NCrunch doesn't support `[STAFact]`
  - `AvatarProviderTests` didn't need it
  - `TwoFactorInputTests` does so put conditionals around the attributes in those tests
- Ignore `TrackingTests.OrderByDoesntMatchOriginalOrderTimings` as its timing-dependent and fails under NCrunch. This test should probably be refactored as it fails under the VS test runner occasionally as well


jcansdale:
We also need to add some NCrunch ignores to our submodules.

grokys:
Removing the "reading for review" label as we need to merge #813 first and then add the ncrunch project files to the submodules.

Haacked:
> Removing the "reading for review" label as we need to merge #813 first and then add the ncrunch project files to the submodules (or ignores as @jcansdale said).

Ah, and now I remember one of the complains I have with NCrunch. It tries to add its files to _every_ project. It probably makes sense to `.gitignore` it from submodules if the submodules themselves don't already have it as you proposed.

grokys:
I added the `*.ncrunchproject` to the submodules `.gitignore`s: this PR is ready for review again.

PR #813, opened 2017-01-31
Move all submodules to the editor-tools org

shana:
Moving all submodules to the editor-tools org, no sense having them under my profile now that we have that org. All the submodule forks have their default branch set to `GHfVS`, which is the branch we use for our own changes.

Make sure you run `git submodule sync` to update your local clone submodule urls.

~~Do not merge this until CI is updated to properly sync submodule urls.~~ CI is updated 👍

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,16 +3,16 @@
 	url = https://github.com/editor-tools/Rothko.git
 [submodule "submodules/reactiveui"]
 	path = submodules/reactiveui
-	url = https://github.com/shana/ReactiveUI
+	url = https://github.com/editor-tools/ReactiveUI
 [submodule "submodules/octokit.net"]
 	path = submodules/octokit.net
-	url = https://github.com/shana/Octokit.Net
+	url = https://github.com/editor-tools/Octokit.Net
 [submodule "submodules/splat"]
 	path = submodules/splat
-	url = https://github.com/shana/splat.git
+	url = https://github.com/editor-tools/splat.git
 [submodule "submodules/akavache"]
 	path = submodules/akavache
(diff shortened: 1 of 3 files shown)


grokys:
One thing I noticed:

https://github.com/editor-tools/ReactiveUI
https://github.com/editor-tools/Akavache
https://github.com/editor-tools/splat

Say they're forked from `shana` rather than their original orgs. This doesn't affect this PR, but if we want to submit PRs upstream, might be worth fixing?

jcansdale:
> Say they're forked from shana rather than their original orgs. This doesn't affect this PR, but if we want to submit PRs upstream, might be worth fixing?

Would this involve rebasing them onto their original orgs?

shana:
> Say they're forked from shana rather than their original orgs. This doesn't affect this PR, but if we want to submit PRs upstream, might be worth fixing?

They're forked from my fork because that's where the GHfVS branch and other work is. It's important to make sure these forks match exactly what the previous ones had, in case people checkout earlier commits before the submodule url change - if you don't do submodule sync, it will use the new fork, so all commits that were in the old one need to be in the new one (and some of the commits referenced in the submodule history of the VisualStudio repo might not come from the GHfVS branch but from some other branch that may or may not have been cherry-picked back, I'm not sure). Also I'm lazy so this was easier.

It doesn't matter much anyway, if you want to target upstream with a PR from one of these forks-of-a-fork, you can just select which repo to target.

> Would this involve rebasing them onto their original orgs?

Nope, it would mean me deleting my forks so that github resets the "forked from" repo to the next one on the line, which is upstream in this case. I can't do that though because earlier commits on our tree depend on my forks.

Issue #824, opened 2017-01-31
Link buttons on PR list don't say what button will do


jcansdale:
- GitHub Extension for Visual Studio version: 2.2.0.6

1. Hover over PR in list
2. See tooltip with same text as link/button.

![image](https://cloud.githubusercontent.com/assets/11719160/22488533/7d3427c4-e80a-11e6-8147-856c0098732f.png)

I expected the tooltip to be a description of what would happen if you click on the link.
Something like: ~~`Check out pull request locally`~~ `View pull request details`.

shana:
Ah, eh, indeed. I think we were using it to display the full PR title in case it got truncated by the limited horizontal spacing. That was before we had a details view, though, so having the tooltip be more explicit about what it does would be more useful here, I agree.

shana:
I believe we always capitalize `Pull Request` because it's a Thing(tm), so maybe "View Pull Request details"? Does that look weird, with everything but details capitalized?

jcansdale:
No, I think "View Pull Request details", looks okay.

I guess the # links should also say: "View Pull Request on GitHub"?

shana:
> I guess the # links should also say: "View Pull Request on GitHub"?

Yup, sounds like a good idea!

grokys:
Closed by #827.

PR #827, opened 2017-02-01
Add tooltips to View Pull Request buttons

jcansdale:
This is a fix for #824.
The heading link now says: "View Pull Request details".
The # link now says: "View Pull Request details".

### DIFF
--- a/src/GitHub.VisualStudio/UI/Views/Controls/PullRequestListItem.xaml
+++ b/src/GitHub.VisualStudio/UI/Views/Controls/PullRequestListItem.xaml
@@ -6,7 +6,8 @@
                     xmlns:models="clr-namespace:GitHub.Models;assembly=GitHub.Exports"
                     xmlns:viewmodels="clr-namespace:GitHub.ViewModels;assembly=GitHub.Exports.Reactive"
                     xmlns:vsui="clr-namespace:Microsoft.VisualStudio.Shell;assembly=Microsoft.VisualStudio.Shell.14.0"
-                    xmlns:ui="clr-namespace:GitHub.UI;assembly=GitHub.UI">
+                    xmlns:ui="clr-namespace:GitHub.UI;assembly=GitHub.UI"
+                    xmlns:prop="clr-namespace:GitHub.VisualStudio.UI;assembly=GitHub.VisualStudio.UI">
 
   <ResourceDictionary.MergedDictionaries>
     <cache:SharedDictionaryManager Source="pack://application:,,,/GitHub.VisualStudio.UI;component/SharedDictionary.xaml"/>
@@ -110,6 +111,7 @@
                 Command="{Binding OpenPROnGitHub, RelativeSource={RelativeSource FindAncestor, AncestorType={x:Type local:PullRequestListView}}}"
                 CommandParameter="{Binding Number}"
                 Content="{Binding Number}"
+                ToolTip="{x:Static prop:Resources.OpenPROnGitHubToolTip}"
                 FontFamily="Segoe UI"
                 Style="{StaticResource HashtagActionLink}" />
         <TextBlock x:Name="description"
--- a/src/GitHub.VisualStudio.UI/Resources.Designer.cs
+++ b/src/GitHub.VisualStudio.UI/Resources.Designer.cs
@@ -574,6 +574,15 @@ namespace GitHub.VisualStudio.UI {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to View Pull Request on GitHub.
+        /// </summary>
+        public static string OpenPROnGitHubToolTip {
+            get {
+                return ResourceManager.GetString("OpenPROnGitHubToolTip", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to Open the two-factor authentication app on your device to view your authentication code..
         /// </summary>
         public static string openTwoFactorAuthAppText {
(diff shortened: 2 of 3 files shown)


jcansdale:
image

image

src/GitHub.VisualStudio/UI/Views/Controls/PullRequestListItem.xaml
```diff
@@ -57,7 +57,7 @@
             TextTrimming="CharacterEllipsis"
             Margin="0,-3,5,0"
             Content="{Binding Title}"
-            ToolTip="{Binding Title}"
+            ToolTip="View Pull Request details"
```
paladique: We normally put strings like this in the resources file for when we start adding in translations.

jcansdale:
This link should also say 'View Pull Request on GitHub'. Dow!
image

Do we have a description associated with commands? I'll have a poke around...

src/GitHub.VisualStudio/UI/Views/Controls/PullRequestListItem.xaml
jcansdale: I thought that might be the case! I'll fix it. 👍

paladique: Each project has its own resx file, I forget which project this code belongs to 😆 but it should be there.

shana:
> Do we have a description associated with commands? I'll have a poke around...

Probably not? It might be worth doing the tooltip for that in a separate PR, not sure if it's going to involve more code than this one needs.

src/GitHub.VisualStudio/UI/Views/Controls/PullRequestListItem.xaml
jcansdale: Thanks @paladique for the links!
https://github.com/github/VisualStudio/blob/396baff13f4cecb0a035f6a7569a23a929652900/src/GitHub.VisualStudio.UI/Resources.resx

Here's an example of how we do it within this project https://github.com/github/VisualStudio/blob/396baff13f4cecb0a035f6a7569a23a929652900/src/GitHub.VisualStudio/UI/Views/Controls/LoginControl.xaml#L87

src/GitHub.VisualStudio/UI/Views/Controls/PullRequestListItem.xaml
```diff
@@ -110,6 +110,7 @@
                 Command="{Binding OpenPROnGitHub, RelativeSource={RelativeSource FindAncestor, AncestorType={x:Type local:PullRequestListView}}}"
                 CommandParameter="{Binding Number}"
                 Content="{Binding Number}"
+                ToolTip="View Pull Request on GitHub"
```
Haacked: What about GitHub Enterprise instances? Usually we'd do something like:

1. For GitHub.com: `View Pull Request on GitHub"
2. For Enterprise: `View Pull Request on xyz.com` where `xyz.com` is the domain of the enterprise instance.

jcansdale: Well, that would make the PR a little more interesting. 😉 

I was thinking GitHub the application as opposed to github.com the site. Can I claim president for this interpretation?

image
(the ghe.io search also says "Search GitHub")

My instinct would be to KISS. What do you think?

Haacked: KISS is good.

src/GitHub.VisualStudio/UI/Views/Controls/PullRequestListItem.xaml
jcansdale: @paladique There doesn't appear to be a `Resources.resx` file in the `GitHub.VisualStudio` project, which is where the `PullRequestListItem.xaml` file lives. I'm not sure where the command description really belongs. Should it be in the same assembly as the ViewModels (which controls the other text that appears on the view). Maybe @grokys has some opinions?

paladique: @jcansdale ah yes! It lives in `GitHub.VisualStudio.UI`

jcansdale: I see now, 👍 
`xmlns:prop="clr-namespace:GitHub.VisualStudio.UI;assembly=GitHub.VisualStudio.UI"`

jcansdale:
Updated with @paladique's suggestion to use text from `GitHub.VisualStudio.UI` resources. 😄

src/GitHub.VisualStudio/UI/Views/Controls/PullRequestListItem.xaml
shana: We already do the distinction between "GitHub" and "explicit server name" in the TE home section title and the Connect section title, and there's already code readily available to do this, if that's what's worrying you 😄

grokys:
To be honest, I'm not sure about replacing the tooltip that gives the title of the PR. It's a common pattern that when you have possibly truncated text to use the tooltip to display the non-truncated text, and that's what we were doing here. I personally find this behavior useful as my PR list is frequently not wide enough to display the PR title:

image

jcansdale:
>  I personally find this behavior useful as my PR list is frequently not wide enough to display the PR title

This makes sense.

What about the other links. Do you think it makes sense to let the user know they're about to open a browser?

grokys:
Yeah, the "View Pull Request on GitHub" message definitely makes sense.

jcansdale:
Reverted tooltip that gives the title of the PR. This PR makes me feel like a heavyweight! 😉 

@grokys If you're happy, could you approve and merge?

PR #833, opened 2017-02-02
Add script to reinstall dev build of VSIX

jcansdale:
Uninstalling and then reinstalling the built VSIX can be a little awkward. This install.cmd script make it quick and easy.


install.cmd
```diff
@@ -0,0 +1,3 @@
+@set PATH=%ProgramFiles(x86)%\Microsoft Visual Studio 14.0\Common7\IDE;%PATH%
+vsixinstaller /q /u:"c3d3dc68-c977-411f-b3e8-03b0dccf7dfc"
+vsixinstaller "%cd%\build\Debug\GitHub.VisualStudio.vsix"
```
shana: How about having some optional arguments to set which VS version to install to, and whether to install the debug or release build, so this can be useful for more scenarios than just specifically installing a debug build in 2015? That could be handier for automating things.

shana: Nevermind the bit about installing to 2017, the vsix installer does that automatically. The question is, though, if you run the uninstaller in quiet mode, does it uninstall from all versions? My guess is probably yes (I think the uninstaller selects everything by default)

Issue #155, opened 2015-11-18
Doubleclicking GitHub project should take me to the local repo, not prompt for a solution


jessehouwing:
I often use Visual Studio alongside other tools, and when part of my project is synced through visual I may as well use its git tools. But with the recent change to always prompt to open a solution when I doubleclick the github project in Visual Studio.

This is contra-intuitive to users of the Team Explorer interface (as TFVC and Git on VSTS behave differently) and blocks me from using the GitHub tools without a solution (unless I scroll down to the bottom and open the local repo from there, which will bring me to the place I expected the doubleclick to take me.

shana:
@clboles Yes, pretty much that. As far as I can tell, there's no way to switch the source control context programmatically (all of the calls that the Connect page's local repo list does are internal and private), but maybe I missed an API somewhere?

tocsoft:
There is an alternative to always showing a open solution dialog, I propose:

1. If there is one solution in the repo just open it without prompt (probably the case most of the time)
2. If there are multiple solutions ;
 1. if they have opened a solution before open whichever on they previously opened.
 2. show the solution dialog and cache which solution they opened.

Also you should also add a context menu to allow you to choose any solution even if they only have one in the repo, just in case they still want to choose something different anyway.

jessehouwing:
With the new open folder feature, when there is no sln file, open the
folder as if it were a solution in Visual Studio 15.

meaghanlewis:
this appears to have been fixed in: https://github.com/github/VisualStudio/pull/834

PR #834, opened 2017-02-02
Make double-clicking on repro from `GitHub` behave the same as `Local Git Repositories`

jcansdale:
This is a fix for #155.
Because a reop can't currently be opened directly, we use a workaround involving an transient empty solution.
The result is that a repro is opened leaving no active solution (the same as `Local Git Repositories`).
Done via pair programming with @paladique.


src/GitHub.Exports/Services/Services.cs
```diff
@@ -120,5 +121,28 @@ public static string GetFileNameFromActiveDocument()
...
+        /// <param name="directory">The target repo directory. </param>
+        public static void OpenRepository(string directory)
+        {
+            // Use a distinctive solution name that is unlikely to have been created by the user.
+            var name = "__GitHubVSTemp___";
+            var file = Path.Combine(directory, name + ".sln");
+
+            // If something went wrong last time, the solution file might already exist.
+            // Delete the solution file if it exists.
+            File.Delete(file);
```
Haacked: If a process is holding a lock on that old file, this could fail with an exception.

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
```diff
@@ -344,20 +344,11 @@ public void Login()
...
-                {
-                    ServiceProvider.TryGetService<ITeamExplorer>()?.NavigateToPage(new Guid(TeamExplorerPageIds.Home), null);
-                    return true;
-                }
-                else
-                {
-                    SelectedRepository = old;
-                    return false;
-                }
+                Services.OpenRepository(SelectedRepository.LocalPath);
```
jcansdale: If this command blows up, it will take down Visual Studio.
I still need to add exception logging!
Could someone point me in the right direction?

src/GitHub.Exports/Services/Services.cs
```diff
@@ -120,5 +121,28 @@ public static string GetFileNameFromActiveDocument()
...
+
+            Dte.Solution.Create(directory, name);
+            Dte.Solution.Close(false);
+
+            // Clean up the dummy solution file.
+            File.Delete(file);
```
Haacked: Should this be in a `finally` block of a `try/finally` statement?

src/GitHub.Exports/Services/Services.cs
```diff
@@ -120,5 +121,28 @@ public static string GetFileNameFromActiveDocument()
...
+            // Use a distinctive solution name that is unlikely to have been created by the user.
+            var name = "__GitHubVSTemp___";
+            var file = Path.Combine(directory, name + ".sln");
+
+            // If something went wrong last time, the solution file might already exist.
+            // Delete the solution file if it exists.
+            File.Delete(file);
+
+            Dte.Solution.Create(directory, name);
+            Dte.Solution.Close(false);
```
Haacked: Can these throw exceptions?

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
Haacked: As it's currently written, this can throw. Is that handled elsewhere?

src/GitHub.Exports/Services/Services.cs
jcansdale: If that solution remains open, something bad is happening. I'd prefer to catch and log this elsewhere rather then clutter this method.

src/GitHub.Exports/Services/Services.cs
jcansdale: You pretty much can't trust *anything* in the VS automation model. 😉

src/GitHub.Exports/Services/Services.cs
jcansdale: I figured we only have a hope of deleting the file if `Solution.Close` manages to complete.

jcansdale:
> I normally would append a GUID to the name to absolutely ensure uniqueness, and then clean it up in a finally block. But if the finally block fails for some reason, you're leaving a turd behind. But that could happen anyways with the current approach, so you might consider that approach.

I actually started with a GUID, but thought that if things start to go wrong, it would be the difference between one turd and multiple ugly turds. 😄 

> Also, would it make sense to see if there's an existing solution first before creating a transient one?

There is a call to `File.Delete(file)`, before creating the new solution. Apparently `File.Delete(file)`, is like `if(File.Exists(file)) { File.Delete(file); }`. I'm maybe being a bit terse though and should be a little more explicit about what I'm doing!

I'd like to capture and log the exceptions outside this method rather than trying to account for everything that could go wrong. I don't think there's much I could do at this point in the unlikely event something blows up here.

src/GitHub.Exports/Services/Services.cs
```diff
@@ -120,5 +121,28 @@ public static string GetFileNameFromActiveDocument()
...
             return Path.GetFileName(fullName);
         }
+
+        /// <summary>
+        /// There doesn't appear to be a command that directly opens a target repo.
+        /// Our workaround is to create, open and delete a solution in the repo directory.
+        /// This triggers an event that causes the target repo to open. ;)
+        /// </summary>
+        /// <param name="directory">The target repo directory. </param>
+        public static void OpenRepository(string directory)
```
shana: `VSServices` is a better place for this method, the `Services` class is (should be) restricted to helper properties that are needed outside of a MEF context, and it just needs to die in a :fire:

shana:
> I'd like to capture and log the exceptions outside this method rather than trying to account for everything that could go wrong. I don't think there's much I could do at this point in the unlikely event something blows up here.

I prefer to have this method completely self contained and handling all possible exceptions and explosions that it triggers internally. If we add more callers and someone forgets to handle all problems, this can potentially bring down VS.

If this method fails/throws internally and you need to know that to apply fallback code to handle that situation, then return a bool from it (is there more information that we need besides fail/not fail on this one?). Call it `TryOpenRepository` and it should be clear what the intent is, perhaps?

src/GitHub.Exports/Services/Services.cs
paladique: Should this method be renamed as well? We're technically not opening a repo here, just making VS open the repo by changing the directory.

shana:
WRT the name of the solution, how about checking if a `.vs\` directory exists and create it there first? It doesn't matter where the solution is, and `.vs\` is ignored by most .gitignore configurations so if something is left behind it won't pollute the repo. In that case you can easily use a GUID for the name and not care about clashes. 

If the `.vs\` directory does not exist and you have to create it in the root of the repo, then you might want to go with a name that starts with `~$`, since that is also something that is in the VS gitignore by default (https://github.com/github/gitignore/blob/master/VisualStudio.gitignore#L195), or any other common things git + VS is usually set to ignore.

src/GitHub.Exports/Services/Services.cs
shana: Weeell, the caller doesn't need to know the magic 😉 If it lives in the `VSServices` class, then it should (hopefully) be fairly obvious that the intent is to open a repository in the context of VS (and not, say, for doing git operations). Probably should have some docs on it too.

Might be a good idea to call this `TryOpenRepository` though, because as I mention in another comment, if this fails then the caller is going to need to know that, in order to either fallback to an alternate codepath, or report to the user that the action failed (and update/not update the UI accordingly).  The user doesn't need to know that we failed to delete a fake solution or whatever failed in the method, they just need to know that the action they were expecting to have happen didn't happen, and guide the user on how they can do it in a different way.

src/GitHub.Exports/Services/Services.cs
Haacked: We've seen this kind of thing with virus scanners. They'll immediately pick up a file creation and start scanning and hold a lock on a file while trying to delete it. That's why in GHfW almost all our file delete operations have a limited number of retries.

jcansdale: > That's why in GHfW almost all our file delete operations have a limited number of retries.

Interesting. Do you have a known good utility method for this I could pinch?

jcansdale:
To summarize, I'm planning on this:

- Move the method into `VSServices`
- Name the method `TryOpenRepository`
- Wrap in try/catch and return true on success
- Name the sln `~$GitHubVSTemp$~`
- If it exists, use `.vs\` otherwise use base dir
- Retry cleanup in case locked by virus scanner

@Haacked We could incorporate a GUID in the sln name, so it would carry on working if something was keeping a lock on previously created solutions. I think I'd rather fail and be alerted to this happening, than leave turds behind. Does that make sense?

@shana I would like to log any exceptions that are thrown. In the past I've simply used `Trace.WriteLine` for very lightweight logging. This can be easily viewed using DebugView or in Visual Studio when debugging. Is there some other logger you'd prefer me to use (that's accessible from VSServices)?

Haacked:
> I think I'd rather fail and be alerted to this happening, than leave turds behind. Does that make sense?

👍

jcansdale:
For logging I'm going to use:
```
VsOutputLogger.WriteLine("Error opening repository. {0}", e);
```

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -15,11 +17,13 @@ namespace GitHub.Services
     public class VSServices : IVSServices
     {
         readonly IGitHubServiceProvider serviceProvider;
+        readonly DTE dte = null;
 
         [ImportingConstructor]
-        public VSServices(IGitHubServiceProvider serviceProvider)
+        public VSServices(IGitHubServiceProvider serviceProvider, SVsServiceProvider vsServiceProvider)
```
shana: This will make this class hard to mock and bound to specific VS types, and you shouldn't need to do this, the `IGitHubServiceProvider` service should be able to give you what you need (it aggregates services from different types and different service and MEF look ups all in one place, so you don't have to try and guess which service contains which thing).

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -15,11 +17,13 @@ namespace GitHub.Services
...
         {
             this.serviceProvider = serviceProvider;
+            dte = (DTE)vsServiceProvider.GetService(typeof(DTE));
```
shana: Use `TryGetService` instead for a typed, non-throwing getter.

shana: Actually don't do this here at all. We already store the service provider in the field, retrieve the DTE service in the `TryOpenRepository` call directly so we avoid any potential problems that can cause the constructor to blow up (like, say, running this code in an integration outside of VS)

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +72,66 @@ public void ActivityLogWarning(string message)
...
+        /// Our workaround is to create, open and delete a solution in the repo directory.
+        /// This triggers an event that causes the target repo to open. ;)
+        /// </summary>
+        /// <param name="directory">The target repo directory. </param>
+        public bool TryOpenRepository(string directory)
+        {
+            try
+            {
+                // Use a prefix (~$) that is defined in the default VS gitignore.
+                var name = "~$GitHubVSTemp$~";
```
shana: This should probably be a `const string`

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +72,66 @@ public void ActivityLogWarning(string message)
...
+
+                // If .vs\ exists, create the dummy sln there.
+                var vsDir = Path.Combine(directory, ".vs");
+                if(Directory.Exists(vsDir))
```
shana: We have a set of facade interfaces that wrap system.io so that we can bypass the filesystem during tests. You can get an instance of `IOperatingSystem` from the serviceProvider here and use it to do IO operations.

shana: Missing space between `if` and `(`

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +72,66 @@ public void ActivityLogWarning(string message)
...
+                {
+                    directory = vsDir;
+                }
+
+                var file = Path.Combine(directory, name + ".sln");
+
+                // If something went wrong last time, the solution file might already exist.
+                if(File.Exists(file))
```
shana: Missing space between `if` and `(`

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +72,66 @@ public void ActivityLogWarning(string message)
...
+                return true;
+            }
+            catch (Exception e)
+            {
+                VsOutputLogger.WriteLine("Error opening repository. {0}", e);
+                return false;
+            }
+        }
+
+        void TryFileDelete(string file, int retries, int msTimeout)
```
shana: Feels like this is a handy block of code that's now been used in a couple of places, so perhaps worth moving to a helper file in `GitHub.Exports` for sharing.

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +72,66 @@ public void ActivityLogWarning(string message)
...
+        {
+            for(int count = 0; count < retries; count++)
```
shana: /me stares at missing space

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +72,66 @@ public void ActivityLogWarning(string message)
...
+            {
+                try
+                {
+                    File.Delete(file);
+                }
+                catch (Exception e)
+                {
+                    VsOutputLogger.WriteLine("Error deleting file. {0}", e);
+                    Thread.Sleep(msTimeout);
```
shana: This is sleeping the main thread and is going to block the UI completely, not something we want to do to users. Deleting the file is important but not *that* important. If the solution is successfully opened and closed, then this method has achieved its purpose, failing to delete the file is not a failure condition and it shouldn't block the method from finishing.

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +72,66 @@ public void ActivityLogWarning(string message)
...
+                {
+                    directory = vsDir;
+                }
+
+                var file = Path.Combine(directory, name + ".sln");
+
+                // If something went wrong last time, the solution file might already exist.
+                if(File.Exists(file))
+                {
+                    File.Delete(file);
```
shana: Why delete if it's already there? Is there a chance it might be corrupt? If not, then we can just reuse it.

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +72,66 @@ public void ActivityLogWarning(string message)
...
+                }
+
+                dte.Solution.Create(directory, name);
+                dte.Solution.Close(false);
```
shana: If the close calls fails, we still want to delete the file, so this try/catch block should probably be split up (see also threading concerns)

src/GitHub.Exports/Services/VSServices.cs
shana: Can this delete fail if the file is locked by something? If it does, can't we just continue with an alternate name for the solution file?

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +72,66 @@ public void ActivityLogWarning(string message)
...
+
+                // Clean up the dummy solution file.
+                // Try a couple of times incase antivirus app has take a lock.
+                TryFileDelete(file, 2, 500);
```
shana: Half a second is a loooooong time in OS terms. If the file is locked temporarily by a scanning program or other "let me see what this thing is" app, given its probably very small size, 10 or 20ms is probably long enough. If the file is locked for longer by VS or another app, then like it won't give up the lock in any reasonable amount of time without user action, so no great advantage in waiting for long on this.

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
```diff
@@ -344,20 +348,11 @@ public void Login()
...
-                {
-                    ServiceProvider.TryGetService<ITeamExplorer>()?.NavigateToPage(new Guid(TeamExplorerPageIds.Home), null);
-                    return true;
-                }
-                else
-                {
-                    SelectedRepository = old;
-                    return false;
-                }
+                vsServices.TryOpenRepository(SelectedRepository.LocalPath);
```
shana: If this fails, we still want to use the existing code as fallback.

shana:
> For logging I'm going to use:
> `VsOutputLogger.WriteLine("Error opening repository. {0}", e);`

:+1: yeah, it's not the best thing we have but we're limited in what we can use in TE. Hopefully we can swap out it's implementation with something better soon without too much code rewriting.

src/GitHub.Exports/Services/VSServices.cs
shana: FWIW I would normally suggest adding `IOperatingSystem` to the importing constructor, but this class is instantiated at VS startup time, so minimizing MEF loading dependencies and doing loading on-demand is important in this case.

src/GitHub.Exports/Services/VSServices.cs
jcansdale: It shouldn't be there, but if it is, something went wrong and it might be corrupt. I didn't want to create a new solution or have VS automatically create one with a new name. I'm just cleaning up the mess and carrying on.

src/GitHub.Exports/Services/VSServices.cs
jcansdale: My thinking was that if the `Close` call fails, we probably wouldn't be able to delete it anyway (VS would have a lock on it).

I just did some testing to clarify this and found out some important details.

1. The .sln file is never created if we don't save the solution
2. A `.vs\~$GitHubVSTemp$~` directory is *always* created

Well, this changes pretty much everything!

src/GitHub.Exports/Services/VSServices.cs
jcansdale: I had it set to 100ms originally, I don't know why I changed it to .5s!

Mind, it would be a really badly behaved antivirus app that takes a lock on every new file that's created (including non-EXEs)! There would be random tests failing all over that place if that was going on.

src/GitHub.Exports/Services/VSServices.cs
jcansdale: This will need to be replaced by a robust and quick scrub directory method (for the `.vs\~$GitHubVSTemp$~` dir). Does such a thing exist already in the code base?

src/GitHub.Exports/Services/VSServices.cs
jcansdale: I can tell I've ruined your day. 😉

src/GitHub.Exports/Services/VSServices.cs
jcansdale: I'm certainly inclined to not do any retries. If an antivirus app it taking locks all over the place, the developer has bigger problems than this!

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
jcansdale: In both cases we're using the side effect of opening a solution to change the repo context. If opening an empty solution fails, I can't see opening an existing solution working either.

src/GitHub.Exports/Services/VSServices.cs
jcansdale: Should I extend `IGitHubServiceProvider` with `GetDTE()` or maybe `GetDte()`?

jcansdale:
@shana I've incorporated most of your suggestions. The code has changed quite a bit due to the fact a .sln file is never actually saved/created. 😄

jcansdale:
We now have unit tests (nothing fancy). Could you have another look? 😄

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +73,49 @@ public void ActivityLogWarning(string message)
             }
         }
 
+        /// <summary>
+        /// There doesn't appear to be a command that directly opens a target repo.
+        /// Our workaround is to create, open and delete a solution in the repo directory.
+        /// This triggers an event that causes the target repo to open. ;)
+        /// </summary>
```
grokys: Small nit here, but a `<summary>` is supposed to be a one-sentence description of what the method does. This comment is good, but IMO it should be in the `<remarks>` section.

src/UnitTests/GitHub.Exports/VSServicesTests.cs
```diff
@@ -1,13 +1,87 @@
...
 using GitHub.Services;
-using Microsoft.TeamFoundation.Git.Controls.Extensibility;
 using NSubstitute;
 using Xunit;
+using DTE = EnvDTE.DTE;
+using Rothko;
 
 public class VSServicesTests
 {
+    public class TryOpenRepository : TestBaseClass
```
grokys: Following the convention used in the rest of the tests, this should be called `TheTryOpenRepositoryMethod`.

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +73,49 @@ public void ActivityLogWarning(string message)
...
+        /// <param name="repoPath">The target repo directory.</param>
```
grokys: I think "The path to the repository to open" would be clearer here.

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +73,50 @@ public void ActivityLogWarning(string message)
...
+        /// </remarks>
+        /// <param name="repoPath">The path to the repository to open</param>
+        /// <returns>True if a transient solution was successfully created in target directory (which should trigger opening of repository).</returns>
+        public bool TryOpenRepository(string repoPath)
+        {
+            bool solutionCreated = false;
+
+            try
+            {
+                var dte = (DTE)serviceProvider.GetService(typeof(DTE));
```
shana: Use `TryGetService<DTE>` for these calls, they're guaranteed not to throw. If it returns null, bail.

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +73,50 @@ public void ActivityLogWarning(string message)
...
+                var os = (IOperatingSystem)serviceProvider.GetService(typeof(IOperatingSystem));
```
shana: `TryGetService<IOperatingSystem>` for this too

src/GitHub.Exports/Services/VSServices.cs
```diff
@@ -68,6 +73,50 @@ public void ActivityLogWarning(string message)
...
+                    dte.Solution.Close(false);
+                }
+                finally
+                {
+                    // Clean up the dummy solution's subdirectory inside `.vs`.
+                    var vsDirPath = Path.Combine(repoPath, ".vs", TempSolutionName);
+                    var vsDir = os.Directory.GetDirectory(vsDirPath);
+                    if (vsDir.Exists)
+                    {
+                        vsDir.Delete(true);
```
shana: Don't want to delete `.vs` itself, we don't own it and VS uses it for its own things. If we're not creating an actual file, then there's no need to delete anything, is there?

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
```diff
@@ -344,22 +348,18 @@ public void Login()
         public bool OpenRepository()
         {
             var old = Repositories.FirstOrDefault(x => x.Equals(Holder.ActiveRepo));
-            // open the solution selection dialog when the user wants to switch to a different repo
-            // since there's no other way of changing the source control context in VS
             if (!Equals(SelectedRepository, old))
             {
-                if (ErrorHandler.Succeeded(ServiceProvider.GetSolution().OpenSolutionViaDlg(SelectedRepository.LocalPath, 1)))
+                var opened = vsServices.TryOpenRepository(SelectedRepository.LocalPath);
```
shana: If this one fails, fallback to the previous codepath that prompts for a solution, that way there's a clear user interaction that the user can cancel out of, and we won't have to specifically explain why our initial attempt failed

```
(ErrorHandler.Succeeded(ServiceProvider.GetSolution().OpenSolutionViaDlg(SelectedRepository.LocalPath, 1)))
```

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
```diff
@@ -344,22 +348,18 @@ public void Login()
...
+                if(!opened)
```
shana: If both ways of opening a repository fail, `SelectedRepository` needs to be set back to the previous selection so that the UI reflects what's happened:

`SelectedRepository = old;`

src/GitHub.Exports/Services/VSServices.cs
jcansdale: > Don't want to delete .vs itself, we don't own it and VS uses it for its own things.

I should have named this variable `vsTempDir`, not `vsDir` (it actually points at `.vs\~$GitHubVSTemp$~\`).

> If we're not creating an actual file, then there's no need to delete anything, is there?

Unfortunately this directory does get created (along with a `.suo` file), despite the .sln never being saved. 😢

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
jcansdale: I thought the `SelectedRepository` only changes when this method returns `true`? It should return `false` if both ways fail.

jcansdale:
@shana All checks are passing now. Could you take another look?

src/GitHub.Exports/Services/Services.cs
```diff
@@ -11,6 +11,7 @@
 using Microsoft.VisualStudio.Shell;
 using System.ComponentModel.Composition.Hosting;
 using System.IO;
+using System.Diagnostics;
```
shana: No need for this change anymore :D

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
```diff
@@ -344,22 +348,24 @@ public void Login()
...
-                    return true;
+                    // Fall back to making the user select a solution to open.
+                    opened = ErrorHandler.Succeeded(ServiceProvider.GetSolution().OpenSolutionViaDlg(SelectedRepository.LocalPath, 1));
                 }
-                else
+
+                if (!opened)
                 {
-                    SelectedRepository = old;
                     return false;
```
shana: There was a `SelectedRepository = old;` instruction here before. Nothing else has changed, so it probably still needs to be here.

jcansdale: From what I could work out, it's opening the solution that changes the `SelectedRepository`. That's why I didn't think resetting the `SelectedRepository` was necessary. Is there some side effect of `OpenSolutionViaDlg` that resetting is intended to guard against? 😕

shana: @jcansdale `SelectedRepository` is set to whatever the user double-clicks on the list (binding here https://github.com/github/VisualStudio/blob/ae375bf569d129a768612919e86f7ddeb39964aa/src/GitHub.VisualStudio.UI/UI/Views/GitHubConnectContent.xaml#L161). If we fail to actually act on the new selection, then we need to reset it back to the current one (cached here https://github.com/github/VisualStudio/blob/ae375bf569d129a768612919e86f7ddeb39964aa/src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs#L350 from the currently active repository) so the UI doesn't look like we actually did open the thing when we didn't.

jcansdale: I added some logging for `SelectedRepository` and it seems to change when the user single-clicks on the list (or scrolls up and down using the up/down arrows). I'm not sure it's terribly meaningful beyond being a cursor for the user. Do you really think it needs to  be reset?
image

Please give me a clue if I'm missing something! 😉

jcansdale:
@grokys I've just tweaked this PR to account for the edge case where a repo directory is moved underneath repo list and the user then asks to open it. I think it's done now. Ready to merge!

PR #835, opened 2017-02-03
Add Blame Command

### LINKED ISSUE #748: Add 'GitHub > Blame' command
I often want to know about the history of a particular line. It would be nice if there was an obvious way to find this information.

The command would be similar to 'GitHub > Open on GitHub', but would open the 'Blame' page instead.

For example, 'GitHub > Open on GitHub' would go here:
https://github.com/github/VisualStudio/blob/76b99f8e1077f63e6f2dccc616a47933ec53af16/src/GitHub.Exports/Helpers/AssemblyResolver.cs#L53

For example, 'GitHub > Blame' would go here:
https://github.com/github/VisualStudio/blame/76b99f8e1077f63e6f2dccc616a47933ec53af16/src/GitHub.Exports/Helpers/AssemblyResolver.cs#L53

I know you can find the same information using 'Source Control > Annotate', but I didn't have a clue what 'Annotate' even meant! Also, the GitHub blame page is much prettier/more functional. 😄

paladique:
Fixes #748 

The only difference between the open/copy links and blame is replacing `blob` with `blame` in the url, so I added an <s>optional bool</s> Enum to generate the blame link.

### DIFF
--- a/src/GitHub.VisualStudio/Menus/BlameLink.cs
+++ b/src/GitHub.VisualStudio/Menus/BlameLink.cs
@@ -9,7 +9,6 @@ namespace GitHub.VisualStudio.Menus
 {
     public class BlameLink : LinkMenuBase, IDynamicMenuHandler
     {
-
         public BlameLink(IGitHubServiceProvider serviceProvider)
             : base(serviceProvider)
         {
@@ -38,6 +37,5 @@ namespace GitHub.VisualStudio.Menus
             var githubRepoCheckTask = IsGitHubRepo();
             return githubRepoCheckTask.Wait(250) ? githubRepoCheckTask.Result : false;
         }
-
     }
 }
--- a/src/GitHub.VisualStudio/Menus/BlameLink.cs
+++ b/src/GitHub.VisualStudio/Menus/BlameLink.cs
@@ -31,11 +31,5 @@ namespace GitHub.VisualStudio.Menus
 
             await UsageTracker.IncrementOpenInGitHubCount();
         }
-
-        public bool CanShow()
-        {
-            var githubRepoCheckTask = IsGitHubRepo();
-            return githubRepoCheckTask.Wait(250) ? githubRepoCheckTask.Result : false;
-        }
     }
 }
--- a/src/GitHub.VisualStudio/Menus/BlameLink.cs
+++ b/src/GitHub.VisualStudio/Menus/BlameLink.cs
@@ -1,4 +1,5 @@
-﻿using GitHub.Services;
+﻿using GitHub.Exports;
+using GitHub.Services;
 using GitHub.VisualStudio.UI;
 using NullGuard;
 using System;
@@ -23,7 +24,7 @@ namespace GitHub.VisualStudio.Menus
             if (!isgithub)
                 return;
 
-            var link = await GenerateLink(true);
+            var link = await GenerateLink(LinkType.Blame);
             if (link == null)
                 return;
             var browser = ServiceProvider.TryGetService<IVisualStudioBrowser>();
--- a/src/GitHub.VisualStudio/Menus/BlameLink.cs
+++ b/src/GitHub.VisualStudio/Menus/BlameLink.cs
@@ -1,0 +1,42 @@
+﻿using GitHub.Services;
+using GitHub.VisualStudio.UI;
+using NullGuard;
+using System;
+using System.Windows;
+
+namespace GitHub.VisualStudio.Menus
+{
+    public class BlameLink : LinkMenuBase, IDynamicMenuHandler
+    {
+
+        public BlameLink(IGitHubServiceProvider serviceProvider)
+            : base(serviceProvider)
+        {
+        }
+
+        public Guid Guid => GuidList.guidContextMenuSet;
+        public int CmdId => PkgCmdIDList.blameCommand;
+
+        public async void Activate([AllowNull]object data = null)
+        {
+            var isgithub = await IsGitHubRepo();
+            if (!isgithub)
+                return;
+
+            var link = await GenerateLink(true);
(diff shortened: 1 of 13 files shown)


paladique:
Still need to add a test or two, but a few questions about this: 
Do we need a new icon?
This was added to the bottom, right under the Gist command, it should probably be placed above it?

Do we need a copy command too, like for the view link command? If so should we rethink the context menu layout? 
Maybe something like:
```
GitHub
    View on Github >
        Copy
            File
            Blame
        Open
            File
            Blame       
    Create Gist
```

src/GitHub.VisualStudio/Menus/BlameLink.cs
```diff
@@ -0,0 +1,42 @@
...
+        public Guid Guid => GuidList.guidContextMenuSet;
+        public int CmdId => PkgCmdIDList.blameCommand;
+
+        public async void Activate([AllowNull]object data = null)
+        {
+            var isgithub = await IsGitHubRepo();
+            if (!isgithub)
+                return;
+
+            var link = await GenerateLink(true);
```
Haacked: Minor nit, but in terms of API design, I generally avoid booleans in this situation. Looking at this, it's not immediately clear what `true` means. Nor what `false`  means.

One way to improve it is to use a named parameter `GenerateLink(blame: true)`. Even so, it's not clear what false would mean in cases where `GenerateLink` is called.

Typically in this case, I think rather than using a boolean, an Enum might make it clear. Especially if there will be other kinds of link. Then it could say `GenerateLink(LinkType.Blame)` or `GenerateLink(LinkType.Blob)`.

If you don't anticipate that, then I'm fine with just specifying the named parameter.

src/GitHub.VisualStudio/Menus/BlameLink.cs
```diff
@@ -0,0 +1,42 @@
...
+            var browser = ServiceProvider.TryGetService<IVisualStudioBrowser>();
+            browser?.OpenUrl(link.ToUri());
+
+            await UsageTracker.IncrementOpenInGitHubCount();
+        }
+
+        public bool CanShow()
+        {
+            var githubRepoCheckTask = IsGitHubRepo();
+            return githubRepoCheckTask.Wait(250) ? githubRepoCheckTask.Result : false;
```
Haacked: What's this 250ms wait for?

src/GitHub.VisualStudio/Menus/BlameLink.cs
paladique: No idea why I went against the named param and removed it. Glad I was on the same track, though!
I added the optional bool to avoid altering the Create/Copy commands. Enum sounds better in the event we start adding more urls, so I think I'll go with that.

src/GitHub.VisualStudio/Menus/BlameLink.cs
paladique: This was from an [old bug fix](https://github.com/github/VisualStudio/pull/334#discussion-diff-65528365) where we added `wait` to avoid blocking the UI if it takes to long to find if a repo is a GitHub repo

Haacked: Ah, probably should document that ~~hack~~workaround here too and mention where it was copied from.

paladique: `CanShow()` is the same between the Open, Copy, and Blame Link commands at this point. 
`CanShow()` is a member of `IDynamicMenuHandler` and I'm curious if there's a better to this potential common pattern, especially if we'd also want a "Copy Blame Link" command. `CanShow()` will just be copied once again. Any ideas? Maybe an extension method? Maybe turning and renaming `IDynamicMenuHandler` into an abstract class?

Haacked: I haven't taken too close a look at this, but it really depends on how different the behaviors are between the different types of links. Typically, you have a few options.

1. If the differences between these three links are pretty minimal, make a simple `GitHubLink` class that is configurable and changes behavior accordingly.
2. Inheritance: Have a base `GitHubLink` class that all these link classes inherit from. This is a good approach if there's some reason these absolutely need to be different types. Otherwise I'd prefer to use a composition approach.
3. Composition: Create a `GitHubLink` class that accepts some configuration parameters but also accepts a `GitHubLinkCommand` or a `GitHubLinkHandler` class. This class would actually handle the behavior of clicking on the link while `GitHubLink` would be responsible for the presentation of the link. This keeps these concerns nicely separated.

paladique:
Since `LinkMenuBase` is the base for any classes that generate or open a github repo link, I moved the method there.

Haacked:
> Since LinkMenuBase is the base for any classes that generate or open a github repo link, I moved the method there.

That seems like a nice safe change and accomplishes the goal of reducing unnecessary duplication. 👍

grokys:
LGTM! I think the main question is if this should go into the next bugfix release (`master`) or the next minor release (`release/2.3`)? It's a feature rather than a bugfix so I'm tempted so say 2.3, though the change is really unlikely to break anything so I'd also be fine with the next bugfix release.

Thoughts @shana @Haacked?

paladique:
Also curious about these [questions](https://github.com/github/VisualStudio/pull/835#issuecomment-277369447) before merging with master

grokys:
TBH to me, things are fine as they stand. I'm not sure copying the link would be so useful here, as it's not likely you'll be doing anything but going to it (the link to the file itself, you may be pasting into a document etc).

jcansdale:
I'd be inclined to keep it really simple.

Most VS menu buttons are 1-3 words that start with a capital.

Something like:
```
GitHub
    Open
    Blame
    Copy Link
    Create Gist
```

PR #836, opened 2017-02-03
GitHub toolbar icon

### LINKED ISSUE #412: Add GitHub icon in the toolbar to open the corresponding pane on the website
For now, probably just always jumping to the PR list page would work.

paladique:
Added the GitHub icon to the toolbar.

- Navigates to PR list on .com when PR list is visible 
- Navigates to PR  on .com when PR detail is visible 
- Disabled when repo is not a git or github repo

Fixes #412


src/GitHub.VisualStudio/UI/Views/GitHubPaneViewModel.cs
```diff
@@ -38,6 +39,9 @@ public class GitHubPaneViewModel : TeamExplorerItemBase, IGitHubPaneViewModel
         Microsoft.VisualStudio.Shell.OleMenuCommand back, forward, refresh;
         int latestReloadCallId;
 
+        [Import]
+        IVisualStudioBrowser VisualStudioBrowser { get; set; }
```
grokys: The rest of the services in this class are imported via the constructor, so probably best to do that for this too.

src/GitHub.VisualStudio/GitHub.VisualStudio.vsct
```diff
@@ -121,6 +121,15 @@
         </Strings>
       </Button>
 
+      <Button guid="guidGitHubToolbarCmdSet" id="githubCommand" type="Button">
+        <Icon guid="guidImages" id="logo" />
+        <CommandFlag>IconIsMoniker</CommandFlag>
+        <CommandFlag>DefaultDisabled</CommandFlag>
+        <Strings>
+          <ButtonText>GitHub</ButtonText>
```
grokys: This should probably be something like "View on GitHub".

paladique:
Feel like there's probably a better way to fix the icon styling, aside from making a new one with the proper foreground, but couldn't come up with a solution. Got the color from the PR icon and set the foreground to that.

Issue #786, opened 2017-01-18
Error message not being cleared after checkout pull request succeeds


jcansdale:
- GitHub Extension for Visual Studio  version: 2.2.04
- Visual Studio version: Enterprise 2015

__What happened__

1. Opened GitHub repo in Visual Studio
2. Clicked 'Pull Requests' button
3. Selected pull request
4. Clicked on 'Checkout [pull-request-name]' link
5. Checkout failed (not sure why this happened):

![checkout](https://cloud.githubusercontent.com/assets/11719160/22077619/c288e9a0-dd69-11e6-9e2f-3cc6e66d5476.png)

6. Checkout subsequently succeeded
7. Error message was not cleared

__Expected Behavior__

Error message cleared when checkout succeeds.

jcansdale:
Here's a different, but possibly related issue:

1. Open a solution with uncommitted changes
2. Open a PR
3. The 'Checkout' link is disabled because of the uncommitted changes
4. Right click and 'undo changes' on changes files in 'Team Explorer'
5. The link remains disabled

![uncommited](https://cloud.githubusercontent.com/assets/11719160/22475822/2596cdc8-e7d8-11e6-8f5e-b234f91ff377.PNG)

It is possible to refresh, but it wasn't immediately obvious.

(just making notes as I go along)

jcansdale:
Trying to repro but came across this: #823

grokys:
Closed by #842.

PR #842, opened 2017-02-08
Clear `OperationError` before attempting to checkout PR

jcansdale:
Fix for #786.

### DIFF
--- a/src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
@@ -137,12 +137,6 @@ namespace GitHub.ViewModels
             DiffFile = ReactiveCommand.Create();
         }
 
-        void SubscribeOperationError(ReactiveCommand<Unit> command)
-        {
-            command.ThrownExceptions.Subscribe(x => OperationError = x.Message);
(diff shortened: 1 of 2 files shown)


grokys:
Seems the change from `Exception` to `FileNotFoundException` made some other tests fail!

jcansdale:
LOL. Oops! I'll fix it.

jcansdale:
Seem to be working now...

grokys:
Thinking about this, we should probably also clear the error on refresh?

jcansdale:
@grokys I'm now clearing `OperationError` on refresh (`Initialize`) and also before Pull and Push.

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
```diff
@@ -137,6 +137,12 @@ public class PullRequestDetailViewModel : BaseViewModel, IPullRequestDetailViewM
             DiffFile = ReactiveCommand.Create();
         }
 
+        void SubscribeOperationError(ReactiveCommand<Unit> command)
```
grokys: Small nit: could you put this with the other private methods down below the public methods?

jcansdale:
@grokys Done. 😄

PR #844, opened 2017-02-08
Making RepositoryCloneViewModel.NoRepositoriesFound more reliable

### LINKED ISSUE #776: Clone dialog displays "No Repositories" message sporadically
No Repositories

StanleyGoldman:
This pull request fixes #776 

It appears that `OriginalCompleted` returns before `UnfilteredCount` is accurately updated.
Adding an `UnfilteredCount` change notification at a more deterministic point resolves the issue.


src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs
```diff
@@ -80,9 +80,24 @@ public class RepositoryCloneViewModel : BaseViewModel, IRepositoryCloneViewModel
                 loading => loading.Value || repositories.UnfilteredCount > 0 && !LoadingFailed)
                 .ToProperty(this, x => x.FilterTextIsEnabled);
 
-            this.WhenAny(x => x.IsLoading, x => x.LoadingFailed,
-                (loading, failed) => !loading.Value && !failed.Value && repositories.UnfilteredCount == 0)
-                .Subscribe(x => NoRepositoriesFound = x);
+            this.WhenAny(
+                x => x.repositories.UnfilteredCount,
```
shana: Tests are broken, and I suspect it may be because `WhenAny` is initially triggered with null values (or some such weirdness). It's weird, `Repositories` is being set before this, but it's very likely this access blowing up.

Probably best to do something like

```
this.WhenAny(x => x.IsLoading, x => x.LoadingFailed, x => x.repositories?.UnfilteredCount,
    (loading, failed, count) => !loading.Value && !failed.Value && count != null && count.Value == 0)
    .Subscribe(x => NoRepositoriesFound = x);
```

StanleyGoldman: Changing the code like such results in..
`Error CS8072 An expression tree lambda may not contain a null propagating operator.`

shana: lol of course

```

Comment (grokys) [inline src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs]: Ok, this is happening because in the unit tests, `repositories.OriginalCompleted` is null.

Previously the unit tests were replacing the tracking collection with another tracking collection in the now removed assignment: one that had a source observable and therefore an OriginalCompleted.

Comment (grokys) [inline src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs]: I've added a fix for this which fixes all but the first of the unit tests. For some reason in that test, 6 items are being added to the tracking collection where only 3 were passed in.

Comment (grokys) [inline src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs]: The failing test appears to be a bug in `TrackingCollection` which duplicates items when the collection is initiated via `Listen` instead of in the ctor, and a `Filter` is set.

Comment (shana) [inline src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs]: Calling the constructor and then Listen is literally the same thing as calling the constructor with an observable - it calls the other constructor and then it calls Listen. If there are duplicated items, then it's because the Comparer is not returning true when two different instances of an Item compare to the same thing. This likely means that the test is incorrectly implementing the necessary comparer to return the item equality.

Comment (grokys) [inline src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs]: Yep, turns out the problem was that NSubstitute's `bool Equals(IRemoteRepositoryModel other)` was always returning false.

File: src/GitHub.App/ViewModels/RepositoryCloneViewModel.cs [hunk 2]
```diff
@@ -80,9 +80,24 @@ public class RepositoryCloneViewModel : BaseViewModel, IRepositoryCloneViewModel
...
+                x => x.LoadingFailed,
+                (unfilteredCount, loading, failed) =>
+                {
+                    if (loading.Value)
+                        return false;
+
+                    if (failed.Value)
+                        return false;
+
+                    return unfilteredCount.Value == 0;
```
grokys: I think I actually found this code to be easier to read before as I could read it as "not loading, not failed and count is 0". Not that I'm requesting a change here though, I don't feel that strongly about it.

Issue #809, opened 2017-01-27
Refactor login process


grokys:
There are various problem with our login process:

1. we assume that if we have a connection object, we're logged in. That's not a bad assumption, but it should be matched with a windows credential lookup whenever things get loaded initially to make sure that there are actual credentials stored that match that connection object. I think we're doing that in the GitHub.Api code (the Simple* stuff). We should verify that this logic is sound
2. We trust that credentials that are stored are valid but don't test this assumption explicitly until the user tries to do something that requires API access. At this point, if the credentials aren't valid, we fail silently at some points and forcebly log out the user at others. This makes for confusing codepaths and potential disconnects of the UI
3. We repeatedly use the stored credentials to ping the API for the Team Explorer buttons and to test that the repo is a valid repo. Since all the Team Explorer Home buttons and sections are loaded in parallel, we do repeated pings with the credentials all at the same time. This was to avoid locking the UI by having a single ping to validate the credentials before letting all the UI do it. The problem here is that if the credentials are not valid, the server might potentially lock out the user due to repeated "hacking" attempts. We have a user report about this.
4. The whole "Login From Cache" and "Login From the API" codepaths are incredibly confusing.
5. We're not being transparent with our login failure messages, either in the login dialog or anywhere else in the UI

Since we're stuck with not using rx for any of the Team Explorer initial stuff, maybe it would be best if we used the GitHub.Api library to consolidate all login code (no observables, just async/await stuff) so that we can make sense of this mess.

jcansdale:
Do we need to wrap every API call in case the token has gone stale and we need to log back in?

grokys:
Perhaps, although I'm not sure that would be feasible for a couple of reasons:

1. VS uses our credentials for git operations, and we can't wrap these as they're not in our codepaths
2. Wrapping every API call with something that can show a dialog box may lead to UI deadlocks and unexpected behavior, particularly when you throw threading into the mix.

In general though tokens shouldn't go stale, maybe we need to identify the causes of invalid credentials first?

shana:
Tokens can be removed on the server side or they can expire/be expired by admins. They may be old, stale, or maybe we updated our scope and had to force everyone to relogin. This isn't something that happens a lot honestly, but if the token is invalid, we need to figure out a way of properly resetting the connection so we really only use credentials that we trust.

PR #845, opened 2017-02-09
Refactor Logins, Part Uno

### LINKED ISSUE #19: 2FA dialog stays busy when network connection is down
Imported from https://github.com/github/VisualStudioInternal/issues/346

Repro steps:
1. Login with an account that has 2fa enabled.
2. Disconnect internet access from your computer.
3. Enter 2fa code and hit verify.

Expected: 2fa dialog shows an error message.
Actual: The verify button stays disabled and busy. Note, the dialog is still responsive, the operation just never completes.

### LINKED ISSUE #829: Login dialog hangs when max 2FA login attempts is exceeded
Hello! Please read the contributing guidelines before submitting an issue regarding the GitHub Extension for Visual Studio.

- Visual Studio version: 2015

- Log in with a user that has 2FA enabled
- Enter an invalid 2FA code multiple times
- When the max login attempts has been exceeded the button stays spinning and no error is displayed

The dialog is still closable in this state but an error should be displayed.

grokys:
This is the first step of #809. I wanted to stop here and submit a PR even if everything isn't finished as there's enough to review already. If need be I can probably split this into separate "refactor" and "bug fix" PRs but I wanted to a) actually get some bug fixes in with the refactoring, and b) test that the refactoring actually helps solve real problems.

# Refactored

- Moved API client configuration into static `ApiClientConfiguration` class in GitHub.Api
- Added `GitHub.Api.LoginManager` service which contains the actual login code (moved out of `RepositoryHost` which is a model and shouldn't contain such code)
- Removed Akavache dependency in `LoginCache` (now `WindowsLoginCache`)
- Moved IO out of `ConnectionManager` and into `IConnectionCache` (`JsonConnectionCache`)
- Made `SimpleApiClientFactory.Create` and `ApiClientFactory.Create` async (this will pave the way for first checking if there is a valid connection)
- Don't use `UserError.Throw` on `LoginTabViewModel` - instead just expose the `UserError` in a property - this removes the need for the changing tabs hack, simplifies things, and more importantly avoids a race condition when moving back to the login view with an error.
- Slightly simplified `TwoFactorDialogViewModel.Show`: https://github.com/github/VisualStudio/commit/78432789f998380d1944d7d7fedcf29fb217534f
- Added unit tests to things I touched

# Fixes

- Check that connection credentials are valid when loading connections
- Fix #19 2FA dialog stays busy when network connection is down
- Fix #829 Login dialog hangs when max 2FA login attempts is exceeded


src/GitHub.Api/ApiClientConfiguration_User.cs
```diff
@@ -1,11 +1,13 @@
+﻿using System;
+
 namespace GitHub.Api
 {
-    public partial class ApiClient : IApiClient
+    static partial class ApiClientConfiguration
```
jcansdale: It would be nice if rather than getting the `ClientId` / `ClientSecret` from source, it got if from an environment variable (`GITHUB_CLIENT_ID` or something). That way the developer wouldn't have to dirty their repo whenever they wanted to try something out (a hassle when changing branches). It could complain with a visible/helpful error when these variables aren't set.

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -130,6 +132,7 @@ protected override Task InitializeAsync(CancellationToken cancellationToken, IPr
         {
             AddService(typeof(IGitHubServiceProvider), CreateService, true);
             AddService(typeof(IUsageTracker), CreateService, true);
+            AddService(typeof(ILoginManager), CreateService, true);
```
jcansdale: Could this be exposed as MEF export rather than a Visual Studio service? If we're ever going to add support for Blend (which also hosts TeamExplorer), we'll need to avoid using custom services.

src/CredentialManagement/CredentialManagement.csproj
```diff
@@ -14,7 +14,7 @@
     <NuGetPackageImportStamp>
     </NuGetPackageImportStamp>
     <TargetFrameworkProfile />
-    <BuildType Condition="Exists('..\..\script\src\ApiClientConfiguration.cs')">Internal</BuildType>
+    <BuildType Condition="Exists('..\..\script\src\ApiClientConfiguration_User.cs')">Internal</BuildType>
```
jcansdale: Just a observation having seen all the .csproj tweaks. It would be nice to factor out these common signing related properties into a separate .props file. I'll do this in a separate PR along with the public key stuff.

src/GitHub.Api/ApiClientConfiguration_User.cs
grokys: Yeah, I was wondering about that as well. We can think about that later I'd say.

src/GitHub.VisualStudio/GitHubPackage.cs
grokys: Actually, I think it could. I think it was originally a service as it had an async initialization, but that's not there any more.

src/CredentialManagement/CredentialManagement.csproj
grokys: 👍

src/GitHub.Api/ApiClientConfiguration_User.cs
shana: These IDs need to be in the code, users can't be configuring this data manually on their systems in order to get the extension working. We can gitignore the file that keeps these variables in the public build so people can configure and build with them without having the repo dirtied or accidentally pushed.

src/GitHub.VisualStudio/GitHubPackage.cs
shana: The services that need to be initialized here are things that we need to use at VS startup. MEF is too slow at loading its catalog and causes a performance hit that we can't afford at this stage. The login manager is not one of these services, so it should definitely be an export.

src/GitHub.Api/ApiClientConfiguration_User.cs
jcansdale: Sounds like a plan!

@grokys can we .`gitignore` the local `ApiClientConfiguration_User.cs` file?

shana: We might need to tweak the csproj and the default codepath a bit so that if it doesn't find the file, it can still build with fake data (right now the file always has to be there). Probably also better to move it to the `common` folder, since it's "common" configuration data and that's an easier location to point to than all the way inside GitHub.Api, and maybe rename it to something more obvious. The steps would probably be something like:

- Set default `ClientId` and `ClientSecret` values in `ApiClientConfiguration.cs` if they're null after `Configure()`
- Delete the public `ApiClientConfiguration_User.cs` file
- Add a `src/common/ApplicationKeys.cs` file to .gitignore
- Include `src/common/ApplicationKeys.cs` in the csproj if it exists and it's not an `internal` build
- Add a little script that asks the user for a client id and client secret and generates the code with them.
- Update the docs with these instructions (and also the template if people want to do it manually)

We can do this on a separate PR, probably 😄

shana:
FYI, @StanleyGoldman and I went to town on how git manages credentials yesterday and found a few interesting things, including how to easily break our credential and login code in GHfVS by logging in with GCM, wincred, Unity and command line git. I suspect some of the issues we are seeing are coming from other providers doing things and we not handling that very well (especially the weird `git:https` message in the login)

I'm 👍 for merging this but we need a comprehensive session on testing all this code and fixing every single paranoid failure codepath before we ship it, I'm thinking it's going to be an agenda item for the mini-summit.

shana:
FYI, @StanleyGoldman and I went to town on how git manages credentials yesterday and found a few interesting things, including how to easily break our credential and login code in GHfVS by logging in with GCM, wincred, Unity and command line git. I suspect some of the issues we are seeing are coming from other providers doing things and we not handling that very well (especially the weird `git:https` message in the login)

I'm 👍 for merging this but we need a comprehensive session on testing all this code and fixing every single paranoid failure codepath before we ship it, I'm thinking it's going to be an agenda item for the mini-summit.

src/GitHub.Api/ApiClientConfiguration_User.cs
jcansdale: A separate PR that fixes a few things for public builds would certainly be good. We could sort out the public / open source signing issue at the same time. Recent versions need to be signed and this isn't happening at the moment. They're all related issues with quite similar fixes.

src/GitHub.App/Factories/ApiClientFactory.cs
```diff
@@ -25,13 +27,19 @@ public ApiClientFactory(ILoginCache loginCache, IProgram program, ILoggingConfig
             config.Configure();
         }
 
-        public IApiClient Create(HostAddress hostAddress)
+        public Task<IGitHubClient> CreateGitHubClient(HostAddress hostAddress)
```
grokys: Tesing leaving comment from GHfVS.

grokys: Does this work?

src/GitHub.VisualStudio/GitHubPackage.cs
StanleyGoldman: This comment is a test.

shana: OOh stuff!

shana: Starting a new conversation

grokys: Hello

grokys: Testing reply.

grokys: Another reply

grokys: A third reply.

grokys: Replying to first message.

PR #852, opened 2017-02-13
Handle all exceptions in PR list view model.

### LINKED ISSUE #823: Refreshing PR list crashes VS when no network connection
- Visual Studio version: 2015 Enterprise

1. Open solution that in GitHub repo
2. Open GitHub tool window
3. Disconnect WiFi
4. Press 'Refresh' icon on GitHub window
5. Visual Studio crashes (sometimes hangs process needs to be killed)

image

grokys:
All exceptions need to be handled in the subscription to `pullRequests.OriginalCompleted` or we will crash VS. We really need to add an error state to our view models, but for the moment just do as we did before and show an empty list if an exception occurs.

Fixes #823.

### DIFF
--- a/src/GitHub.App/ViewModels/PullRequestListViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestListViewModel.cs
@@ -6,7 +6,6 @@ using System.Linq;
 using System.Reactive.Linq;
 using System.Windows.Input;
 using System.Windows.Media.Imaging;
-using System.Windows.Threading;
 using GitHub.App;
 using GitHub.Collections;
 using GitHub.Exports;
@@ -14,6 +13,7 @@ using GitHub.Models;
 using GitHub.Services;
 using GitHub.Settings;
 using GitHub.UI;
+using NLog;
 using NullGuard;
 using ReactiveUI;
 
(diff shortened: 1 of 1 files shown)


src/GitHub.App/ViewModels/PullRequestListViewModel.cs
```diff
@@ -112,9 +112,9 @@ public override void Initialize([AllowNull] ViewWithData data)
                     // TODO: Do some decent logging here
                     return repositoryHost.LogOut();
                 })
-                .Catch<System.Reactive.Unit, Octokit.NotFoundException>(ex =>
+                .Catch<System.Reactive.Unit, Exception>(ex =>
                 {
-                    //this is caused when repository was deleted on github
+                    // Occurs on network error, when the repository was deleted on GitHub etc.
                     return Observable.Empty<System.Reactive.Unit>();
```
jcansdale: Should we be logging the exception somewhere?

grokys: Yeah, we should be... There was previously a TODO to add logging there. I might as well add it!

grokys: Logging added!

PR #859, opened 2017-02-15
Hide copy/open link when file not hosted on repo.

### LINKED ISSUE #530: Hide copy/open link when file not hosted on repo
Steps to reproduce (all in Visual Studio)

1) Open a solution that is hosted in a GitHub repo (Visual Studio should switch git context to hosted repo)
2) Close solution
3) Open a file that doesn't exist in the repo of the Git context
4) Highlight over the file's text & right click to navigation to "Open in GitHub/Copy link"

Link contains last commit of repo. 

Not sure the option to close/exit out of the context or repo when the solution is closed is available, might be a Visual Studio thing, but we could check that the current file exists in the repo and hide it if it doesn't for now.

grokys:
Check to see if the current file is located in the active repository and only show "Open on GitHub" and "Copy link to clipboard" context menus if it is.

Fixes #530.

### DIFF
--- a/src/GitHub.VisualStudio/Menus/CopyLink.cs
+++ b/src/GitHub.VisualStudio/Menus/CopyLink.cs
@@ -42,9 +42,8 @@ namespace GitHub.VisualStudio.Menus
 
         public bool CanShow()
         {
-            var githubRepoCheckTask = IsGitHubRepo();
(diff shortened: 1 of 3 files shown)


grokys:
@jcansdale yeah, we don't currently consider submodules when generating links. We should fix this, but it should be a separate PR IMO. Could you add an issue to track?

PR #862, opened 2017-02-15
Remove copyright year from LICENSE file

Phineas:


Haacked:
In US copyright law, a copyright notice is not required to be protected under copyright law. Thus this notice is entirely educational. And in general, for a changing work such as this, it should be a range of years from the year it was first published to the current year.

However, I'm inclined to suggest we just remove the year. The important part for anyone looking is who owns the copyright.

shana:
After much pondering and discussion, the conclusion is that yeah, we should nuke the year altogether!

shana:
celebrate

marlazinger200:
1

PR #868, opened 2017-02-16
Refactor Button styling.

### LINKED ISSUE #796: Publish to GitHub button in TE does not support high contrast
Steps
1. Set Windows 10 to High Contrast mode
2. Open VS
3. Click publish button in the bottom right corner of VS
4. Look at the GitHub option 

scr_rc5lf4d0 zuh

grokys:
- Remove `GitHubButton` style - this was the default style for all our buttons, but wasn't used anywhere. It appears to be the default style in GHfW but doesn't fit with Visual Studio. By removing this default style, buttons in Team Explorer use the default button style - fixes #796.
- Make `GitHubVsButton` the default style for buttons in the GitHub pane. We only currently have one button that's not a primary button ("Cancel" in the PR creation view) but it should be the default style for subsequent buttons there, so removed the style from that individual button and made it apply to all buttons.


shana:
Can we get some screenshot action here? 😄

grokys:
Sure! There's not much to show really, everything looks the same except the high-contrast used to look like this:

image

And now looks like this:

image

(They're suing different text colors in the theme, but you can see the "Publish to GitHub" button is now styled correctly.

Issue #655, opened 2016-10-28
Missing Icon on VS Start Page


abuchholtzau:
Hello! Please read the contributing guidelines before submitting an issue.
- GitHub Extension version: 2.0.15.1
- Visual Studio version: VS15 Preview 5

**What happened** (with steps, logs and screenshots, if possible)
On the Start Page, GitHub shows up correctly in the Open section but the icon is missing. The icon had previously shown up. 

![github_icon_missing](https://cloud.githubusercontent.com/assets/3174849/19816781/8964dd52-9cfd-11e6-9d65-90781728d412.PNG)

abuchholtzau:
This is only happening on the first run after installing the extension. I just restarted VS and the icon showed up.

shana:
@abuchholtzau I'd love to know why this is happening, actually. The image is coming from the image service (https://github.com/github/VisualStudio/search?utf8=%E2%9C%93&q=27841f47 and https://github.com/editor-tools/StartPage/blob/master/preview5/StartPagePackage.cs#L26), I have no clue why it wouldn't work the first time (VS registering something late?)

@prasethu @aarnott @clboles Do you know why this would be happening?

AArnott:
I've asked Andrew Sarratori to help out if he can.

Newrad0603:
I did some debugging after repro-ing the issue. The reason this fails is because at the time the image would be displayed, the GitHub.UI.dll has not been loaded into the app domain, so when the ImageLibrary attempts to load the image, it fails. Toggling the theme a short time after launch causes the image to display. There seems to be a timing issue where sometimes the UI dll is loaded in time, so the image request succeeds. It appears a change needs to be made to make sure that the UI dll is loaded when the new Start Page is instantiated.

shana:
@Newrad0603 Thank you for tracking this down! This makes sense, we'll see how we can fix it.

AArnott:
@Newrad0603: is this a case of a missing codebase path for their github.ui.dll assembly? Is the image service willing to load the assembly as long as the CLR can find it?

Newrad0603:
@AArnott Yes, the Image Service should load any dll that VS can find. A few other people have hit codebase issue, so I imagine that the missing codebase entry would likely be the cause.

shana:
@AArnott @Newrad0603 What is this codebase path issue? How can we fix it? Is adding a reference to the dll in the package that needs the icon enough to get it to load early enough, or is there something else we're missing?

[...]

PR #875, opened 2017-02-21
Fix assembly resolving issues and start page icon

jcansdale:
It turns out the missing start page icon #655 was a symptom of much larger assembly resolving issues.

- What happened depended on the order in which different packages were loaded.
- If an assembly containing WPF resources loaded first, WPF is able to resolve it from there.
- If the `GitHubPackage` loaded first, its custom assembly resolver was able to resolve resources.
- If an assembly is loaded via binding/codebase in .config file, its dependencies fail to resolve.
- More issues that I don't understand yet. 😉  

This PR is an attempt to come up with a solution that is deterministic and robust!


jcansdale:
This version fixes the icon the very first time Visual Studio loads after installing the extension. I believe this is because the `GitHub.UI` / `GitHub.VisualStudio.UI` are loaded as the extension installs itself and WPF is able to find them in memory. The icon vanishes on subsequent loads.

jcansdale:
The last commit makes the icon consistently appear. Unfortunately clicking on the icon/GitHub link didn't always work.  It doesn't seem to be completely consistent and might depend on whether the `Team Explorer` package loaded first.

Here is an example error:

image

Being more liberal with our assembly resolver's assembly loading seems to help...

jcansdale:
If we combine defining a codebase using `ProvideBindingRedirection` with our assembly resolver, the assembly resolver ends up having to resolve the following assemblies:

```
GitHub.VisualStudio.UI, Culture=neutral
System.Reactive.Core, Version=2.2.5.0, Culture=neutral, PublicKeyToken=62aa029873c516b4
NLog, Version=3.1.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c
Octokit, Version=0.22.0.0, Culture=neutral, PublicKeyToken=5525be5bc50478ea 
LibGit2Sharp, Version=0.22.0.0, Culture=neutral, PublicKeyToken=7cbde695407f0333 
Rothko, Version=0.0.1.0, Culture=neutral, PublicKeyToken=9f664c41f503810a 
GitHub.UI.Reactive, Version=2.2.0.8, Culture=neutral, PublicKeyToken=bc1bd09f2901c82e 
GitHub.VisualStudio.UI, Version=2.2.0.8, Culture=neutral, PublicKeyToken=bc1bd09f2901c82e 
GitHub.VisualStudio.UI, Version=2.2.0.8, Culture=neutral, PublicKeyToken=bc1bd09f2901c82e 
ReactiveUI, Version=6.3.1.0, Culture=neutral, PublicKeyToken=31f8874b2ce7ed44 
System.Reactive.Linq, Version=2.2.5.0, Culture=neutral, PublicKeyToken=62aa029873c516b4 
System.Reactive.Interfaces, Version=2.2.5.0, Culture=neutral, PublicKeyToken=62aa029873c516b4
```

Visual Studio 2017 then crashes! It seems to matter where some of these assemblies resolve to (even if the assembly version matches).

Note, there is no `PublicKeyToken` defined when the `GitHub.VisualStudio.UI` assembly is resolved. This means a binding redirect/codebase could never be used to resolve this assembly.

When the explicit codebases aren't defined, the assembly resolved is only required to resolve the following:

```
GitHub.UI, Version=2.2.0.8, Culture=neutral, PublicKeyToken=bc1bd09f2901c82e 
GitHub match: GitHub.VisualStudio.UI, Culture=neutral
```

jcansdale:
I've extended to assembly resolver to resolve any assembly that it finds an exact match for in our extension directory. This should be safe because the first assembly that is resolved from the `LoadFrom` context is the one that will be shared between all extensions. If an extension can't deal with this, it's a bug! 😉 

After this rule, I think the only extension that really needs to go in the `ourAssemblies` list is `GitHub.VisualStudio.UI` (because it's referenced from somewhere without a `PublicKeyToken`). All the other assemblies are referenced using their full names.

Here is the current list:

```
        // list of assembly names that should always be loaded from extension dir
        static readonly string[] ourAssemblies =
        {
            "GitHub.Api",
            "GitHub.App",
            "GitHub.CredentialManagement",
            "GitHub.Exports",
            "GitHub.Exports.Reactive",
            "GitHub.Extensions",
            "GitHub.Extensions.Reactive",
            "GitHub.UI",
            "GitHub.UI.Reactive",
            "GitHub.VisualStudio",
            "GitHub.TeamFoundation",
            "GitHub.TeamFoundation.14",
            "GitHub.TeamFoundation.15",
            "GitHub.VisualStudio.UI",
            "System.Windows.Interactivity"
        };
```

If we ever want to allow other packages to integrate with GHfVS, it would make sense to included other assemblies in this list. That way our version number could change without always breaking them (as long as they don't package the assembly). We would need to be careful about what other assemblies were part of the interface (e.g. `Octokit`).

jcansdale:
I've changed it to only relax assembly resolving for:
```
GitHub.VisualStudio.UI
GitHub.Exports
GitHub.Exports.Reactive
```
(They can be resolved by simple name only)

This is only necessary for `GitHub.VisualStudio.UI`, but could be useful for `GitHub.Exports` and `GitHub.Exports.Reactive` in future.

All other assemblies will only be resolved to the extension directory when their full name matches.

jcansdale:
@grokys It appears the assembly full name can be inferred from the referenced assemblies. This vastly simplifies the attribute definitions! 😄 

```

[assembly: ProvideCodeBase(AssemblyName = "GitHub.UI", CodeBase = @"$PackageFolder$\GitHub.UI.dll")]
[assembly: ProvideCodeBase(AssemblyName = "GitHub.VisualStudio.UI", CodeBase = @"$PackageFolder$\GitHub.VisualStudio.UI.dll")]
[assembly: ProvideCodeBase(AssemblyName = "GitHub.Exports", CodeBase = @"$PackageFolder$\GitHub.Exports.dll")]
[assembly: ProvideCodeBase(AssemblyName = "GitHub.Extensions", CodeBase = @"$PackageFolder$\GitHub.Extensions.dll")]
[assembly: ProvideCodeBase(AssemblyName = "Octokit", CodeBase = @"$PackageFolder$\Octokit.dll")]
[assembly: ProvideCodeBase(AssemblyName = "LibGit2Sharp", CodeBase = @"$PackageFolder$\LibGit2Sharp.dll")]
[assembly: ProvideCodeBase(AssemblyName = "Splat", CodeBase = @"$PackageFolder$\Splat.dll")]
```

We really need to audit the dependencies of these assemblies. All referenced assemblies should appear in the GAC, `devenv.exe.config` or on this list. For example, `GitHub.UI` references `NullGuard`. I'm guessing that should appear on this list?

jcansdale:
Oh, bugger. This happened when I opened VS 2017 with the `GitHub` tool window open:  😭 

```
Exception details:
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> Microsoft.VisualStudio.Composition.CompositionFailedException: An exception was thrown while initializing part "GitHub.VisualStudio.UI.Views.GitHubPaneView". ---> System.Windows.Markup.XamlParseException: 'Set property 'GitHub.VisualStudio.UI.Helpers.SharedDictionaryManager.Source' threw an exception.' Line number '7' and line position '10'. ---> System.IO.FileNotFoundException: Could not load file or assembly 'GitHub.UI, Culture=neutral' or one of its dependencies. The system cannot find the file specified.
   at System.Reflection.RuntimeAssembly._nLoad(AssemblyName fileName, String codeBase, Evidence assemblySecurity, RuntimeAssembly locationHint, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.RuntimeAssembly.nLoad(AssemblyName fileName, String codeBase, Evidence assemblySecurity, RuntimeAssembly locationHint, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.RuntimeAssembly.InternalLoadAssemblyName(AssemblyName assemblyRef, Evidence assemblySecurity, RuntimeAssembly reqAssembly, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   at System.Reflection.Assembly.Load(AssemblyName assemblyRef)
   at System.Windows.Navigation.BaseUriHelper.GetLoadedAssembly(String assemblyName, String assemblyVersion, String assemblyKey)
   at MS.Internal.AppModel.ResourceContainer.GetResourceManagerWrapper(Uri uri, String& partName, Boolean& isContentFile)
   at MS.Internal.AppModel.ResourceContainer.GetPartCore(Uri uri)
   at System.IO.Packaging.Package.GetPartHelper(Uri partUri)
   at System.IO.Packaging.Package.GetPart(Uri partUri)
   at System.IO.Packaging.PackWebResponse.CachedResponse.GetResponseStream()
   at System.IO.Packaging.PackWebResponse.GetResponseStream()
   at System.IO.Packaging.PackWebResponse.get_ContentType()
   at MS.Internal.WpfWebRequestHelper.GetContentType(WebResponse response)
   at MS.Internal.WpfWebRequestHelper.GetResponseStream(WebRequest request, ContentType& contentType)
   at System.Windows.ResourceDictionary.set_Source(Uri value)
   at GitHub.VisualStudio.UI.Helpers.SharedDictionaryManager.set_Source(Uri value) in C:\Source\github.com\github\VisualStudio\src\GitHub.VisualStudio.UI\Helpers\SharedDictionaryManager.cs:line 51
```

jcansdale:
@grokys I've fixed some cases where assembly resolver is still needed. This does however feel like a band-aid and not a long term solution.

I noticed that when I view the `View > Other Windows` menu, the `GitHub.VisualStudio.UI` assembly gets loaded by the resolver! I think what we need to do is ensure that when an `.imagemanifest` contains XAML, the XAML should only depend on resources in that same assembly.

Does that sound feasible?

grokys:
> what we need to do is ensure that when an .imagemanifest contains XAML, the XAML should only depend on resources in that same assembly.

Yep, sounds totally reasonable to me. We can link files from one project to another to prevent duplicating code.

jcansdale:
I've squashed all of the experimental commits and removed the `ProvideCodeBase` attributes.

We weren't actually gaining anything from the `ProvideCodeBase` attributes (the assembly resolver was still required). When we add a `ProvideCodeBase` to a project/assembly, we need to make sure we also `ProvideCodeBase` for *all* of its dependencies. It will blow up if we miss one and the assembly resolver can't save us (resolve events won't be fired)!

I'm still keen to consolidate the XAML / .imagemanifest dependencies. It's possible we won't need either the assembly resolver or `ProvideCodeBase` attributes when we do this. If it turns out all we need is the `ProvideCodeBase` attributes, then that would be a good time to re-add them. We can look at this in a separate PR. 😄 

Does that make sense?

grokys:
Yep, that makes sense @jcansdale!

src/GitHub.VisualStudio/AssemblyResolverPackage.cs
```diff
@@ -0,0 +1,88 @@
...
+    [ProvideAutoLoad(VSConstants.UICONTEXT.NoSolution_string)]
+    public class AssemblyResolverPackage : Package
+    {
+        // list of assemblies that should be resolved by name only
+        static readonly string[] ourAssemblies =
+        {
+            "GitHub.UI",
+            "GitHub.VisualStudio.UI",
+            "System.Windows.Interactivity"
+        };
```
shana: If I understand the code correctly, then we're loading things from this list if they're referenced from xaml, which means they're referenced by partial name. If that's the case, then this list should probably also include `GitHub.UI.Reactive`, since that's also referenced from xaml. Perhaps audit all our xaml files to make sure this list includes all our referenced?

shana:
Also, make sure to test the GitHub options dialog to make sure it also loads properly (Tools / Options / GitHub), this also impacts that.

jcansdale:
@shana,

> If I understand the code correctly, then we're loading things from this list if they're referenced from xaml, which means they're referenced by partial name.

That's part of it. In fact, it only happens when both of these are true:

1. If they're referenced from XAML, which means they're referenced by partial name.
2. If the XAML `x:Class` doesn't depend on the referenced assembly.

We're hoping to eliminate this situation by only referencing XAML resources in the same assembly as loose XAML files (e.g. `mark_github.xaml` should reference its home assembly).

I've run through various loading scenarios to seen when the resolver is actually triggered. Visual Studio is opened with no tool windows active and I navigate directly to one of the following:

### Start Page

```
GitHub.UI, Version=2.2.0.8, Culture=neutral, PublicKeyToken=bc1bd09f2901c82e
GitHub.VisualStudio.UI, Culture=neutral
```

### View > Other Windows

```
GitHub.UI, Version=2.2.0.8, Culture=neutral, PublicKeyToken=bc1bd09f2901c82e
GitHub.VisualStudio.UI, Culture=neutral
```

### Options > GitHub for Visual Studio

```
GitHub.VisualStudio.UI, Version=2.2.0.8, Culture=neutral, PublicKeyToken=bc1bd09f2901c82e
```

### 'GitHub' tool window

```
GitHub.UI, Culture=neutral
```

### Team Explorer

```
Resolver not used at all.
```

You can see from this list, the resolver is only used to resolve `GitHub.VisualStudio.UI` and `GitHub.VisualStudio` by partial name.

It also resolves, `GitHub.UI, Version=2.2.0.8, Culture=neutral, PublicKeyToken=bc1bd09f2901c82e` and `GitHub.VisualStudio.UI, Version=2.2.0.8, Culture=neutral, PublicKeyToken=bc1bd09f2901c82e` by full name, which I'm a bit confused about (it won't be getting the full name from XAML). 😕

> If that's the case, then this list should probably also include GitHub.UI.Reactive, since that's also referenced from xaml. Perhaps audit all our xaml files to make sure this list includes all our referenced?

I haven't been able to find a scenario where the resolver is actually required for `GitHub.UI.Reactive` or `System.Windows.Interactivity`. I'm hopeful that soon the resolver won't be required at all. Maybe we could allow the resolver to resolve these assemblies but log an error if it ever does?

PR #898, opened 2017-03-01
Clear the IsProductComponent flag.

grokys:
This should only be set on a vsix that is distributed with the VS installer.


jcansdale:
I hope you don't mind, but this is now included in #884. I've grouped some build flavor related stuff together.

PR #906, opened 2017-03-03
Add VSIX to AppVeyor artifacts

jcansdale:
Now we have clearly marked 'Experimental' builds of the VSIX, I think it makes sense to make these visible on AppVeyor builds. They're unsigned and similar to builds any developer can create using the public repo.

This will make it easy to test the funcationality of an incomming PR, having reviewed it on the website. Having to build a version locally after minor code changes always seemed like a waste of time!

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,3 +26,6 @@ test_script:
     scripts\Run-Nunit.ps1 TrackingCollectionTests 180 Release -AppVeyor
 
     scripts\Run-Xunit.ps1 UnitTests 180 Release -AppVeyor
+    
+artifacts:
+- path: build\Release\GitHub.VisualStudio.vsix


shana:
If we're publishing random builds that people can install, we need to figure out a way of bumping the version so that we don't have two vsix packages with different code both marked with the same version. There's a difference between someone checking out the code and building locally, and us providing builds that can be considered "official" (because they're coming from our CI). We should make sure each of these builds can be uniquely identified. Even for our testing purposes, every build that is a non-local build should have a unique version.

jcansdale:
@shana,

AppVeyor has some handy environment variables that we could use:
```
APPVEYOR_REPO_COMMIT 
APPVEYOR_BUILD_NUMBER 
APPVEYOR_PULL_REQUEST_TITLE 
```

How about:
image

Or better still:
image
(we could lookup the SHA via the build number, so human readable makes sense)

PR #907, opened 2017-03-03
Refactored View/ViewModel hierarchy

grokys:
Previously our view model base interface/class (`IViewModel` and `BaseViewModel`) had a bunch of properties that weren't relevant to many view models, namely:

`IViewModel`:
- A `Title` property
- An `IsShowing` property
- An `IsBusy` property
- A `Cancel` command

`IVIew`:
- A `Done` observable
- A `Cancel` observable
- An `IsBusy` observable

These have been separated out:
- `IDialogViewModel`/`DialogViewModelBase` for view models that can be dismissed by OK/Cancel
- `IPanePageViewModel`/`PanePageViewModelBase` for view models that are displayed in the GitHub pane
- `IDialogView` for views that are displayed in the dialog

These all still implement `IView`/`IViewModel`.

There are still numerous things that could be done to clean this up, but I wanted to keep this PR focused:

- `Done`, `Cancel`, `IsBusy` should be moved out of `IDialogView` and into the view model
- `IsShowing` should calculated by the parent controller
- etc...


src/GitHub.App/SampleData/PullRequestDetailViewModelDesigner.cs
```diff
@@ -70,6 +70,7 @@ public PullRequestDetailViewModelDesigner()
         public IPullRequestModel Model { get; }
         public string SourceBranchDisplayName { get; set; }
         public string TargetBranchDisplayName { get; set; }
+        public bool IsBusy { get; }
```
shana: Probably want to add `IHasBusy` to the interface list for this one, just to make things match?

src/GitHub.TeamFoundation.14/Sync/GitHubPublishSection.cs
```diff
@@ -113,7 +113,7 @@ public void ShowPublish()
 
             controller.TransitionSignal.Subscribe(data =>
             {
-                var c = data.View;
+                var c = (IDialogView)data.View;
```
shana: Since we only care about the busy signal here, perhaps just cast to `IHasBusy`?

shana:
This is making the UIController only responsible for handling `IDialogView`, from what I can tell. `Done`, however, is not something that is exclusive to something running on a dialog - case in point, the PR creation view/viewmodel, which has a done state and a cancel state, but doesn't run in a dialog. I'm totally for moving signals out of the views and into viewmodels, but perhaps with interfaces that attach behaviours to the views or viewmodels, instead of attaching the behaviour to the way the thing is going to be displayed on screen?

i.e., yes most if not all things that are displayed in a dialog have a "done/cancel" state, but that doesn't mean only those things have that type of behaviour, which means that the "done/cancel" state behaviour is not exclusive to the dialog and therefore shouldn't be attached to it. Probably the dialog interface should implicitly inherit from the `IHasDone` (or `IHasActionState` implying done/cancel states (or whatever name we'd like to give this)), and then UIController can check for those/that interface and hook up things accordingly

grokys:
Yeah, I was wondering about this myself; initially `IDialogView` was indeed limited to views running in the dialog, but when I got to e.g. PR Creation, I came upon the issue you mentioned and repurposed `IDialogView` to mean "anything with an OK/Cancel button". I wasn't entirely happy with this, but as I mentioned I was trying to keep changes minimal.

It sounds like you think I should go a bit deeper with this refactor however, so I'll make the changes you suggest as they are pretty much what I was considering ;)

grokys:
Ok, I've done that - I remember why I decided not to do it that way originally ;)

From the commit message:

>Refactor the `Done` and `Cancel` signals from `IDialogView` into
`IHasDone` and `IHasCancel` interfaces so that views can define their
capabilities in a more granular fashion. This has the slightly
unfortunate side-effect of leaving `IDialogView` with only a single
`IsBusy` signal which can't be refactored out into an interface because
we already have an incompatible `IHasBusy` which is implemented by the
view models. This will be resolved in the next step of refactoring where
the signals are moved from the views to the view models.

src/GitHub.App/SampleData/PullRequestDetailViewModelDesigner.cs
grokys: `IHasBusy` is already on `IPullRequestDetailViewModel` so is already being implemented.

src/GitHub.TeamFoundation.14/Sync/GitHubPublishSection.cs
grokys: `IHasBusy` only has a `bool HasBusy` property, not an observable. I could use `.WhenAny` here but this is already addressed in #933 so might as well leave it for now?

grokys:
This was merged in #933.

PR #911, opened 2017-03-06
Fix installing 2.2.0.9-Experimental on top of 2.2.0.8

### LINKED ISSUE #910: Extension breaks when 2.2.0.9-Experimental is installed on top of 2.2.0.8
- GitHub Extension for Visual Studio version: 2.2.0.8 and 2.2.0.9-Experimental
- Visual Studio version: 2015/2017

1. Install 2.2.0.8 via `Extensions and Updates`.
2. Install 2.2.0.9-Experimental via VSIX (local build)
3. Open Visual Studio.

Both versions are loaded at the same time and terrible things start happening!

image

See also #884.

jcansdale:
Due to quirks of how Experimental versions are installed, the GUIDs of packages mustn't change between versions.

Because `AssemblyResolverPackage` wasn't being referenced from anywhere else, it didn't need an explicit GUID. The default GUID is a hash of the type name and full assembly name. This meant that when the assembly version was bumped, the package GUID also changed.

This wouldn't have been a problem, except for the fact that `Experimental` versions are overlayed on top of `AllUsers` versions (rather than replacing them entirely). The package GUID from the `AllUsers` version leaked through, causing both versions to get loaded!

Fixes #910

### DIFF
--- a/src/GitHub.VisualStudio/AssemblyResolverPackage.cs
+++ b/src/GitHub.VisualStudio/AssemblyResolverPackage.cs
@@ -1,11 +1,12 @@
 ﻿using System;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 using System.Diagnostics;
 using System.Globalization;
+using System.Runtime.InteropServices;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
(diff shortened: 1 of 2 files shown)


grokys:
gif

PR #914, opened 2017-03-07
Remove the need for assembly resolver

### LINKED ISSUE #885: Remove AppDomain.AssemblyResolve event handler
The https://github.com/github/VisualStudio/pull/875 change removed (and re-added) a custom assembly resolver. But this is an anti-pattern in VS for several reasons, including the very one that caused your old one to be so problematic. This new one is also problematic for (at least) the reason that it forces your new package to be loaded on startup, which slows down VS start time.

No VS extension should have to do this, AFAIK. Instead, simply supply a pkgdef with codebase paths for each of your assemblies so that the CLR can find them when it looks for them.

jcansdale:
A number of tweaks to remove the need of an assembly resolver.

This PR depends on the GitHubVSAutomationIDs assembly having a string name (being signed).

Fixes #885


jcansdale:
Depends on #918.

GitHubVS.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.0.0
```
shana: Avoid committing these, there's no point.

src/GitHub.UI/Helpers/LoadingResourceDictionary.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Diagnostics;
+using System.Collections.Generic;
+
+namespace GitHub
+{
+    public class LoadingResourceDictionary : ResourceDictionary
+    {
+        static Dictionary<string, Assembly> assemblyDicts = new Dictionary<string, Assembly>();
+
+        public new Uri Source
```
shana: For this to not break the xaml designer, it needs to be in an `#ifdef !XAML_DESIGNER` block (the xaml designer breaks if the `Source` property is overriden). Since that will cause assemblies not to be found (because now this code is handling both caching and finding assemblies), we'll probably need to fallback to using an assembly resolver when we're in the `XamlDesign` configuration that sets the `XAML_DESIGNER` define, so that VS can find things while on that mode.

src/GitHub.UI/Helpers/LoadingResourceDictionary.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            var path = packUri.LocalPath;
+            if(!path.StartsWith("/"))
+            {
+                return null;
+            }
+
+            var component = ";component/";
+            int componentIndex = path.IndexOf(component, 1);
+            if(componentIndex == -1)
```
shana: Missing space after `if`

src/GitHub.UI/Helpers/LoadingResourceDictionary.cs
```diff
@@ -0,0 +1,75 @@
...
+                    Trace.WriteLine("Couldn't find assembly at: " + assemblyFile);
+                    return;
+                }
+
+                var assembly = Assembly.LoadFrom(assemblyFile);
+                assemblyDicts.Add(assemblyFile, assembly);
+            }
+            catch(Exception e)
+            {
+                Trace.WriteLine($"Error loading assembly for '{value}': {e}");
```
shana: Logging should go to `VsOutputLogger`

src/GitHub.UI/Helpers/LoadingResourceDictionary.cs
```diff
@@ -0,0 +1,75 @@
...
+                var baseDir = Path.GetDirectoryName(GetType().Assembly.Location);
+                var assemblyFile = Path.Combine(baseDir, assemblyName + ".dll");
+                if (assemblyDicts.ContainsKey(assemblyFile))
+                {
+                    return;
+                }
+
+                if (!File.Exists(assemblyFile))
+                {
+                    Trace.WriteLine("Couldn't find assembly at: " + assemblyFile);
```
shana: Logging should go through `VsOutputLogger`

src/GitHub.UI/Helpers/LoadingResourceDictionary.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        void EnsureAssemblyLoaded(Uri value)
+        {
+            try
+            {
+                var assemblyName = FindAssemblyNameFromPackUri(value);
+                if (assemblyName == null)
+                {
+                    Trace.WriteLine("Couldn't find assembly name in: " + value);
```
shana: Logging should go through `VsOutputLogger`

src/GitHub.UI/Helpers/LoadingResourceDictionary.cs
```diff
@@ -0,0 +1,75 @@
...
+            catch(Exception e)
+            {
+                Trace.WriteLine($"Error loading assembly for '{value}': {e}");
+            }
+        }
+
+        static string FindAssemblyNameFromPackUri(Uri packUri)
+        {
+            var path = packUri.LocalPath;
+            if(!path.StartsWith("/"))
```
shana: Missing space after `if`

src/GitHub.UI/Helpers/LoadingResourceDictionary.cs
jcansdale: This was getting messy, so I thought I'd have a crack at finding out why the XAML designer doesn't like `SharedDictionaryManager`. With any luck `XAML_DESIGNER` will no longer be needed! See here #945.

I've added the `#ifdef !XAML_DESIGNER` anyway as part of this PR.

jcansdale:
> The reason why we're doing Trace.WriteLine in other places is because the VsOutputLogger actually hooks up a TraceListener, so it might just work automatically. 

It looks like this code is commented out.
```
//Debug.Listeners.Add(new VSTraceListener());
```

I'll log directly to `VsOutputLogger`.

jcansdale:
@shana I think this one is done now. Changes incorporated. 😄

shana:
@jcansdale All of this looks fine except for that automation project GUID change that we were talking about. Fix that and then merge this?

shana:
This looks good to go, just needs rebasing on #947 once that's in, right?

Issue #926, opened 2017-03-20
Use ProvideBindingRedirection rather than ProvideCodeBase


jcansdale:
- GitHub Extension for Visual Studio version: 2.2.0.9 (dev)
- Visual Studio version: All

In versions of the extension up until 2.2.0.8, assemblies are loaded into Visual Studio using `Assembly.LoadFrom` and end up in the `LoadFrom` context (along with their dependencies). The first application to load a particular version of an assembly gets to set its CodeBase and the search path for its dependencies.

This means another extension can wreck havoc by simply referencing one of our assemblies and including it in their VSIX. This happens by default when a VSIX project references an assembly and actually takes work stop it being included. Suddenly our extension will either work or error in unexpected ways depending on package load order! (see #923)

We can use the VS SDK `ProvideCodeBase` attribute to ensure that a particular version of an assembly will load from a specific CodeBase (into the `Load` context). For example:
```
[assembly: ProvideCodeBase(AssemblyName = "GitHub.Exports",
    CodeBase = @"$PackageFolder$\GitHub.Exports.dll")]
```

This can help when an extension references the exact version of an assembly that has been installed (with a `ProvideCodeBase` attribute). If however a different version is referenced, another assembly will be loaded with a different codebase, static fields and types. Unless the assembly has been designed with side-by-side execution in mind, the results could be unpredictable.

A resolution for this would be to sweep up multiple versions of an using the `ProvideBindingRedirection` attribute. For example:
```
[assembly: ProvideBindingRedirection(AssemblyName = "GitHub.Exports",
    CodeBase = @"$PackageFolder$\GitHub.Exports.dll",
    OldVersionLowerBound = "0.0.0.0", OldVersionUpperBound = "2.2.0.9")]
```

This would prevent old versions of an assembly being loaded at the same time as the current one (they would be redirected to the current one). This would also allowing 3rd party extensions that could work with multiple newer versions of GHfVS (rather than having to be built against a specific version).

jcansdale:
Fixed by #928.

Issue #923, opened 2017-03-14
Visual Studio crashes when I click on Team Explorer


jcansdale:
- GitHub Extension for Visual Studio version: 2.2.0.8
- Visual Studio version: 2017 Enterprise

1. Double-click on .sln file
2. Visual Studio opens with Solution Explorer window active
3. Click on 'Team Explorer' tab
4. Visual Studio crashes

![image](https://cloud.githubusercontent.com/assets/11719160/23901808/c15e9cfa-08b6-11e7-9971-1bb22b1f80d2.png)

The following was sent to debug output:
```
[38932] <TraceRecord xmlns="http://schemas.microsoft.com/2004/10/E2ETraceEvent/TraceRecord" Severity="Critical"><TraceIdentifier>http://msdn.microsoft.com/TraceCodes/System/ActivityTracing/2004/07/Reliability/Exception/Unhandled</TraceIdentifier><Description>Unhandled exception</Description><AppDomain>DefaultDomain</AppDomain><Exception><ExceptionType>System.TypeInitializationException, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</ExceptionType><Message>The type initializer for 'LibGit2Sharp.Core.NativeMethods' threw an exception.</Message><StackTrace>   at LibGit2Sharp.Core.NativeMethods.RemoveHandle() 
[38932]    at LibGit2Sharp.Core.NativeMethods.LibraryLifetimeObject.Finalize()</StackTrace><ExceptionString>System.TypeInitializationException: The type initializer for 'LibGit2Sharp.Core.NativeMethods' threw an exception. ---&amp;gt; System.DllNotFoundException: Unable to load DLL 'git2-785d8c4': The specified module could not be found. (Exception from HRESULT: 0x8007007E) 
[38932]    at LibGit2Sharp.Core.NativeMethods.git_libgit2_init() 
[38932]    at LibGit2Sharp.Core.NativeMethods.LibraryLifetimeObject..ctor() 
[38932]    at LibGit2Sharp.Core.NativeMethods..cctor() 
[38932]    --- End of inner exception stack trace --- 
[38932]    at LibGit2Sharp.Core.NativeMethods.RemoveHandle() 
[...]

PR #928, opened 2017-03-20
Use ProvideBindingRedirection instead of ProvideCodeBase

jcansdale:
In PR #914 I used the VS SDK `ProvideCodeBase` attribute to ensure that a particular version of an assembly will load from a specific CodeBase (into the `Load` context). For example:
```c#
[assembly: ProvideCodeBase(AssemblyName = "GitHub.Exports",
    CodeBase = @"$PackageFolder$\GitHub.Exports.dll")]
```

This can help if another extension references the exact version of an assembly that has been installed (with a `ProvideCodeBase` attribute). If however a different version is referenced, another assembly will be loaded with a different codebase, static fields and types. Unless our assembly has been designed with side-by-side execution in mind, the results could be unpredictable.

A resolution for this would be to sweep up multiple versions of an using the `ProvideBindingRedirection` attribute. For example:
```c#
[assembly: ProvideBindingRedirection(AssemblyName = "GitHub.Exports",
    CodeBase = @"$PackageFolder$\GitHub.Exports.dll",
    OldVersionLowerBound = "0.0.0.0", OldVersionUpperBound = "2.2.0.9")]
```

This would prevent old versions of an assembly being loaded at the same time as the current one (they would be redirected to the current one). This would also allowing 3rd party extensions that could work with multiple newer versions of GHfVS (rather than having to be built against a specific version).

This PR depends on #914.
See #926 and #923.

### DIFF
--- a/src/GitHub.VisualStudio/Properties/AssemblyInfo.cs
+++ b/src/GitHub.VisualStudio/Properties/AssemblyInfo.cs
@@ -1,4 +1,5 @@
-﻿using System.Reflection;
+﻿using System;
+using System.Reflection;
 using System.Runtime.InteropServices;
 using Microsoft.VisualStudio.Shell;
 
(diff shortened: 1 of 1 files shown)


jcansdale:
There appear to be some bugs when using the `ProvideBindingRedirection` attribute, which I had to work around. The following doesn't appear to work:

```
[assembly: ProvideBindingRedirection(AssemblyName = "GitHub.Exports",
    CodeBase = @"$PackageFolder$\GitHub.Exports.dll",
    OldVersionLowerBound = "0.0.0.0", OldVersionUpperBound = "Current")]
```
(IIRS, this *only* redirects `0.0.0.0` to the current version 😭)

I had more luck with the following:
```
[assembly: ProvideBindingRedirection(AssemblyName = "GitHub.Exports",
    CodeBase = @"$PackageFolder$\GitHub.Exports.dll",
    OldVersionLowerBound = "LowestMajor", OldVersionUpperBound = "Current")]
```

This redirects all assemblies with the same major version number to the current version.

We could use this for `Splat`, `Rothko`, etc. who's assembly versions we don't have constants defined for. For the `GitHub.*` assemblies, we could take advantage of `AssemblyVersionInformation.Version`.

I believe the following would work:
```
[assembly: ProvideBindingRedirection(AssemblyName = "GitHub.Exports",
    CodeBase = @"$PackageFolder$\GitHub.Exports.dll",
    OldVersionLowerBound = "0.0.0.0", OldVersionUpperBound = AssemblyVersionInformation.Version)]
```

shana:
> We could use this for Splat, Rothko, etc. who's assembly versions we don't have constants defined for.

Do we need it for those? We will likely never bump versions on our dependencies (except Octokit), so if we need to hardcode things for them, that should be fine.

jcansdale:
> Do we need it for those? We will likely never bump versions on our dependencies (except Octokit), so if we need to hardcode things for them, that should be fine.

Using the following form (using `LowestMajor` and `Current`) works fine if we're only interested in a single major version number:
```
[assembly: ProvideBindingRedirection(AssemblyName = "Octokit",
    CodeBase = @"$PackageFolder$\Octokit.dll",
    OldVersionLowerBound = "LowestMajor", OldVersionUpperBound = "Current")]
```
Since we're not likely to bump versions, this certainly applies here. 😄 

It just confused me for a while that it isn't possible to use `0.0.0.0` in combination with `Current` (it failed silently and only bound `0.0.0.0`) IIRC! 😕

Here are the current versions we're using:
```
LibGit2Sharp, Version=0.22.0.0, Culture=neutral, PublicKeyToken=7cbde695407f0333
Rothko, Version=0.0.1.0, Culture=neutral, PublicKeyToken=9f664c41f503810a
Splat, Version=1.6.2.0, Culture=neutral, PublicKeyToken=79cb2fa33d9108a3
Octokit, Version=0.22.0.0, Culture=neutral, PublicKeyToken=5525be5bc50478ea
```
Can I double-check these are use keys generated by us for GHfVS?

shana:
All the submodule projects have their own keys checked in, so you can check those. Libgit2sharp is a nuget package so that's not signed by us. There's also ReactiveUI and Akavache on the list of things we depend on and sign ourselves, I believe?

jcansdale:
> Libgit2sharp is a nuget package so that's not signed by us.

I've changed it to only provide redirections for GitHub.* assemblies. People shouldn't be using a different version of these inside the same VS instance.

Doing this for Libgit2sharp or any other assembly that might be used by another extension would be a bad thing.

I'm still wondering about Octokit, because it is part of our interface and the version is likely to be bumped in future. It probably should be redirected as well...

shana:
I assume we have to wait for #914 to get merged before this one can be merged?

jcansdale:
> I assume we have to wait for #914 to get merged before this one can be merged?

@shana Yes, and ideally #947 as well. I tried merging this one in as well, but it was proving to be a pain (the combination of a .sln and submodule merge 😕). It obviously can be done, but I'm sure there is less painful way! I've suggested a zoom chat with Jasmine to discuss.

jcansdale:
Fixes #926

PR #934, opened 2017-03-22
Display indeterminate progress bar when updating PR details view.

grokys:
Added a new `IHasLoading` interface in addition to `IHasBusy` and use `IsLoading` when the PR details pane is initially loading and then    `IsBusy` afterwards. When `IsBusy` display an indeterminate progress bar at the top of the view instead of hiding the content and displaying a spinner.

Depends on #933


shana:
There's some duplicated code between `SimpleViewUserControl` and `BusyStateView`, we should dedupe this into one class that rules them all.

Issue #936, opened 2017-03-23
Select repository if one already exists when using `Open with Visual Studio`


jcansdale:
When opening Visual Studio with the `/GitClone` open, it would be nice if we could select an already cloned repo (if one exists). This way the user could avoid ending up with multiple clones of the same repo on their machine.

I've prototyped this functionality here:
https://github.com/editor-tools/GitHub.VisualStudio.Contrib/pull/2

PR #937, opened 2017-03-24
WIP: Automatically select repo (if one already exists) when using 'Open in Visual Studio'

jcansdale:
When using 'Open in Visual Studio' button it's easy to end up with many duplicate repos on your hard drive.

image

The idea behind this PR is to automatically select a repo (if one already exists), so that it's obvious to the user if they're about to create a duplicate. This also makes the 'Open in Visual Studio' button useful if the user wants to open an existing repo.

image

This functionality could be expanded to allow the automatic opening of a PR (like you can with GitHub Desktop). BTW, this seemed to be broken last time I checked. 😕 

image

For example, something similar to the current `github-windows` protocol:
```github-windows://openRepo/https://github.com/github/VisualStudio?branch=jcansdale%2Fimprove-OpenInVS```

Could be used in `git-client` (which Visual Studio uses), e.g:
```git-client://clone?repo=https%3A%2F%2Fgithub.com%2Fgithub%2FVisualStudio%3Fbranch%3Djcansdale%252Fimprove-OpenInVS```

Although the URL couldn't be cloned directly, we could intercept it and offer to clone, checkout and open the corresponding PR.

We could also open a specific file (like GitHub Desktop):
github-windows://openRepo/https://github.com/jcansdale/StaticMocks?branch=master&filepath=example.cs

### CODE REVIEW

File: GitHubVS.sln [hunk 1]
```diff
@@ -2,6 +2,8 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
 VisualStudioVersion = 14.0.25420.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.26228.4
```
shana: Let's avoid doing these changes, they're just going to pop up again for the next person.

src/GitHub.Exports/Services/GitCloneCommandLineService.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                return null;
+            }
+
+            return optionValue;
+        }
+
+        public bool TryOpenRepository(string cloneUrl)
+        {
+            if(TryOpenLocalClonePath(cloneUrl))
```
shana: Space missing after the `if`

src/GitHub.Exports/Services/GitCloneCommandLineService.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                return true;
+            }
+
+            if(TryOpenKnownRepository(cloneUrl))
```
shana: Space missing after the `if`

src/GitHub.Exports/Services/GitCloneCommandLineService.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        bool TryOpenKnownRepository(string cloneUrl)
+        {
+            var repos = vsGitServices.GetKnownRepositories();
+            foreach (var repo in repos)
+            {
+                if (cloneUrl == repo.CloneUrl)
+                {
+                    if(vsServices.TryOpenRepository(repo.LocalPath))
```
shana: Space missing after the `if`

src/GitHub.Exports/Services/GitCloneCommandLineService.cs
```diff
@@ -0,0 +1,139 @@
...
+                }
+            }
+
+            return false;
+        }
+
+        bool TryOpenLocalClonePath(string cloneUrl)
+        {
+            var cloneUri = FindGitHubCloneUri(cloneUrl);
+            if(cloneUri == null)
```
shana: Space missing after the `if`

src/GitHub.Exports/Services/GitCloneCommandLineService.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                return false;
+            }
+
+            var clonePath = vsGitServices.GetLocalClonePathFromGitProvider();
+            if(clonePath == null)
```
shana: Space missing after the `if`

src/GitHub.Exports/Services/GitCloneCommandLineService.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            try
+            {
+                var uriString = new UriString(cloneUrl);
+                if (uriString.Host != "github.com")
+                {
+                    return null;
+                }
+
+                if(string.IsNullOrEmpty(uriString.Owner) || string.IsNullOrEmpty(uriString.RepositoryName))
```
shana: Space missing after the `if`

src/GitHub.Exports/Services/GitCloneCommandLineService.cs
```diff
@@ -0,0 +1,139 @@
...
+                {
+                    return null;
+                }
+
+                return uriString;
+            }
+            catch(Exception e)
+            {
+                Trace.WriteLine(e);
```
shana: The `VsOutputLogger` class provides logging services for code running at startup and in Team Explorer. It doesn't do Trace logging, but there's a trace listener there if you want to hook it up (or call Trace from the methods there if you want to do it the other way around).

For other logging, we're using NLog right now, btw. PR #573 introduces Serilog for logging purposes, which is a much lighter weight solution for logging that we could use for everything. Maybe that's something worth revisiting and cleaning up so we can fix up our logging once and for all.

src/GitHub.VisualStudio/GitCloneCommandLinePackage.cs
```diff
@@ -0,0 +1,36 @@
...
+        public const string PackageGuidString = "379dad24-c111-4d73-89ff-e8f91cab68db";
+
+        protected override void Initialize()
+        {
+            try
+            {
+                var sp = (IGitHubServiceProvider)GetService(typeof(IGitHubServiceProvider));
+                var gitCloneCommandLineService = sp.ExportProvider.GetExportedValue<IGitCloneCommandLineService>();
+                var cloneUrl = gitCloneCommandLineService.FindGitCloneOption();
+                if(cloneUrl != null)
```
shana: Space missing after the `if`

src/GitHub.VisualStudio/GitCloneCommandLinePackage.cs
```diff
@@ -0,0 +1,36 @@
...
+                {
+                    gitCloneCommandLineService.TryOpenRepository(cloneUrl);
+                }
+            }
+            catch(Exception e)
+            {
+                Trace.WriteLine(e);
```
shana: See previous comment on logging

src/GitHub.Exports/Services/GitCloneCommandLineService.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+
+            if (vsAppCommandLine == null)
+            {
+                return null;
+            }
+
+            int isPresent;
+            string optionValue;
+            if (ErrorHandler.Failed(vsAppCommandLine.GetOption("GitClone", out isPresent, out optionValue)))
```
shana: This "GitClone" string and the one in the `ProvideAppCommandLine` package atributes should probably be a const shared by both, so that it doesn't go out of sync. Not that it will, but given that it's likely we'll implement more of these and this code will likely be copy pasted around, it's good practice to avoid mistakes.

src/GitHub.Exports/Services/GitCloneCommandLineService.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            if (isPresent == 0)
+            {
+                return null;
+            }
+
+            return optionValue;
+        }
+
+        public bool TryOpenRepository(string cloneUrl)
```
shana: Wouldn't this method be better in `IVSServices`? We already have a `TryOpenRepository` there, it would make sense to have an overload that takes a `UriString` (which is what this is really taking in), and run this logic there.

jcansdale: I tried moving things around a bit and `IVSGitServices` seems to be an even better fit.  `IVSServices` feels like a lower level service than `IVSGitServices` (it doesn't know anything about Git).

jcansdale:
Fixes #936

jcansdale:
This PR uses Visual Studio's built in GitClone command line argument. Unfortunately this isn't expressive enough to also target a PR. We will likely want to tie this kind of functionality with some kind of open from GitHub URL command (which could be passed on the command line). 

It was a good experiment, but it's now stale and won't be merged.

PR #941, opened 2017-03-28
WIP: Add links to documentation.

grokys:
Adds a welcome message with a link to docs to team explorer (but only if Team Explorer isn't already displaying the "Install 3rd Party Tools"
message) and a Help button to the GitHub pane.

TODO:
- [x] Decide where the docs are to be hosted and update links

### DIFF
--- a/src/GitHub.TeamFoundation.14/Home/GitHubHomeSection.cs
+++ b/src/GitHub.TeamFoundation.14/Home/GitHubHomeSection.cs
@@ -13,6 +13,9 @@ using System.Diagnostics;
 using GitHub.Extensions;
 using System.Windows.Input;
 using ReactiveUI;
+using GitHub.VisualStudio.UI;
+using GitHub.Settings;
+using System.Windows.Threading;
 
 namespace GitHub.VisualStudio.TeamExplorer.Home
 {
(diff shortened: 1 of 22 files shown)


src/GitHub.TeamFoundation.14/Home/GitHubHomeSection.cs
```diff
@@ -107,6 +147,36 @@ void DoOpenOnGitHub()
             visualStudioBrowser?.OpenUrl(ActiveRepo.CloneUrl.ToRepositoryUrl());
         }
 
+        void ShowWelcomeMessage()
+        {
+            var welcomeMessageGuid = new Guid("C529627F-8AA6-4FDB-82EB-4BFB7DB753C3");
```
shana: Probably a good idea to put this and other GUID next to the other GUIDs we have in the Guids class, so they're easy to keep track of

src/GitHub.TeamFoundation.14/Home/GitHubHomeSection.cs
```diff
@@ -21,22 +24,59 @@ namespace GitHub.VisualStudio.TeamExplorer.Home
...
+                        ShowWelcomeMessage();
+                    }
+                };
+                timer.Start();
+            }
+        }
+
+        bool IsGitToolsMessageVisible()
+        {
+            return teamExplorerServices.IsNotificationVisible(new Guid("DF785C7C-8454-4836-9686-D1C4A01D0BB9"));
```
shana: Probably a good idea to put this and other GUID next to the other GUIDs we have in the Guids class, so they're easy to keep track of

src/GitHub.VisualStudio/UI/Views/GitHubPaneViewModel.cs
```diff
@@ -125,6 +128,15 @@ public override void Initialize(IServiceProvider serviceProvider)
                 },
                 true);
 
+           serviceProvider.AddCommandHandler(GuidList.guidGitHubToolbarCmdSet, PkgCmdIDList.helpCommand,
+                () => true,
+                () =>
+                {
+                    browser.OpenUrl(new Uri("https://visualstudio.github.com/docs"));
```
shana: There's a const string for this url defined somewhere else. Maybe these urls should go into one of the shared classes we have for defining const data, that way we don't have to hunt down all the instances when we want to change something.

paladique:
Per internal discussion we decided to point to the documentation in the repo for now.

PR #942, opened 2017-03-29
Merge SimpleUserControl and BusyStateView.

grokys:
Merges `SimpleUserControl` and `BusyStateView` into a single `ViewBase` control.

This control will:
- Display a spinner if the view model implements `IHasLoading` and `IsLoading` is true
- Display an indeterminate progress bar if the view model implements `IHasBusy` and `IsBusy` is true.

If the view model implements neither of these interfaces, then it will act the same as a standard `UserControl`.

Initially my intention was to keep 2 controls, with `BusyStateView` inheriting from `SimpleUserControl` but this turned out not to be possible, as `BusyStateView` would have needed to be a generic class with a control template, which is not possible in WPF.


src/GitHub.VisualStudio/UI/Views/PullRequestListView.xaml.cs
```diff
@@ -59,8 +59,10 @@ public PullRequestListView(IGitHubServiceProvider serviceProvider)
 
         bool disposed;
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA1816:CallGCSuppressFinalizeCorrectly")]
```
shana: Can probably kill this warning since `Dispose()` is not defined here anymore

src/GitHub.VisualStudio/UI/Views/PullRequestListView.xaml.cs
```diff
@@ -59,8 +59,10 @@ public PullRequestListView(IGitHubServiceProvider serviceProvider)
...
-        public void Dispose()
+        protected override void Dispose(bool disposing)
         {
+            base.Dispose(disposing);
+
             if (disposed) return;
```
shana: For the pattern to be correctly implemented, this whole block should be inside `if (disposing)`

src/GitHub.UI.Reactive/Properties/AssemblyInfo.cs
```diff
@@ -1,6 +1,16 @@
...
 [assembly: Guid("885a491c-1d13-49e7-baa6-d61f424befcb")]
+
+[assembly: ThemeInfo(
+    ResourceDictionaryLocation.None, //where theme specific resource dictionaries are located
+                                     //(used if a resource is not found in the page, 
+                                     // or application resource dictionaries)
+    ResourceDictionaryLocation.SourceAssembly //where the generic resource dictionary is located
+                                              //(used if a resource is not found in the page, 
+                                              // app, or any theme specific resource dictionaries)
+    )]
```
shana: Oh this is interesting!

src/GitHub.UI.Reactive/Controls/ViewBase.cs
```diff
@@ -0,0 +1,220 @@
...
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Releases the managed or unmanaged resources held by the control.
+        /// </summary>
+        protected virtual void Dispose(bool disposing)
+        {
+            subscriptions?.Dispose();
```
shana: This should be in a `if (disposing)` block and a `if (!disposed)` block

shana:
Cross-linking #943 (depends on this)

src/GitHub.UI.Reactive/Controls/ViewBase.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        /// <summary>
+        /// Releases the managed or unmanaged resources held by the control.
+        /// </summary>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                subscriptions?.Dispose();
+                subscriptions = null;
```
shana: This code should only be called once, and it is possible to call Dispose() multiple times, so it should be in a `if (!disposed)` block (I don't usually throw exceptions if Dispose is called multiple times because I figure it's a hard thing to control, and it's nice to be able to do it and not worry about whether some other code has already done it). Alternatively, you can do `if (disposed) return` right before this.

src/GitHub.VisualStudio/UI/Views/PullRequestListView.xaml.cs
shana: For types with finalizers, you only want to access managed data if you're inside the `disposing` block. Not that it will happen here, but with finalizers in play, there is a (very very very) small chance that managed instance data is no longer valid when they're running, so accessing instance fields could crash (it only really matters if you're disposing managed and native data, really). `disposing` is the flag that tells you whether it's safe to access managed data during this process, so the `if (disposed)` check should be inside that block (and since `disposed` is only set inside that block, doing the check here or there doesn't change things in practice - it's just the safest pattern for when you're disposing managed and native data and there's finalizers in play)

PR #943, opened 2017-03-30
Display busy indicator in PR list.

grokys:
- Implemented `IHasBusy` on `PullRequestListViewModel`
- Renamed `IsLoaded` to `IsBusy` (and inverted the logic)

Depends on #942


shana:
Cross-linking #944 (depends on this)

PR #944, opened 2017-03-30
Fix PR list filters

### LINKED ISSUE #843: Pull request list filters disabled for extended period of time
In version 2.2.0.7, when switching to the pull request list in either direction (cancelling creation form, refreshing, pressing the PR toolbar icon, clicking TE Pull Request button) the filters at the top of the list stay disabled well after the list is loaded. They eventually are enabled, around 10 -15 seconds later. 

The problem in detail is this: 

> We save the current filter setting for loading the next time the list is shown, however, that setting can't be applied until the PR list is loaded because the filter choices aren't known until the list is loaded; we can only set the filter to something that's in the dropdown.
> 
> However the problem in #450 was that if the user changed the filter setting whilst the list was loading, then that choice would be overridden when the list finishes loading!
> 
> A better solution might be to allow the user to change the filters while loading, but if this happens, don't restore the state for that filter once load finishes.

I was able to take my time and open the snipping tool and get a screenshot before they were enabled, with a few more seconds to spare:

capture

Looks like this when loaded:
loaded

### LINKED ISSUE #880: Pull Requests Author filter keeps getting reset
- GitHub Extension for Visual Studio version: 
- Visual Studio version: 2.2.0.8 (dev build)

At one point I had the Author filter set to myself. Now when I set it to `[None]`, it keeps getting reset. It can take a good few seconds, so maybe it's when the PR list finishes loading?

PS. Should `[None]` really be `[Any]`?

grokys:
Save changes to filters settings as soon as they are changed. This means that they don't have to be disabled while loading.

Fixes #843
Fixes #880

Depends on #943


shana:
Feels like the changes in this PR are going to conflict with #943 once that one is merged. Perhaps rebasing this one off of #943 is preferable?

grokys:
This one depends on #943, I mentioned it in the description :)

PR #945, opened 2017-03-31
Make `SharedDictionaryManager` work inside XAML Designer

jcansdale:
The custom `ResourceDictionary` implementation `SharedDictionaryManager`, was failing when running in design mode. This caused resources not to be found in the designer and errors when looking at the XAML source.

image

This PR fixes the `SharedDictionaryManager` so that it can execute without errors in design mode.

image

It needs to convert `file` URIs that the designer uses to `pack` URIs that contain an assembly context. This allows the designer to resolve custom types used in XAML (i.e. `SharedDictionaryManager`) to their home assembly. This seems to fix the `Cannot create unknown type` errors.

This PR also automatically disables Visual Studio styles when running inside the designer.

If it works as hoped, there shouldn't be any more need for `#if !XAML_DESIGNER`. 😉 

Depends on #914.


src/GitHub.UI/Helpers/SharedDictionaryManagerBase.cs
```diff
@@ -0,0 +1,46 @@
...
+            var assemblyPrefix = "/src/";
+            var assemblyIndex = url.LastIndexOf(assemblyPrefix);
+            if(assemblyIndex == -1)
+            {
+                return inUri;
+            }
+
+            assemblyIndex += assemblyPrefix.Length;
+            var pathIndex = url.IndexOf('/', assemblyIndex);
+            if(pathIndex == -1)
```
shana: Missing space after `if`

src/GitHub.UI/Helpers/SharedDictionaryManagerBase.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            if (inUri.Scheme != "file")
+            {
+                return inUri;
+            }
+
+            var url = inUri.ToString();
+            var assemblyPrefix = "/src/";
+            var assemblyIndex = url.LastIndexOf(assemblyPrefix);
+            if(assemblyIndex == -1)
```
shana: Missing space after `if`

src/GitHub.VisualStudio.UI/Helpers/ThemeDictionaryManager.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                baseThemeUri = null;
+                if (!isInDesignMode)
+                {
+                    VSColorTheme.ThemeChanged -= OnThemeChange;
+                }
+            }
+        }
+
+        void InitTheme(Uri themeUri)
```
grokys: You know what would be nice? Being able to change the theme in the designer. We could expose a `ThemeName` property or similar on this class which would allow you to toggle the theme at design time and would have no effect at runtime. This should probably go in a different PR though ;)

jcansdale: Definitely! I wonder if there is any way to infer the current VS theme from the designer? It's running in a different process which likely complicates things.

shana: There might be a way of hacking it by sampling a known color to figure out which theme it is, perhaps?

grokys: Personally I'd rather just set a property so that I can see e.g. how my design looks in the dark theme without having to go to the options menu, set the theme and then change it back when I've finished...

grokys: It could even be a design-time attached property which could be set on the `UserControl` which would be ignored at runtime I think.

PR #947, opened 2017-03-31
Fix automation id submodule naming

paladique:
- Renamed to `GitHubVSTestAutomationIDs`
- Updated submodule so it's using the right GUID
- Renamed the submodule's path in the .sln file & required .csproj files


jcansdale:
@paladique As discussed, I've moved the AutomationIDs into GitHub.UI (which is referenced by all UI projects that need them). I've also incorporated the unit test from GitHubVSTestAutomationIDs. This lets us remove the submodule and avoid changing the .sln file (sparing us some friction in the future).

The low tech approach would be to build `GitHub.UI.dll` and copy it to libs folder inside `GitHubVSTestAutomation` (instead of maintaining the `GitHubVSTestAutomationIDs` submodule).

Alternatively you could write a script that pulls down the following file (after its merged) and incorporates it in the test automation project:
```
https://github.com/github/VisualStudio/blob/fixes/master/src/GitHub.UI/TestAutomation/AutomationIDs.resx
https://github.com/github/VisualStudio/blob/fixes/master/src/GitHub.UI/TestAutomation/AutomationIDs.Designer.cs
```

Personally I'd go with the low tech approach! 😉

paladique:
👍 on low tech approach. On less submodule to worry about!

PR #950, opened 2017-04-04
Make `SharedDictionaryManager` work inside XAML Designer

jcansdale:
This is same as #945 except a PR to master.


jcansdale:
@shana, This is the same as #950, except a PR with master. Let me know if you. 👍

GitHubVS.sln
```diff
@@ -544,6 +436,18 @@ Global
 		{50E277B8-8580-487A-8F8E-5C3B9FBF0F77}.XamlDesign|Any CPU.Build.0 = Release|Any CPU
 		{50E277B8-8580-487A-8F8E-5C3B9FBF0F77}.XamlDesign|x86.ActiveCfg = Release|Any CPU
 		{50E277B8-8580-487A-8F8E-5C3B9FBF0F77}.XamlDesign|x86.Build.0 = Release|Any CPU
+		{9453C37F-44CF-463D-B563-5C3110CEF5E7}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
shana: This project isn't actually in the solution, is there something missing?

src/GitHub.UI/Helpers/SharedDictionaryManagerBase.cs
```diff
@@ -0,0 +1,115 @@
...
+        static SharedDictionaryManagerBase()
+        {
+            // Avoid leaking resources at design time.
+            sharedDictionaries = GetAppDomainSharedDictionaries();
+            disposables = GetAppDomainDisposables();
+        }
+
+        public virtual new Uri Source
+        {
+            get { return source; }
```
shana: NullGuard will throw a NRE if the getter is called when `source` hasn't been set yet. Mark this with `[return: AllowNull]` to allow a null return, or make sure `source` is initialized to something to avoid the NRE.

(technically it won't be called before the setter, but the XamlDesigner might decide to do something wonky, and throwing here will cause it to go boom, so probably better safe than sorry)

src/GitHub.UI/Helpers/SharedDictionaryManagerBase.cs
```diff
@@ -0,0 +1,115 @@
+﻿using System;
+using System.Windows;
+using System.Collections.Generic;
+
+namespace GitHub.Helpers
+{
+    public class SharedDictionaryManagerBase : ResourceDictionary
+    {
+        static IDictionary<Uri, ResourceDictionary> sharedDictionaries;
+        static IList<IDisposable> disposables;
```
shana: You can use a `CompositeDisposable` type here, it's designed to handle disposables in an optimal way (you can just call `Dispose()` on it and it will handle disposing everything it holds.

src/GitHub.UI/Helpers/SharedDictionaryManagerBase.cs
```diff
@@ -0,0 +1,115 @@
...
+            var assemblyName = url.Substring(assemblyIndex, pathIndex - assemblyIndex);
+            var path = url.Substring(pathIndex + 1);
+
+            return new Uri($"pack://application:,,,/{assemblyName};component/{path}");
+        }
+
+        static IDictionary<Uri, ResourceDictionary> GetAppDomainSharedDictionaries()
+        {
+            var name = typeof(SharedDictionaryManagerBase).FullName + ".SharedDictionaries";
+            var sharedDictionaries = GetAppDomainData<Dictionary<Uri, ResourceDictionary>>(name);
```
shana: The `sharedDictionaries` and the `disposables` list are already static across the domain, why save them in the appdomain?

src/GitHub.VisualStudio.UI/Helpers/ThemeDictionaryManager.cs
```diff
@@ -0,0 +1,70 @@
...
+            get { return base.Source; }
+            set
+            {
+                InitTheme(value);
+                base.Source = GetCurrentThemeUri();
+            }
+        }
+
+        public void Dispose()
+        {
```
shana: The pattern for `Dispose` should be:

```
bool disposed;
void Dispose(bool disposing)
{
    if (disposed) return;
    disposed = true;
    if (disposing)
    {
        // dispose managed resources here
    }
}

public void Dispose()
{
    Dispose(true);
    // GC.SuppressFinalize(this); // if there's a finalizer in this class or any subclasses
}
````

File: src/GitHub.UI/Helpers/SharedDictionaryManagerBase.cs [hunk 6]
```diff
@@ -0,0 +1,115 @@
+﻿using System;
+using System.Windows;
+using System.Collections.Generic;
+
+namespace GitHub.Helpers
+{
+    public class SharedDictionaryManagerBase : ResourceDictionary
```
shana: Given that this class needs to manage disposable resources, it should be `IDisposable` (if you have a `CompositeDisposable` instead of a `List<IDisposable>`, the code analysis will let you know this 😉

src/GitHub.UI/Helpers/SharedDictionaryManagerBase.cs
shana: Nevermind, `CompositeDisposable` is an rx thing and not available here

src/GitHub.UI/Helpers/SharedDictionaryManagerBase.cs
jcansdale: The designer does some shadow assembly tricks, so the same assembly can end up being multiple times. I'm doing this to prevent stale caches from being left behind.

src/GitHub.UI/Helpers/SharedDictionaryManagerBase.cs
jcansdale: As discussed, this is a Rx class, so will avoid taking this dependency. 😄

src/GitHub.UI/Helpers/SharedDictionaryManagerBase.cs
```diff
@@ -0,0 +1,115 @@
...
+        static IDictionary<Uri, ResourceDictionary> GetAppDomainSharedDictionaries()
+        {
+            var name = typeof(SharedDictionaryManagerBase).FullName + ".SharedDictionaries";
+            var sharedDictionaries = GetAppDomainData<Dictionary<Uri, ResourceDictionary>>(name);
+
+            sharedDictionaries.Clear();
+            return sharedDictionaries;
+        }
+
+        static IList<IDisposable> GetAppDomainDisposables()
```
shana: Feels like an odd name for a method that disposes of a cached list before returning it empty. Perhaps `InitializeAppDomainDisposables()` would be more descriptive of what this is doing (and when it should be called)?

src/GitHub.UI/Helpers/SharedDictionaryManagerBase.cs
```diff
@@ -0,0 +1,115 @@
...
+                return inUri;
+            }
+
+            var assemblyName = url.Substring(assemblyIndex, pathIndex - assemblyIndex);
+            var path = url.Substring(pathIndex + 1);
+
+            return new Uri($"pack://application:,,,/{assemblyName};component/{path}");
+        }
+
+        static IDictionary<Uri, ResourceDictionary> GetAppDomainSharedDictionaries()
```
shana: `InitializeAppDomainSharedDictionaries` or something similar is probably a better name for this method?

src/GitHub.VisualStudio.UI/Helpers/ThemeDictionaryManager.cs
jcansdale: I tried to sidestep this by marking the class as `sealed`. Maybe this is a bit antisocial, but it wasn't designed with extenders in mind. It kept CodeAnalysis from complaining anyway. 😉

jcansdale:
@shana I think it's time to stick of fork in this one. I've incorporated your suggestions, including lifecycle management using a single disposable class. Test coverage also also much better now.

Finally, I've done the grand consolidation of the 3 `SharedDictionaryManager`s. This touched a bunch of XAML files, but they're minor tweaks.

PR #955, opened 2017-04-19
Normalize Dispose implementations across the codebase

shana:
Went through our various `Dispose()` implementations and normalized them to the same pattern of:

```
bool disposed;
private [virtual/override] void Dispose(bool disposing)
{
  if (disposed) return;
  if (disposing)
  {
    disposed = true;
    // dispose managed stuff
  }
  // dispose native stuff
}

public void Dispose()
{
  Dispose(true);
  GC.SuppressFinalize(this);
}
```

### DIFF
--- a/src/CredentialManagement/Credential.cs
+++ b/src/CredentialManagement/Credential.cs
@@ -63,12 +63,12 @@ namespace GitHub.Authentication.CredentialManagement
         bool disposed;
         void Dispose(bool disposing)
         {
+            if (disposed) return;
             if (disposing)
             {
-                if (disposed) return;
+                disposed = true;
                 SecurePassword.Clear();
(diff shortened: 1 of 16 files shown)


grokys:
I notice some of the `void Dispose(bool disposing)` methods aren't virtual. Isn't the point of the dispose pattern to allow for derived classes to add their own dispose implementation? Without `virtual` there are no advantage of using the dispose pattern over just implementing `Dispose()`.

shana:
@grokys yes indeed, good point. Also, I seem to have broken the credential cache tests with this one, so it might be that this pattern is still not completely correct, I'll take another crack at it.

sguthals:
@shana - want to create an issue for this and re-visit when there is time to think about it?

PR #956, opened 2017-04-19
Add `GitHubVS.proj` script to build `Debug`, `Release` and `ProductComponent` flavors

jcansdale:
The `GitHubVS.proj` MSBuild file is intended to be an alternative target for our CI build scripts. It will build build `Debug`, `Release` and `ProductComponent` flavors.

The artifacts will be placed in the following locations:

This is an 'Experimental' build that will install for the current user.
`build\Debug\GitHub.VisualStudio.vsix`

This is a 'Release' build that will install for all users.
`build\Release\GitHub.VisualStudio.vsix`

This is a 'Release build with 'IsProductComponent=true'.
`build\ProductComponent\GitHub.VisualStudio.zip`
(which contains 'GitHub.VisualStudio.vsix' and 'GitHub.VisualStudio.vsix')

When the CA scripts are updated to target `GitHubVS.proj` instead of `GitHubVS.sln`, they shouldn't set `Configuration`, `IsExperimental` or `IsProductComponent` properties.

### DIFF
--- a/GitHubVS.proj
+++ b/GitHubVS.proj
@@ -1,20 +1,23 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">  
     <PropertyGroup>
       <SolutionFile>GitHubVS.sln</SolutionFile>
-      <OutputDirectory>$(MSBuildProjectDirectory)</OutputDirectory>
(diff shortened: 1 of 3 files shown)


jcansdale:
This is now all done in scripts.
See: https://github.slack.com/archives/C110888DQ/p1492680640233071

PR #960, opened 2017-04-26
Use Timer in usage tracker.

grokys:
Use Timer instead of DispatcherTimer in usage tracker. Because `TimerTick` does IO, we should be running on a b/g thread.

### DIFF
--- a/src/GitHub.VisualStudio/Services/UsageTracker.cs
+++ b/src/GitHub.VisualStudio/Services/UsageTracker.cs
@@ -11,21 +11,21 @@ using Task = System.Threading.Tasks.Task;
 using GitHub.Extensions;
 using System.Threading.Tasks;
 using GitHub.Helpers;
+using System.Threading;
 
 namespace GitHub.Services
 {
-    public class UsageTracker : IUsageTracker
+    public sealed class UsageTracker : IUsageTracker, IDisposable
     {
         const string StoreFileName = "ghfvs.usage";
         static readonly Calendar cal = CultureInfo.InvariantCulture.Calendar;
 
         readonly IGitHubServiceProvider gitHubServiceProvider;
-        readonly DispatcherTimer timer;
-
         IMetricsService client;
         IConnectionManager connectionManager;
         IPackageSettings userSettings;
         IVSServices vsservices;
+        Timer timer;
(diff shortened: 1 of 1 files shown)


src/GitHub.VisualStudio/Services/UsageTracker.cs
```diff
@@ -11,21 +11,21 @@
 using GitHub.Extensions;
 using System.Threading.Tasks;
 using GitHub.Helpers;
+using System.Threading;
 
 namespace GitHub.Services
 {
-    public class UsageTracker : IUsageTracker
+    public sealed class UsageTracker : IUsageTracker, IDisposable
```
jcansdale: Sealing your disposable classes? I see what you did there. 😉

grokys: Ha yeah, easier than implementing the disposable pattern and getting it wrong.

grokys: Also this class is a singleton, so `Dispose` won't even be called.

jcansdale: Inline comments working again. Can I reply as well from inside VS? :-)

src/GitHub.VisualStudio/Services/UsageTracker.cs
```diff
@@ -11,21 +11,21 @@
...
     {
         const string StoreFileName = "ghfvs.usage";
         static readonly Calendar cal = CultureInfo.InvariantCulture.Calendar;
 
         readonly IGitHubServiceProvider gitHubServiceProvider;
-        readonly DispatcherTimer timer;
```
grokys: Testing comments.

src/GitHub.VisualStudio/Services/UsageTracker.cs
Haacked: LOL! I do this all the time in apps. Less so in libraries. But it's very convenient to just seal them until you need them not to be sealed. 😄

marlazinger200: 1

marlazinger200:
1

PR #961, opened 2017-04-27
Add support for Visual Studio Team Explorer 2017

### LINKED ISSUE #938: Extension doesn't work with Visual Studio Team Explorer 2017
- GitHub Extension for Visual Studio version: 2.2.0.9
- Visual Studio version: Visual Studio Team Explorer 2017

I installed GHfVS 'Extensions and Updates', but the extension didn't appear after relaunch.

See here:
https://www.visualstudio.com/en-us/news/releasenotes/vs2017-preview-relnotes

jcansdale:
From the Visual Studio website:

"Visual Studio Team Explorer 2017 is a rich, standalone client for accessing Team Foundation Server and Visual Studio Team Services and is free for any user. This install only includes Team Explorer so you can access version control and work item tracking without other IDE components."
https://www.visualstudio.com/en-us/news/releasenotes/vs2017-relnotes#te

At the very least, adding support for this could be useful for testing (a minimal install of VS with just the packages we need). It's also an obvious target for our extension!

image

At the moment all this PR does is allows the extension to load when the `InstallationTarget` is "Microsoft.VisualStudio.IntegratedShell".

We will likely want to check what components are available when loading the extension so we don't fail in any IntegratedShells that don't have Team Explorer available.

Fixes #938


meaghanlewis:
@jcansdale earlier today I got the same errors with PRs in certain repos while using the latest master build so I don't think what you're seeing is related to this PR.

I was going to open up an issue about it, but then I logged out and back in again and haven't reproduced this problem since.

jcansdale:
@meaghanlewis.

> I got the same errors with PRs in certain repos while using the latest master build so I don't think what you're seeing is related to this PR.

Ah, thanks for letting me know! Will you open an issue about it?  You can use the screenshot above.

PR #966, opened 2017-04-28
[WIP] Experimental feature: synchronize forks

shana:
When you fork a repo, it's hard to keep it in sync with the upstream repository. This experimental feature goes through all the remote branches of a configured "upstream" remote, grab all local branches which match by name, have not diverged in history but are behind the upstream remote branch, updates the local branches via a fast forward fetch, and pushes the updated local branches to the "origin" remote, so that the fork is synchronized on the server with upstream.

This is a really really old branch that's been lying around. The button is in Team Explorer Home which is probably not the best place (probably in Connect? Context menu somewhere? unsure), and I'm sure it needs some cleaning up.


sguthals:
@shana Is this the Fork PR you will pick up again after GDC? Or is this an older one that should be closed?

jcansdale:
Too many conflicts! 😉

PR #970, opened 2017-05-02
[SPIKE] Stop TeamFoundation MEF exceptions when installing

### LINKED ISSUE #954: Lots of "Catalog construction errors" in Microsoft.VisualStudio.Default.err
- GitHub Extension for Visual Studio  version: 2.2.0.8
- Visual Studio version: VS 2015/2017

There are lots of exceptions in the error log file (7 pages!) from when Visual Studio constructs its MEF catalog. Here is the `Microsoft.VisualStudio.Default.err` log from VS 2017:
[Microsoft.VisualStudio.Default.err.txt](https://github.com/github/VisualStudio/files/927920/Microsoft.VisualStudio.Default.err.txt)

This is a known issue (and workaround) caused by a version change to the `Microsoft.TeamFoundation.Controls` assembly between VS 2015 and 2017.

The `GitHub.TeamFoundation.14` assembly is compiled against `Microsoft.TeamFoundation.Controls, Version=14.0.0.0` and intended to rune inside VS 2015. The `GitHub.TeamFoundation.15` assembly is compiled against `Microsoft.TeamFoundation.Controls, Version=15.0.0.0` and intended to rune inside VS 2017.

Unfortunately it isn't possible to specify different assemblies for different versions of Visual Studio inside a VSIX file, so Visual Studio attempts to install both assemblies which results in the errors.

*Update: is is possible in recent versions of Visual Studio 2017 See* https://github.com/github/VisualStudio/commit/c9703cd2a284840474b191449c52bb67b181a5e4

I wonder if it's possible to somehow forward MEF exports to the `GitHub.TeamFoundation.14` and `GitHub.TeamFoundation.15` assemblies and avoid these errors?

jcansdale:
This PR stops MEF exceptions being thrown as the extension installs. After installing the `Microsoft.VisualStudio.Default.err` file should be completely empty. This makes it a lot easier to stop MEF issues when working on other aspects of the extension.

All `GitHub.TeamFoundation` related discovery has been moved to the `GitHub.VisualStudio` assembly. MEF attributes have been left on the `GitHub.TeamFoundation.*` assemblies, but are no longer used by Visual Studio for discovery. I wanted to touch these projects as little as possible as part of this PR.

The `GitHub.TeamFoundation.15` is now used inside VS 2015 and 2017. The enhanced VS 2017 features are enabled in code rather than using `#if` directives. The `GitHub.TeamFoundation.14` project is no longer necessary.

Enabling support for VS 2015 and 2017 in the same assembly is done using an assembly resolver. The resolver is installed deterministically and doesn't race to be installed on time (which was an issue when we used an assembly resolver for the extension as a whole).

Fixes #954.


shana:
If the `GitHub.TeamFoundation.14` project is no longer necessary, then we should probably have one `GitHub.TeamFoundation` project and not a `GitHub.TeamFoundation.15` project, since the number in the name no longer matters?

src/GitHub.TeamFoundation.14/Services/VSGitServices.cs
```diff
@@ -60,20 +60,31 @@ public string GetLocalClonePathFromGitProvider()
...
 
         /// <inheritdoc/>
-        public async Task Clone(
-            string cloneUrl,
-            string clonePath,
-            bool recurseSubmodules,
-            object progress = null)
+        public async Task Clone(string cloneUrl, string clonePath, bool recurseSubmodules, object progress = null)
+        {
+#if (!TEAMEXPLORER14)
```
shana: If we're only using one project now, do we need these ifdefs?

src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
```diff
@@ -109,13 +109,18 @@
...
-    <Reference Include="Microsoft.TeamFoundation.Controls">
-      <HintPath>..\..\lib\14.0\Microsoft.TeamFoundation.Controls.dll</HintPath>
-    </Reference>
     <Reference Include="Microsoft.Expression.Interactions, Version=4.5.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
       <HintPath>..\..\packages\Expression.Blend.Sdk.WPF.1.0.1\lib\net45\Microsoft.Expression.Interactions.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="Microsoft.TeamFoundation.Controls, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\lib\15.0\Microsoft.TeamFoundation.Controls.dll</HintPath>
```
shana: TE doesn't ship backwards compatible DLLs, so how does referencing version 15 and calling it 14 work, exactly?

src/GitHub.VisualStudio/TeamFoundation/TeamFoundationResolver.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.IO;
+using System.Reflection;
+using System.Diagnostics.CodeAnalysis;
+using System.ComponentModel.Composition;
+
+namespace GitHub.VisualStudio.TeamFoundation
+{
+    [Export]
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public sealed class TeamFoundationResolver : IDisposable
```
shana: This and the other new MEF exports should be exporting interfaces and not the actual types, for easier testing.

src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
jcansdale: It doesn't. I'm not sure what's happening there. We'll only be referencing the `15.0` version.

src/GitHub.TeamFoundation.14/Services/VSGitServices.cs
jcansdale: Indeed, we can get rid of all `#if` and consolidate to a single project. I just wanted a commit where it worked but hardly touched the `GitHub.TeamFoundation.*` projects.

I'll now strip out the unnecessary stuff. 😄

src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
shana: Have you tried this in VS2015? I'm really curious how this will work given that the sections and navigation items implement Team Explorer interfaces that are not ABI compatible (because DLL versions)

jcansdale:
@shana I've addressed all of the issues you highlighted. I also did some refactoring/separating of concerns. Could you take another look? 😄

shana:
@jcansdale Looks like a good time to revisit this PR?

One thing that's important to ensure is that this works properly on a machine that only has Visual Studio 2015 installed. Machines with both will cause VS2015 to correctly locate assemblies that are only shipped with 2017 and might mask problems with a VS2015-only installation.

grokys:
@jcansdale what is the status of this PR?

jcansdale:
@grokys I think this is worth revisiting some time, but it certainly isn't ready to merge. The situation with the two tests of TeamFoundation assemblies completely messes up the MEF log. I wonder if the next version of VS will introduce a 3rd set?

I'm going to mark this as `[SPIKE]` the the time being, signifying that it's an experiment / proof of concept.

jcansdale:
There was a simpler fix for this issue merged in https://github.com/github/VisualStudio/pull/1705 for VS 2017.

Fixing for VS 2015 proved to be extremely difficult or maybe impossible.

PR #979, opened 2017-05-04
Add a delay to the login process.

grokys:
We're getting a lot of reports of failed logins, and looking at the logs it appears that using a token immediately after it's given out can fail. Add a delay of 1s (I tried less but this seems to give the best results) before reading `/user` after authorization.


shana:
triple-facepalm-no-words

marlazinger200:
1

PR #980, opened 2017-05-05
Show docs only when in GitHub repo once a week

### LINKED ISSUE #971: GitHub docs are promoted even when I am connected to TFVC Project
- GitHub Extension for Visual Studio  version:
image

- Visual Studio version:
image

- Connected to a TFVC Project on my local server
- Got prompt to create a workspace
- After the workspace got created I see some git docs:

image

paladique:
Updating work from #974 to branch from 2.2.0.11

An info panel with links to docs and training displays in Team Explorer, but can be misleading when not in a GitHub repo. Only display the panel when the user is in a GitHub hosted repo.

Fixes #971


paladique:
Not sure if I did this right 😅 . Just made a new branch from the 2.2.0.11, picked the one commit, and fixed the conflicts

grokys:
I changed the PR to point to `release/2.2.0.11` and there are 213 changes :(

src/GitHub.TeamFoundation.14/Home/GitHubHomeSection.cs
```diff
@@ -65,7 +65,7 @@ public class GitHubHomeSection : TeamExplorerSectionBase, IGitHubHomeSection
                 timer.Tick += (s, e) =>
                 {
                     timer.Stop();
-                    if (!IsGitToolsMessageVisible())
+                    if (!IsGitToolsMessageVisible() && IsVisible)
```
shana: Instead of letting the timer tick and then check if it's visible here, move the whole if block to line 109 where we already have a `IsVisible` check and don't bother with the timer, that way we know for sure that we are really visible. This timer might kick in before VS triggers the event that makes us visible, so if we have this here, we might not show the message even if we're on a github repo, if VS happens to be slow.

shana:
In case it gets missed with a force-push, added a comment https://github.com/github/VisualStudio/pull/980#discussion_r114989904

src/GitHub.TeamFoundation.14/Home/GitHubHomeSection.cs
```diff
@@ -106,8 +88,19 @@ bool IsGitToolsMessageVisible()
...
+                // already displaying the "Install 3rd Party Tools" message and the current repo is hosted on GitHub. 
+                if (!settings.HideTeamExplorerWelcomeMessage && !IsGitToolsMessageVisible())
+                {
+                    ShowWelcomeMessage();
+                }
                 IsLoggedIn = IsUserAuthenticated();
             }
+            else
+            {
+                teamExplorerServices.HideNotification(welcomeMessageGuid);
```
shana: I assume this doesn't do anything if our message isn't showing?

paladique: Nope, does nothing! Also checked the logs.

paladique:
This now displays once a week. I copied some code from `UsageTracker` and `DateTimeOffset` to compare weeks. Not sure where to place it, but it probably shouldn't be in `GitHubHomeSection`

src/GitHub.TeamFoundation.14/Home/GitHubHomeSection.cs
```diff
@@ -175,8 +170,28 @@ void ShowWelcomeMessage()
                 }),
                 false,
                 welcomeMessageGuid);
+
+            settings.WelcomeMessageLastSeen = DateTime.UtcNow;
+            settings.Save();
         }
 
+        //todo:temporary
+        static int GetIso8601WeekOfYear(DateTimeOffset time)
```
grokys: Yeah, we should probably move this into a utils class somewhere as it's used in two places now.

src/GitHub.TeamFoundation.14/Home/GitHubHomeSection.cs
```diff
@@ -93,12 +76,24 @@ bool IsGitToolsMessageVisible()
                 RepoName = ActiveRepoName;
                 RepoUrl = ActiveRepoUri.ToString();
                 Icon = GetIcon(false, true, false);
+               
+                var weekElapsed = settings.WelcomeMessageLastSeen.GetIso8601WeekOfYear() != DateTimeOffset.Now.GetIso8601WeekOfYear();
```
shana: If you save on a sunday and then run this on a monday (assuming sunday is the first day of the week), it'll pop up on both days (also friday/monday, etc). You probably want to just check that x days have elapsed since the last time the message was seen instead?

src/GitHub.TeamFoundation.14/Home/GitHubHomeSection.cs
shana: :+1:

src/GitHub.TeamFoundation.14/Home/GitHubHomeSection.cs
paladique: Since `GetIso8601WeekOfYear()` returns an int I could just change that to compare
`Current.GetIso8601WeekOfYear() > WelcomeMessageLastSeen.GetIso8601WeekOfYear() = showmessage`?

paladique:
@grokys I save the settings [here ](https://github.com/github/VisualStudio/pull/980/files#diff-ef451518c29038bbc123814920f25b5cR175) in `ShowMessage()` This the wrong place?

paladique:
So because `RepoChanged()` fires every time you change a view in TE back to Home, it always checks and updates the time the message was last viewed, which gets saved, which is why it doesn't get shown. `WelcomeMessageLastSeen` wasn't getting loaded properly (fixed).

src/GitHub.TeamFoundation.14/Home/GitHubHomeSection.cs
```diff
@@ -93,12 +76,24 @@ bool IsGitToolsMessageVisible()
                 RepoName = ActiveRepoName;
                 RepoUrl = ActiveRepoUri.ToString();
                 Icon = GetIcon(false, true, false);
+
+                var weekElapsed = DateTimeOffset.Now.GetIso8601WeekOfYear() > settings.WelcomeMessageLastSeen.GetIso8601WeekOfYear();
```
shana: This won't give you an elapsed week, it will only tell you whether a day is on a different week of the year than another day - i.e., Saturday and Sunday will be on different weeks.

To check whether one day is more than 7 days from another day, do `(DateTimeOffset.Now - settings.WelcomeMessageLastSeen).Days > 6`

shana:
@paladique Got a little comment on the week check, otherwise looks good! 🏈

paladique:
This currently displays once and will disappear when you change the view in Team Explorer from Home view to anything else, and back to the Home View. This could be resolved if it were possible to know when the user clicks on the `x` in the top right of the panel. There's no obvious way to do this at the moment.

If the above isn't possible, we could:
- Add an additional button in the panel that says something along the lines of "show later"
- Add showing the docs as a setting so that someone who has previously asked to never show this again can decide to have it show up again.

paladique:
👋 will close this for now, as I'm sure a lot has changed since!

PR #990, opened 2017-05-12
Set ExtensionInstallationFolder

grokys:
To make sure that the extension is always installed to the same directory.


src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
```diff
@@ -747,4 +749,8 @@
   <Target Name="AfterBuild">
   </Target>
   -->
-</Project>
\ No newline at end of file
+  
+  <!-- Update CreateVsixContainer target to support InstallationRoot and InstallationFolder -->
+  <Import Project="..\..\lib\14.0\Microsoft.VisualStudio.Sdk.BuildTasks.14.0.CreateVsixContainer.targets" />
```
shana: This should only be included when building with VS2015

grokys:
Closing as I believe this has been addressed.

PR #1002, opened 2017-05-25
Add missing metrics back in

paladique:
The "Pull Requests Opened" Metric is missing from the details view. Adding it back in. Also cleared the help docs metrics counter, as it seems that they weren't being cleared.


src/GitHub.VisualStudio/Services/UsageTracker.cs
```diff
@@ -353,6 +353,9 @@ static void ClearCounters(UsageModel usage, bool weekly, bool monthly)
             usage.NumberOfForkPullRequestsCheckedOut = 0;
             usage.NumberOfForkPullRequestPulls = 0;
             usage.NumberOfForkPullRequestPushes = 0;
+            usage.NumberOfGitHubPaneHelpClicks = 0;
+            usage.NumberOfWelcomeTrainingClicks = 0;
+            usage.NumberOfWelcomeDocsClicks = 0;
```
grokys: Oh! How embarrassing! This means these metrics up until now are completely wrong...

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
```diff
@@ -362,6 +361,8 @@ public override void Initialize([AllowNull] ViewWithData data)
             }
 
             IsLoading = IsBusy = false;
+            IsBusy = false;
```
grokys: This line just repeats what the line above is already doing.

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
```diff
@@ -362,6 +361,8 @@ public override void Initialize([AllowNull] ViewWithData data)
...
+            usageTracker.IncrementPullRequestOpened().Forget();
```
grokys: This method actually gets called when the pane is refreshed too, we probably don't want to increment the counter there. I think you can tell if it's a new pull request when `Model == null` on entry to the method. So I'd set a flag there and only increment if the flag is set.

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
paladique: ahhh a cherry pick/merge mistake!

src/GitHub.VisualStudio/Services/UsageTracker.cs
marlazinger200: 1

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
marlazinger200: 1

marlazinger200:
1

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
marlazinger200: 1

PR #1004, opened 2017-06-05
Add inline comments

grokys:
This PR adds inline comments for Pull Requests to editor and compare views in Visual Studio.

This PR implements the MVP for this feature, namely:

- [x] Number of comments next to files in the tree
     - [x] When # of comments is clicked on, file is opened and first inline comment thread navigated to
- [x] A margin displaying icons where comments are located, and showing where new comments can be left
- [x] Comments displayed in peek view
- [x] Existing comments can be replied to
- [x] New comments can be left where allowed
- [x] Navigating next/previous comment in file
    - [x] Menu items
    - [x] Keyboard shortcuts
    - ~~Toolbar buttons for text editor~~ don't think this is necessary, toolbar buttons aren't very discoverable
    -  ~~Toolbar buttons already exist in the diff view - can we hook into them?~~ can't find a way to do this
- [x] Buttons for next/previous comment in peek view
- ~~Existing comments displayed as marks on scrollbars~~ punting to a later date
- [x] When a file has local changes, display a message to commit/push before comments can be left
- [x] Refreshing the PR in the PR details view updates the inline comments


jcansdale:
Ever so often, this starts happening:
image

It happens for this and other files in the PR. Other times this same file works and both sides are loaded. There is no error message or anything in the status bar.

Any idea what might be causing it? My connection was a little flaky today, so maybe it was that.

src/GitHub.InlineReviews/GitHub.InlineReviews.csproj
```diff
@@ -247,13 +252,59 @@
...
+    </Reference>
+    <Reference Include="System.IO.FileSystem, Version=4.0.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\System.IO.FileSystem.4.3.0\lib\net46\System.IO.FileSystem.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="System.IO.FileSystem.Primitives, Version=4.0.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\System.IO.FileSystem.Primitives.4.3.0\lib\net46\System.IO.FileSystem.Primitives.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="System.Net.Http, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
```
jcansdale: Looks like we're referencing `System.Net.Http, Version=4.1.1.0` from NuGet, rather than the built in system version.

src/UnitTests/app.config
```diff
@@ -34,6 +34,18 @@
         <assemblyIdentity name="Microsoft.VisualStudio.CoreUtility" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-14.0.0.0" newVersion="14.0.0.0" />
       </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="System.Xml.ReaderWriter" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.1.0.0" newVersion="4.1.0.0" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="System.Net.Http" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.1.1.0" newVersion="4.1.1.0" />
```
jcansdale: It looks like this was the issue. For some reason it's redirecting `System.Net.Http` to `4.1.1.0` (which doesn't exist).

jcansdale:
Many PR diff files show the top line as having changed.

image

The equivalent line doesn't appear to have changed on https://github.com/grokys/PullRequestSandbox/pull/1#pullrequestreview-43200786

image

I'm guessing this is something to do with one file having a UTF-8 preamble and the other not. I know VS likes to add the UTF-8 preamble, but in this case the files are coming from Git (the right hand file isn't checked out).

We might need to tweak something here.
https://github.com/github/VisualStudio/blob/01b79cf3b1c3a5ff62c4370548ce487d7596a147/src/GitHub.App/Services/GitClient.cs#L206

Investigating...

jcansdale:
Both blobs that are being diffed appear to have the UTF8 preamble:
```
blob1: 239 187 191 117 115 105 110 103 32 83 
blob2: 239 187 191 117 115 105 110 103 32 83 
```

```cs
System.Text.Encoding.UTF8.GetPreamble()
{byte[3]}
    [0]: 239
    [1]: 187
    [2]: 191
```

Now I'm more confused! 😉

jcansdale:
Mast have got the wrong file, here are the byte headers and the diff:

```
blob1: *239* *187* *191* 117 115 105 110 103 32 83 
blob2: 117 115 105 110 103 32 83 121 115 116 
```

It looks like `blob1` has the preamble. Mystery solved! 😄 

```
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Linq;
 using System.Text;
@@ -8,8 +8,16 @@ namespace PullRequestSandbox
 {
     class Program
     {
+        // Comment that requires a merge.
         static void Main(string[] args)
         {
+            // Made a comment.
+            // And another.
+            // 3rd comment
+            // 4th comment
+            // 5th comment
+            // 6th comment
+            // 7th comment
         }
     }
 }
```

src/GitHub.InlineReviews/Services/PullRequestSession.cs
```diff
@@ -0,0 +1,276 @@
...
+        }
+
+        async Task UpdateFile(PullRequestSessionFile file)
+        {
+            var content = await GetFileContent(file);
+            var unmodified = await service.IsUnmodifiedAndPushed(Repository, file.RelativePath, content);
+
+            file.BaseSha = PullRequest.Base.Sha;
+            file.CommitSha = await CalculateCommitSha(file, content);
+            file.Diff = await service.Diff(Repository, file.BaseSha, file.RelativePath, content);
```
jcansdale: I think we need to calculate the `merge base` here, rather than simply use the `file.BaseSha`. Wonder if we should have an extra `MergeBaseSha` property?

grokys: Yep, we can add that if that would help!

jcansdale:
Opening a pull request seems to have slowed on this branch (not sure since exactly when).

I'm doing the following timing.

1. Wait until PR list has loading loading.
2. Click on `Add inline comments`
3. Stop timer when PR appears

- On `2.2.0.11` it takes ~3 sec
- On this branch it takes ~15 sec

Here is a profile (using CodeTrack) that might be relevant:
image

jcansdale:
It seems caching `IRepository` object in `GitService.GetRepository` brings the load down from ~15 sec to ~9 sec.

src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
```diff
@@ -0,0 +1,142 @@
...
+            if (mergeBase == null)
+            {
+                throw new FileNotFoundException($"Couldn't find merge base between {baseSha} and {headSha}.");
+            }
+
+            mergeBaseCache[key] = mergeBase;
+            return mergeBase;
+        }
+
+        IRepository GetCachedRepository(ILocalRepositoryModel model)
```
grokys: Are we sure this will be safe? The `GitClient` uses a threadpool thread for executing its operation, meaning that an essientially random thread will be used. Given that `Repository` cannot be shared between threads, this looks likely to me to cause problems. Am I missing anything?

jcansdale: Because we're reading from the repo doing blob-to-blob comparisons, carlosmn seemed to think this would be safe:
https://github.slack.com/archives/C110888DQ/p1497616923751759

I also did some sanity checks where I diffed using the same repo object on 100 thread pool tasks. Not at all definitive, to still good to see nothing broke. 😉 

Even is this brings it down from ~15 sec to ~9 sec, this is still pretty painful. 
How feasible do you think it would be to show the tree view immediately and only update the comment status when it becomes available?

I know this is pretty similar to your calculate the diff just-in-time suggestion, but I imagine it would involve some changes to the tree view.

grokys: Ok, that makes sense, but this still worries me a bit... What if we add another git operation in here and it's _not_ safe to do on a random thread?

We can definitely lazily update the comment status in the background though. Question is, do we do it in this PR or in a later PR? For ease of review, I'd be tempted to take out performance fixes from this PR and address them in another - the PR is already pretty huge!

jcansdale: I agree it adds risk and complexity without completely solving the performance issue. Shall I revert it and mention it in an issue?

grokys: Yeah, I say lets move it to an issue and try to address the performance problems in another PR, starting with loading the comment statuses in the background.

jcansdale: I reverted it and tried running it again. Part of the issue is that creating the repo object is being done on the Main thread. You can see when it's crunching away because the spinner starts to stutter and a wheel appear on the mouse pointer.

I'm going to try creating the repo object on the thread pool and see if that helps. We'd still be using a separate repo object per thread, but we'd be doing less work on the Main thread.

jcansdale: Creating Repository objects on the thread pool makes for a much smoother experience when loading a PR and might even improve time to display the PR. I've reverted the Repository caching.

jcansdale:
Just got this exception, which is a bit strange.

image

```
Exception thrown: 'LibGit2Sharp.LibGit2SharpException' in LibGit2Sharp.dll

Additional information: Could not open 'C:/Source/github.com/grokys/PullRequestSandbox/.git/objects/b6/6d829f78ae49e932f3684997ae25477a54890c': The process cannot access the file because it is being used by another process.
```

src/GitHub.InlineReviews/Commands/IVsCommand.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace GitHub.InlineReviews.Commands
+{
+    /// <summary>
+    /// Represents a Visual Studio command that does not accept a parameter.
+    /// </summary>
+    public interface IVsCommand : IVsCommandBase
```
grokys: This framework for VS commands is similar to the one that we already have in GitHub.VisualStudio for these commands: https://github.com/github/VisualStudio/tree/master/src/GitHub.VisualStudio/Menus.

I used a different framework however because I need to support command arguments. It might be worth merging these two frameworks in future.

src/GitHub.InlineReviews/Glyph/GlyphData.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Windows;
+using System.Windows.Controls;
+using Microsoft.VisualStudio.Text;
+
+namespace GitHub.InlineReviews.Glyph.Implementation
+{
+    internal class GlyphData<TGlyphTag>
```
grokys: Could we get some documentation comments on the `GitHub.InlineReviews.Glyph` classes? Just a brief description of what they represent.

src/GitHub.InlineReviews/Glyph/GlyphMargin.cs
```diff
@@ -0,0 +1,215 @@
...
+            {
+                tagAggregator.Dispose();
+                marginVisual = null;
+                isDisposed = true;
+            }
+        }
+
+        public ITextViewMargin GetTextViewMargin(string marginName)
+        {
+            if (marginName == this.marginName)
```
grokys: Nit: a `?:` operator would be more succinct here.

src/GitHub.InlineReviews/Glyph/GlyphMargin.cs
```diff
@@ -0,0 +1,215 @@
...
+            return null;
+        }
+
+        void OnLoaded(object sender, RoutedEventArgs e)
+        {
+            RefreshMarginVisibility();
+
+            tagAggregator.BatchedTagsChanged += OnBatchedTagsChanged;
+            textView.LayoutChanged += OnLayoutChanged;
+            if (handleZoom)
```
grokys: What would be the use-case for not handling zoom?

src/GitHub.InlineReviews/Glyph/GlyphMargin.cs
```diff
@@ -0,0 +1,215 @@
...
+                foreach (var span in e.Spans)
+                {
+                    list.AddRange(span.GetSpans(textView.TextSnapshot));
+                }
+
+                if (list.Count > 0)
+                {
+                    var span = list[0];
+                    int start = span.Start;
+                    span = list[0];
```
grokys: Don't this just happen 2 lines above?

src/GitHub.InlineReviews/Glyph/GlyphMargin.cs
```diff
@@ -0,0 +1,215 @@
...
+                    int end = span.End;
+                    for (int i = 1; i < list.Count; i++)
+                    {
+                        span = list[i];
+                        start = Math.Min(start, span.Start);
+                        span = list[i];
```
grokys: And again...

src/GitHub.InlineReviews/Glyph/GlyphMargin.cs
```diff
@@ -0,0 +1,215 @@
...
+
+        void ThrowIfDisposed()
+        {
+            if (isDisposed)
+            {
+                throw new ObjectDisposedException(marginName);
+            }
+        }
+
+        public bool Enabled
```
grokys: This might be personal preference, and I know we don't stick to it elsewhere, but personally I like methods to grouped together by access visiblility - when I looking through the code I assume the public API has finished when I get to private methods...

src/GitHub.InlineReviews/Glyph/GlyphMarginVisualManager.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            // set background color for children
+            var properties = editorFormatMap.GetProperties(marginPropertiesName);
+            if (properties.Contains("BackgroundColor"))
+            {
+                var backgroundColor = (Color)properties["BackgroundColor"];
+                ImageThemingUtilities.SetImageBackgroundColor(glyphMarginGrid, backgroundColor);
+            }
+
+            // set background color for margin
```
grokys: Why is the code block commented out? I prefer not to have commented out code blocks in source especially if it's unclear _why_ they're commented out...

src/GitHub.InlineReviews/Glyph/GlyphMargin.cs
jcansdale: Maybe if the margin was being reused as a map view (under the scroll bar)? It isn't being used at the moment. I'm in two minds about removing it.

src/GitHub.InlineReviews/Glyph/GlyphMargin.cs
jcansdale: Yup, I've been trying to stick to that as well. I'll fix.

src/GitHub.Exports/Models/DiffUtilities.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                for (var i = source.Lines.Count - 1; i >= 0; --i)
+                {
+                    if (source.Lines[i].Content == target[j].Content)
+                    {
+                        if (++j == target.Count) return source.Lines[i + j - 1];
+                    }
+                    else
+                    {
+                        j = 0;
```
jcansdale: I wonder if we could do second, more forgiving pass that still matches if lines have been added? That would address the inconsistency where you can add lines above the chunk, but if you happen to add a line within the chunk the comment dissapears (for me this is the most disconcerting bit).

I'm sure there are fancier things we could do later, but this would be simple, consistent and predictable. What do you think?

jcansdale: Actually, one really simple thing we could do is treat `/` as white space when matching lines. That way a developer could `//` out lines and not immediately loose their inline comments.

jcansdale: Simply ignoring white-space and comments actually works really pretty well. It means you can comment out the line with an inline-comment and add stuff below. For example:

image

You can see I've added `j = 777`, but the inline-comment is still there for reference.

src/GitHub.App/Services/GitClient.cs
```diff
@@ -239,36 +291,107 @@ public Task<Remote> GetHttpRemote(IRepository repo, string remote)
...
 
         [return: AllowNull]
-        public async Task<string> ExtractFile(IRepository repository, string commitSha, string fileName)
+        public Task<string> ExtractFile(IRepository repository, string commitSha, string fileName)
         {
-            var commit = repository.Lookup<Commit>(commitSha);
+            return Task.Factory.StartNew(() =>
+            {
+                var commit = repository.Lookup<Commit>(commitSha);
+                if(commit == null)
```
shana: Space missing!

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -385,6 +413,23 @@ string CreateUniqueRemoteName(IRepository repo, string name)
...
+        }
+
+        static string CreateTempFile(string fileName, string commitSha, string contents, Encoding encoding)
+        {
+            var tempDir = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+            var tempFileName = $"{Path.GetFileNameWithoutExtension(fileName)}@{commitSha}{Path.GetExtension(fileName)}";
+            var tempFile = Path.Combine(tempDir, tempFileName);
+
+            Directory.CreateDirectory(tempDir);
+            File.WriteAllText(tempFile, contents, encoding);
```
shana: Which thread is this happening in? If it's on the main thread, IO is slow for some reason and the file is large, this may block the UI. All IO should be off on a side thread to avoid this.

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -242,93 +271,92 @@ public IObservable<Unit> SwitchToBranch(ILocalRepositoryModel repository, IPullR
...
 
-                if (left == null)
+                var baseSha = pullRequest.Base.Sha;
+                var headSha = pullRequest.Head.Sha;
+                var baseRef = pullRequest.Base.Ref;
+                string mergeBase = await gitClient.GetPullRequestMergeBase(repo, remote.Name, baseSha, headSha, baseRef, pullRequest.Number);
+                if (mergeBase == null)
                 {
-                    throw new FileNotFoundException($"Could not retrieve {fileName}@{pullRequest.Base.Sha}");
+                    throw new FileNotFoundException($"Couldn't find merge base between {baseSha} and {headSha}.");
```
shana: The exception type doesn't match the actual exception, this isn't an IO operation. Perhaps a custom exception type could be useful here for conveying meaning?

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -385,6 +413,23 @@ string CreateUniqueRemoteName(IRepository repo, string name)
...
+        async Task<string> ExtractToTempFile(IRepository repo, string commitSha, string fileName, Encoding encoding)
+        {
+            var contents = await gitClient.ExtractFile(repo, commitSha, fileName) ?? string.Empty;
+            return CreateTempFile(fileName, commitSha, contents, encoding);
+        }
+
+        static string CreateTempFile(string fileName, string commitSha, string contents, Encoding encoding)
+        {
+            var tempDir = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+            var tempFileName = $"{Path.GetFileNameWithoutExtension(fileName)}@{commitSha}{Path.GetExtension(fileName)}";
```
shana: We shouldn't be trusting that the filename coming from the PR is valid under the current locale and encoding or even valid to Windows. There are certain filenames that can be created in other OSs that the Windows fs can't deal with, and there are certain characters and encodings that might not match the local. In this case, ignoring the string formatting warning is not a good option, and using the original filename to create a temp file is probably not a good option either.

We're already creating a GUID in the directory name, if we use the GUID instead for the filename we avoid both issues entirely. Alternatively, if you want to reuse the cached temp file, you could encode the filename with `Convert.ToBase64String(Encoding.UTF8.GetBytes(filename))` and put it in a known temp directory.

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -385,6 +413,23 @@ string CreateUniqueRemoteName(IRepository repo, string name)
...
+            var tempFile = Path.Combine(tempDir, tempFileName);
+
+            Directory.CreateDirectory(tempDir);
```
shana: We have an OS facade to abstract the filesystem out for these things. We should use it everywhere, for consistency.

src/GitHub.Extensions/StringExtensions.cs
```diff
@@ -180,6 +180,21 @@ public static string Wrap(this string text, int maxLength = 72)
             return sb.ToString();
         }
 
+        public static bool EqualsIgnoringLineEndings(this string a, string b)
```
shana: This isn't being used anywhere, is it still needed?

src/GitHub.InlineReviews/Commands/IVsCommand.cs
shana: Yeah, the menus support parameters but are not nicely typed. We can kill the menu types or extend/replace bits in them with these interfaces.

src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml.cs
```diff
@@ -168,5 +196,44 @@ void TreeView_ContextMenuOpening(object sender, ContextMenuEventArgs e)
                 }
             }
         }
+
+        private void ViewCommentsClick(object sender, RoutedEventArgs e)
```
shana: :eyes: private :eyes:

src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml.cs
```diff
@@ -168,5 +196,44 @@ void TreeView_ContextMenuOpening(object sender, ContextMenuEventArgs e)
...
+        {
+            var model = (object)ViewModel.Model;
+            Services.Dte.Commands.Raise(
+                GlobalCommands.CommandSetString,
+                GlobalCommands.ShowPullRequestCommentsId,
+                ref model,
+                null);
+        }
+
+        private async void ViewFileCommentsClick(object sender, RoutedEventArgs e)
```
shana: :eyes: private :eyes:

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -242,93 +271,92 @@ public IObservable<Unit> SwitchToBranch(ILocalRepositoryModel repository, IPullR
...
+            return Encoding.Default;
+        }
+
+        static bool HasPreamble(string file, Encoding encoding)
+        {
+            using (var stream = File.OpenRead(file))
+            {
+                foreach (var b in encoding.GetPreamble())
+                {
+                    if(b != stream.ReadByte())
```
shana: if space!

src/GitHub.InlineReviews/Glyph/GlyphMarginVisualManager.cs
```diff
@@ -0,0 +1,196 @@
...
+                    if (startingLine != null)
+                    {
+                        data.SetTop(startingLine.TextTop - textView.ViewportTop);
+                    }
+                }
+
+                glyphs = dictionary;
+            }
+        }
+        static ITextViewLine GetStartingLine(IList<ITextViewLine> lines, Span span)
```
shana: Missing newline

src/GitHub.InlineReviews/Glyph/GlyphMarginVisualManager.cs
```diff
@@ -0,0 +1,196 @@
...
+                {
+                    var element = new Canvas();
+                    element.ClipToBounds = true;
+                    glyphMarginGrid.Children.Add(element);
+                    visuals[type] = element;
+                }
+            }
+        }
+
+        public FrameworkElement MarginVisual => glyphMarginGrid;
```
shana: Properties should be defined after methods

src/GitHub.InlineReviews/Peek/InlineCommentPeekResult.cs
```diff
@@ -0,0 +1,35 @@
...
+        public InlineCommentPeekViewModel ViewModel { get; }
+
+        public IPeekResultDisplayInfo DisplayInfo { get; }
+            = new PeekResultDisplayInfo("Review", null, "GitHub Review", "GitHub Review");
+
+        public Action<IPeekResult, object, object> PostNavigationCallback => null;
+
+        public event EventHandler Disposed;
+
+        public void Dispose()
```
shana: This needs implementing

src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
```diff
@@ -0,0 +1,212 @@
...
+        public IAccount User { get; }
+
+        /// <inheritdoc/>
+        public ReactiveCommand<object> BeginEdit { get; }
+
+        /// <inheritdoc/>
+        public ReactiveCommand<object> CancelEdit { get; }
+
+        /// <inheritdoc/>
+        public ReactiveCommand<Unit> CommitEdit { get; }
```
shana: All of these properties should be after methods

src/GitHub.InlineReviews/ViewModels/InlineCommentPeekViewModel.cs
```diff
@@ -0,0 +1,204 @@
...
+                _ => previousCommentCommand.Execute(new InlineCommentNavigationParams
+                {
+                    FromLine = peekService.GetLineNumber(peekSession, triggerPoint).Item1,
+                }));
+        }
+
+        /// <summary>
+        /// Gets the thread of comments to display.
+        /// </summary>
+        public ICommentThreadViewModel Thread
```
shana: This should probably called something other than `Thread`, since there's a `Thread` type that can easily be confused with it, and makes the code harder to read. We call a series of related comments a `Discussion` on .com.

src/GitHub.InlineReviews/Peek/InlineCommentPeekableItemSource.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                var viewModel = new InlineCommentPeekViewModel(
+                    apiClientFactory,
+                    peekService,
+                    session,
+                    sessionManager,
+                    nextCommentCommand,
+                    previousCommentCommand);
+                viewModel.Initialize().Forget();
+                peekableItems.Add(new InlineCommentPeekableItem(viewModel));
```
shana: The viewmodel may not be finished initializing by the time it's added to this list. What is the impact of that?

src/GitHub.InlineReviews/Services/PullRequestSession.cs
```diff
@@ -0,0 +1,289 @@
...
+        /// <inheritdoc/>
+        public IPullRequestModel PullRequest { get; private set; }
+
+        /// <inheritdoc/>
+        public ILocalRepositoryModel Repository { get; }
+
+        IEnumerable<string> FilePaths
+        {
+            get { return PullRequest.ChangedFiles.Select(x => x.FileName); }
+        }
```
shana: Properties after methods

src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
```diff
@@ -0,0 +1,133 @@
...
+            var repo = await GetRepository(repository);
+
+            try
+            {
+                return await gitClient.ExtractFileBinary(repo, sha, relativePath);
+            }
+            catch (FileNotFoundException)
+            {
+                var pullHeadRef = $"refs/pull/{pullRequestNumber}/head";
+                await gitClient.Fetch(repo, "origin", sha, pullHeadRef);
```
shana: This needs to use an https remote, so it needs to do `var remote = await gitClient.GetHttpRemote(repo);` and call fetch with the returned remote.

src/GitHub.Extensions/StringExtensions.cs
grokys: Nope!

src/GitHub.InlineReviews/ViewModels/InlineCommentPeekViewModel.cs
grokys: I'd not considered this; I've used `Thread` in the names of quite a few types as well.

src/GitHub.InlineReviews/Peek/InlineCommentPeekableItemSource.cs
grokys: It should be fine - the view model will update itself in-place. The reason I did it like this was to make clear that there's async work to be done - the other options were to do the async work in the ctor or make `Initialize` an `async void` method.

src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
grokys: Already fixed this in the work associated with the PR details right-click menu.

src/GitHub.InlineReviews/Peek/InlineCommentPeekResult.cs
grokys: I don't think there's actually anything we need to dispose here. The reason this method appears is that `IPeekResult` implements `IDisposable` so we need to implement it too.

src/GitHub.App/Services/PullRequestService.cs
jcansdale: At the moment I don't think the file path is surfaced anywhere, but it is visible in the #1036 PR (which isn't ideal).

image

Maybe we should find a way to sanitize the file name and not include the `commitSha`? Would URL-encoding the filename work? I think we definitely need to need to keep the file extension for automatic syntax highlighting.

src/GitHub.App/Services/PullRequestService.cs
jcansdale: I was reluctant to create a new exception type for something that really shouldn't happen. I've changed it to `InvalidOperationException` which we're using elsewhere in this class. Still not ideal, but at least it isn't an IO exception.

grokys: In #1036 [I changed it to `LibGit2Sharp.NotFoundException`](https://github.com/github/VisualStudio/pull/1036/files#diff-a8710ac5b8611c56a154b8e06799ec67R322)

src/GitHub.InlineReviews/Peek/InlineCommentPeekResult.cs
jcansdale: I've marked it as `sealed` because it isn't intended to be extended.

src/GitHub.App/Services/PullRequestService.cs
jcansdale: Thanks for the heads up. I'll revert my other commit. Did you fix a test for that as well?

grokys: Yep, I changed the test to expect `NotFoundException` too.

PR #1005, opened 2017-06-05
Stop TryOpenRepository tests spamming the log

jcansdale:
The log file getting spammed when some unit tests are run. For the log to be useful it shouldn't be full of junk. 😄 Rather than allowing tests to write to the log, check that calls are made to `ILogger`.

This PR does the following:

- Make `SolutionCreateThrows_ReturnsFalse` check calls to `ILogger`
- Remove unused `ActivityLog*` methods from `IVSServices`


jcansdale:
There were a bunch of exceptions like this:
```
Couldn't clean up x:\repo\.vs\~$GitHubVSTemp$~. System.IO.IOException: I/O error occurred.
   at VSServicesTests.TheTryOpenRepositoryMethod.<>c.<DeleteThrowsIOException_ReturnTrue>b__3_1(CallInfo ci) in C:\Source\github.com\github\VisualStudio\src\UnitTests\GitHub.Exports\VSServicesTests.cs:line 64
   at NSubstitute.Core.CallActions.CallAction.Invoke(CallInfo callInfo)
   at NSubstitute.Core.CallActions.InvokeMatchingActions(ICall call)
   at NSubstitute.Routing.Handlers.DoActionsCallHandler.Handle(ICall call)
   at NSubstitute.Routing.Route.<>c__DisplayClass3.<Handle>b__0(ICallHandler x)
   at System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()
   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
   at NSubstitute.Routing.Route.Handle(ICall call)
   at NSubstitute.Core.CallRouter.Route(ICall call)
   at NSubstitute.Proxies.CastleDynamicProxy.CastleForwardingInterceptor.Intercept(IInvocation invocation)
   at Castle.DynamicProxy.AbstractInvocation.Proceed()
   at Castle.Proxies.IDirectoryInfoProxy.Delete(Boolean recursive)
   at GitHub.Services.VSServices.TryCleanupSolutionUserFiles(IOperatingSystem os, String repoPath, String slnName) in C:\Source\github.com\github\VisualStudio\src\GitHub.Exports\Services\VSServices.cs:line 134
```
This was being logged from inside the NCover runner.

I've disabled logging for this test using: `VsOutputLogger.SetLogger(_ => { });`

This is obviously a hack. I'm not sure the best way to disable/redirect logging when unit tests run.

grokys:
@jcansdale this is obviously a bit stale, but I think now we've moved to serilog we have a central place to enable/disable this depending on whether we're in a test runner: `LogManager` can check the splat `ModeDetector` I think.

jcansdale:
@grokys In this case having logging enabled would have flagged up a problem with some of the tests. I'd rather have the test explicitly turn off logging when they are expected than turn off logging for all tests. Do you know if Serilog would let me temporarily disable logging?

jcansdale:
@grokys I've changed it to use dependency injection of `ILogger`, which lets me check that something is actually being logged. I think this is a cleaner approach. What do you reckon?

PR #1021, opened 2017-06-28
Allow selection of fork/parent pull requests.

### LINKED ISSUE #863: Can't see any pull request if origin is a fork
- GitHub Extension for Visual Studio  version: 2.2.0.6
- Visual Studio version: 2015.3

When I open a Github project, I can't see any pull request in GHVS. I suspect this is because my repo is a fork, not the source repo where the pull requests are, but this doesn't fit my workflow at all...

Here's my use case :
- I maintain a project that is owned by an organization (in this case, [FakeItEasy](https://github.com/FakeItEasy/FakeItEasy/)
- I don't commit directly to the source repo; instead, I work on a fork, and submit pull requests from there
- In my local clone, I have 2 main remotes: `origin` is my fork, and `upstream` is the source repo

If I rename `origin` to something else, and `upstream` to `origin`, I can see the pull requests, so it seems GHVS uses `origin` to decide in which repo it looks for pull requests

There should be a way to select the remote for which I want to see the pull requests. Or maybe look at the source repo if `origin` is a fork. In the current state, this feature is useless unless my setup is exactly the one GHVS expects.

grokys:
When working on a fork, previously the PR list would show PRs for that fork, which is often not what is wanted. This PR adds a selector dropdown to select the repository from which to display the pull requests, defaulting to the parent repository if the repository is a fork. If the repository is _not_ a fork, the selector will be hidden.

image

This exposes a similar problem to #734 in that there quickly ends up being no room along the top row of filters for the repository name, meaning it wraps to a new line, causing ugly. We need to think of a new way to display filters for this and #734. cc: @donokuda 

Depends on #1004, #1023
Fixes #863


tierninho:
@grokys Tried testing this PR, but the PR page in the github window just hangs, with "Unable to complete read operation after process has exited" in the Output window. Not sure how to get around this.

screen shot 2017-07-11 at 10 08 58 am

grokys:
@tierninho could you try now?

tierninho:
@grokys same result, still hangs.

jcansdale:
Strange. This PR has been working well for me.  😕

tierninho:
@grokys and I got this working on my machine after tweaking my setup 👍

src/GitHub.App/Services/GitClient.cs
```diff
@@ -359,20 +359,36 @@ public Task<bool> IsModified(IRepository repository, string path, [AllowNull] by
...
-        public async Task<string> GetPullRequestMergeBase(IRepository repo, string remoteName, string baseSha, string headSha, string baseRef, int pullNumber)
+        public async Task<string> GetPullRequestMergeBase(IRepository repo,
+            UriString baseCloneUrl, UriString headCloneUrl, string baseSha, string headSha, string baseRef, int pullNumber)
         {
             var mergeBase = GetMergeBase(repo, baseSha, headSha);
             if (mergeBase == null)
             {
-                var pullHeadRef = $"refs/pull/{pullNumber}/head";
-                await Fetch(repo, remoteName, baseRef, pullHeadRef);
+                // TODO: Optimize and error check.
```
jcansdale: I need to fix this.

src/GitHub.App/Services/GitClient.cs
```diff
@@ -394,6 +401,20 @@ public Task<bool> IsHeadPushed(IRepository repo)
             });
         }
 
+        Task Fetch(IRepository repo, UriString cloneUrl, params string[] refspecs)
+        {
+            var tempRemoteName = $"{cloneUrl.Owner}-{Guid.NewGuid()}";
```
jcansdale: This should only create temporary remotes when necessary (when cloneUrl != origin).

src/GitHub.App/Services/GitClient.cs
```diff
@@ -422,6 +430,26 @@ public Task<bool> IsHeadPushed(IRepository repo)
...
+
+            var tempRemoteName = cloneUrl.Owner + "-" + Guid.NewGuid();
+            repo.Network.Remotes.Add(tempRemoteName, httpsUrl);
+            try
+            {
+                return Fetch(repo, tempRemoteName, refspecs);
+            }
+            finally
+            {
+                repo.Network.Remotes.Remove(tempRemoteName);
```
jcansdale: This is a test comment. Let's see if the PR gets blocked.

PR #1023, opened 2017-06-29
Added a version to host cache.

grokys:
In #1021 we require the repository parent (in the case of forked repositories) - this data is returned from the API but wasn't getting stored in our cache, meaning that we have invalid data when data is returned from the cache as opposed to from the API.

This adds a `cacheVersion` object to host caches. This version is checked when the host cache is first loaded, and if it doesn't match the expected version, the cache is invalidated. In this way each time we make a change as required by #1021 we can bump `HostCacheFactory.CacheVersion` to ensure the cache won't be returning invalid data.


jcansdale:
I tried merging in #1021 and the cache seemed to clear itself okay.

grokys:
Ok, lets merge this and get it tested.

PR #1025, opened 2017-07-03
Preview inline-comments as tooltip

jcansdale:
This PR enables support for previewing inline-comments as a tool-tip.

It was originally intended as a quick way to find comment threads of interest, but it turned out to be convenient way to read inline-comments without necessarily opening the peek window.

image

Loitering over a comment will preview the comment thread in a tooltip. This is particularly useful on the diff view, where the peek window can be a little cramped.

For example, here's a thread shown in a tooltip:

image

Here's the same thread in the peek window:

image

There will usually be enough space to show all comments in a thread. If however the comment thread is too long to display fully, the preview will be shown with the following status message:
`Click to reply or view more (showing most recent of 6 comments)`

A scroll bar will also be displayed to give an indication of how much of the thread is off screen (it's visual indicator not a functional scroll bar).

image

The rational for showing the most recent comments is that on an active PR, the developer will be most likely to want to read the latest comments to arrive. It would be a shame for this functionality to suddenly stop being useful as soon as a thread grows to a certain length.


src/GitHub.InlineReviews/Resources.resx
```diff
@@ -0,0 +1,126 @@
...
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="CommentTooltipStatusMessage" xml:space="preserve">
+    <value>Click to reply</value>
+  </data>
+  <data name="CommentTooltipStatusOverflowMessage" xml:space="preserve">
+    <value>Click to reply or view more (showing most recent of {0} comments)</value>
+  </data>
+</root>
```
paladique: `ShouldParsePr960` in [DiffServiceTests](https://github.com/github/VisualStudio/blob/master/test/GitHub.InlineReviews.UnitTests/Services/DiffServiceTests.cs#L21) fails cause when this resx is introduced as a public class, `Resources.pr_960_diff.txt` is assumed to be in the `GitHub.InlineReviews`  resource file. Changing this class to internal makes the test. Not sure if there's a better solution, or if changing to internal would break something else.

grokys:
I think this is a great feature to have, but there are 2 things that worry me about this particular implementation:

1. The tooltip can grow to be _huge_ and fill the whole screen vertically - this feels wrong for a tooltip to me
2. When the tooltip grows to the vertical size of the screen, it shows scrollbars - when scrollbars appear I assume I can scroll, which I can't

I don't want to block this PR based on these issues, but I would like to know what @donokuda and @shana think. It may well be that this isn't a problem to anyone but me ;)

donokuda:
> 1. The tooltip can grow to be _huge_ and fill the whole screen vertically - this feels wrong for a tooltip to me
> 2. When the tooltip grows to the vertical size of the screen, it shows scrollbars - when scrollbars appear I assume I can scroll, which I can't

These seem like valid concerns to me. When I think of tooltips, I think of terse information that's enough to inform what's going on or what will be happening.

For example:



Instead of displaying the entire comment thread, what information is worth surfacing to let a developer know that this thread might be of interest? Thinking back to when you're either reviewing a pull request or addressing feedback in a PR, what would you like to understand first in order to know when to get started?

Brainstorming a couple of things off the top of my head:

- Number of comments in a thread
- Number of new comments since last read
- Participants (number and names?) in a review
- Timestamp of last comment

paladique:
I do like the idea of having some form of preview, but also agree with the above points.

It's very easy to miss context from the tooltip since the whole conversation may not be visible. The preview is leading the user to the comments anyway so they can see the entire conversation. Comments, specifically large ones, might just be too difficult to work with when it comes to incorporating them into tooltips. @donokuda's suggestions could be some good alternatives. Out of that list I would say that knowing when a comment thread was last updated is a big one.

donokuda:
Another idea: we could surface or (try to) summarize the parent comment and then show other information like number of replies and participants.

That way I have an understanding of what feedback was left and how large of a discussion a thread is.

paladique:
> Another idea: we could surface or (try to) summarize the parent comment [...]

If we did this, we'd probably be better off summarizing because it could be a very large comment that would require scrolling and get cut off. Summarizing may be better off as a future enhancement if we took that route.

src/GitHub.InlineReviews/Resources.resx
jcansdale: Good catch. Unfortunately changing it to internal would break somethign else. Grrrr! :-(

jcansdale:
I appreciate everyone's concern about it being an overly large tooltip. When I started implementing this functionality, I had in mind a regular sized tooltip.

I started off with just the first comment, but found I wanted to read replies as well. Then I changed it to a cropped thread (with a max height), but it was frustrating when all the comments didn't fit.

I found letting it expand to the maximum height works well for reading a whole thread. The more comments there are, the more likely you are to re-read them. I found keeping the peek window open while working on code doesn't work well, but having the tooltip/preview nearly is very handy.

Part of the issue is the resizing, scrolling and jumping around that seems to happen/be required with the peek window. This is fine when you're actively writing a comment, but it all you want to do is re-read a comment it's too disruptive.

Think of it more as a preview than a tooltip (like on the windows toolbar). Try it on some PRs with a lot of discussion and see what you think then. 😉

src/GitHub.InlineReviews/Resources.resx
jcansdale: @paladique I've fixed this now and it seems to be passing. Just waiting for appveyor to complete.

paladique:
I didn't have a actual long PR comment thread to try this out so I did [this](https://github.com/evilpaladique/new-test/pull/1#pullrequestreview-47175812) on a test repo and below was the result of a very long comment with a lot of lorem ipsum text. This is an uncommon scenario, but confirms that the preview could cut off large comments or individual comments of large threads. 

With that said, we're making the assumption that someone who wants to preview the thread has read the parts of the discussion that are not visible. Even though the preview says "showing most recent of 6 comments", it would be hard to assume someone is keeping track of the number of comments they've read, especially if that thread is long.

capture

jcansdale:
@paladique, My main argument for previewing the end of the thread is that the last few comments are likely to be the most revealing about its status. Has an action plan been decided upon or have chances been accepted? The end of the thread is likely to tell you that in a way that the opening comment certainly won't! Does that make sense?

donokuda:
> With that said, we're making the assumption that someone who wants to preview the thread has read the parts of the discussion that are not visible. Even though the preview says "showing most recent of 6 comments", it would be hard to assume someone is keeping track of the number of comments they've read, especially if that thread is long.

I'm going to agree with @paladique here. Not only are we assuming that developers have read part of the discussion that's not visible, but we're also assuming that they _remember_ the entire context of the discussion.

At this moment, I'm not convinced that this is the right approach but clarifying with what we trying to help developers do with this tooltip might help me understand this better...

From the original post:

> It was originally intended as a quick way to find comment threads of interest

It might be worthwhile to explore other ways to surface interesting threads before we decide on this approach. One thing we could consider doing is defining what we believe meet the criteria of an interesting thread, then calling them out in the margin (potentially with an indicator.) This is just one potential solution out of many, I'm sure.

> it turned out to be convenient way to read inline-comments without necessarily opening the peek window

If the peek window isn't doing a good job for inline comments (it seemed like space is one problem), then we should also explore to see what other ways we can show inline comments / feedback in general. I think that the tooltip can be useful for showing comments, but I'm not convinced that showing as much as the space persist is the best approach.

For the tooltip itself: We might want to think about what information is needed to establish enough context for a developer before they decide whether they want to read an entire thread. At this moment, I think that both the first and most recent comment are important. The first because it establishes the context of the thread / feedback (why the feedback was left). The most recent comment _might_ be important but I think it probably depends on the individual commenting and (obviously) the comment itself.

Dotcom does something interesting with pull requests and issues with a lot of content:

image

Instead of trying to show as many comments as possible, we could do something along the lines of

```
+------------------------------+
|   (First comment truncated)  |
| ~~~ 5 comments not shown ~~~ |
|   (Last comment truncated)   |
+------------------------------+
```

From there, we'll let the developer decide whether or not they want to click in and read the thread.

jcansdale:
@donokuda,

> At this moment, I think that both the first and most recent comment are important. The first because it establishes the context of the thread / feedback (why the feedback was left). The most recent comment might be important but I think it probably depends on the individual commenting and (obviously) the comment itself.

I agree that both the first and most recent comments are important. The first comment should work standalone as there is no other context at that point. The last comment might not work standalone as there will be the context of the previous comments. For example, someone might suggest alternative solution to the original comment and the last comment might simply be "Done" (not terribly revealing).

My preference would be to show the first comment and last n-comments with a splitter as suggested:

image

> From there, we'll let the developer decide whether or not they want to click in and read the thread.

I think there are definitely usability issues with the peek window. The longer the comment thread, the more intrusive the peek window tends to be. It causes the position in the code window to jump around and there's too much stretching, resizing and scrolling for my taste.

If a comment thread can fit on the preview window, it makes for a much nicer reading experience IMHO. Maybe I re-read comments more than average, but I hate having to constantly re-open the peek window (or split my code by leaving it open).

jcansdale:
I've rationalized the MVVM types roughly as @grokys suggested. I'm reusing `CommentViewModel` rather than creating new `TooltipCommentView/Model/Designer` implementations. I'm also using `CommentThreadViewModel` as a base class for `TooltipCommentThreadViewModel` which simplifies things further.

There is now just a `TooltipCommentThreadView` and `TooltipCommentThreadViewModel` without the ambiguous names. I've reused the `ICommentThreadViewModel` interface, but felt creating a simple `TooltipCommentThreadViewModel` class was cleaner than passing nulls to `CommentThreadViewModel`.

jcansdale:
@paladique, I've rationalized the MVVM types and resolved some merge conflicts. Shall we merge this with a view to refining the UI in a separate PR?

paladique:
Merging to continue to iterate on the UI.

PR #1027, opened 2017-07-05
[wip] Cleaning up some bits of the PR review and inline comments code

shana:
Scratching some itches on formatting and code neatness, maybe killing some bugs in the process.


sguthals:
@shana - I think this one is very stale. Close if you agree?

PR #1031, opened 2017-07-05
Show comment bubble when hovering over the current mouse line

shana:
Making commenting more discoverable by always showing the bubble when hovering over a line


shana:
@jcansdale Been testing this one out and it works great. I agree that we should only do it on the diff view and not on the editor view.

jcansdale:
@shana I've updated it so that it only tracks the mouse on the diff views.

ungb:
One minor thing I found is if you hover between the line number and the comment gutter the comment icon doesn't show up.

image

jcansdale:
@ungb Thanks for the catch! Could you try again with the latest push?

PR #1032, opened 2017-07-05
WIP: Improve tracking of lines with inline-comments

jcansdale:
This PR allows new lines to be entered near to inline-comments without the comment mysteriously disappearing.   It also allows lines with inline-comments to be commented out and still track the same line.

For posterity here is a pre-built VSIX from this branch: [GitHub.VisualStudio.zip](https://github.com/github/VisualStudio/files/1847652/GitHub.VisualStudio.zip)


jcansdale:
This was a naive "solution" to a hard problem. 😉 

I hope one day we can find a decent solution!

PR #1034, opened 2017-07-06
Close peek session when Cancel clicked on new inline comment thead.

grokys:
When Cancel is clicked on a peek session that represents a new comment thread, close the peek session.


tierninho:
@jcansdale Can you confirm the functionality: 

1) When clicking `cancel` on a *new* comment in an empty thread, the comment form collapses. Status: working.

2) When clicking `cancel` on a *existing* comment in a populated thread, the comment form reverts to reply only. Status: working. Note comment form remains open.

Tested both of these scenarios, but unsure the the second one is correct.

grokys:
@tierninho yes the second one is correct - does that behavior feel right to you?

tierninho:
Yes, looks good. 👍

PR #1040, opened 2017-07-07
Removed NullGuard and some Debug.Assert/Fail calls.

### LINKED ISSUE #774: Remove NullGuard and Debug.Asserts

grokys:
- Removed `NullGuard` and from the classes where it was removed, added `Guard` checks
- Remove the `Debug.Fail`s from `Guard` as this causes problems with ncrunch, and an exception will be thrown afterwards anyway.
- Replaced `Debug.Assert`s with throwing `GitHubLogicException` where an exception would be encountered after the `Debug.Assert` anyway.

Fixes #774


grokys:
@jcansdale I don't think we can remove them from Rothko easily, at least not without causing future merge problems.

shana:
We can nuget package rothko up instead of building it ourselves, we change it so rarely that we don't really need it as a submodule honestly

Issue #1060, opened 2017-07-12
Failed Null Check on 'session'


jcansdale:
- GitHub Extension for Visual Studio version: 2.3.0 (dev)
- Visual Studio version: 2015

__What happened__ (with steps, logs and screenshots, if possible)

I was pulling changes from a PR with VS open and the following happened.

![image](https://user-images.githubusercontent.com/11719160/28116047-3bc7e0c2-66ff-11e7-9fc7-690b0d68e528.png)

jcansdale:
@grokys I've had this quite a few times now.

grokys:
I think what's happening here is that we have some re-entrancy going on. It shouldn't be possible for a null session to be passed to the `AddInlineCommentTag` ctor in `InlineCommentTagger.GetTags` as there's a `file != null` check around that code, and if `file != null` then `session != null`.

However we're returning the `TagSpan`s from that method using `yield` so it's possible that the calling VS code is `await`ing while enumerating the returned collection, allowing `SessionChanged` to be called in the middle. I guess we should build a `List` in `GetTags` and return that rather than using `yield`.

grokys:
Another possibility is that `SessionChanged` is being called on a background thread. We need to make sure we switch to the UI thread there if that can happen.

PR #1066, opened 2017-07-13
Attempt to prevent NullReferenceExceptions in InlineCommentTagger

grokys:
The cause of #1060 isn't entirely clear, so this PR attempts to mitigate any possibility of it happening:

- Make sure `PullRequestSessionManager.CurrentSession` is set on main thread (it should already be afaics, this is a "just in case"; maybe not needed?)
- Don't yield in `InlineCommentTagger.GetTags` - in case the caller `awaits` while enumerating the results, in which case `session` could change while running
- Handle `session` being null in `Rebuild` - this is an actual NRE i came across, though it doesn't seem to be the one in #1060


jcansdale:
I still seem to be having issue with this one. I'll post some exception detail next time I see it.

PR #1068, opened 2017-07-13
Make diff line matching more robust

### LINKED ISSUE #1052: Adding a comment to a line in green gutter scroll section works, however the comment is not found if I were to click on that line again.
Adding a comment to a line in green gutter scroll section works, however the comment is not found if I were to click on that line again. The comment count in the GH window does not increment either.

screen shot 2017-07-07 at 10 19 06 am

I can see the comment in PR on dotcom though. Not sure if this is the correct functionality. I would expect the comment to be sticky in VS and the comment count to increment.

screen shot 2017-07-07 at 10 20 18 am

(contact @tierninho for questions)

jcansdale:
This PR matches diff lines based on their original line number and content.

- Diff lines can be matched based on a single line (multiple lines of context aren't required)
- Lines can be inserted between and near to comments without them vanishing

Fixes #1052

### DIFF
--- a/src/GitHub.Exports/Models/DiffUtilities.cs
+++ b/src/GitHub.Exports/Models/DiffUtilities.cs
@@ -96,7 +96,10 @@ namespace GitHub.Models
                 {
                     if (source.Lines[i].Content == target[j].Content)
                     {
-                        if (++j == target.Count) return source.Lines[i + j - 1];
+                        if (++j == target.Count || i == 0)
+                        {
+                            return source.Lines[i + j - 1];
+                        }
                     }
                     else
                     {
--- a/test/GitHub.InlineReviews.UnitTests/Models/DiffUtilitiesTests.cs
+++ b/test/GitHub.InlineReviews.UnitTests/Models/DiffUtilitiesTests.cs
@@ -173,14 +173,20 @@ namespace GitHub.InlineReviews.UnitTests.Models
         {
             [Theory]
             [InlineData(" 1", " 1", 0)]
-            [InlineData(" 1\n 2", " 2", 1)]
-            [InlineData(" 1\n 1", " 1", 1)] // match the later line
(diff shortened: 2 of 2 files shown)


grokys:
There are problems with this approach. For example, I created this PR: https://github.com/grokys/PullRequestSandbox/pull/29

As you can see, I added a comment on line 22, but in VS with this PR, this is showing up on line 15:

image

jcansdale:
Well spotted. I've been wondering about how to make it more robust. Could you try this latest push. It will match using all available context lines before falling back to matching using old line number.

Still not perfect, but I wonder if this will end up being about getting the balance right (false positives vs false negatives).

grokys:
It now shows the unedited file with the comments in the correct place, but when I delete the `Guard.NotNull` as the comment suggests, the comment jumps to line 15 again.

jcansdale:
From @grokys on the Slack:
```
https://gist.github.com/as-cii/d63737d490d007f0c5f2cd3c0afa0669
i think something like that will give much better results
but obviously, it's more involved which is why i was punting it until we've got the basic stuff working...
tbh i think the only _real_ way to do it would be to track the position of the comments as edits are made and store them locally
then use an algorithm like the above to match comments if the file is e.g. edited outside VS
we also need a "outdated" visual state to show that the comment may no longer be valid
to me it feels like a separate minor release
```

I'm going to push a version that fixes the original issue. It was trying to match 5 lines of context when the diff only had 3.

jcansdale:
I've just pushed more of a point fix that should resolve just this issue. Hopefully 3rd time's a charm? 😉

jcansdale:
I've done some more testing on .com this seems to behave very similarly. Sometimes .com requires less than 5 lines of context, but when I noticed this the API returned only 4 lines so the extension actually behaved in a similar way.

PR #1074, opened 2017-07-13
Allow navigation to comment on GitHub.

grokys:
When the date in a comment is clicked, navigate to the comment on GitHub.


donokuda:
I pulled down the branch it looks and behaves well for me. I think it would be interesting to look at how we can utilize the toolbar for viewing discussions and comments on GitHub:

PR #1080, opened 2017-07-18
Polish peek view for inline reviews

donokuda:
This pull request starts bringing in support for themes to inline reviews.

Doesn't look 100% great so I'll be following up with some custom color choices in a series of future pull requests. Here's what it looks like so far:

**Light**

**Dark**

**Blue**


src/GitHub.InlineReviews/Tags/ShowInlineCommentGlyph.xaml
```diff
@@ -4,21 +4,38 @@
...
 
-    <UserControl.Resources>
-        <!-- This will be the color of an added/deleted word in the diff view or the indicator margin color -->
-        <SolidColorBrush x:Key="DiffChangeBackground" Color="DarkSlateGray" />
-    </UserControl.Resources>
+    <UserControl.Style>
+        <Style TargetType="UserControl">
+            <Style.Triggers>
+                <DataTrigger Binding="{Binding DiffChangeType}" Value="Add">
+                    <Setter Property="Background" Value="{DynamicResource DiffChangeBackground.Add}" />
```
jcansdale: I've created 3 new dynamic brushes that will change according the the theme. They're called `DiffChangeBackground.Add`, `DiffChangeBackground.Delete` and `DiffChangeBackground.None` to mirror the `DiffChangeType` enum.

The old `DiffChangeBackground` is no longer being set for each glyph. Everything is now done via the `DiffChangeType` property.

jcansdale:
I noticed the arrows won't showing up here:
image

Strange, because they show in your screen grabs. I did merge from `master`. Maybe something got messed up in the process?

src/GitHub.InlineReviews/Tags/ShowInlineCommentGlyph.xaml
```diff
@@ -4,21 +4,38 @@
...
         <Viewbox>
             <Path Stroke="Black"
-          Data="M13 2H1c-0.55 0-1 0.45-1 1v8c0 0.55 0.45 1 1 1h2v3.5l3.5-3.5h6.5c0.55 0 1-0.45 1-1V3c0-0.55-0.45-1-1-1z m0 9H6L4 13V11H1V3h12v8z"/>
+                  Data="M13 2H1c-0.55 0-1 0.45-1 1v8c0 0.55 0.45 1 1 1h2v3.5l3.5-3.5h6.5c0.55 0 1-0.45 1-1V3c0-0.55-0.45-1-1-1z m0 9H6L4 13V11H1V3h12v8z"/>
         </Viewbox>
     </Grid>
+
+    <UserControl.ToolTip>
+        <ToolTip
+            Background="{DynamicResource GitHubVsToolWindowBackground}"
```
jcansdale: You can now access the GitHub dynamic/themed resources from the glyph XAML files (this wouldn't have worked before). You can see I'm using `GitHubVsToolWindowBackground` here.

src/GitHub.InlineReviews/Properties/DesignTimeResources.xaml
```diff
@@ -0,0 +1,15 @@
+﻿<ResourceDictionary
+        xmlns:cache="clr-namespace:GitHub.UI.Helpers;assembly=GitHub.UI"
+        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+
+    <SolidColorBrush x:Key="DiffChangeBackground.Add" Color="Green" />
```
jcansdale: If you want any resources to show up in the designer, you can add there here.

You can see below, I'm also adding the 3 `SharedDictionary.xaml` files. At the moment this is only needed for the `Show/AddInlineCommentGlyph.xaml` files (because they're added by `BrushesManager.cs` rather then an explicit `MergedDictionaries` element in each XAML file).

donokuda:
>Strange, because they show in your screen grabs. I did merge from master. Maybe something got messed up in the process?

🤔 Hmm, that is strange.. I'll take a lookie loo at this.

donokuda:
Hmm, not running into that problem here...

image

donokuda:
Here's where I'm at right now:

blue
dark
light pn

Most of the focus is around the margin in these screenshots (nothing has changed for the comment boxes.)

This is ready for review 🙏

jcansdale:
@donokuda,

At the moment there are a whole bunch of different shades of red/green on the diff view. I wonder if the comment bubbles could borrow the colors from the scroll bar indicators?

image

I think the colors might work and make it look a little less busy? What do you think?

jcansdale:
When I was taking a bunch of screen-grabs for the tooltip issue, I noticed that I mistook the tracking comment for a real comment on a number of occasions. I wondered if there might be a might be a way to de-emphasize tracking comment glyph (with the `+`) to make this less likely to happen?

image

donokuda:
> At the moment there are a whole bunch of different shades of red/green on the diff view. I wonder if the comment bubbles could borrow the colors from the scroll bar indicators?

I like this idea and will give it a shot.

> When I was taking a bunch of screen-grabs for the tooltip issue, I noticed that I mistook the tracking comment for a real comment on a number of occasions. I wondered if there might be a might be a way to de-emphasize tracking comment glyph (with the +) to make this less likely to happen?

Yup. I think one idea is to use a square with the plus instead like what we do on dotcom:

donokuda:
New fill colors added:

**Light**
image

**Blue**
image

**Dark**
image

donokuda:
Changed the "Add comment" glyph to a square icon and adjusted the blue for the dark theme.

**Dark**
image

**Blue**
image

**Light**
image

jcansdale:
> I think one idea is to use a square with the plus instead like what we do on dotcom

I've been having a play with this iteration. I have a few reservations.

- Having the icon change color as it tracks up and down can look a bit psychedelic
- Because the `+` isn't visible when there's already a comment, it flashes on and off
- The add comment glyph is now more prominent than the comment here glyph

Do you think showing just the `+` might work?

jcansdale:
> One thing to point out: the different bg colored margin indicates that the line can be commented on. It does not imply that the line has changed in that branch. So even with the goal of "When I run into a bug while working in a branch, I want to know which lines have changed so I understand if I introduced the bug" the implementation as it stands doesn't hold up to that goal.

I realized we hadn't talked about this and it is an important part of the equation. Here are some pertinent notes from the mini-summit (I'll discuss in a separate comment):

# Margin

We're currently displaying the comment icons in the glyph margin, but this has a number of problems:

- We interfere with existing glyph margin functionality such as breakpoints
- We need to create a glyph for every line where a comment can be left
- It's not clear on which lines comments can be left

The plan is to add our own margin which will consist of:

- Icons for existing comments
- A method of visually indicating where comments can be left
- When the mouse is hovered over a line where a comment can be left, an "add comment" icon will appear

Regarding point 2 - in future when we move to the new GraphQL reviews API then it will be possible to leave comments anywhere, so rather than say highlighting the lines where comments can be left we felt it better to highlight lines where comments _can't_ be left - this way once the restriction is removed there will be no re-learning the interface for the user. We thought that maybe a subtle crosshatch or strikeout pattern in the margin might be the best visual indication for "you can't leave comments here".

A rough mockup of what I'm envisioning, where the cross-hatched areas represent areas that can't be commented on:

image

jcansdale:
> the different bg colored margin indicates that the line can be commented on. It does not imply that the line has changed in that branch.

At the moment only lines within 3 lines of a change can be commented on (in its current incarnation, but this is likely to change at some point). In the mockup above, the cross-hatched areas represent areas that can't be commented on.

Without any other visual cues, the non-cross-hatched areas do absolutely imply that a line has changed in that branch. The user can see that one or more lines have changed, but it isn't clear which lines have changed. Hence the need to disambiguate changed lines from commentable lines.

I'm be happy with a subtle que indicating which lines have changed (doesn't have to be red/green). This will become all the more important if/when all lines become commentable.

shana:
I think we should probably focus on the UI for the diff view only for now, and consider the UI for the editor itself to be a separate case - we've already seen that they have different context and different use cases, and considering them together is increasingly complicating things. We should have a separate issue for discussing the design of the inline comments UI in the editor itself, I think.

So, theming! In .com we always show a blue comment icon regardless of the type of change that happened in the line. What do you think of doing the same here @donokuda?

jcansdale:
> So, theming! In .com we always show a blue comment icon regardless of the type of change that happened in the line. What do you think of doing the same here @donokuda?

This is basically what I prototyped in the latest commit (at least in the light themes). I stopped the glyphs from differentiating between their added/deleted line status. @donokuda can back this out if he wants but I wanted to dogfood it for a while. Having used it for a bit I do prefer the more sober colors.

I think we still need to consider how best to represent non-commentable lines (keeping in mind they'll probably go away eventually). In this latest commit, I found it could be confusing when there's equal mix of commentable and non-commentable lines.  Anyway, just an observation.

src/GitHub.InlineReviews/Tags/AddInlineCommentGlyph.xaml
```diff
@@ -2,36 +2,55 @@
...
+                </DataTrigger>
+
+                <DataTrigger Binding="{Binding DiffChangeType}" Value="None">
+                    <Setter Property="Background" Value="{DynamicResource GitHubDiffChangeBackground.None}" />
+                </DataTrigger>
+            </Style.Triggers>
+        </Style>
+    </UserControl.Style>
+
+    <Grid>
```
shana: Just doing a little comment to see how it looks...

shana:
@donokuda master merged in. this is how it looks to me right now:

image

src/GitHub.InlineReviews/Tags/AddInlineCommentGlyph.xaml
shana: ... and replying to myself in the extension!

donokuda:
@shana Thank you! 🙇 

Looks like there's some styling stuff I need to fix but it should be pretty straightforward



Or at least more straightforward than me attempting to fix merge conflicts 😄

grokys:
Just taking a look at this now: looks great in general, a few things though:

- The white background to the peek view doesn't look great on the blue theme. Could we use the gray that the light theme uses?
- On the light theme, the text box background is the same gray as the background and so the text box looks disabled
- I think we still need some visual indication of where you can leave comments. There's a gray background on lines that were add/removed but IMO it would be better if that were removed and used for lines that can't be commented on

grokys:
@donokuda merged master which now includes #1112 for allowing for the peek view to auto-size to its content.

donokuda:
> - The white background to the peek view doesn't look great on the blue theme. Could we use the gray that the light theme uses?
> - On the light theme, the text box background is the same gray as the background and so the text box looks disabled

Good calls on both. For the textbox, I updated to use the searchbox background color that VS provides us. Here's what the peek view looks like right now:

donokuda:
> I think we still need some visual indication of where you can leave comments. There's a gray background on lines that were add/removed but IMO it would be better if that were removed and used for lines that can't be commented on

Updated the margins:

donokuda:
Updated codeblock styles:

image

image

image

@grokys @jcansdale I think this is ready for another review. This branch now has better support for themes and fixes up the peek view so that the experience is better than it was before (by removing and reduce the sizes of some elements.)

src/GitHub.InlineReviews/Tags/AddInlineCommentGlyph.xaml
```diff
@@ -2,16 +2,55 @@
...
+        <Border Background="{DynamicResource GitHubGlyphMarginCommentableBackground}" BorderThickness="0,0,1,0" />
+        <Viewbox x:Name="AddViewbox" Margin="1,1,0,0">
+            <Canvas Width="14" Height="14">
+                <Rectangle Width="13" Height="13" />
```
grokys: Is this `Rectangle` needed?

src/GitHub.InlineReviews/Tags/AddInlineCommentGlyph.xaml
```diff
@@ -2,16 +2,55 @@
...
+                <Rectangle Stroke="#313131" Fill="Transparent" Opacity="0.163" Width="13" Height="13" />
+                <Path Canvas.Top="-0.5" Canvas.Left="-0.5" Fill="#FFF" Data="M11 8H8v3H6V8H3V6h3V3h2v3h3z"/>
+
+                <Canvas.Resources>
+                    <Style TargetType="Rectangle">
+                        <Style.Triggers>
```
grokys: Are these triggers still used? AFAICS the add glyph is now always the same color?

src/GitHub.InlineReviews/Tags/ShowInlineCommentGlyph.xaml
```diff
@@ -3,18 +3,54 @@
...
+            <Canvas Width="16" Height="15">
+                <Path Canvas.Top="1" Canvas.Left="0.85"
+                    Data="M1,0 C0.45,0 0,0.45 0,1 L0,9 C0,9.55 0.45,10 1,10 L3,10 L3,13.5 L6.5,10 L13,10 C13.55,10 14,9.55 14,9 L14,1 C14,0.45 13.55,0 13,0 L1,0 Z" />
+                <Path Canvas.Top="1" Canvas.Left="0.85" Opacity="0.163" Stroke="#313131" 
+                      Data="M3.5,12.2928932 L6.29289322,9.5 L13,9.5 C13.2738576,9.5 13.5,9.27385763 13.5,9 L13.5,1 C13.5,0.726142375 13.2738576,0.5 13,0.5 L1,0.5 C0.726142375,0.5 0.5,0.726142375 0.5,1 L0.5,9 C0.5,9.27385763 0.726142375,9.5 1,9.5 L3.5,9.5 L3.5,12.2928932 Z" />
+
+                <Canvas.Resources>
+                    <Style TargetType="Path">
+                        <Style.Triggers>
+                            <DataTrigger Binding="{Binding DiffChangeType}" Value="Add">
```
grokys: Again, are these triggers needed?

src/GitHub.InlineReviews/Views/GlyphMarginGrid.xaml
```diff
@@ -4,67 +4,17 @@
...
-                            </GeometryDrawing.Geometry>
-                        </GeometryDrawing>
-                    </DrawingGroup.Children>
-                </DrawingGroup>
-            </DrawingBrush.Drawing>
-        </DrawingBrush>
-        -->
-        
-    </Grid.Background>
+    <Border Background="{DynamicResource GitHubPeekViewBackground}" BorderBrush="{DynamicResource GitHubPeekViewBackground}" BorderThickness="0,0,1,0"/>
```
grokys: Do we need to specify a border when the background is the same color?

donokuda: Hmm, maybe not. I'll check on this.

donokuda:
Here's what I got so far as of 2ba07ba. Good call on giving a little more spacing between the author information and the comment.

image

src/GitHub.InlineReviews/Tags/AddInlineCommentGlyph.xaml
donokuda: Yup! The background is being set with a style.

donokuda:
@grokys Ready for another 👀

PR #1083, opened 2017-07-19
Add an AutomationPeer to custom controls and views

### LINKED ISSUE #969: Custom controls no longer have Automation IDs
While running GHfVS on master, I noticed that things like the GitHubPaneView and the PullRequestListView no longer have automation ids like they did in 2.2.0.8. They also seem to be missing in 2.2.0.10. Controls within those views (ie: Create new button in the PR list view) still have their automation ids.

paladique:
At some point, custom controls and views no longer had their automation ids, even though their `automationid` properties were set. This was possibly due to some recent changes to the code base (can't pinpoint to where). This causes issues with test automation, as the framework cannot find the controls since majority of them are searched by `AutomationId`.

Now, these controls require an `AutomationPeer` to expose these properties. Adding this to the ViewBase will create the peer and add the AutomationProperties.AutomationId that is set in xaml.

This can be tested by examining the WPF tree with a tool like Inspect or WPF Inspector.

Fixes #969


src/GitHub.UI.Reactive/Controls/ViewBase.cs
```diff
@@ -8,6 +8,7 @@
 using GitHub.ViewModels;
 using NullGuard;
 using ReactiveUI;
+using System.Windows.Automation.Peers;
```
jcansdale: Tiny nit. I'd move this up to be with the other System usings. ;-)

PR #1088, opened 2017-07-26
Disable PR comment margin in editor by default

jcansdale:
This PR makes showing PR comments/changes in the editor an option that's off by default. ~~The comment margin won't be constructed at all unless the option is explicitly enabled.~~ The editor margin visibility is controled via the option below.

This option looks like this (updated).

image

I'm pained to hide it behind an option because it's a feature that I use all the time (more so than in the diff view). Once a PR has been created, I find the margin helps be track its status. I realize that some others (@shana 😉) might not work this way.

Some suggestions

At the moment there isn't a UI element that shows the current PR and comment status of file being edited. This could be a good place to navigate to the PR detail view and show comments for active document. You currently need to know which PR you're on and hunt for it in the PR list view.

- The comment margin could be toggled via some UI element
- The margin could be auto-hidden if there are no comments/changes in the current file
- The margin could be displayed if there are active comments on the current file


shana:
The text should have an `experimental` note on it and we should probably explain that it may change in future versions. The editor comments feature is something that we should design separately from the diff view comments and it might dramatically change depending on what we end up wanting to do there.

jcansdale:
@shana something like this?

image

jcansdale:
I've updated the PR description with some suggestions.

shana:
> I'm pained to hide it behind an option because it's a feature that I use all the time (more so than in the diff view). Once a PR has been created, I find the margin helps be track its status. I realize that some others (@shana 😉) might not work this way.

@jcansdale This is not about how I work or not work or my personal preferences, this is about what problem we're trying to solve and whether this solution fits, and all of the work that goes into figuring that out. In the same vein, you should look at features not from the point of view of solving your own personal preferences, but how they fit into the overall workflow for our users, how they augment specific tasks, what problems they solve, and especially what problems they don't solve and how they can be detrimental to other tasks that the user may perform.

All of the discussions that we've had on how to do inline comments in the code editor and all the problems that come with it have made it pretty clear that we should be treating this as a separate feature in its own right, and not as a side effect of having inline comments in the diff view, I think. Therefore, we should not be shipping it as-is. It's not complete and we should be dedicating time to spec'ing it out, mocking it up, identifying the problems we want to solve and how this feature fits into the solution, and implementing that.

src/GitHub.InlineReviews/ViewModels/GlyphMarginViewModel.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        public void Dispose()
+        {
+            packageSettings.PropertyChanged -= PackageSettings_PropertyChanged;
+        }
+
+        void PackageSettings_PropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            if(e.PropertyName == nameof(packageSettings.EditorComments))
```
shana: Spacing

src/GitHub.InlineReviews/ViewModels/GlyphMarginViewModel.cs
```diff
@@ -0,0 +1,92 @@
...
+            this.wpfTextViewHost = wpfTextViewHost;
+            this.isDiffView = isDiffView;
+
+            Visible = IsVisible();
+            packageSettings.PropertyChanged += PackageSettings_PropertyChanged;
+        }
+
+        public void Dispose()
+        {
+            packageSettings.PropertyChanged -= PackageSettings_PropertyChanged;
```
shana: This should follow the established Dispose pattern that we use everywhere else in the codebase

src/GitHub.InlineReviews/InlineCommentMarginProvider.cs
```diff
@@ -30,68 +30,59 @@ internal sealed class InlineCommentMarginProvider : IWpfTextViewMarginProvider
...
-                router.Add(host.HostControl, marginElement);
-            }
-        }
+            var margin = new GlyphMargin<TGlyphTag>(wpfTextViewHost, glyphFactory, glyphMarginView, tagAggregator, editorFormatMap,
+                MarginPropertiesName, MarginName, true, 17.0);
 
-        bool IsMarginVisible(ITextBuffer buffer)
-        {
-            if (sessionManager.GetTextBufferInfo(buffer) != null)
+            if(IsDiffView(wpfTextViewHost))
```
shana: Spacing

src/GitHub.InlineReviews/InlineCommentMarginProvider.cs
```diff
@@ -30,68 +30,59 @@ internal sealed class InlineCommentMarginProvider : IWpfTextViewMarginProvider
...
-        {
-            var textView = host.TextView;
-            if (textView.Roles.Contains("DIFF"))
-            {
-                var router = new MouseEnterAndLeaveEventRouter<AddInlineCommentGlyph>();
-                router.Add(host.HostControl, marginElement);
-            }
-        }
+            var margin = new GlyphMargin<TGlyphTag>(wpfTextViewHost, glyphFactory, glyphMarginView, tagAggregator, editorFormatMap,
+                MarginPropertiesName, MarginName, true, 17.0);
```
shana: Is this the width of the margin? If so, it should be a const value somewhere else, so we don't have to hunt it down in the code.

src/GitHub.InlineReviews/ViewModels/GlyphMarginViewModel.cs
jcansdale: Do we need to follow that pattern even if the class is sealed and isn't designed to be extended? I thought that pattern about about allowing desiging with extending classes in mind.

shana: It's also to protect from Dispose being called twice and generally defensive programming. If we keep it consistent, we don't have to think about it.

jcansdale:
> We should probably not be doing any comment processing at all if the margin isn't visible, especially the InlineCommentTagger - it shouldn't be processing data if the UI is disabled, that introduces a needless performance penalty for no benefit. Looking at the tagger, it looks like it's always on regardless?

It's still very much a work in progress. I was planning to migrate more stuff into the `GlyphMarginViewModel` and be sensible about when to construct the tagger. Certainly don't want an active tagger for no reason!

src/GitHub.InlineReviews/ViewModels/GlyphMarginViewModel.cs
```diff
@@ -0,0 +1,104 @@
...
+            var buffer = wpfTextViewHost.TextView.TextBuffer;
+            if (sessionManager.GetTextBufferInfo(buffer) != null)
+            {
+                return true;
+            }
+
+            InlineCommentTagger inlineCommentTagger;
+            if (buffer.Properties.TryGetProperty(typeof(InlineCommentTagger), out inlineCommentTagger))
+            {
+                if (inlineCommentTagger.ShowMargin)
```
jcansdale: Should maybe make `ShowMargin` fire a propetty change event rather than listening to `BatchedTagsChanged`.

src/GitHub.InlineReviews/InlineCommentMarginProvider.cs
```diff
@@ -30,68 +30,59 @@ internal sealed class InlineCommentMarginProvider : IWpfTextViewMarginProvider
...
         {
             var tagAggregator = tagAggregatorFactory.CreateTagAggregator<TGlyphTag>(wpfTextViewHost.TextView);
-            var margin = new GlyphMargin<TGlyphTag>(wpfTextViewHost, glyphFactory, gridFactory, tagAggregator, editorFormatMap,
-                IsMarginVisible, MarginPropertiesName, MarginName, true, 17.0);
 
-            TrackCommentGlyphOnDiffView(wpfTextViewHost, margin.VisualElement);
-            return margin;
-        }
+            var isDiffView = IsDiffView(wpfTextViewHost);
+            var glyphMarginViewModel = new GlyphMarginViewModel(packageSettings, sessionManager, wpfTextViewHost, isDiffView);
```
jcansdale: Should pass the `GlyphMarginViewModel` an instance of `InlineCommentTagger` for the view if one exists. The code creating the `InlineCommentTagger` should look at `IPackageSettings.EditorComments` when deciding to create a tagger (rather than doing this check in `GlyphMarginViewModel`). That way we can defer creating the TagAggregator until it's actually used.

src/GitHub.InlineReviews/ViewModels/GlyphMarginViewModel.cs
jcansdale: I'm no longer refreshing visiblility on BatchedTagsChanged.

PR #1089, opened 2017-07-26
Switch to a custom nuget package of Rothko

shana:
This should hopefully remove the last references of fody on our codebase and make ncrunch happy.


jcansdale:
It's getting a lot further than last time. 😄 

Alas it looks like the `UnitTests` project itself uses Fody. I wonder how tricky it would  be to remove the dependency?

Here's what I'm getting:
```
NCrunch: This project was built on server 'ORANGE'
Fody: Could not find a weaver named 'EntryExitDecorator'. 
If you have nuget package restore turned on you probably need to do a build to download the weavers. 
Alternatively you may have added a weaver to your 'FodyWeavers.xml' and forgot to add the appropriate nuget package. 
Perhaps you need to run 'Install-Package EntryExitDecorator.Fody'. This url may provide more information http://nuget.org/packages/EntryExitDecorator.Fody/ .
```

jcansdale:
It required a little extra tweak because the `UnitTests` project still uses Fody. The required files are now copied to the correct place on the remote NCrunch workspace. A few tests are failing when executed remotely, but at least they're now running!

Thanks for this fix. 👍

rubenscholfield:
Everything Ok
Bankstuff etc. ready
Trainingthings on doing
will respond

Sec. General Fleet Comander Vice Admiral Scholfield

Am 26.07.2017 6:35 PM schrieb "Jamie Cansdale" <notifications@github.com>:

shana:
Fody is being used in the unit tests to inject Entry and Exit method calls to each test, since xunit doesn't believe in having methods that trigger at the start and end of each test. By the time I needed this in all the tests, there were way too many tests to manually add calls to all of them. The source for that fody plugin is https://github.com/shana/EntryExitDecorator.Fody

grokys:
Regarding Fody in the unit tests - afaics it only uses it to call `Splat.ModeDetector.Current.SetInUnitTestRunner(true);` on entry. Can't this be done in the `TestBaseClass` constructor?

Issue #1077, opened 2017-07-17
Surfacing interesting inline comments threads


donokuda:
Opening up this discussion based on some discussion in https://github.com/github/VisualStudio/pull/1025

The tooltip that's used for previewing inline comments could be improved, so let's have a discussion on how we can improve it and decide on a plan moving forward.

The original purpose for the tooltip was stated as:

> a quick way to find comment threads of interest

Given that, what makes a comment thread interesting to a developer? Particularly one who is either looking for feedback or giving feedback in a pull request. Few things that came to my mind:

- **A new comment thread was created:** If I'm looking for feedback, then any new thread means that I might 
- **A new comment to an existing thread was created:**
  - If I'm a pull request author, then a new comment from someone else is interesting to me because it might influence what changes I plan on doing next.
  - If I'm a reviewer and someone has commented on my thread (whether it's the author or not), I might be interested in the thread because it could be a response that requires a reply from me.
- **A thread has a lot of participates**
  - Could this indicate that a lot of people have varying opinions on something? If I'm the author, is this a thread that I should be focused on the most?

These were the first couple that came to my head. I'll most likely add more as I think of them. I find it useful to state these down so we can start digging into exactly what type of thread is useful and how we can surface that to our users. One solution that comes to my mind is calling out (maybe with an indicator) of these threads that might useful/relevant to a developer.

One suggestion I had for improving the tooltip (in https://github.com/github/VisualStudio/pull/1025#issuecomment-315436535), is trying to summarize the first and last comment while displaying how many comments aren't shown:

```
+------------------------------+
|   (First comment truncated)  |
| ~~~ 5 comments not shown ~~~ |
|   (Last comment truncated)   |
+------------------------------+
```

I was suggesting that we show the first comment (and truncate for space) since I assume the pull request description would be the best place to understand why the PR is needed and what is being changed. The last comment is assumes that it would be the best place to understand the next step of the feedback process. Last, showing the amount of comments not shown (or some indication with the amount activity) could indicate that there's might be a lot of opinions in the thread worth reading.

☝️ All that said above, I'm suggesting this approach purely based on assumptions about what developers would need if they were to hover over the 💬 in the margin.

To move forward with designing, I'd like to better understand what developers are looking for when they are an author of a pull request (and getting their code reviewed) or when they are responsible for reviewing code. What are examples of comments in a thread that help you understand what action to take next, and where do they appear (at the end, in the middle, etc?)

donokuda:
[...]

PR #1091, opened 2017-07-27
[wip] Preview inline comments as a tooltip, part deux

shana:
Going back to the drawing board on #1025, which was prematurely merged to master and reverted on #1090 

We should address all the concerns raised in #1025 and #1077 before considering this done.


sguthals:
I think this one is along the lines of what we were talking about in our PR Reviews chat on March 12th here: https://github.com/github/VisualStudio/pull/1415#issuecomment-372379990

I would say we should have an issue tracking the use of popup tooltips?

PR #1100, opened 2017-08-01
Fix submodule link generation.

mayuki:
When a current document is under submodule repository, "Open on GitHub" context menu opens a wrong link. I think the base repo should be determined by the path of the active document.

## Repro
1. Clone `github/VisualStudio` repo and `git submodule init/update`
2. Open `GitHubVS.sln` and `submodules\octokit.net\Octokit\GitHubClient.cs`
3. Show ContextMenu and select **"Open on GitHub"**
4. Open https://github.com/github/VisualStudio/blob/2436b20db91387ab51a5b6456fc1af4aa15fd59d/submodules/octokit.net/Octokit/GitHubClient.cs in browser but it's 404 page


src/GitHub.VisualStudio/Menus/MenuBase.cs
```diff
@@ -46,6 +46,24 @@ protected MenuBase(IGitHubServiceProvider serviceProvider)
...
         }
 
+        protected ILocalRepositoryModel GetRepositoryByPath(string path)
+        {
+            try
+            {
+                if (!string.IsNullOrEmpty(path))
+                {
+                    var repo = ServiceProvider.TryGetService<IGitService>().GetRepository(path);
+                    return new LocalRepositoryModel(repo.Info.WorkingDirectory.TrimEnd('\\'));
```
grokys: Interesting that `TrimEnd('\\')` is needed here. Seems to be a failure in the logic in `LocalRepositoryModel.GenerateUrl`. Getting a common subpath in C# isn't easy and seems we're not doing it correctly there. However, for the moment trimming the path here is a good workaround - we should fix that method in another PR.

grokys: CI is failing due to the assembly signing in `GitHub.InlineReview.csproj` - we should merge that PR first to get CI working.

mayuki: Thank you for your comment!

This workaround is not smart. Visual Studio's Git Provider returns directory path ends **without** "`\`", however when calling `IGitService.GetRepository(LibGit2Sharp)` with a full path of the document returns a working dir path ends **with** "`\`".

I couldn't decide where to be consistent as the best. I think that should fix as you say.

grokys:
Merged! Thanks for this @mayuki!

PR #1101, opened 2017-08-01
Fix inline comment position in inline diff view.

### LINKED ISSUE #1099: Peek view sometimes shows up in wrong place in inline diff mode
For example this comment:

https://github.com/AvaloniaUI/Avalonia/pull/827#discussion_r126401935

Appears on the wrong line:

image

grokys:
We need to map between projection buffers: we were previously doing it for deleted lines, but we need to do it for all lines in the inline diff view.

Fixes #1099


shana:
Currently getting this behaviour:

pr-1101

rubenscholfield:
Canada in Come around course of work

rubenscholfield:
Can in come around course of workk

rubenscholfield:
???
need your adress

rubenscholfield:
Email rubenscholfield@gmail.com

rubenscholfield:
Greatings from germany

src/GitHub.InlineReviews/Services/InlineCommentPeekService.cs
```diff
@@ -51,18 +51,18 @@ class InlineCommentPeekService : IInlineCommentPeekService
             {
                 if (diffModel.ViewType == DifferenceViewType.InlineView)
                 {
-                    // If we're displaying a diff in inline mode, then we're in the left buffer if
```
grokys: // If we're displaying a diff in inline mode, then we're in the left buffer if

src/GitHub.InlineReviews/Services/InlineCommentPeekService.cs
```diff
@@ -51,18 +51,18 @@ class InlineCommentPeekService : IInlineCommentPeekService
...
-                    // the point can be mapped down to the left buffer.
+                    // If we're displaying a diff in inline mode, then we need to map the point down
```
grokys: // If we're displaying a diff in inline mode, then we need to map the point down

src/GitHub.InlineReviews/Services/InlineCommentPeekService.cs
```diff
@@ -51,18 +51,18 @@ class InlineCommentPeekService : IInlineCommentPeekService
...
+                    // to the left or right buffer.
                     var snapshotPoint = point.GetPoint(point.TextBuffer.CurrentSnapshot);
-                    var mappedPoint = session.TextView.BufferGraph.MapDownToBuffer(
+                    var mappedPoint = session.TextView.BufferGraph.MapDownToFirstMatch(
                         snapshotPoint,
                         PointTrackingMode.Negative,
```
grokys: PointTrackingMode.Negative,

jcansdale:
Testing *without* this PR on the `PullRequestsSandbox`. Everything is fine up until line 14, when it appears on the line above where it's supposed to.

image

Testing with this PR, it appears to be fixed!

image

PR #1108, opened 2017-08-03
Fix View.GitHub command and consolidate Guids/PkgCmdIDList

jcansdale:
This fixes the `View > Other Windows > GitHub` command. The command was simply not being called.

It also removes the reference from `GuidList` to `GlobalCommands` and instead consolidates our solution's constants in `Guids.cs` / `PkgCmdID.cs` (in the `GitHub.Exports` project). This lets us access these constants from anywhere without having to reference `GitHub.VisualStudio`.


src/GitHub.Exports/Settings/PkgCmdID.cs
```diff
@@ -18,5 +19,10 @@ public static class PkgCmdIDList
         public const int githubCommand = 0x320;
         public const int helpCommand = 0x321;
         public const int blameCommand = 0x500;
```
jcansdale: This ID is referenced by `OpenPullRequests` but isn't actually defined in the `GitHub.VisualStudio.vsct` file. Probably worth fixing in a different PR.

jcansdale: Arg this isn't the line I thought I commented on.  I tried to comment on the following line:
```
        public const int openPullRequestsCommand = 0x201;
```

src/GitHub.Exports/Settings/PkgCmdID.cs
```diff
@@ -18,5 +19,10 @@ public static class PkgCmdIDList
```
jcansdale: comment on addConnectionCommand

src/GitHub.Exports/Settings/PkgCmdID.cs
```diff
@@ -4,6 +4,7 @@ namespace GitHub.VisualStudio
 {
     public static class PkgCmdIDList
     {
+        // IDs defined in GitHub.VisualStudio.vsct
```
grokys: Adding comment on comment.

src/GitHub.Exports/Settings/PkgCmdID.cs
```diff
@@ -4,6 +4,7 @@ namespace GitHub.VisualStudio
...
         public const int addConnectionCommand = 0x110;
         public const int idGitHubToolbar = 0x1120;
         public const int showGitHubPaneCommand = 0x200;
```
jcansdale: Commenting on the line with  `showGitHubPaneCommand`.

PR #1113, opened 2017-08-03
Fix building and signing and all this mess

shana:
The csproj files, particularly the GitHub.VisualStudio one, were a total mess, probably due to merges.

This fixes internal and external build signing and normalizes how we do signing in a common `signing.props` file that all projects requiring signing include it.

This also adds a fallback in case of external builds to support user signing keys, with instructions in the readme on how to add one. Ideally we'd have a public key for users to use and not bother adding one manually, but I'd rather do that once we have checks in place to verify that the internal builds and packages are signed correctly.

This does not fix the broken markdig problem in master, that's on a separate PR.

### DIFF
--- a/src/common/signing.props
+++ b/src/common/signing.props
@@ -10,9 +10,19 @@
     <DelaySign>false</DelaySign>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(BuildType)' != 'Internal' and Exists('..\..\publickey.snk')">
+    <AssemblyOriginatorKeyFile>..\..\publickey.snk</AssemblyOriginatorKeyFile>
+    <SignAssembly>true</SignAssembly>
+    <DelaySign>false</DelaySign>
+  </PropertyGroup>
+
   <ItemGroup>
     <None Include="..\..\script\Key.snk" Condition="'$(BuildType)' == 'Internal'">
       <Link>Key.snk</Link>
+    </None>
+
+    <None Include="..\..\publickey.snk" Condition="'$(BuildType)' != 'Internal' and Exists('..\..\publickey.snk')">
+      <Link>publickey.snk</Link>
     </None>
   </ItemGroup>
 </Project>
--- a/src/common/signing.props
+++ b/src/common/signing.props
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <BuildType Condition="Exists('..\..\script\src\MetricsService.cs')">Internal</BuildType>
+    <BuildType Condition="Exists('..\..\script\Key.snk')">Internal</BuildType>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(BuildType)' == 'Internal'">
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Beta releases will have `(beta)` in their title in the gallery, following the ve
 
 ## Build requirements
 
-* Visual Studio 2015
+* Visual Studio 2015+
 * Visual Studio SDK
 
 ## Build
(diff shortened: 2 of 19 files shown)


src/common/signing.props
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <BuildType Condition="Exists('..\..\script\src\MetricsService.cs')">Internal</BuildType>
```
grokys: One thing I've always wondered: why not test for `..\..\script\Key.snk` to detect an internal build as that's the thing we need for signing?

shana: lol yeah that would probably be smarter, I just randomly picked one of the files in there that weren't likely to move :P

jcansdale:
I mentioned in an inline comment about issues when the assemblies aren't signed. Here is a symptom if this issue. This is the most obvious place to see when it's not working, but there are more insidious and confusing things what will also go wrong:

image

To reproduce this:

1. Delete `build` and `script` directories.
2. msbuild GitHubVS.sln /t:Clean
3. build.cmd
4. install.cmd
5. Launch Visual Studio 2015
6. Close the `Team Explorer` and `GitHub` tool windows (if you can see their tabs)
7. Close Visual Studio
8. Launch Visual Studio
9. Check the icon for `View > Other Windows > GitHub` has gone. 😭 

Now create the `publickey.snk` and repeat (icon should now appear).

MSBuild won't rebuild the assemblies if `publickey.snk` has been added or removed (which is why 1. and 2. are necessary).

jcansdale:
When the signing it working correctly, the `AssemblyResolverPackage` is only required as mitigation for #911. In its current form the `AssemblyResolverPackage.cs` has two responsibilities.

1. As mitigation for #911 
2. For diagnostics when we're faced with possible assembly resolving issues

I think we should clearly separate these responsibilities. 

1. Include a stub package to as mitigation for #911 
```
    [Guid(GuidList.guidAssemblyResolverPkgString)]
    public class AssemblyResolverPackage : Package
    {
        // This empty packge should be kept as a fix for:
        // https://github.com/github/VisualStudio/pull/911
    }
```

2. Comment out the original `AssemblyResolverPackage` with `/* */`, but uncomment the following attributes (which are necessary for it to work):
```
    [ProvideAutoLoad(Guids.GitSccProviderId)]
    [ProvideAutoLoad(VSConstants.UICONTEXT.NoSolution_string)]
```

I'm generally not a fan of commented out code, but assembly resolving issues are a common enough challenge that it's useful having this resolver to hand.

What are your thoughts? Of these two suggestions, I think 1. is the more important one.

grokys:
Rather than commented out code, how about something like `#if DEBUG_ASSEMBLY_LOADING` or something? That way it's clear what the code is doing.

jcansdale:
> Rather than commented out code, how about something like #if DEBUG_ASSEMBLY_LOADING? That way it's clear what the code is doing.

That is a much better idea! 👍

shana:
> Also <Import>s are usually towards the bottom of .csproj files, might be easier to find the signing.props import if it were moved with the others? I don't have a strong opinion on this though.

csproj organization is generally properties, references, includes, targets, so `.props` files are imported at the top of the file, and `.targets` files at the bottom. The `BuildType` property is defined in the `signing.props` file, therefore it needs to be with the other property definitions, before anything else tries to use it.

shana:
@jcansdale Are you sure that's not a symptom of running an experimental build in the non-experimental hive? If you debug the instance of VS that is failing, where are the DLLs being loaded from?

src/common/signing.props
```diff
@@ -0,0 +1,28 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(BuildType)' == 'Internal'">
+    <AssemblyOriginatorKeyFile>..\..\script\Key.snk</AssemblyOriginatorKeyFile>
+    <SignAssembly>true</SignAssembly>
+    <DelaySign>false</DelaySign>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(BuildType)' != 'Internal' and Exists('..\..\publickey.snk')">
```
shana: That means changing our CI to generate a key before end for public builds so that we can actually build external PRs, otherwise builds will fail. I've added a note on the README saying that the public key is required, and this has been broken since we disabled the assembly resolver, so I think external builds can survive another week or two until we get our release in shape, and then we can add the checks we need to make sure all our builds are correct.

jcansdale:
The main issue is that the `[ProvideCodeBase]` attributes are ignored (when the assembly is unsigned) and the assemblies are being loaded into the wrong context. I've created a separate issue for this. #1116

PR #1114, opened 2017-08-03
Fix markdig signing

shana:
- [x] Depends on #1113

StrongNameSigner is not doing enough when signing unsigned assemblies to allow xaml to properly reference signed assemblies, and things fail at runtime.

Markdig has a signed version of their package, so we're reverting to a custom build of Markdig.Wpf - https://github.com/editor-tools/markdig-wpf/releases/tag/v0.2.1 - that references the Markdig.Signed package and is also signed. Also adds ProvideCodeBase to ensure that these DLLs are found at runtime.

Also, markdown.xaml should not be included in the GitHub.UI shared dictionary. The SharedDictionary in GitHub.UI includes the lowest level of controls and is included in all the Team Explorer views (Home, Connect, etc). Therefore it should not bring in any external dependencies that cause other libraries to get loaded. Since markdown.xaml requires Markdig now, it should be referenced only in the views that use it.


jcansdale:
> StrongNameSigner is not doing enough when signing unsigned assemblies to allow xaml to properly reference signed assemblies, and things fail at runtime.

I've noticed that when I run unit tests that use XAML resources, the resources don't get updated unless I explicitly build the project that contains them. I suspect that XAML resources don't get compiled as part of the regular build pipeline and this is why the StrongNameSigner changes aren't getting incorporated (it happens too late).

grokys:
Will do a proper review when #1113 is merged (diff is too noisy right now) but I note that we're still signing the Markdig packages here:

https://github.com/github/VisualStudio/blob/c21fa052067621377b0cd1320dbc732c989b36dc/src/common/StrongNameSign.targets#L4-L5

We should probably remove that `.`targets` file now it's not being used.

I'm also curious as to what StrongNameSigner isn't doing!

shana:
I left the targets file in to make it easier to include it back for figuring out why it's not working properly. I was kinda "should I leave it, shouldn't I?" on this one so I'm good with removing if you prefer. It's not being included anywhere so it doesn't run, and the nuget package for it is not in the build.

PR #1120, opened 2017-08-04
Prevent null reference exception in PullRequestSessionManager.RepoChanged

### LINKED ISSUE #1119: Opening a solution not in a git repository cause NRE in PullRequestSessionManager.RepoChanged
If a solution that isn't in a git repo is opened then the `ILocalRepositoryModel` passed to `PullRequestSessionManager.RepoChanged` will be null. We're not handling this (although it doesn't cause a crash as there's a `try` block around it.

grokys:
If a solution that isn't in a git repo is opened then the ILocalRepositoryModel passed to PullRequestSessionManager.RepoChanged will be null. Similarly a the repository has no remote then `ILocalRepositoryModel.CloneUrl` will be an empty string, These were causing exceptions.

There was a `try` block around the code that was throwing so the end result before and after this PR is the same - this just avoids the exception.

Fixes #1119


jcansdale:
I just tried this on a solution not on github. I got following exception:

image

I don't know if this is related at all?

jcansdale:
It looks like it just wrote something to the log. No crash or anything.

grokys:
@jcansdale - well spotted, new fixed in this PR in 96bc6e4cce83b51a17b73ca6d8d8af1555e34f6d

PR #1123, opened 2017-08-07
Add message for pending review comments.

### LINKED ISSUE #1044: clarify validation message derived from review process
While comments are pending as part of the review process, the user on VS cannot post a comment. We do say "Validation pending" but perhaps this could be a little clearer. I suggest tailoring the message or just putting a error message and not allowing the user to even reply.

screen shot 2017-06-30 at 10 56 31 am
above has to wait for this to be cleared via review process:
screen shot 2017-06-30 at 10 56 46 am

(contact @tierninho for questions)

grokys:
If the user has pending review comments on the server then we can't post new comments. The correct way to test for this would be to make a request to `/repos/:owner/:repo/pulls/:number/reviews` and check for comments with a `PENDING` state and disable the comment box with a message. For the moment however we'll just modify the error message displayed to the user, as a rather non-helpful "Validation Failed" message is returned from the server.

This is a bit of a hack and I'm not sure about the wording of the message, but it's the best we can do for now without writing a fair bit of code to check for the pending comments.

Fixes #1044


jcansdale:
This works fine.

image

PR #1102, opened 2017-08-01
Show current PR on status bar and allow navigation to PR details MVP

jcansdale:
This adds some UI that displays information about any PR associated with the current branch. At the moment it simply shows the PR number and has a tooltop for its title. This was to show what data it is currently populated with.

Maybe we could allow navigation the the PR details pane or let the user toggle modes for adding comments?

This primarily a branch for experimenting with different UI ideas.

image

Fixes #1097

PR #1128, opened 2017-08-09
Add VS command to show current PR details

jcansdale:
This is just the command from #1102 without the UI.

The comand is called `GitHub.ShowCurrentPullRequest` and can be bound to a keyboard shortcut. It can be later be surfaced on a UI.


jcansdale:
@shana Do you think this modest little PR could make its way into the beta? 😉 

I'm finding not having an easy way to navigate the to PR details view a pain. The UI from #1102 certainly needs more work, but the command itself works very well.

jcansdale:
Even simply binding to key is useful, but I can't wait for a UI as well! 😄

PR #1130, opened 2017-08-10
Make it easier to add new usage metrics.

grokys:
Currently, adding a new usage metric is a bit of a PITA. You need to:

- Add a new field to `UsageModel`
- Add a new method to `IUsageTracker`
- Add a new method to `UsageTrackerDispatcher`
- Add a new method to `UsageTracker`
- Add the field to `UsageTracker.ClearCounters`
- Add the field to `UsageModel.Clone`

This PR:

- [x] Removes the individual `IncrementX` methods from `IUsageTracker` and replaces them with an `IncrementCounter` method which takes a lambda describing the metric to update.
- [x] Uses reflection to reset the counters in `UsageTracker.ClearCounters`
- [x] Uses reflection to clone the counters in `UsageModel.Clone`
- [x] Moves IO and date comparison operations out of `UsageTracker` so it can be unit tested
- [x] Adds unit tests for `UsageTracker`


src/GitHub.VisualStudio/Services/UsageService.cs
```diff
@@ -0,0 +1,136 @@
...
+            using (var s = File.OpenRead(path))
+            using (var r = new StreamReader(s, Encoding.UTF8))
+            {
+                return await r.ReadToEndAsync();
+            }
+        }
+
+        async Task WriteAllTextAsync(string path, string text)
+        {
+            using (var s = File.OpenWrite(path))
```
jcansdale: We should be creating a new file, not overwriting the old one.

src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
```diff
@@ -11,6 +11,7 @@
     <NuGetPackageImportStamp>
     </NuGetPackageImportStamp>
     <TargetFrameworkProfile />
+    <ApplicationVersion>2.3.1.0</ApplicationVersion>
```
shana: This change shouldn't be in here. Bad merge somewhere?

src/GitHub.VisualStudio/Services/UsageService.cs
```diff
@@ -0,0 +1,136 @@
...
+                storePath = Path.Combine(
+                    Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
+                    program.ApplicationName,
+                    StoreFileName);
+            }
+        }
+
+        async Task<string> ReadAllTextAsync(string path)
+        {
+            using (var s = File.OpenRead(path))
```
shana: You can use rothko for IO services that can be mocked for unit testing

src/GitHub.VisualStudio/Services/UsageTracker.cs
```diff
@@ -1,319 +1,105 @@
...
-        }
-
         async Task TimerTick()
         {
             await Initialize();
 
-            if (firstRun)
+            if (firstTick)
             {
                 await IncrementLaunchCount();
```
shana: `IncrementLaunchCount` calls `LoadUsage` which calls `await Initialize` which has already been called right before this. Do we need these double `await Initialize` calls?

src/GitHub.VisualStudio/Services/UsageTracker.cs
```diff
@@ -1,319 +1,105 @@
...
                 connectionManager = gitHubServiceProvider.GetService<IConnectionManager>();
                 userSettings = gitHubServiceProvider.GetService<IPackageSettings>();
                 vsservices = gitHubServiceProvider.GetService<IVSServices>();
-
-                var program = gitHubServiceProvider.GetService<IProgram>();
-                storePath = System.IO.Path.Combine(
-                    Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
-                    program.ApplicationName,
-                    StoreFileName);
+                initialized = true;
```
shana: Is there a possibility of a race condition here where `await Initialize` gets called at the same time?

src/GitHub.VisualStudio/Services/UsageService.cs
grokys: Already thought of that, but Rothko doesn't have the methods we need as far as I can tell, which is partly why I split this I/O functionality out into a separate service apart from the usage tracker itself.

src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
grokys: Yep, bad merge - well spotted!

src/GitHub.VisualStudio/Services/UsageTracker.cs
grokys: There is a small possibility, if a counter is initialized on a background thread and while switching to the main thread, another counter is initialized. However all that would happen in that scenario is that services would get loaded twice, which I didn't think was a big enough problem to warrant putting a `SemaphoreSlim` or similar around.

src/GitHub.VisualStudio/Services/UsageTracker.cs
grokys: Actually, yes we can remove the `Initialize` at the top of `TimerTick`! 👍

grokys:
@shana any chance of another review of this? I seem to remember you expressing surprise that this hadn't already gone in a few weeks back.

PR #1133, opened 2017-08-11
InfoPanel Fixes.

### LINKED ISSUE #1131: GitHub pane InfoPanel is blank in 2.3.0.16 beta
I created a new PR from VS, which succeeded but the following was displayed on completion:

image

Looks like the `InfoPanel` got broken.

grokys:
A few `InfoPanel` fixes:

- Bind `MarkdownViewer.Markdown` in `InfoPanel.xaml` - the panel won't display a message if we don't bind the property!
- Use `VsBrush.InfoBackground` as the background color, previously it was just `Beige`

Fixes #1131

### DIFF
--- a/src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
+++ b/src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
@@ -51,13 +51,6 @@ namespace GitHub.VisualStudio.UI.Controls
             private set { iconColor = value; RaisePropertyChanged(nameof(IconColor)); }
         }
 
-        ICommand linkCommand;
-        public ICommand LinkCommand
-        {
-            get { return linkCommand; }
-            set { linkCommand = value; RaisePropertyChanged(nameof(LinkCommand)); }
-        }
-
         static InfoPanel()
         {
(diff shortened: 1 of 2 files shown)


src/GitHub.VisualStudio.UI/UI/Controls/InfoPanel.xaml.cs
```diff
@@ -82,12 +75,14 @@ public InfoPanel()
...
-            {
-                if (!String.IsNullOrEmpty(x as string))
-                    Browser.OpenUrl(new Uri((string)x));
-            });
+        void OpenHyperlink(object sender, ExecutedRoutedEventArgs e)
+        {
+            var url = e.Parameter.ToString();
+
+            if (!string.IsNullOrEmpty(url))
+                Browser.OpenUrl(new Uri(url));
```
jcansdale: Similar to other PR. What is `url` is malformed?

shana: We set the urls in the info panels, it's not from user input, so they should never be malformed. If they are and throw, we should catch that in testing (we are clicking on every infopanel that we have during testing to make sure nothing regressed, right? 😉)

PR #1135, opened 2017-08-11
Fix links in PR details view.

### LINKED ISSUE #1134: Links in Pull Request description not clickable.
In 2.3.0.16, links in PR details view are not clickable.

grokys:
The command wasn't being handled so links were disabled.

Fixes #1134

### DIFF
--- a/src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml.cs
+++ b/src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml.cs
@@ -309,11 +309,11 @@ namespace GitHub.VisualStudio.UI.Views
 
         void OpenHyperlink(object sender, ExecutedRoutedEventArgs e)
         {
-            var url = e.Parameter.ToString();
+            Uri uri;
 
-            if (!string.IsNullOrEmpty(url))
+            if (Uri.TryCreate(e.Parameter?.ToString(), UriKind.Absolute, out uri))
             {
-                VisualStudioBrowser.OpenUrl(new Uri(url));
+                VisualStudioBrowser.OpenUrl(uri);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/GitHub.VisualStudio/UI/Views/PullRequestDetailView.xaml.cs
```diff
@@ -306,5 +306,15 @@ void ViewCommentsClick(object sender, RoutedEventArgs e)
...
             catch { }
         }
+
+        void OpenHyperlink(object sender, ExecutedRoutedEventArgs e)
+        {
+            var url = e.Parameter.ToString();
+
+            if (!string.IsNullOrEmpty(url))
+            {
+                VisualStudioBrowser.OpenUrl(new Uri(url));
```
jcansdale: What would happen if `url` was malformed?

grokys: As @shana said in https://github.com/github/VisualStudio/pull/1133#discussion_r133485930 - these links are created by us, so they shouldn't be malformed.

grokys: Oh actually sorry, they're not created by us here. If they're malformed they'll be passed to the browser malformed, as they would on .com I imagine.

grokys: Actually thinking more about this, this would be worth checking...

jcansdale: I think `new Uri(...)` might throw.  Try `new Uri("boom://oops")`.

shana: Probably good to use `Uri.TryParse()` on this one

PR #1138, opened 2017-08-14
Don't add a byte order mark to editor contents.

### LINKED ISSUE #1137: Pull Request shows as modified even when it isn't

grokys:
This was causing #1137; this code was obviously added for a reason in 6b81f3ff but I can no longer reproduce VS erroneously signalling a file as changed if we don't add the BOM. Maybe we handle this elsewhere now?

I could do with some 👀 on this @jcansdale as I'm worried I'm missing something.

Fixes #1137.

### DIFF
--- a/src/GitHub.InlineReviews/Tags/InlineCommentTagger.cs
+++ b/src/GitHub.InlineReviews/Tags/InlineCommentTagger.cs
@@ -157,7 +157,9 @@ namespace GitHub.InlineReviews.Tags
 
         Task<byte[]> IEditorContentSource.GetContent()
         {
-            return Task.FromResult(GetContents(buffer.CurrentSnapshot));
+            var text = buffer.CurrentSnapshot.GetText();
+            var bytes = document.Encoding.GetBytes(text);
+            return Task.FromResult(bytes);
         }
 
         void Initialize()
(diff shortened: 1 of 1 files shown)


jcansdale:
I think the potential issue here is when a file does have an explicit BOM. Won't this new code strip it off and the file will now look like it has changed?

jcansdale:
> I can no longer reproduce VS erroneously signalling a file as changed if we don't add the BOM.

The stuff I added to to make sure the temporary file we extract from Git has the same preamble as the working directory file we compare it to. This prevents the diff view showing the ugly warning message. Was this what you means by VS erroneously signalling?

grokys:
> I think the potential issue here is when a file does have an explicit BOM. Won't this new code strip it off and the file will now look like it has changed?

This si what i was wondering, and what I was trying to repro but couldn't. I added some test PRs [here](https://github.com/grokys/PullRequestSandbox/pull/32) and [here](https://github.com/grokys/PullRequestSandbox/pull/33) but couldn't get a false changed state. Am I missing anything?

grokys:
Looking further this doesn't seem to be right. The problem seems to be caused by the fact that we add a UTF8 BOM to extracted temp files, which causes `IsModified` to think the file has been modified.

PR #1140, opened 2017-08-14
Ensure that CacheIndex.IndexKey != null.

### LINKED ISSUE #1139: Crash in Clone dialog.
Getting VS crash when the clone dialog finishes loading, caused by this exception:

```
System.ArgumentNullException occurred
  HResult=-2147467261
  Message=Value cannot be null.
  Source=System.Reactive.PlatformServices
  StackTrace:
       at System.Reactive.PlatformServices.ExceptionServicesImpl.Rethrow(Exception exception)
       at System.Reactive.ExceptionHelpers.ThrowIfNotNull(Exception exception)
       at System.Reactive.Subjects.AsyncSubject`1.GetResult()
       at GitHub.Extensions.AkavacheExtensions.<>c__DisplayClass4_0`1.<<GetAndFetchLatestFromIndex>b__13>d.MoveNext() in D:\projects\VisualStudio\src\GitHub.App\Extensions\AkavacheExtensions.cs:line 231
  InnerException: 
```

It appears that it's being caused by  `CacheIndex.IndexKey` being `null`.

This happens on 2.3.0.16 but I can also repo on 2.2.0.11. It goes away if I clear my cache and so is most likely something getting into a strange state which is probably why nobody else has reported it.

grokys:
This is a strange one. In debugging #1139, I started by trying to make sure that ` CacheIndex.IndexKey` is not null by removing its setter, setting the property in the constructor and adding a `Guard`.

Somehow this fixed #1139. I have no idea how. Reverting this PR causes the crash to show up again.

(I also gave some properties in `CacheIndex` private setters or removed setters where they were only set by the class itself or in the constructor.)

### DIFF
--- a/src/GitHub.App/Caches/CacheIndex.cs
+++ b/src/GitHub.App/Caches/CacheIndex.cs
@@ -18,11 +18,14 @@ namespace GitHub.Caches
 
         public static CacheIndex Create(string key)
         {
-            return new CacheIndex { IndexKey = key };
+            return new CacheIndex(key);
         }
 
-        public CacheIndex()
+        public CacheIndex(string key)
         {
+            Guard.ArgumentNotEmptyString(key, nameof(key));
+
+            IndexKey = key;
             Keys = new List<string>();
             OldKeys = new List<string>();
         }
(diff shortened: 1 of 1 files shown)


grokys:
Ah, I've just noticed the test failures, damn. I think this PR just caused the cache to become unserializable, therefore fixing it but not caching stuff. Closing.

PR #1147, opened 2017-08-16
Use correct encoding for temp diff files.

### LINKED ISSUE #1137: Pull Request shows as modified even when it isn't

grokys:
The encoding logic in `PullRequestDetailView` was completely wrong - it was extracting the right file with UTF8 encoding and then using that for the left file. Moved the encoding detection logic to `PullRequestDetailViewModel`.

Fixes #1137

### DIFF
--- a/src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
@@ -477,18 +477,12 @@ namespace GitHub.ViewModels
         /// </param>
         /// <param name="encoding">The encoding to use.</param>
         /// <returns>The path to a temporary file.</returns>
-        public Task<string> ExtractFile(IPullRequestFileNode file, bool head, Encoding encoding)
+        public Task<string> ExtractFile(IPullRequestFileNode file, bool head)
         {
             var path = Path.Combine(file.DirectoryPath, file.FileName);
+            var encoding = pullRequestsService.GetEncoding(path);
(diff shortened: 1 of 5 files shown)


src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
```diff
@@ -61,9 +61,10 @@ class PullRequestSessionService : IPullRequestSessionService
         public async Task<bool> IsUnmodifiedAndPushed(ILocalRepositoryModel repository, string relativePath, byte[] contents)
         {
             var repo = await GetRepository(repository);
+            var modified = await gitClient.IsModified(repo, relativePath, contents);
+            var pushed = await gitClient.IsHeadPushed(repo);
```
grokys: Moved these checks to separate variables like this for ease of debugging.

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
```diff
@@ -477,20 +477,14 @@ public override void Initialize(ViewWithData data)
         /// </param>
         /// <param name="encoding">The encoding to use.</param>
```
jcansdale: This should be `head` not `encoding`.

grokys: The comment for `head` is already there. It's just `encoding` needs removing.

jcansdale:
We're currently special casing UTF8, but I think this is fine because UTF8 is the only text format with a signature that VS currently supports.

Here is the `Advanced Save Options...`.
image

PR #1150, opened 2017-08-17
Use correct path to get file encoding.

### LINKED ISSUE #1137: Pull Request shows as modified even when it isn't

grokys:
`PullRequestService.GetEncoding` required a full path but we were passing it a relative path. Make that method take a repository and a relative path to try to make the API clearer.

Fixes #1137 (again)

### DIFF
--- a/src/GitHub.App/Services/PullRequestService.cs
+++ b/src/GitHub.App/Services/PullRequestService.cs
@@ -328,12 +328,14 @@ namespace GitHub.Services
             });
         }
 
-        public Encoding GetEncoding(string path)
+        public Encoding GetEncoding(ILocalRepositoryModel repository, string relativePath)
         {
-            if (File.Exists(path))
+            var fullPath = Path.Combine(repository.LocalPath, relativePath);
+
+            if (File.Exists(fullPath))
             {
                 var encoding = Encoding.UTF8;
-                if (HasPreamble(path, encoding))
+                if (HasPreamble(fullPath, encoding))
                 {
                     return encoding;
                 }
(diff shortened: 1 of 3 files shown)


src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -328,12 +328,14 @@ public IObservable<Unit> SwitchToBranch(ILocalRepositoryModel repository, IPullR
             });
         }
 
-        public Encoding GetEncoding(string path)
+        public Encoding GetEncoding(ILocalRepositoryModel repository, string relativePath)
         {
-            if (File.Exists(path))
+            var fullPath = Path.Combine(repository.LocalPath, relativePath);
+
+            if (File.Exists(fullPath))
```
jcansdale: We probably shouldn't silently return `Encoding.Default` if the file doesn't exist. I'm guessing this is what allowed the isssue to slip through in the first place.

Is it ever normal for the file not to exist? Do you think we should throw a file not found exception or something?

src/GitHub.Exports.Reactive/Services/IPullRequestService.cs
```diff
@@ -124,9 +124,10 @@ public interface IPullRequestService
         /// <summary>
         /// Gets the encoding for the specified file.
         /// </summary>
-        /// <param name="path">The path to the file.</param>
+        /// <param name="repository">The repository.</param>
+        /// <param name="relativePath">The relative path to the file in the repository.</param>
         /// <returns>The file's encoding</returns>
```
jcansdale: Should mention behaviour if file doesn't exist (return `Encoding.Default` or throw?).

src/GitHub.App/Services/PullRequestService.cs
grokys: Yeah, if a file has been deleted in the PR then it won't exist on disk.

grokys: Also if a file has been added in a PR, but that PR is not checked out then it won't exist on disk.

PR #1156, opened 2017-08-18
Fix implementation of IsHeadPushed. 

### LINKED ISSUE #1155: "You must commit and push your changes" message shown when there are no commits to push, but instead commits to pull
The following [check](https://github.com/github/VisualStudio/blob/master/src/GitHub.App/Services/GitClient.cs#L429-L433) is incorrect:

```
if (repo.Head.IsTracking)
{
	var trackedBranchTip = repo.Head.TrackedBranch.Tip;
	if (trackedBranchTip != null)
	{
		return repo.Head.Tip.Sha == trackedBranchTip.Sha;
	}
}
```

This is checking that head is at the _same_ commit as the tracked branch, so will return `false` when there are commits to pull, not only when there are commits to push.

grokys:
Previously we were checking if head and the remote tracking branch were on the same commit. Instead check whether head is not ahead by any commits.

Fixes #1155
Depends on #1150

### DIFF
--- a/src/GitHub.App/Services/GitClient.cs
+++ b/src/GitHub.App/Services/GitClient.cs
@@ -424,16 +424,7 @@ namespace GitHub.Services
 
             return Task.Factory.StartNew(() =>
             {
-                if (repo.Head.IsTracking)
-                {
-                    var trackedBranchTip = repo.Head.TrackedBranch.Tip;
-                    if (trackedBranchTip != null)
-                    {
-                        return repo.Head.Tip.Sha == trackedBranchTip.Sha;
-                    }
-                }
-
-                return false;
+                return repo.Head.TrackingDetails.AheadBy == 0;
             });
         }
 
--- a/src/UnitTests/GitHub.App/Services/GitClientTests.cs
+++ b/src/UnitTests/GitHub.App/Services/GitClientTests.cs
@@ -14,45 +14,28 @@ public class GitClientTests
     public class TheIsHeadPushedMethod : TestBaseClass
     {
         [Theory]
-        [InlineData(true, "sha", "sha", true)]
-        [InlineData(true, "xxx", "yyy", false)]
-        [InlineData(true, "headSha", null, false)]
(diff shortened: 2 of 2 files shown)


src/GitHub.App/Services/GitClient.cs
```diff
@@ -424,16 +424,7 @@ public Task<bool> IsHeadPushed(IRepository repo)
...
-                {
-                    var trackedBranchTip = repo.Head.TrackedBranch.Tip;
-                    if (trackedBranchTip != null)
-                    {
-                        return repo.Head.Tip.Sha == trackedBranchTip.Sha;
-                    }
-                }
-
-                return false;
+                return repo.Head.TrackingDetails.AheadBy == 0;
```
jcansdale: I'm guessing that `repo.Head.TrackingDetails` might be `null` if the remote branch has been deleted (e.g when the PR you're on has been closed). I believe that `repo.Head.TrackedBranch` will be `null` in this case.

Should we return true or false if the remote beanch has been deleted?

grokys: If you look at the documentation for [`TrackingDetails.AheadBy`](https://github.com/libgit2/libgit2sharp/blob/master/LibGit2Sharp/BranchTrackingDetails.cs#L30-L31) it says:

> This property will return <c>null</c> if this local branch has no upstream configuration or if the upstream branch does not exist

`TrackingDetails.AheadBy` will be null if there's no remote branch, and thus `== 0` will return `false`. `TrackingDetails` itself [will not be null]( https://github.com/libgit2/libgit2sharp/blob/9a817484df4a4025fa0b004c5718350f63f1a522/LibGit2Sharp/Branch.cs#L98).


So I believe this code is correct.

jcansdale: Ah, I didn't realize `AheadBy` was an `int?`!

Maybe do an explicit `if (repo.Head.TrackingDetails.AheadBy == null) return false;` for clarity?

PR #1172, opened 2017-08-21
Prevent crash when exception is thrown from InlineCommentTagger.SessionChanged

### LINKED ISSUE #1158: VS exits with `Couldn't find merge base` exception
Hello! Please read the contributing guidelines before submitting an issue regarding the GitHub Extension for Visual Studio.

- GitHub Extension for Visual Studio version: 2.3.0.18
- Visual Studio version: 2015

- Ended up on PR branch that had been deleted.
- VS crashes with the exception below:

image

jcansdale:
This PR changes `SessionChanged` from being an `async void` to `async Task`.
If an exception is thrown it will now be logged.

This fixes the crash from #1158 and will help us find what sometimes causes the merge base not to be found.

If the merge base can't be found again, it should appear like this:
image

Fixes #1158.

### DIFF
--- a/src/GitHub.InlineReviews/Tags/InlineCommentTagger.cs
+++ b/src/GitHub.InlineReviews/Tags/InlineCommentTagger.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
@@ -15,8 +14,7 @@ using Microsoft.VisualStudio.Text.Projection;
 using Microsoft.VisualStudio.Text.Tagging;
 using ReactiveUI;
 using System.Collections;
-using GitHub.Helpers;
-
+using GitHub.VisualStudio;
 namespace GitHub.InlineReviews.Tags
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


grokys:
Perhaps a more explanatory error then "Boom!" might be useful? ;) Especially as I can't see the "Boom!" string in any of the changed code...

jcansdale:
@grokys oh, it doesn't just say "Boom!". That was just my forced exception for testing. 😉

src/GitHub.InlineReviews/Tags/InlineCommentTagger.cs
```diff
@@ -187,16 +185,21 @@ void Initialize()
...
             else
             {
-                SessionChanged(session);
+                ForgetWithLogging(SessionChanged(session));
             }
 
             initialized = true;
         }
 
+        static void ForgetWithLogging(Task task)
```
grokys: Any reason we can't make `Forget()` do this by default?

jcansdale: I looked into this but the project with the `VsOutputLogger` didn't appear to be referenced.  That would however be my preference (I don't like ignoring exceptions). Suggestions welcome?

grokys: Ah yeah, that problem keeps coming up. We need to make logging the lowest level of our architecture so we can log from anywhere. We should handle this in #573 when we get time to finish that PR off.

For the moment, this seems like a workable hack. Should maybe add a note to #573 explaining that we have this hack here?

PR #1178, opened 2017-08-21
Fix diff line matching when there's a partial match

### LINKED ISSUE #1149: Comments added to a line in diff are not displaying in VS
I added some new comments in a diff via inline comments, it was working fine until one certain line did not display the latest comment, even though it showed up on dotcom fine.

When you try adding a comment to line 5, it just disappears in VS.

Public repo:
https://github.com/tierninho-tester/TestRepo/pull/1

Build: 2.3.0.18

screen shot 2017-08-17 at 9 44 16 am

jcansdale:
Needed to implement backtracking when there's a partial match of the diff (e.g. 1 or more lines matched but the rest didn't). Before it would fail rather backtracking and trying again for the next line. This resulted in some lines that couldn't show an inline-comment (when there was repeated lines in the text).

Fixes #1149

### DIFF
--- a/test/GitHub.InlineReviews.UnitTests/Models/DiffUtilitiesTests.cs
+++ b/test/GitHub.InlineReviews.UnitTests/Models/DiffUtilitiesTests.cs
@@ -193,9 +193,16 @@ namespace GitHub.InlineReviews.UnitTests.Models
             [InlineData("", " x", -1)]
             [InlineData(" x", "", -1)]
 
-            [InlineData(" 1. 2.", " 2. 1.", 1)] // matched full context
-            [InlineData(" 1. 2.", " 2. 3.", -1)] // didn't match full context
-            [InlineData(" 2.", " 2. 1.", 0)] // match if we run out of context lines
+            [InlineData(" 1. 2.", " 1. 2.", 1)] // matched full context
+            [InlineData(" 1. 2.", " 3. 2.", -1)] // didn't match full context
+            [InlineData(" 2.", " 1. 2.", 0)] // match if we run out of context lines
+
+            // Tests for https://github.com/github/VisualStudio/issues/1149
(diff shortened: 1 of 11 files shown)


test/GitHub.InlineReviews.UnitTests/Models/DiffUtilitiesTests.cs
```diff
@@ -193,18 +193,26 @@ public class TheMatchMethod
...
             public void MatchLine(string lines1, string lines2, int skip /* -1 for no match */)
             {
                 var header = "@@ -1 +1 @@";
                 lines1 = lines1.Replace(".", "\r\n");
                 lines2 = lines2.Replace(".", "\r\n");
                 var chunks1 = DiffUtilities.ParseFragment(header + "\n" + lines1).ToList();
                 var chunks2 = DiffUtilities.ParseFragment(header + "\n" + lines2).ToList();
-                var expectLine = (skip != -1) ? chunks1.First().Lines.Skip(skip).First() : null;
-                var targetLines = chunks2.First().Lines;
+                var expectLine = (skip != -1) ? chunks1.First().Lines[skip] : null;
```
grokys: Nit: this being called `skip` now doesn't make much sense (didn't really make much sense before...). `expectedLineNumber` might be better?

jcansdale: You're right. Those tests aren't very clear. I'll improve them a bit..

jcansdale: @grokys I hope https://github.com/github/VisualStudio/pull/1178/commits/2b73ec5ae804536b66b8c6a7b2a9dd2aff417000 clarifies the tests a bit?

I don't want to leave a maintenance nightmare for the next person. 😉

PR #1193, opened 2017-08-24
Updating docs for inline comments

grokys:


meaghanlewis:
@grokys I added 2 screenshots. Let me know if they work.

Otherwise, not sure what else to add about inline comments, so LGTM

meaghanlewis:
@grokys think this is ready to be merged?

PR #1198, opened 2017-08-25
Post comments to correct repository when on a fork.

### LINKED ISSUE #1195: Can't comment on PR from parent repo from fork
- GitHub Extension for Visual Studio version: 2.3.0.24 (published)
- Visual Studio version: 2015/2017

1. Opened a solution in a forked repo `jcansdale/PullRequestSandbox`.
2. Selected a PR in list of PRs from parent repo
3. Opened a PR file (diff view appeared)
4. Tried to add inline comment to file
5. When I clicked `Comment` button `Not Found` appeared

image

It looks like it's trying to publish the comment to a PR on the forked repo.

image

grokys:
If the user had a forked repo checked out and was viewing the parent repo's PRs, posting a comment was erroneously trying to post the comment to the _fork_ repo, resulting in a "Not Found" exception.

Fixes #1195.

### DIFF
--- a/src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
+++ b/src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
@@ -7,6 +7,7 @@ using System.Threading.Tasks;
 using GitHub.Extensions;
 using GitHub.Models;
 using GitHub.UI;
+using NLog;
 using Octokit;
 using ReactiveUI;
 
@@ -17,6 +18,7 @@ namespace GitHub.InlineReviews.ViewModels
     /// </summary>
     public class CommentViewModel : ReactiveObject, ICommentViewModel
     {
+        static readonly Logger log = LogManager.GetCurrentClassLogger();
         string body;
         string errorMessage;
         bool isReadOnly;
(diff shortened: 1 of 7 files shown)


jcansdale:
Could we log the exception as well as writing `Not Found` to the GUI?

image

grokys:
Good point! Will do.

src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
```diff
@@ -167,6 +169,7 @@ void DoCancelEdit(object unused)
                 }
 
                 ErrorMessage = message;
+                log.Error("Error posting inline comment.", e);
```
jcansdale: Just an observation. In most places we don't seem ends the message with a `.`.

It is a bit all over the place though. I've even seen one like:
`log.Error("foo {0}. ", ex);`

I guess we should do an audit at some point. 😉

grokys: Actually i started out with no "." but then looked in the log and the messages immediately before it had a closing "." so I added it!

jcansdale: lol. LGTM!

meaghanlewis:
Thanks for the fix @grokys 🎉 and for adding unit tests and logging on this! ✅

Issue #1190, opened 2017-08-23
'Couldn't find merge base' displayed when opening pull request


meaghanlewis:
__Version__
- GitHub Extension for Visual Studio  version: 2.3.0.22
- Visual Studio version: 2017

__What happened__ 
I encountered this 'Couldn't find merge base..' exception after cloning the repo and then trying to open this PR: AvaloniaUI/Avalonia#1127
![couldnt find merge base](https://user-images.githubusercontent.com/10103121/29629911-510b1fce-87ef-11e7-8977-6e9c81d3f5fa.png)

This issue was consistently reproducible, but now I can't reproduce it anymore. Will add more details to this issue if I encounter this problem again.

jcansdale:
Hi @meaghanlewis,

Once #1199 has been merged, you should see a message like, "The Pull Request failed to load. Please check your network connection and click refresh to try again" is this happens again. It will also write more detail to the logs.

If you do see this, could you send me the logs from:
`\Users\<user>\AppData\Local\GitHubVisualStudio\extension.log` (or 
`\Users\<user>\AppData\Local\GìtHūbVisualStudio\extension.log` for local builds).

Hopefully that will show me where the issue is!

meaghanlewis:
Thanks for the note @jcansdale! I'll keep an eye out for when this is ready to be tested.

jcansdale:
#1191 and this issue shouldn't have been closed! A bug in github.com perhaps?  Reopening.

btw, #1199  is now in master so should make it into the next release.

meaghanlewis:
So I noticed this again while trying to view this PR: https://github.com/meaghanlewis/TestWebAppProject/pull/7

Here's what I see from extension.log
`2017-08-28 14:33:19.6917|ERROR|thread: 1|PullRequestDetailViewModel|Error loading PullRequestModel
LibGit2Sharp.NotFoundException: The Pull Request failed to load. Please check your network connection and click refresh to try again. If this issue persists, please let us know at support@github.com ---> LibGit2Sharp.NotFoundException: Couldn't find 8ea5d6a4c0b8202398614ef13a50167aadb00914 after fetching from https://github.com/meaghanlewis/TestWebAppProject.git:test-branch-2.
   at GitHub.Services.GitClient.<GetPullRequestMergeBase>d__24.MoveNext() in c:\jenkins\workspace\GHfVS\src\GitHub.App\Services\GitClient.cs:line 408
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #1199, opened 2017-08-25
Write to log when error is shown in PR detail view

jcansdale:
These commits were originally intended to help track down the 'Couldn't find merge base' issue #1190. This issue hasn't been resolved yet, but I think having the extra logging and exception detail will be useful stand alone.

When the following error is displayed, the exception detail will also be written to the log.
couldnt find merge base

### DIFF
--- a/src/GitHub.App/Services/GitClient.cs
+++ b/src/GitHub.App/Services/GitClient.cs
@@ -98,7 +98,7 @@ namespace GitHub.Services
                     var remote = repository.Network.Remotes[remoteName];
                     repository.Network.Fetch(remote, refspecs, fetchOptions);
                 }
-                catch(Exception ex)
+                catch (Exception ex)
                 {
                     log.Error("Failed to fetch", ex);
 #if DEBUG
@@ -394,7 +394,7 @@ namespace GitHub.Services
                 baseCommit = repo.Lookup<Commit>(baseSha);
                 if (baseCommit == null)
                 {
-                    return null;
+                    throw new NotFoundException($"Couldn't find {baseSha} after fetching from {baseCloneUrl}:{baseRef}.");
                 }
             }
 
@@ -405,14 +405,14 @@ namespace GitHub.Services
                 headCommit = repo.Lookup<Commit>(headSha);
                 if (headCommit == null)
                 {
-                    return null;
+                    throw new NotFoundException($"Couldn't find {headSha} after fetching from {headCloneUrl}:{headRef}.");
                 }
             }
 
             var mergeBaseCommit = repo.ObjectDatabase.FindMergeBase(baseCommit, headCommit);
-            if(mergeBaseCommit == null)
+            if (mergeBaseCommit == null)
             {
-                return null;
+                throw new NotFoundException($"Couldn't find merge base between {baseCommit} and {headCommit}.");
             }
 
             return mergeBaseCommit.Sha;
--- a/src/GitHub.App/Services/PullRequestService.cs
+++ b/src/GitHub.App/Services/PullRequestService.cs
@@ -321,7 +321,7 @@ namespace GitHub.Services
                     }
                     catch (NotFoundException ex)
                     {
-                        throw new NotFoundException($"The Pull Request file failed to load. Please check your network connection and click refresh to try again.", ex);
+                        throw new NotFoundException($"The Pull Request file failed to load. Please check your network connection and click refresh to try again. If this issue persists, please let us know at support@github.com", ex);
                     }
                 }
 
--- a/src/GitHub.App/Services/PullRequestService.cs
+++ b/src/GitHub.App/Services/PullRequestService.cs
@@ -316,11 +316,6 @@ namespace GitHub.Services
                         pullRequest.Head.Sha,
                         pullRequest.Base.Ref,
                         pullRequest.Head.Ref);
-
-                    if (sha == null)
-                    {
-                        throw new NotFoundException($"Couldn't find merge base between {pullRequest.Base.Sha} and {pullRequest.Head.Sha}.");
-                    }
                 }
 
                 var file = await ExtractToTempFile(repo, pullRequest.Number, sha, fileName, encoding);
@@ -350,7 +345,7 @@ namespace GitHub.Services
             {
                 foreach (var b in encoding.GetPreamble())
                 {
-                    if(b != stream.ReadByte())
+                    if (b != stream.ReadByte())
                     {
                         return false;
                     }
@@ -473,7 +468,7 @@ namespace GitHub.Services
         {
             var prConfigKey = $"branch.{branchName}.{SettingGHfVSPullRequest}";
             var value = ParseGHfVSConfigKeyValue(await gitClient.GetConfig<string>(repo, prConfigKey));
-            return value != null && 
+            return value != null &&
                 value.Item1 == pullRequest.Base.RepositoryCloneUrl.Owner &&
                 value.Item2 == pullRequest.Number;
         }
--- a/src/GitHub.App/Services/PullRequestService.cs
+++ b/src/GitHub.App/Services/PullRequestService.cs
@@ -308,14 +308,21 @@ namespace GitHub.Services
                 }
                 else
                 {
-                    sha = await gitClient.GetPullRequestMergeBase(
-                        repo,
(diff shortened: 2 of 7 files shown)


src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -316,11 +316,6 @@ public IObservable<Unit> SwitchToBranch(ILocalRepositoryModel repository, IPullR
                         pullRequest.Head.Sha,
                         pullRequest.Base.Ref,
                         pullRequest.Head.Ref);
-
```
jcansdale: This exception is now thrown from `GetPullRequestMergeBase` (with more info about what went wrong).

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -350,7 +345,7 @@ static bool HasPreamble(string file, Encoding encoding)
             {
                 foreach (var b in encoding.GetPreamble())
                 {
-                    if(b != stream.ReadByte())
+                    if (b != stream.ReadByte())
```
jcansdale: Look @shana, a space! ;-)

src/GitHub.App/ViewModels/PullRequestDetailViewModel.cs
```diff
@@ -351,6 +354,7 @@ public override void Initialize(ViewWithData data)
                 .ObserveOn(RxApp.MainThreadScheduler)
                 .Catch<IPullRequestModel, Exception>(ex =>
                 {
+                    log.Error("Error observing GetPullRequest", ex);
```
jcansdale: Add logs when exception is displayed on UI.

src/GitHub.App/Services/PullRequestService.cs
jcansdale: It appears we still have an issue when commenting in inline mode. This wasn't the line I commented on! 😭

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -316,11 +316,6 @@ public IObservable<Unit> SwitchToBranch(ILocalRepositoryModel repository, IPullR
                         pullRequest.Head.Sha,
                         pullRequest.Base.Ref,
                         pullRequest.Head.Ref);
-
-                    if (sha == null)
```
jcansdale: This is the line with `sha == null`. How does it look?

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -316,11 +316,6 @@ public IObservable<Unit> SwitchToBranch(ILocalRepositoryModel repository, IPullR
...
-                    {
-                        throw new NotFoundException($"Couldn't find merge base between {pullRequest.Base.Sha} and {pullRequest.Head.Sha}.");
```
jcansdale: This exception is now thrown from GetPullRequestMergeBase (with more info about what went wrong).

src/GitHub.App/Services/PullRequestService.cs
shana: rejoice

src/UnitTests/GitHub.App/Services/GitClientTests.cs
```diff
@@ -194,7 +194,11 @@ public class TheGetPullRequestMergeBaseMethod : TestBaseClass
             repo.Network.Remotes.Add(null, null).ReturnsForAnyArgs(remote);
             var gitClient = new GitClient(Substitute.For<IGitHubCredentialProvider>());
 
-            await gitClient.GetPullRequestMergeBase(repo, baseUri, headUri, baseSha, headSha, baseRef, headRef);
+            try
+            {
+                await gitClient.GetPullRequestMergeBase(repo, baseUri, headUri, baseSha, headSha, baseRef, headRef);
+            }
+            catch (NotFoundException) { }
```
shana: Ignoring an exception in a test is weird. Is it supposed to throw or not?

src/GitHub.App/Services/GitClient.cs
```diff
@@ -394,7 +394,7 @@ public Task<bool> IsModified(IRepository repository, string path, byte[] content
                 baseCommit = repo.Lookup<Commit>(baseSha);
                 if (baseCommit == null)
                 {
-                    return null;
+                    throw new NotFoundException($"Couldn't find {baseSha} after fetching from {baseCloneUrl}:{baseRef}.");
```
shana: So I understand why you'd want the details of the shas in the log, but this error message is incredibly cryptic, there is no way anyone is going to know what is going on with an error like this, and this is going to be displayed directly to a user where it's going to be completely useless.

We should have a message for the user, and another with more details for the log. If the log is only showing the message and stacktrace, we can set InnerException with the log details and have the outer exception be for the user.

shana: Also, are all the callers of this method expecting an exception to be thrown and handling it properly?

src/UnitTests/GitHub.App/Services/GitClientTests.cs
jcansdale: It's certainly a bit unusual. The method it's calling would previously return a null when a comment couldn't be fetched/found. It now throws an exception.

The test is to check whether or not `Fetch` was called. It doesn't care about the exception or exception message. There is another test for that.

src/GitHub.App/Services/PullRequestService.cs
jcansdale: What happens if an exception is thrown from a `Observable.Defer(async () =>`?

src/GitHub.App/Services/GitClient.cs
jcansdale: Callers of this method previously turned the null into an exception themselves. I've also tested it by making this method always throw to check that nothing bad happens. The exception message is displayed as expected.

jcansdale: I wonder what the user facing message should say? I suspect that for some reason, fetch doesn't always work or there is some race condition. I guess we could suggest thry try again or maybe refresh. I'll investagate...

src/UnitTests/GitHub.App/Services/GitClientTests.cs
shana: Ah ok, in that case we should probably document that with a comment, so we know in a year from now what the assumption here is.

src/GitHub.App/Services/GitClient.cs
shana: Yeah, an indication of what the user action should be in the case of this error would be great!

jcansdale: Something like this.

image

Or maybe simply, "Please check your network connection and try again."?

jcansdale: "Couldn't find merge base. Please check your network connection and try again."

I think this is maybe the right balance of revealing what actually went wrong, but giving the user a hint that has a good chance of fixing it.

jcansdale: Here we go:
image

shana: So "merge base" may be meaningless to a user, and try again might not be helpful either if the action that the user performed to get here isn't available from this screen.

What is the action that failed here? Opening a file? Showing the PR details? The first part of the message should say what was the action that failed.

What is try again? Refresh? Click on the PR button in the toolbar and then clicking on the PR again? The second part should say what the user can do on this screen to attempt to make it work.

Assuming it was the PR that failed to load, then we should probably say something like

`The Pull Request failed to load. Please check your network connection and click refresh to try again.`

Adjust accordingly for the failed action.

jcansdale: Okay, I've chained it to variants of your suggestion above.

My concern was that we hide the real issue behind a workaround. I'm still keen to get to the bottom of what's going wrong. My hunch is that it's a race condition in gitlib2sharp. I'll see if I can reproduce the issue with some standalone code... 😄

shana: Yes, we don't want to hide the issue in a workaround. If you're concerned about people not reporting it, you can add "If this issue persists, please let us know at support@github.com"

jcansdale: Done. Also confirmed that nested exceptions appear in logs.

PR #1200, opened 2017-08-25
Remove duplicate credential classes.

### LINKED ISSUE #1176: Duplicate ILoginCaches.
Pull request #845 introduced a new [`ILoginCache`](https://github.com/github/VisualStudio/blob/master/src/GitHub.Api/ILoginCache.cs) that uses `Task`s instead of `IObservables`. However the PR didn't remove the existing [`ILoginCache`](https://github.com/github/VisualStudio/blob/master/src/GitHub.Exports.Reactive/Caches/ILoginCache.cs).

The old `ILoginCache` should be removed.

grokys:
Previously credentials were read by 4 separate interfaces/classes:

- `GitHub.Api.ILoginCache`/`IWindowsLoginCache`
- `GitHub.Api.ILoginCache`/`GitHub.Caches.LoginCache`
- `SimpleCredentialStore`
- `CredentialCache`

This PR consolidates all of these separate classes into `IKeychain`/`WindowsKeychain`.

Fixes #1176


src/GitHub.App/Services/GitHubCredentialProvider.cs
```diff
@@ -34,21 +34,21 @@ public Credentials HandleCredentials(string url, string username, SupportedCrede
...
-        }
-
-        static Credentials CreateCredentials(Tuple<string, SecureString> data)
-        {
-            Guard.ArgumentNotNull(data, nameof(data));
 
-            return new SecureUsernamePasswordCredentials
+            try
+            {
+                var credentials = ThreadHelper.JoinableTaskFactory.Run(async () => await keychain.Load(host));
```
grokys: Previously we were doing a `.Wait` on an observable here, which is Bad. I've changed it to use the pattern described in 
[the 3 Threading Rules of VS](https://github.com/Microsoft/vs-threading/blob/master/doc/threading_rules.md#rule-2-when-an-implementation-of-an-already-shipped-public-api-must-call-asynchronous-code-and-block-for-its-completion-it-must-do-so-by-following-this-simple-pattern).

shana: Oh yeah, nice catch, we should beat whoever wrote that with a large 🐟

shana: *whistles*

PR #1212, opened 2017-08-31
Find merge base when external PR branch has been deleted

### LINKED ISSUE #1190: 'Couldn't find merge base' displayed when opening pull request

jcansdale:
There was an issue when a PR has been submitted from a fork, the PR had been accepted and the owner deleted the PR branch. GitHub for Visual Studio was unable to find the original (now deleted) branch to show the PR details. This should allow those old PRs to still be displayed.

There are likely other edge cases that I need to think about (hence WIP for the moment).

Fixes #1190

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,20 @@
 # GitHub Extension for Visual Studio
 
-**The location of the submodules has changed as of 31-01-2017.** If you have an existing clone, make sure to run `git submodule sync` to update your local clone with the new locations for the submodules.
+## Notices
+
+### VS 2017 v15.3 users, please read
+
+If you need to downgrade or uninstall the extension, **do not use** ***Revert*** in Visual Studio 2017 15.3. Instead, manually uninstall the extension with the steps listed in https://github.com/github/VisualStudio/issues/1206#issuecomment-326558902
+
+If you have a corrupted extension cache, steps for fixing it are in https://github.com/github/VisualStudio/issues/1206#issuecomment-326053090
+
+The Visual Studio 2017 15.3 installer [has a bug](https://github.com/github/VisualStudio/issues/1206) that causes a corruption of the installed extensions data when you revert an installation of the extension (see also [this MS issue](https://developercommunity.visualstudio.com/content/problem/102178/error-installing-github-extension.html)). Until VS 2017 15.4 comes out, **do not use Revert in** ***Extensions and Updates***.
+
+### The location of the submodules has changed as of 31-01-2017
+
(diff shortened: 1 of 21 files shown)


meaghanlewis:
I tested this with some PR's that I previously wasn't able to view, and they appear to load now with your fix. 👍

jcansdale:
@meaghanlewis,

> I tested this with some PR's that I previously wasn't able to view, and they appear to load now with your fix. 

That is good news. I think this is by far the most common scenario. Someone submits a PR from and fork and deletes the branch once the PR has been accepted.

I'm trying to think what other possible edge cases there might be.  In the above case I was able to take advantage of the special `refs/pull/<PR>/head` branch name. It seems these remain available even after the PR has been closed and its branch deleted.

I'm wondering what would happen if the PR base branch has been deleted (there isn't an equivalent `refs/pull/<PR>/base`). Obviously this won't happen very often with the default `master` branch.

Could you try the following:

1. Create a PR from a fork onto a branch other than master.
2. Accept and close the PR.
3. Delete the branch the PR was merged onto.

- I'm interested to know if you can still view the PR details.
- There might be a difference between using an account that owns the repo and one that can only view it.
- Be sure to create a fresh clone of the repo (if the commits already exist, it won't need to fetch from the deleted branch)

jcansdale:
> I'm wondering what would happen if the PR base branch has been deleted (there isn't an equivalent refs/pull/<PR>/base).

I've done some further testing with this. If the PR base branch has been deleted, we can's simply do a fetch using the base branch name (or SHA). This seems to be the case whether or not the user has permissions to undelete the branch.

As a workaround for this, I'm fetching the head branch using `refs/pull/<PR>/head`, before checking if the required base commits are available. This means that if the base is directly upstream of the head, there is no need to fetch the head branch (which will very often be the case).

The problematic case would be if commits have been added to the base branch of the PR, before the PR is closed and the base branch deleted (not just the head branch, which would be the usual case).

jcansdale:
> The problematic case would be if commits have been added to the base branch of the PR, before the PR is closed and the base branch deleted (not just the head branch, which would be the usual case).

I've so far failed to produce this problematic case. Maybe it isn't an issue after all? 😕

PR #1219, opened 2017-09-05
Refactoring Pull Request Sessions

grokys:
Previously, our treatment of pull request inline comments was a little confused, in that we were not properly distinguishing between files as they are in the working directory (i.e. their content can change) and files as they are in a pull request (i.e. not changing). In addition we were using the `InlineCommentTagger` to update the file with the current contents of the editor, which was messy and meant that there was shared state between different views of a file.

The problem here can be seen (on `master`) if you:

- Check out a PR branch
- Make a change to a file, but don't commit it
- "Compare Changes" on a file
- Try to add a new comment
- See that it says that you must commit your changes, even when the file state in the diff view represents an already committed state

This PR fixes that by:

- Making `IPullRequestSession.GetFile` always represent a file at a fixed commit. Files returned in this manner aren't tied to editor contents
- Adding `IPullRequestSessionManager.GetLiveFile` which represents the state of a file in a Visual Studio text buffer compared to the currently checked out PR


meaghanlewis:
This looks good to me now following the repro steps you provided above.
view diff from pr


One thing I notice on the same branch, is when I make a change on a particular file, go to Team Explorer - Changes, and select 'Compare with Unmodified...' then I see the same message "You must commit and push..."
view diff from changes
And then a question I have is should I even be able to make a comment here at all if this isn't a pull request?

Branch used for testing: Testing this here: https://github.com/meaghanlewis/TestWebAppProject/tree/test-pr-creation

PR #1220, opened 2017-09-07
Enable assembly resolver to ensure that `Start Page` and other icons can be found

### LINKED ISSUE #1217: bindingRedirects can become out of sync with installed extension
- GitHub Extension for Visual Studio version:  v2.3.1.30 (iirc)
- Visual Studio version: 2017 15.3

There have been a few instances where it looks like the `devenv.exe.config` file has become out of whack with the installed version of the extension. This has been a problem because we use `ProvideCodeBase` / `ProvideBindingRedirection` attributes to ensure that our assemblies can be resolved (without resorting to assembly resolve event hacks).

This was particularly noticeable when I changed to an earlier version of the extension. There was a `bindingRedirect` in the `devenv.exe.config`, that was looking for an assembly that no longer existed. The extension failed to work completely.

There have also been reports of the GitHub icon missing from the Start Page. This used to happen in the past when the assembly containing the icon couldn't be resolved by its fully qualified name. This might happen if the `devenv.exe.config` hasn't been updated and the assembly version on disk doesn't match the assembly range in the `bindingRedirect`.

Here's an example of one of the GitHub extension's entries in `devenv.exe.config`.

```
<dependentAssembly>
    <assemblyIdentity name="GitHub.UI" publicKeyToken="bc1bd09f2901c82e" culture="neutral"/>
    <bindingRedirect oldVersion="0.0.0.0-2.3.2.33" newVersion="2.3.2.33"/>
    <codeBase version="2.3.2.33" href="C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\Extensions\GitHub\GitHub\GitHub.UI.dll"/>
</dependentAssembly>
```

If this wan't updated and the version on disk changed to say, `2.3.3.44`, Visual Studio wouldn't know where to find the `GitHub.UI.dll` assembly.

@tinaschrepfer @Michael-Eng,
Is this a known issue or a likely symptom of what has been reported in #1213?

jcansdale:
This PR fixes a few potential issues:

1. Ensures that the icons in `GitHub.VisualStudio.imagemanifest` icons can be found
2. Allows assemblies to be resolved when running an unsigned build of the extension
3. Keeps the extension working even if `devenv.exe.config` failed to update

The most likely scenario is when someone is using using Visual Studio 2017 (15.3) and has never used the extension before. In this case the Start Page icon next to GitHub will be blank. This is unfortunate because it's when we most want people to notice it!

The reason it doesn't appears is because `GitHub.VisualStudio.imagemanifest` references the `GitHub.VisualStudio` by its simple name. In order for the icons to be found, the `GitHub.VisualStudio` assembly must be loaded before any of them are used. The assembly resolver is in this assembly and will be auto-loaded on `UICONTEXT.NoSolution` (which should happen before the icons are needed.

Rather than having to maintain a separate list of assemblies that might need resolving, the assembly resolver loads the `ProvideCodeBase` and `ProvideBindingRedirection` attributes (from `GitHub.VisualStudio`) and resolves according to these rules. The idea is to behave as if the `dependentAssembly` elements had been installed.

The previous implementation of the assembly resolver would resolve any assembly in the extension folder when there was a full name match. This avoided false matches, but also meant it needed to call `AssemblyName.GetAssemblyName(file)` to see if there was a match.

This implementation will only hit the disk if the assembly being resolved has a matching `ProvideCodeBase` or `ProvideBindingRedirection` attribute. It will only resolve an assembly when there is a full name patch.

Fixes #1217


src/GitHub.VisualStudio/GitHub.VisualStudio.imagemanifest
```diff
@@ -1,7 +1,10 @@
 ﻿<ImageManifest xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns="http://schemas.microsoft.com/VisualStudio/ImageManifestSchema/2014">
 
   <Symbols>
+    <!-- We need to ensure that `GitHub.VisualStudio` has loaded before any of these resources are used. -->
+    <!-- This should happen when `AssemblyResolverPackage` auto-loads on `UICONTEXT.NoSolution`. -->
     <String Name="Resources" Value="/GitHub.VisualStudio;component/Resources/icons" />
```
jcansdale: This seems the culprit was actually here. It tries to find an assembly by its simple name before that assembly has been loaded. I think when the `Team Explorer` related assemblies get initialized might have changed in VS 15.3. This can cause `GitHub.VisualStudio` not to get pre-loaded and the icons fail to load.

There is a possible workaround for this:
https://github.com/Microsoft/PTVS/commit/ec1b895f9410c4fa51d05dc62b47ae0ab854d3e5

I did a spike to see if this might work for us, but it would involve moving a bunch of resources into `GitHub.UI` and fixing a bunch of other issues (not a point fix). This might be worth investigating at some future time.

jcansdale:
@meaghanlewis, I thought a simple test plan might be useful.

1. Start from a fresh Visual Studio 2017 (15.3) install.
2. Install an older version of the extension (`2.3.2.32` or earlier).
3. Open Visual Studio 2017 and confirm that the two circled icons are blank.
image
(the menu path is `View > Other Windows > GitHub`)
4. Install the build from this PR.
5. Open Visual Studio 2017 and confirm that the two circled icons are now visible.

The `Start Page doesn't exist in VS 2015, but it would be worth doing the same but checking the `View > Other Windows > GitHub` icon.

It would also be good to get a VM with Visual Studio 2017 (15.2) installed on it. I'm not sure how to install older versions, but it must be possible somehow? They keep updating Visual Studio 2017 and things sometimes break.

It's possible this was also an issue on Visual Studio 2017 (15.2), but no one noticed. It requires a machine where the user hasn't been using the GitHub extension, which makes it very easy to miss! I'm very interested to see what you find...

src/GitHub.VisualStudio/AssemblyResolverPackage.cs
```diff
@@ -1,90 +1,124 @@
...
 
         protected override void Dispose(bool disposing)
         {
-            AppDomain.CurrentDomain.AssemblyResolve -= LoadAssemblyFromExtensionDir;
+            AppDomain.CurrentDomain.AssemblyResolve -= ResolveAssemblyFromPackageFolder;
+
+            if (resolvingAssemblies.Count > 0 || resolvingExceptions.Count > 0)
+            {
+                // Avoid executing any logging code unless there is something to log.
+                WriteToLog();
```
grokys: Why do we log the resolved assemblies and resolve errors when VS is shutdown rather than when they occur?

src/UnitTests/GitHub.VisualStudio/AssemblyResolverPackageTests.cs
```diff
@@ -0,0 +1,157 @@
...
+using Microsoft.VisualStudio.Shell;
+using System.Reflection;
+using System.IO;
+using Xunit;
+
+namespace UnitTests.GitHub.VisualStudio
+{
+    public class AssemblyResolverPackageTests
+    {
+        class TheResolveDependentAssemblyMethod
```
grokys: These tests aren't being run under ncrunch because this class needs to be `public`.

src/GitHub.VisualStudio/AssemblyResolverPackage.cs
jcansdale: For a few reasons:

1. Since this package is always loaded, I wanted to avoid loading any extra assemblies as VS starts.
2. Avoid the possability of recursive assembly resolving (in a previous version there was an issues with `Trace.WriteLine`).
3. Avoid doing any file creation or writes as VS starts.

I've tried to make it as predictable and light weight as possible. :-)

meaghanlewis:
Thanks for the testing notes 👍 

I verified this using VS2017 with versions 15.3 and VS 15.0. On a fresh install of an older version of the extension (v2.3.0.24) the GitHub icon was not present, but on updating to the most recent version of the extension (v2.3.2.32) the GitHub icon was present.

The icon is present in both 15.3 and 15.0 with a fresh install of the extension from this PR build.

I'll check the `View > Other Windows > GitHub` option shows the icon in VS2015, but this looks good to me so far 🎉

jcansdale:
@meaghanlewis,

> On a fresh install of an older version of the extension (v2.3.0.24) the GitHub icon was not present, but on updating to the most recent version of the extension (v2.3.2.32) the GitHub icon was present.

Was this with VS2017 15.0? I'm hoping the icon wasn't present on 15.3 until this PR build. 🤞

meaghanlewis:
@jcansdale I started with a fresh install of 15.3 and then installed the extension version `v2.3.0.24` and saw no icon, then updated the extension to version `v2.3.2.32` and the icon was visible. (Not exactly what you were hoping for...)

Next, I did an uninstall/reinstall of 15.3, and installed extension `v2.3.0.24` again. there was no icon at this point but once I updated to the PR build extension the icon was visible.

jcansdale:
@meaghanlewis,

> I started with a fresh install of 15.3 and then installed the extension version v2.3.0.24 and saw no icon, then updated the extension to version v2.3.2.32 and the icon was visible. (Not exactly what you were hoping for...)

Lol, it's complicated! I wouldn't be surprised if this happened with VS 15.0. Did you start with a fresh install of 15.0 or 15.3? 😉 

> Next, I did an uninstall/reinstall of 15.3, and installed extension v2.3.0.24 again. there was no icon at this point but once I updated to the PR build extension the icon was visible.

This is the main thing! 😄 👍

PR #1224, opened 2017-09-12
Update install script to work with VS 15.3

jcansdale:
The previous install script started hanging after Visual Studio 2017 was updated to 15.3.

This PR updates `VsixUtil.exe` to a version that works with 15.3.
https://github.com/jaredpar/VsixUtil/pull/8

Calling `install.cmd` will now re-install the extension for all installed instances of Visual Studio (including preview versions).


jcansdale:
It seems this only happens if the same version is already installed for all-users. If an earlier or no all-users version is installed, it works okay.

PR #1226, opened 2017-09-13
Only show comment margin if there are changes in current file

jcansdale:
When the experimental, `Show PR comments on editor margin` feature is enabled, only show the comment margin if there are changes in the current file. If the file hasn't been touched, there is no need for the margin to take up space.


src/GitHub.InlineReviews/Tags/InlineCommentTagger.cs
```diff
@@ -50,7 +50,14 @@ public sealed class InlineCommentTagger : ITagger<InlineCommentTag>, IDisposable
             this.sessionManager = sessionManager;
         }
 
-        public bool ShowMargin => file != null;
+        public bool ShowMargin
+        {
+            get
+            {
+                var diff = file?.Diff;
```
grokys: Nit: this could be rewritten more simply: `file?.Diff?.Count > 0`

jcansdale: Thanks. I learn something every day! 👍

PR #1243, opened 2017-09-30
Change submodule to use https

Haacked:
It's weird that this one submodule uses ssh rather than https like the rest of them. I'm assuming that's not intentional, but if it is, feel free to reject this PR.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,4 +12,4 @@
 	url = https://github.com/editor-tools/Akavache
 [submodule "script"]
 	path = script
-	url = git@github.com:github/VisualStudioBuildScripts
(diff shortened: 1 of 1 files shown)


Haacked:
Ok, the build failure tells me this is how we securely build GHVS by using SSH for that one submodule. I'll close this. :smile:

jcansdale:
The SSH submodule seems to hang Visual Studio 2017. I believe there are reasons we use SSH, but it's definitely a pain! Maybe @shana will remind us?

shana:
For CI (appveyor and jenkins) to checkout the submodule securely, it needs to go through ssh. VS doesn't support ssh, so it might be doing something silly there. You can locally change the submodule to https so VS works for you.

PR #1267, opened 2017-10-12
Functionality to easily add new metrics

grokys:
Updates the metrics reporting:
- Store the metrics in a different format, with separate Measures and Dimensions
- Store reports for each day separately
- Change the serialization functionality to serialize all dimensions as strings

Depends on #1130


src/GitHub.VisualStudio/Services/UsageService.cs
```diff
@@ -79,9 +113,9 @@ public IDisposable StartTimer(Func<Task> callback, TimeSpan dueTime, TimeSpan pe
                 var json = SimpleJson.SerializeObject(data);
```
grokys: Note that this is going to serialize the fields with PascalCasing. I guess we need to write a `CamelCaseSerializerStrategy` if we want to serialize it using camelCase.

shana: It doesn't matter much here, the method that post the data serializes the model into camel case.

src/GitHub.Exports/Models/UsageModel.cs
```diff
@@ -1,65 +1,61 @@
...
-            foreach (var property in properties)
+            return new UsageModel
             {
-                var cloneValue = property.PropertyType == typeof(int);
-
-                if (property.Name == nameof(result.NumberOfStartupsWeek))
-                    cloneValue = includeWeekly;
-                else if (property.Name == nameof(result.NumberOfStartupsMonth))
-                    cloneValue = includeMonthly;
+                Guid = guid,
```
shana: I think the GUID needs to be in the dimensions model and not here.

jcansdale:
Was there a plan to somehow exclude metrics from our development usage? Would we turn off metrics for debug builds or include an `IsDebugBuild` field in the metrics?

grokys:
@jcansdale sending metrics is already disabled in debug builds.

StanleyGoldman:
https://github.com/github/central/pull/354

This WIP pull request will absorb the changes on the receiving side.

src/GitHub.VisualStudio/Services/UsageTrackerDispatcher.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.VisualStudio.Shell;
+using System.Threading.Tasks;
+using GitHub.Models;
+using System.Linq.Expressions;
+
+namespace GitHub.Services
+{
+    [ExportForProcess(typeof(IUsageTracker), "devenv")]
+    [PartCreationPolicy(CreationPolicy.NonShared)]
+    public class UsageTrackerDispatcher : IUsageTracker
```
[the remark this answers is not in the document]
StanleyGoldman: It wasn't even included in the project file.

PR #1269, opened 2017-10-13
Fix new inline comments not showing

grokys:
This PR fixes problems introduced in #1230 where new inline comments weren't displayed when a diff was opened via "View Changes in Solution".

Also adds a number of unit tests for `InlineCommentTagger`.


src/GitHub.Exports.Reactive/Models/IPullRequestSessionFile.cs
```diff
@@ -41,5 +41,11 @@ public interface IPullRequestSessionFile : INotifyPropertyChanged
         /// Gets the inline comments threads for the file.
         /// </summary>
         IReadOnlyList<IInlineCommentThreadModel> InlineCommentThreads { get; }
+
+        /// <summary>
+        /// Gets an observable that is raised with a collection of 0-based line numbers when the
```
jcansdale: What do the line numbers represent? Are they where new comments have been added or deleted?

grokys: They're the lines that have have a changed inline comment state, so either a comment was added or removed on that line. Feel free to suggest a better wording!

meaghanlewis:
Tested this build and it LGTM ✅ 

Thanks for adding `InlineCommentTaggerTests.cs`. The tests look very thorough!

src/GitHub.Exports.Reactive/Models/IPullRequestSessionFile.cs
jcansdale: That makes sense. :thumbsup:

jcansdale:
I don't know if this is related, but I'm getting the following exception when viewing a PR file in the editor:

image

In the above example, Line number is 12 and the snapshot content looks like this:

```c#
using System;
using System.Linq;
using System.Text;
using System.Threading.Tasks;

namespace PullRequestSandbox
{
}
```

1. `Show PR comments on editor margin`
2. Open PullRequestSandbox PR 42
3. `Testing PR Commants on LHS of Diff`
4. Right click on `Program.cs` and `Open File in Solution`

This is when it blocks up with the above exception.

jcansdale:
I can'r repro the above issue at all now. 😕  Everything seems to be working fine. 👍

PR #1268, opened 2017-10-12
Fix diff parsing when file contains loose carriage return

jcansdale:
This is a fix for an issue where a diff contains one or more loose carriage returns.

Previous versions show the following when *any* file in the PR contained a loose loose carriage return. You can use the following PR as a repro: https://github.com/grokys/PullRequestSandbox/pull/39.

image

The above exception has been fixed as well as diff line matching. For example, take the following string: `"foo\r\r\nbar\n\nfo\ro\r\nbar\n"`.

When rendered in the GitHub diff view it looks like this:

image

However, when viewed using the GitHub file view it looks like this:

image

You can see that although there are only 5 numbered lines, the file actually takes up 7 lines (with two extra line breaks).

This is also how it appears when rendered on the Visual Studio diff (and normal) view. The only difference is that Visual Studio shows numbers for the rendered lines (as opposed to the line numbers when carriage return are ignored).

image

We need to account for this when the `OldLineNumber`  and `NewLineNumber` is calculated for a `DiffLine` (otherwise the lines that can be commented on won't match the lines displayed in Visual Studio).

This is a potential issue when lines are added to or deleted from a diff.

image

You can see on the two screen grabs above that the lines that can be commented on (the light colored ones) match the lines from the first screen grabs from GitHub's diff view. The line numbers are different, but the start of each line is consistent. This is how it's supposed to work in the fixed version.

NOTE: I wonder what other control chars we might be missing?

Fixes #1241

PR #1270, opened 2017-10-16
Only scan the diff once when processing carriage returns

shana:
The previous solution still scans the string 3 times, one for each `IndexOf` and `Substring` call. Given that we have to go through the string at least once to find the boundaries, we should grab all the information we need along the way immediately. Also given that there is a custom `LineReader` class, we can take advantage of that and return all the data we need in one go.

Also add null checks for invalid data being passed into the constructor. The current caller of `LineReader` probably never calls it with null, but since `LineReader` is a public class, other code in the future might decide to call it (or it could be expanded), and we should make sure the assumptions we make are documented.

### DIFF
--- a/src/GitHub.Exports/Models/DiffUtilities.cs
+++ b/src/GitHub.Exports/Models/DiffUtilities.cs
@@ -3,6 +3,8 @@ using System.IO;
 using System.Collections.Generic;
 using System.Text.RegularExpressions;
 using System.Diagnostics.CodeAnalysis;
+using System.Text;
+using GitHub.Extensions;
 
 namespace GitHub.Models
 {
@@ -13,14 +15,15 @@ namespace GitHub.Models
         public static IEnumerable<DiffChunk> ParseFragment(string diff)
         {
             var reader = new LineReader(diff);
-            string line;
+            LineReader.LineInformation lineInfo;
             DiffChunk chunk = null;
             int diffLine = -1;
             int oldLine = -1;
             int newLine = -1;
 
-            while ((line = reader.ReadLine()) != null)
+            while ((lineInfo = reader.ReadLine()).Line != null)
             {
+                var line = lineInfo.Line;
                 var headerMatch = ChunkHeaderRegex.Match(line);
 
                 if (headerMatch.Success)
@@ -55,8 +58,7 @@ namespace GitHub.Models
                             Content = line,
                         });
 
-                        var lineCount = 1;
-                        lineCount += CountCarriageReturns(line);
+                        var lineCount = lineInfo.CarriageReturns + 1;
 
                         switch (type)
                         {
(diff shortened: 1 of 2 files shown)


jcansdale:
> The previous solution still scans the string 3 times, one for each `IndexOf` and `Substring` call. 

I did consider scanning through it one char at a time, but took a punt that since `IndexOf` and `Substring` are native methods, doing it this way would be faster. Scanning for a specific char in native code can remarkably fast. Does that make sense?

Would you be interested in seeing them profiled in BenchmarkDotNet?

jcansdale:
I've just done a couple of ad-hoc benchmarks.

One for this implementation:

```
            static void Benchmark_ReadLineAndCountCarriageReturns_100000()
            {
                // use current file as test data
                var file = new System.Diagnostics.StackFrame(true).GetFileName();
                var text = File.ReadAllText(file);

                for (int count = 0; count < 100000; count++)
                {
                    var lineReader = new DiffUtilities.LineReader(text);
                    DiffUtilities.LineReader.LineInformation line;
                    while ((line = lineReader.ReadLine()).Line != null)
                    {
                        int crs = line.CarriageReturns;
                    }
                }
            }
```

A similar one for the original implementation:

```
            static void Benchmark_ReadLineAndCountCarriageReturns_100000()
            {
                var file = new System.Diagnostics.StackFrame(true).GetFileName();
                var text = File.ReadAllText(file);

                for (int count = 0; count < 100000; count++)
                {
                    var lineReader = new DiffUtilities.LineReader(text);
                    string line;
                    while ((line = lineReader.ReadLine()) != null)
                    {
                        DiffUtilities.CountCarriageReturns(line);
                    }
                }
            }
```

The first one completes in ~26.91 seconds, the second one in ~6.27 seconds. I think `IndexOf` and `Substring` being native methods is what makes the difference (unless I've got the benchmarking wrong).

jcansdale:
We could also adapt this implementation to use, `text.IndexOfAny(new[] {'\r', '\n'}, index)`.

src/GitHub.Exports/Models/DiffUtilities.cs
```diff
@@ -116,61 +118,72 @@ public static DiffLine Match(IEnumerable<DiffChunk> diff, IList<DiffLine> target
...
 
-                var startIndex = index;
-                index = text.IndexOf('\n', index);
-                var endIndex = index != -1 ? index : text.Length;
-                var length = endIndex - startIndex;
-
-                if (index != -1)
+                var carriageReturns = 0;
+                StringBuilder sb = new StringBuilder();
+                for (; index < length; index++)
```
jcansdale: We could maybe adapt this implementation to use, `text.IndexOfAny(new[] {'\r', '\n'}, index)`? With `new[] {'\r', '\n'}`, stored in a static. 😉

jcansdale:
Here are some alternative implementations we tried:
https://gist.github.com/shana/200e4719d4f571caab9dbf5921fa5276
Scanning with `text.IndexOf('\n', index)` appears to the the best compromise for average .diff files.
It's likely that `text.IndexOfAny(new [] {'\r', '\n'}, index)` would be faster if lines were much longer.

Merged #1268

PR #1275, opened 2017-10-19
Fix possible null ref on PR details window

### LINKED ISSUE #1274: Object reference null in PR details view
- GitHub Extension for Visual Studio version: `v2.3.3.38`

image

```
2017-10-17 19:21:40.3574|ERROR|thread: 1|PullRequestDetailViewModel|Error loading PullRequestModel
System.NullReferenceException: Object reference not set to an instance of an object.
   at GitHub.Services.GitClient.<>c__DisplayClass23_0.<IsModified>b__0() in Y:\jenkins\workspace\GitHub for Visual Studio speakeasy jenkinskeep\src\GitHub.App\Services\GitClient.cs:line 364
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
```

jcansdale:
It seems `repository.Head[path]` can return null, even when `repository.RetrieveStatus(path) == FileStatus.Unaltered` is `true`. I'm not sure why that would happen.

This PR updates the `IsModified` method to make it more robust. I've also added unit tests, because we've had problems with this method before.

Fixes #1274


src/GitHub.App/Services/GitClient.cs
```diff
@@ -405,19 +405,20 @@ public Task<bool> IsModified(IRepository repository, string path, byte[] content
             {
                 if (repository.RetrieveStatus(path) == FileStatus.Unaltered)
                 {
-                    var head = repository.Head[path];
-                    if (head.TargetType != TreeEntryTargetType.Blob)
```
shana: Given that you always want to return false if it's not a blob, and not having a head is equivalent to not being a blob, how about just doing `if (head?.TargetType != TreeEntryTargetType.Blob)` here and no other change?

jcansdale: Updated to do it like this.

PR #1279, opened 2017-10-20
Removed `IConnection.Repositories`.

grokys:
Removed `IConnection.Repositories` and `IConnectionManager.RefreshRepositories` and replaced them with `LocalRepositories` which acts as a central place to find known repositories.

As part of this, also added `IReadOnlyObservableCollection` and `ObservableCollectionEx` because the BCL's `ReadOnlyObservableCollection` sucks. Made `TrackingCollection` implement `IReadOnlyObservableCollection`.


src/GitHub.App/Services/LocalRepositories.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    /// <summary>
+    /// Stores a collection of known local repositories.
+    /// </summary>
+    /// <remarks>
+    /// The list of repositories exposed here is the list of local repositories known to Team
+    /// Explorer.
+    /// </remarks>
+    [Export(typeof(ILocalRepositories))]
+    public class LocalRepositories : ILocalRepositories
```
shana: I don't this this implementation should be here, this will bring all the Rx dependencies in when there's no need for it. The reason this was in ConnectionManager to begin with was meant to be light weight and require minimal dependencies. Perhaps GitHub.Exports or some other place would be better? Maybe even GitHub.VisualStudio?

grokys: Isn't the fact that its export is in `GitHub.Exports` good enough? If you look at some of our models, e.g. `PullRequestModel` - that is defined in `GitHub.App` with its interface in `GitHub.Exports` meaning that it's available to the non-rx parts of the extension. Rx shouldn't be loaded until a _type_ that references rx is loaded, right?

PR #1280, opened 2017-10-23
Refactoring ConnectionManager/RepositoryHost(s)

### LINKED ISSUE #1242: Cannot click on Github Enterprise on Sign in prompt
Hi!

I have a problem similar to https://github.com/github/VisualStudio/issues/12 but reversed.
I cannot click on GitHub Enterprise both on VS2015 and VS2017.

Any ideas?

grokys:
This PR is the first step on the way to removing `RepositoryHosts`/`RepositoryHost`.

Ever since #845 we've been "logging in" cached connections when `ConnectionManager` starts up to make sure we have valid credentials, and then "logging in" again from `RepositoryHost`.

This PR moves all of the logic around connections into `ConnectionManager` and makes `RepositoryHosts` a simple "wrapper" around `ConnectionManager.Connections` that exposes a `GitHubHost` and an `EnterpriseHost`.

As this PR is a first step, there are a few hacks needed. See inline PR comments.

Fixes #1277 
Fixes #1242

Depends on #1279


src/GitHub.App/Models/RepositoryHost.cs
```diff
@@ -1,168 +1,39 @@
...
-        [SuppressMessage("Microsoft.Reliability", "CA2000:Dispose objects before losing scope")]
-        public IObservable<AuthenticationResult> LogInFromCache()
-        {
-            Func<Task<AuthenticationResult>> f = async () =>
-            {
-                try
-                {
-                    var user = await loginManager.LoginFromCache(Address, ApiClient.GitHubClient);
-                    var accountCacheItem = new AccountCacheItem(user);
-                    usage.IncrementLoginCount().Forget();
```
grokys: TODO: Need to work out if a login from cache counts as a login, and if so put this in `ConnectionManager.

src/GitHub.App/Models/RepositoryHost.cs
```diff
@@ -1,168 +1,39 @@
...
-        public IObservable<AuthenticationResult> LogIn(string usernameOrEmail, string password)
-        {
-            Guard.ArgumentNotEmptyString(usernameOrEmail, nameof(usernameOrEmail));
-            Guard.ArgumentNotEmptyString(password, nameof(password));
-
-            return Observable.Defer(async () =>
-            {
-                var user = await loginManager.Login(Address, ApiClient.GitHubClient, usernameOrEmail, password);
-                var accountCacheItem = new AccountCacheItem(user);
-                usage.IncrementLoginCount().Forget();
```
grokys: TODO: Need to put this in `ConnectionManager`.

src/GitHub.App/Models/RepositoryHosts.cs
```diff
@@ -45,78 +40,20 @@ public class RepositoryHosts : ReactiveObject, IRepositoryHosts
...
-                {
-                    var host = LookupHost(x.HostAddress);
-                    if (host.Address != x.HostAddress)
-                        host = await RepositoryHostFactory.Create(x.HostAddress);
-                    return host;
-                })
-                .Select(h => LogOut(h))
-                .Merge().ToList().Select(_ => Unit.Default).Subscribe());
-
+            connectionManager.ConnectionCreated = ConnectionAdded;
```
grokys: This is a temporary hack, that performs the same purpose as `connectionManager.DoLogin`, but renamed (as it no longer does a login) and using `Task` instead of an observable.

It's needed because creating a RepositoryHost is async meaning that for a short time a connection can exist without
a repository host, but other older parts of the code get confused by this. This callback allows RepositoryHosts to hook into the creation of the connection to make sure a host is available by the time the connection is made available.

The longer-term goal is to remove `RepositoryHost(s)` so this can go away.

src/GitHub.App/Models/RepositoryHosts.cs
```diff
@@ -45,78 +40,20 @@ public class RepositoryHosts : ReactiveObject, IRepositoryHosts
...
 
-            // Wait until we've loaded (or failed to load) an enterprise uri from the db and then
-            // start tracking changes to the EnterpriseHost property and persist every change to the db
-            disposables.Add(persistEntepriseHostObs.Subscribe());
+            loaded = new TaskCompletionSource<object>();
+            disposables.Add(connectionManager.Connections.ForEachItem(_ => { }, ConnectionRemoved, ConnectionsReset));
```
grokys: Here we'd ideally just call `ConnectionAdded` as the first callback in `ForEachItem` which would also handle initialization, but because of the hack mentioned above we have to initialize explicitly and use `ConnectionCreated` to create the `RepositoryHost`.

src/GitHub.App/ViewModels/PullRequestListViewModel.cs
```diff
@@ -148,11 +148,11 @@ public override void Initialize(ViewWithData data)
...
-                .Catch<System.Reactive.Unit, Octokit.AuthorizationException>(ex =>
-                {
-                    log.Info("Received AuthorizationException reading pull requests", ex);
-                    return repositoryHost.LogOut();
-                })
+                ////.Catch<System.Reactive.Unit, Octokit.AuthorizationException>(ex =>
+                ////{
+                ////    log.Info("Received AuthorizationException reading pull requests", ex);
+                ////    return repositoryHost.LogOut();
+                ////})
```
grokys: I think this code should be removed @shana - as far as I can see there's no good reason for the PR list to be messing with removing connections?

src/GitHub.Exports.Reactive/Extensions/ConnectionManagerExtensions.cs
```diff
@@ -12,33 +15,40 @@ public static class ConnectionManagerExtensions
         public static IObservable<bool> IsLoggedIn(this IConnectionManager cm, IRepositoryHosts hosts)
```
grokys: These methods should be changed to return `async Task<T>`  but I didn't want to have to update all the call-sites in the PR.

src/UnitTests/GitHub.App/Models/RepositoryHostTests.cs
```diff
@@ -16,115 +16,4 @@
 
 public class RepositoryHostTests
 {
-    public class TheLoginMethod : TestBaseClass
```
grokys: `RepositoryHost` no longer has any functionality, so removed these tests.

src/GitHub.App/Models/RepositoryHost.cs
grokys: Done this now in `ConnectionManager`.

src/GitHub.App/Models/RepositoryHost.cs
grokys: Done this now in `ConnectionManager`.

grokys:
Merged https://github.com/github/VisualStudio/pull/1283 which was based on and superseded this PR - not sure why this PR wasn't automatically marked as merged when I merged that one.

PR #1278, opened 2017-10-20
WIP: Consolidate unit tests and test runners

jcansdale:
This PR moves consolidate all unit tests into the `test` folder. It also updates the NUnit and xUnit scripts so they know where to find tests to execute.

The version of the NUnit console runner has been updated to 3.7.0. This allows it to execute the tests from `GitHub.UI.UnitTests`.  The version of NUnit used by `TrackingCollectionTests` has been bumped to `3.6.1`. This allows the same NUnit console runner to execute all tests.

The NUnit adapter for VS has been updated to NUnit3TestAdapter. This lets all NUnit tests be executed using the built in test runner for VS.

PR #1282, opened 2017-10-26
Moving the build scripts out to the public repo and other cleanups

shana:
It's much easier for everyone if the build scripts are out in the public so everyone can use them, and all CI systems use the same scripts so that things aren't missed.

This:
- Adds a public key to the repo, with a different name from what we suggested before so it doesn't conflict with any existing keys user may have locally
- Adds all build scripts to the `scripts` directory
- Sets the version of the vsix to a timestamp when building locally
- Sets the version of DLLs and vsix when building in CI, but doesn't commit them
- Cleans up csproj files so all have the same CA and warnings settings
- Adds solution configurations for different variations of building with CA enabled and disabling vsix creation/deployment
  - Debug now builds with warnings as errors enabled but CA disabled
- Builds in CI from our own PR branches and from master produced ready-to-release signed vsix packages as artifacts
- Deployments are enabled in AppVeyor to S3 and GitHub releases (go to a branch with artifacts and click Deploy to deploy them). S3 deploys become available in https://ghfvs-installer.github.com/releases/X.X.X.X/[name of vsix/md5/manifest file] and https://ghfvs-installer.github.com/releases/ghfvs-X.X.X.X.zip


jcansdale:
You'll notice there's a failing test:

```
1) Error : GitHub.UI.Helpers.UnitTests.SharedDictionaryManagerTests+TheGetCurrentDomainCachingFactoryMethod.InvokeMethodOnNewAssembly_DisposeCalled
NSubstitute.Exceptions.ReceivedCallsException : Expected to receive exactly 1 call matching:
```

I have a fix for this in #1278

jcansdale:
@shana, the build is now failing with:

image

There is still an intermittently failing test when I test locally. I'll continue trying to fix it.

src/GitHub.InlineReviews/SampleData/DiffCommentThreadViewModelDesigner.cs
```diff
@@ -1,12 +1,15 @@
...
 using GitHub.InlineReviews.ViewModels;
 
 namespace GitHub.InlineReviews.SampleData
 {
+    [SuppressMessage("Microsoft.Performance", "CA1812:AvoidUninstantiatedInternalClasses")]
     class DiffCommentThreadViewModelDesigner : IDiffCommentThreadViewModel
     {
         public string DiffHunk { get; set; }
         public int LineNumber { get; set; }
+        public event EventHandler Disposed = delegate { };
```
grokys: This member can just be removed: it's not implementing an interface or anything.

src/GitHub.InlineReviews/Models/PullRequestSessionFile.cs
```diff
@@ -35,6 +35,27 @@ public PullRequestSessionFile(string relativePath)
...
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposed)
+            {
+                disposed = true;
+
+                if (disposing)
+                {
+                    linesChanged.Dispose();
```
grokys: A `PullRequestSessionFile` can be a shared instance between multiple clients, disposing of `linesChanged` here would unsubscribe all clients. IMO this class shouldn't be disposable, or at least the `Dispose` method shouldn't dispose `linesChanged` as this will break things.

src/GitHub.InlineReviews/Peek/InlineCommentPeekResult.cs
```diff
@@ -22,10 +22,11 @@ public InlineCommentPeekResult(InlineCommentPeekViewModel viewModel)
 
         public Action<IPeekResult, object, object> PostNavigationCallback => null;
 
-        public event EventHandler Disposed;
+        public event EventHandler Disposed = delegate { };
 
         public void Dispose()
         {
+            Disposed?.Invoke(this, EventArgs.Empty);
```
grokys: I don't think `= delegate { };` is needed as `Disposed` is actually called.

src/GitHub.InlineReviews/Peek/InlineCommentPeekResultPresentation.cs
```diff
@@ -40,9 +40,9 @@ private set
             }
         }
 
-        public event EventHandler IsDirtyChanged;
-        public event EventHandler IsReadOnlyChanged;
-        public event EventHandler<RecreateContentEventArgs> RecreateContent;
+        public event EventHandler IsDirtyChanged = delegate { }; // IsDirty is always false.
```
grokys: I think I'd prefer these unused events to be declared as explicit interface implementations with empty `add` and `remove` blocks, even though it's a bit more verbose - this syntax doesn't make clear to me what is happening...

src/GitHub.InlineReviews/Services/PullRequestSession.cs
```diff
@@ -21,9 +21,8 @@ namespace GitHub.InlineReviews.Services
     /// It takes the pull request model and updates according to the current state of the
     /// repository on disk and in the editor.
     /// </remarks>
-    public class PullRequestSession : ReactiveObject, IPullRequestSession
+    public class PullRequestSession : ReactiveObject, IPullRequestSession, IDisposable
```
grokys: Again, this is a shared object so shouldn't be getting disposed.

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
```diff
@@ -308,20 +308,20 @@ void HandleCreatedRepo(ILocalRepositoryModel newrepo)
         {
             Guard.ArgumentNotNull(newrepo, nameof(newrepo));
 
-            var msg = string.Format(CultureInfo.CurrentUICulture, Constants.Notification_RepoCreated, newrepo.Name, newrepo.CloneUrl);
-            msg += " " + string.Format(CultureInfo.CurrentUICulture, Constants.Notification_CreateNewProject, newrepo.LocalPath);
+            var msg = String.Format(CultureInfo.CurrentCulture, Constants.Notification_RepoCreated, newrepo.Name, newrepo.CloneUrl);
```
grokys: Nit: the capitalization of `string` probably shouldn't have been changed here.

src/GitHub.InlineReviews/Peek/InlineCommentPeekResultPresentation.cs
jcansdale: Indeed. It's concise, but not exactly clear!

src/GitHub.InlineReviews/Peek/InlineCommentPeekResult.cs
jcansdale: All use of `delegate { }` gone now.

appveyor.yml
```diff
@@ -17,12 +16,13 @@ install:
...
 build_script:
-- cmd: msbuild "GitHubVS.sln" /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll" /p:Configuration=Release /p:DeployExtension=false /verbosity:minimal /p:VisualStudioVersion=14.0
+- ps: scripts\build.ps1 -AppVeyor -BuildNumber:$env:APPVEYOR_BUILD_NUMBER
 test_script:
-- ps: >-
-    scripts\Run-Nunit.ps1 TrackingCollectionTests 180 Release -AppVeyor
-
-    scripts\Run-Xunit.ps1 UnitTests 180 Release -AppVeyor
\ No newline at end of file
+- ps: scripts\test.ps1 -AppVeyor
```
grokys: @shana do we want to use appveyor's automatic test detection? https://www.appveyor.com/docs/running-tests/#selecting-assemblies-andor-categories-to-test

shana: Yeah, we just need to consolidate all the tests in one place so we can have AppVeyor run them with the correct categories ignored. We also need to consolidate the test categories to be the same for everyone.

PR #1283, opened 2017-10-27
Remove RepositoryHost(s)

grokys:
This PR removes `RepositoryHosts` and `RepositoryHost` and makes all clients that used them use `ConnectionManager` instead.

Current state: I think mostly done but not tested much.

Depends on #1280


jcansdale:
After removing the contents of the: `C:\Users\<user>\AppData\Local\GìtHūbVisualStudio` directory, the login stuck on `GitHub Enterprise` and I wasn't able to log into my `GitHub` account. I installed this PR version and tried again. After this I was able to log in successfully. 👍

src/GitHub.App/Factories/ModelServiceFactory.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        readonly IApiClientFactory apiClientFactory;
+        readonly IHostCacheFactory hostCacheFactory;
+        readonly IAvatarProvider avatarProvider;
+        readonly Dictionary<IConnection, ModelService> cache = new Dictionary<IConnection, ModelService>();
+        readonly SemaphoreSlim cacheLock = new SemaphoreSlim(1);
+
+        [ImportingConstructor]
+        public ModelServiceFactory(
+            IApiClientFactory apiClientFactory,
```
shana: It would be better to use `ISimpleApiClientFactory` here, so we can slowly move away from `IApiClientFactory` and get rid of it.

grokys: Yes, but `ModelService` still relies on `IApiClient` rather than `ISimpleApiClient` and needs the reactive stuff, so if we want to make that change it will have to be later I think.

meaghanlewis:
this PR looks good to me 👍

PR #1284, opened 2017-10-31
[wip] Getting versioning and releases working

shana:
Based on #1282, getting CI to version and publish package artifacts


scripts/modules.ps1
```diff
@@ -0,0 +1,244 @@
...
+                Write-Host "``$Command`` failed" $output
+            } else {
+                Die $exitCode "``$Command`` failed" $output
+            }
+        }
+        $output
+    }
+
+    <#
+    function Run-Job([string[]]$Command, [switch]$Fatal, [switch]$Quiet, [int]$Timeout = 0) {
```
grokys: Commented-out code.

scripts/modules/WiX.ps1
```diff
@@ -0,0 +1,43 @@
+Set-StrictMode -Version Latest
```
grokys: Do we still need this now we're not building a WiX installer?

src/GitHub.InlineReviews/Peek/InlineCommentPeekResult.cs
```diff
@@ -22,10 +22,11 @@ public InlineCommentPeekResult(InlineCommentPeekViewModel viewModel)
 
         public Action<IPeekResult, object, object> PostNavigationCallback => null;
 
-        public event EventHandler Disposed;
+        public event EventHandler Disposed = delegate { };
```
grokys: What does the `= delegate { }` do here considering the `Disposed` event is being called?

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
```diff
@@ -308,20 +308,20 @@ void HandleCreatedRepo(ILocalRepositoryModel newrepo)
         {
             Guard.ArgumentNotNull(newrepo, nameof(newrepo));
 
-            var msg = string.Format(CultureInfo.CurrentUICulture, Constants.Notification_RepoCreated, newrepo.Name, newrepo.CloneUrl);
-            msg += " " + string.Format(CultureInfo.CurrentUICulture, Constants.Notification_CreateNewProject, newrepo.LocalPath);
+            var msg = String.Format(CultureInfo.CurrentCulture, Constants.Notification_RepoCreated, newrepo.Name, newrepo.CloneUrl);
```
grokys: Why change `string` to `String` here?

src/GitHub.InlineReviews/Peek/InlineCommentPeekResult.cs
jcansdale: It stops CA complaining that `Disposed` isn't being used.

The alternatives are:
1. An attribute to ignore this error.
2. Explicitly defining `add { } and remove { }`.

This seemed like the least worst option. I'm open to alternatives.

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
shana: I was copy-pasting the culture change and I guess I copy-pasted from somewhere that was uppercase

scripts/modules.ps1
shana: Yes, I had a hard time getting this to work, powershell sucks

src/GitHub.InlineReviews/Peek/InlineCommentPeekResult.cs
grokys: What I meant was that on line 29 we're calling `Disposed?.Invoke(this, EventArgs.Empty);`, so `Disposed` _is_ being used. Will `Disposed` even work with the event defined like this?

PR #1292, opened 2017-11-06
Making TeamExplorerServiceHolder testable

shana:
This makes TeamExplorerServiceHolder testable by moving/abstracting all the dependent VS types. 

~~This also reverts the change to have GitHub.TeamFoundation.15 reference Shell.15 - we rely on types from Shell.14 in GitHub.Exports and other areas (Shell.14 is available in VS 2017), namely UIContextChangedEventArgs.~~


grokys:
When running this I get:

```
Exception thrown: 'GitHub.GitHubLogicException' in GitHub.Exports.dll

Additional information: Could not locate view model for GitHubPane.
```

jcansdale:
> This also reverts the change to have GitHub.TeamFoundation.15 reference Shell.15

The problem is that `Microsoft.TeamFoundation.Git.Provider, Version 15.0.0.0` (which we reference from `GitHub.TeamFoundation.15`), references the `Microsoft.VisualStudio.Shell.15.0` assembly.

This causes the following warning.

```
C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1820,5):
warning MSB3277: Found conflicts between different versions of the same dependent assembly that
could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed.
[C:\source\github.com\github\VisualStudio\src\GitHub.TeamFoundation.15\GitHub.TeamFoundation.15.csproj]
```

These warnings can be difficult to track down and it isn't at all clear what the issue is whether it's serious or not.

jcansdale:
I keep getting the following on this branch:

image

I wonder what's going on?

shana:
This should be working now, MEF crashes are fixed.

Also, this branch is now loading the following DLLs up front when starting VS in Team Explorer Home:

image

Note that `GitHub.Exports.Reactive`, `GitHub.UI.Reactive` and `GitHub.Extensions.Reactive` are not on the list, which is good! But rx is still getting loaded, probably via `GitHub.App`, so we should continue spelunking to find out what's loading `GitHub.App` so we stop doing that (this might be the credential handler, in which case #1277 might fix that.

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        void CheckAndUpdate(object sender, System.ComponentModel.PropertyChangedEventArgs e)
+        {
+            Guard.ArgumentNotNull(e, nameof(e));
+            if (e.PropertyName != "ActiveRepositories" || gitService == null)
+                return;
+            ActiveRepositoriesChanged?.Invoke();
+        }
+
+        public IEnumerable<ILocalRepositoryModel> ActiveRepositories => gitService?.ActiveRepositories.Select(x => IGitRepositoryInfoExtensions.ToModel(x));
```
grokys: `IGitRepositoryInfoExtensions.ToModel` is an extension method so why not just say `x.ToModel()`?

PR #1302, opened 2017-11-07
Don't block checking out a PR if there are untracked files

### LINKED ISSUE #1271: Can't checkout PR when there are untracked files
- GitHub Extension for Visual Studio version: 2.3.3.38
- Visual Studio version: Add

The extension often won't let me checkout PR branches.

image

This can be due to trivial things like build artifacts being added to the staging area.

image

This can be caused by build artifacts/tools (NCrunch) adding files to the working directory. Untracked files usually wouldn't be a problem when using Git from the command line, they can be a real nuisance when using the extension.

This is particularly problematic when the untracked files appear in submodules. The Git command line arguments necessary to clean them up aren't at all obvious (to me at least).

The current check is done here:
https://github.com/github/VisualStudio/blob/a480116a45053b43a3175d37603e117caa0234c5/src/GitHub.App/Services/PullRequestService.cs#L89

It uses `IsDirty`, which checks for any differences between the working directory/index and the last comment. By default, command line Git will allow a checkout if there are untracked files in the working directory.

This issue is even more insidious when an untracked file gets added to a submodule. Here I'ved added `foo.txt` to the `splat` submodule.

Visual Studio (below) and the Git command line doesn't give much of a clue as to what's going wrong. You can see that something might be up with the `splat` submodule, but it doesn't give any indication of exactly what.

image

jcansdale:
Loosen slightly when a PR can be checked out.

We don't allow checkout when:
-    A file has been added to index
-    A file has been modified in working dir
-    A file has been staged
-    A file is missing from working dir
-    A file has been removed from index
-    A file has been renamed in index
-    A file has been renamed in working dir

We do allow checkout when:
-    The index and working dir is unaltered since last commit
-    An ignored file has changed
-    There are untracked files

This isn't quite a loose as when checking out from the command line, but it looser than is was before (when only Ignored and Unaltered files were alowed).

Fixes #1271


src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -89,7 +89,22 @@ public IObservable<string> GetPullRequestTemplate(ILocalRepositoryModel reposito
         public IObservable<bool> IsWorkingDirectoryClean(ILocalRepositoryModel repository)
         {
             var repo = gitService.GetRepository(repository.LocalPath);
-            return Observable.Return(!repo.RetrieveStatus().IsDirty);
+            var isClean = !IsFilthy(repo.RetrieveStatus());
+            return Observable.Return(isClean);
+        }
+
+        static bool IsFilthy(RepositoryStatus status)
```
grokys: Is this actually needed though? Can't we just let libgit2sharp error out if it doesn't manage the operation? I honestly can't remember why I felt it necessary to do a check for cleanliness here :/

Also LOL at the name ;)

jcansdale: > Is this actually needed though? Can't we just let libgit2sharp error out if it doesn't manage the operation?

I wanted to start with a simple drop-in replacement for what we've been using (a simple bug fix). Letting libgit2sharp error out might be better and give the user more info, but would also require more testing.

> I honestly can't remember why I felt it necessary to do a check for cleanliness here :/

I know command line Git lets you checkout when there modified files, as long as they don't conflict with changes in the branch you're checking out. I'm not sure this is something we would want to allow in this context. Maybe this is an option, I'll investigate.

If you're happy with this simple fix, I'd prefer to start with that and improve later. Can I open a separate PR that uses the letting libgit2sharp error out technique?

src/UnitTests/GitHub.App/Services/PullRequestServiceTests.cs
```diff
@@ -17,6 +17,241 @@
...
+                var repositoryModel = CreateLocalRepositoryModel(repo);
+                var path = "file.txt";
+                var renamedPath = "renamed.txt";
+                var file = Path.Combine(repo.Info.WorkingDirectory, path);
+                var renamedFile = Path.Combine(repo.Info.WorkingDirectory, renamedPath);
+                File.WriteAllText(file, "contents");
+                Commands.Stage(repo, path);
+                repo.Commit("foo", Author, Author);
+                File.Move(file, renamedFile);
+                // NOTE: Renamed files appear as Missing and Untracked rather than RenamedInWorkingDir. Is this a bug?
```
jcansdale: @ethomson I'm surprised that renaming a (committed) file didn't make it appear as `RenamedInWorkingDir`. It instead appears as `Missing` and `Untracked`. This also makes sense, but I wonder what `RenamedInWorkingDir` is actually for? Might this be a bug or am I missing something obvious?

ethomson: You need to query for working directory renames explicitly in your Status, they're not included by default.  Since they look at _every_ untracked file, they're particularly expensive to calculate and I wouldn't recommend it, tbqh.

We added it to libgit2/LibGit2Sharp for Visual Studio 2013's Team Explorer when working in the mode where it hid the stage.  (If you have no staging area, you can't calculate renames between the HEAD and the index, so it calculated unstaged renames in the working directory).

Note that this default (_not_ detecting renames in the working directory) matches the behavior of https://github.com/git/git where status only shows staged renames (and in fact has no option to enable unstaged renames in the working directory).

jcansdale: Thanks for clearing up my confusion!

src/GitHub.App/Services/PullRequestService.cs
jcansdale: Here is a PR that lets libgit2sharp error out rather than simply allowing un-tracked files: #1304. I'm going to document the implications of this further and get more eyes on it.

Until #1304 is approved, I'm keen to get this PR merged as a point fix. I don't mind reverting it later, but I'm sick of not being able to check-out PRs! 😉  

Does that make sense?

Issue #826, opened 2017-02-01
Submodules cause all kinds of issues when checking out a PR


jcansdale:
- GitHub Extension for Visual Studio version: 2.2.0.6
- Visual Studio version: 2015

There are many ways a submodule can become dirty/changed without the user intending to touch the module. Here is one example:

1. Open the `github/VisualStudio` solution and GitHub tool window
2. Click the title of PR 813
3. Click `Checkout fixes/move-submodules`
4. Click `Pull Requests` icon.
5. Click the title of PR 734
6. Click `Checkout fixes/move-submodules`
7. Hovering over the `Checkout` button shows the following:
![image](https://cloud.githubusercontent.com/assets/11719160/22509052/5d36e7b6-e884-11e6-8b6e-0aaf30f5142e.png)

This is disconcerting because the user hasn't made any changes. To fix it from inside Visual Studio, they'll need to.

1. Open the `Team Explorer` tool window
2. Click on the `Changes` icon
3. Right-click on the `script` module and `Submodule Update`.
4. On the GitHub tool window, click the refresh icon
5. The `Checkout` button will now be clickable

This seems like a lot of ceremony to go though have not intentionally changed anything! Now, I know the user should have done a `Submodule Update` (or submodule sync) after checking out PR 813, but they won't necessarily know they have to on other projects.

jcansdale:
Another scenario is if someone enables NCrunch on a solution that contains projects in a submodule (not an uncommon situation). If you you do that with the `github\VisualStudio` solution, you end up in a superficially similar situation to the one above.

![image](https://cloud.githubusercontent.com/assets/11719160/22510503/1e2be19c-e88a-11e6-8f09-811aad128cb3.png)

Except, this time there is no obvious way to checkout or undo changes in the module from inside Visual Studio. The user will need to drop to the command line and `git submodule foreach git clean -xdf`. This isn't at all obvious for someone who hasn't knowingly touched a submodule and isn't familiar with now to maintain them!

jcansdale:
Obviously these are limitations of Visual Studio's Git support, but they're issues that will prevent our GitHub users from easily moving between PRs.

I wonder if we could offer some kind of submodule nuke functionality that would get submodules into a clean state so that another PR can be checked out?

Maybe an option to clean submodules here?
![image](https://cloud.githubusercontent.com/assets/11719160/22510989/f6e7dbe8-e88b-11e6-99be-16d1420b4553.png)

Haacked:
[...]

PR #1304, opened 2017-11-08
[SPIKE] Don't do "clean" check before attempting to check-out PR

### LINKED ISSUE #1271: Can't checkout PR when there are untracked files

jcansdale:
NOTE: This PR is intended for experimenting with functionality rather than necessarily merging.

In previous versions, the PR `Checkout` link was only active if the working directory and latest commit where exactly the same. There isn't this restriction when checking out using command line Git.

In #1302 this restriction is relaxed a bit to allow untracked files in the staging area (often build artifacts or log files). This PR changes it to behave more like the command line. It will always attempt a checkout and inform the user if it fails.

- For example, if the user makes a local change.

image

The `Checkout` command will work and bring that change to the PR branch. There currently isn't any indication that there are local changes when using the PR pane.

- If there are local changes that conflict with the PR being checked out, an error message will be displayed.

image

It doesn't currently give any hints how to fix the conflicts. Maybe we should offer the user a link to the `Team Explorer - Changes` pane?

Fixes #1271.


grokys:
> It doesn't currently give any hints how to fix the conflicts. Maybe we should offer the user a link to the Team Explorer - Changes pane?

Yes, I think that would be the best solution.

jcansdale:
Here are some related issues:

Can't checkout PR when there are untracked files: https://github.com/github/VisualStudio/issues/1271
Submodules cause all kinds of issues when checking out a PR: https://github.com/github/VisualStudio/issues/826

meaghanlewis:
Hey @jcansdale thanks for making these changes! I was testing this and have a few questions.

- Where does the message `1 conflict prevents checkout` come from? I can't find it in the codebase anywhere.
- I get the conflict message sometimes even when I've changed files that aren't related to a PR. For example, I changed a file (`src\GitHub.UI\GitHub.UI.csproj`) that wasn't included in the files changed in this PR and then tried to check out this branch `fixes/1271-just-try-PR-checkout`. This is unexpected, right?
- Is clicking `[GitChanges]` supposed to take me to Team Explorer - Changes? Currently it's not doing anything for me.
git changes

- Unrelated to functionality, but I noticed that tests were removed for this PR but no new ones were added. Should they be added? I'd be up to pair to add a test or two (especially because I currently have no idea how to write tests for this project)!

jcansdale:
> Where does the message 1 conflict prevents checkout come from? I can't find it in the codebase anywhere.

This is a the message from a libgit2sharp exception that gets thrown when it fails to checkout (kind of like when command line Git can't do a checkout).

> I get the conflict message sometimes even when I've changed files that aren't related to a PR. For example, I changed a file (src\GitHub.UI\GitHub.UI.csproj) that wasn't included in the files changed in this PR and then tried to check out this branch fixes/1271-just-try-PR-checkout. This is unexpected, right?

I think changing files that are included in the PR might be okay, it's changing files that are changed on the target branch that is an issue.

> Is clicking [GitChanges] supposed to take me to Team Explorer - Changes? Currently it's not doing anything for me.

Yup, that is how it's supposed to work! I'll check.

> Unrelated to functionality, but I noticed that tests were removed for this PR but no new ones were added. Should they be added?

In this case the tests were for a method that has been removed. To test the new functionality, we would need to do integration tests against a real Git repo. This might be a little involved, so I think it would make sense to flesh out exactly how we want it to work before tackling this. It definitely makes sense to add them though.

jcansdale:
This PR was an experiment at addressing #1271. This has since been fixed by the following PRs:

- Don't block checking out a PR if there are untracked files #1302 
- [Feature] Reviewing PR with submodule changes #1392

PR #1308, opened 2017-11-09
Reference Serilog assemblies for NCover

jcansdale:
NCover needs to copy these asemblies to test directory.


grokys:
The version of Serilog used elsewhere is 2.5.0 but here version 2.6.0-dev-00922 has been installed. Should use the same version everywhere.

jcansdale:
@grokys Okay, fixed now!

PR #1312, opened 2017-11-13
Pull Request Filtering

### LINKED ISSUE #722: Add filtering and sorting to PR list.

grokys:
Adds a filter box to the pull request list.

image

This PR supersedes #732 - moving it to a branch in this repository so we can get artifacts.

Fixes #722


meaghanlewis:
These changes LGTM ✅ . The filter box is really nice.

jcansdale:
I've been using this and it's really useful functionality!

One thing that feels a little wrong. I want to be able to enter a PR number and hit return to open it (like you can with some other VS search boxes). This functionality would also be useful when in the PR details view, where you can see the search box but can't use it. Would either of these be appropriate for this PR?

grokys:
Yes the search could be useful in the PR details view too, but we'd need to decide what it did. Filter the changed files list? Highlight the search term in the description?

For these reasons I felt it best just to disable it in the PR details for now, we can address searching there in another PR.

grokys:
Hmm, hitting enter should work - it works for me at least.

jcansdale:
@grokys,

> What other VS search boxes use hitting enter to navigate rather than submit the query immediately?

image

- The `Quick Launch` performs the action at the top of the list when you hit enter
- The toolbar `Search` searches the active document when you hit enter

grokys:
Hmm, ok. To me these search bars are slightly different, in that they're not filters, but lets discuss this separately I'd say.

PR #1314, opened 2017-11-13
 Fix packaging of startpage and inlinereviews vsix's

### LINKED ISSUE #1310: Add GitHub option back to start page in VS2017
__Tested on:__
Visual Studio version: VS2017, 15.4.3

This option was apparent in the last GHfVS release `v2.3.4.54` but stopped working after that. 
I notice the icon disappear after https://github.com/github/VisualStudio/pull/1282 was merged to master, but am not sure about the exact commit

__What happened__ 
Open VS2017 with the latest dev build on master and look on the start page.

**Expected: GitHub option is present**
screen shot 2017-11-09 at 10 25 45 am

**Actual: GitHub option is not present**
screen shot 2017-11-09 at 10 23 13 am

shana:
The build system overhaul removed some key parts of the entries that control what gets packaged into the VSIX so pkgdef files for GitHub.StartPage and GitHub.InlineReviews weren't being included. Additionally, both these projects should not be deployable as VSIXs because they're not standalone extensions (they can be for testing purposes but that should be done separately and manually on a case-by-case basis as needed). The combination of these two things means that default local builds would cause these packages to be deployed to the VS experimental instance and override the loading directory for DLLs, which is Not A Good Thing(tm)

Also take the opportunity to fix the publisher and identity of the InlineReviews package to match our other packages.

Fixes #1310

### DIFF
--- a/src/GitHub.InlineReviews/source.extension.vsixmanifest
+++ b/src/GitHub.InlineReviews/source.extension.vsixmanifest
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
   <Metadata>
-    <Identity Id="GitHub.InlineReviews.Steven Kirk.56b6fdce-e5b1-4ec3-9837-af4545ba933e" Version="1.0" Language="en-US" Publisher="Steven Kirk" />
+    <Identity Id="GitHub.InlineReviews.56b6fdce-e5b1-4ec3-9837-af4545ba933e" Version="2.0" Language="en-US" Publisher="GitHub, Inc" />
     <DisplayName>GitHub Inline Reviews</DisplayName>
     <Description xml:space="preserve">Inline reviews for GitHub pull requests</Description>
   </Metadata>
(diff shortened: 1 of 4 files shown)


src/GitHub.InlineReviews/source.extension.vsixmanifest
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
```
jcansdale: This `source.extension.vsixmanifest` shouldn't be used (we're not creating a .vsix for this project). Can we delete it from this project?  The same would apply to the on in `GitHub.StartPage`.

Am I missing something and there's a reason why they can't be deleted?

shana: The vsix file has an entry exporting the pkgdef. I'm not sure you can kill the manifest and still have the project packaging properly. I think you may have to change the project type, and/or maybe have another project type for the pkgdef-but-no-vsix logic. There used to be a project type for this at some point iirc, but honestly I can't be sure so it would need testing.

jcansdale: That's what I thought as well, but I tried deleting the `.vsixmanifest` and it still seemed to compile. Maybe this has changed in a recent VSIX project type update? Dunno.

shana: Could have, I wouldn't be surprised. I remember the file being mandatory but if they updated the targets file to skip it since then, then that would make sense.

PR #1325, opened 2017-11-17
Stop fighting NuGet

jcansdale:
Every time we update NuGet packages, Visual Studio / NuGet likes to add
`dependentAssembly` elements to `App.config` files and `NuGetPackageImportStamp` elements to `.csproj` files. These can be a pain to tidy up after every change.

This PR simply adds the stuff NuGet wanted to add when the `libgit2sharp` package was updated. This will avoid having to clean up again next time a NuGet package is updated. The idea is to avoid fighting NuGet all the time.

Alternatively we might find that `App.config` files aren't actually nessesary and we can remove them from the solution. NuGet would probably still add them, but deleting them might be easier than editing them. Thoughts?


shana:
Pretty sure app.config files are not used by extensions. The only app.config file that would be used is the one in DesignTimeStyleHelper, since that's an actuall app. I'd say delete them and add them to .gitignore so if they get added again they won't get accidentally pushed to the repo.

jcansdale:
>  I'd say delete them and add them to .gitignore so if they get added again they won't get accidentally pushed to the repo.

Good plan. I'll try that and see if anything breaks.

grokys:
The problem with deleting the `app.config` files and adding them to `.gitignore` is that they'll get added to the `.csproj` anyway. I guess the compile error this will cause will at least mean then get deleted again, but it's not ideal.

If I'm reading [this issue](https://github.com/NuGet/Home/issues/1147) right, there should be a way to disable writing them?

shana:
@grokys oh nice, yeah, let's do that!

jcansdale:
I've confirmed that you can change the version of a NuGet package and the only changes in the diff are the version numbers in `package.config` / `*.csproj`.

PR #1334, opened 2017-11-21
Check scopes on login.

### LINKED ISSUE #1332: Validate scopes for login.
From: https://github.com/github/VisualStudio/pull/1277#pullrequestreview-77416497

> Validating the token for permission scopes. If other git clients replace the token, they may not create one with the correct scopes, we need to handle that.

grokys:
When performing a login, check that the scopes returned are what we expect, otherwise throw an exception causing a new login to be required.

Fixes #1332


grokys:
The changes in this PR have been rolled into #1373, so closing this.

Issue #1345, opened 2017-11-27
Checked out status of PR doesn't update until refresh


jcansdale:
- GitHub Extension for Visual Studio version: 2.3.7
- Visual Studio version: 2015

__What happened__ (with steps, logs and screenshots, if possible)

1. Open PR detail view from PR list
2. Click on `Checkout - branch-name` button
3. PR is checked out, but button doesn't update
4. Click refresh button on GitHub pane
5. Button updates to show `Local branch up to date`

Notes:
- This works fine on current live version (`v2.3.6.391`).
- This doesn't work on branch `fixes/testable-teserviceholder`.

@grokys any thoughts?

grokys:
I think this was caused by https://github.com/github/VisualStudio/pull/1292. This PR also fixes a bug that we were relying on in `TeamExplorerServiceHolder`, in that it was previously firing `RepoChanged` when the _status_ of the repo were changed where it was meant to only be fired when _moving to a new repo_.

I suspect we'll need to add a `StatusChanged` signal or similar to `TeamExplorerServiceHolder`.

PR #1346, opened 2017-11-28
MVVM-ize things

### LINKED ISSUE #1105: Navigating to the PR list causes a reload of all PRs
Whenever one navigates from a PR details pane back to the PR list pane, the list of pull requests is refreshed. In the case of a repository with a lot of pull requests, this can be a lot of data, for example https://github.com/aspnet/mvc has 2.6MB of PR data.

This also happens when switching branches.

We probably shouldn't be doing this.

### LINKED ISSUE #1341: Replace UIController
`UIController` has various problems:

- It's hard to add new views, and how to do this isn't well documented
- It's not very flexible (a fixed set of triggers to move from state to state)
  - Even the Login view hacks around this internally
- It mixes view and view model concerns so is hard to test
- Relies on passing a **global** `IConnection` to get the connection from the `UIController` to view models
- Adding views involves adding entries to a long enum (`UIViewType`)
- `GitHubPaneViewModel` has to be implemented in `Views`
- #1105 When used in the GitHub pane there's no differentiation between load and refresh

MVVM is the accepted solution for such things, we should rip out `UIController` and replace it with a more standard architecture.

### LINKED ISSUE #1357: GitHub Pane shows "No repository" while initializing
While the GitHub pane is initializing after loading or when the active repo has changed, we show:

image

It would be better to show a spinner here.

grokys:
This pull request migrates our code to use the [MVVM pattern](https://msdn.microsoft.com/en-us/library/ff798384.aspx) throughout. MVVM is generally considered to be the "right" way to do WPF and greatly improves the unit-testability of code by separating application layer and view layer code.

This is a big PR, sorry! To ease reviewing, the following attempts to explain the major changes and concepts.

## UIController Removed

`UIController` was previously responsible for creating views and view models based on the `UIViewType` enumeration.

- There is no concept equivalent to `UIController` after the refactor; it's view models all the way down!
- Instead of `UIViewType`, view models and views are keyed on the type of the view model interface

## NavigationController Removed

`NavigationController` was built on `UIController` to display the browser-style user interface in the GitHub pane. The equivalent now is `NavigationViewModel`, however this simply maintains a list of pages and a position in that list, it knows nothing about the pages it will display. That responsibility is handled by:

## GitHubPaneViewModel

`GitHubPaneViewModel` exposes a `Content` property which:

- When the user is logged-in and a GitHub repository is open, will be a `NavigationViewModel`
- When the user is not logged in or there's not a GitHub repository open then the `Content` will be a view model representing this state such as an `ILoggedOutViewModel`

`GitHubPaneViewModel` supports navigation via URLs, such as `github://pane/owner/repo/pull/5` (via `NavigateTo(Uri)`) or by calling a method such as `ShowPullRequest("owner", "repo", 5)`.

## Asynchronous Initialization

Previously view model initialization methods were all of the signature `void Initialize(ViewWithData)`  (this was defined in the `IViewModel` base interface). This had three downsides:

- Type-safely was lost as everything had to be cast to an `object` and placed in `ViewWithData.Data`
- Asynchronous initialization was difficult
- It mixed view and view model concerns

The base `IViewModel` interface no longer defines a one-fits-all initialization method and view models are now expected to define a (probably `async`) initialization method in their interface which must be called following creation.

Many of the changes in this PR consist of moving things from the ViewModel constructor to an async initialization method.

## ViewLocator

At the root of each GitHub top-level control (i.e. dialog and the GitHub pane) there is a `DataTemplate` typed on `ViewModelBase`. This `DataTemplate` has a `ViewLocator` as its converter and is used to convert view models to their equivalent views. Using this technique the application layer code need have no concept of the view layer: the application state is built purely with view models which are converted to equivalent views by the view layer.

See the documentation [here](https://github.com/github/VisualStudio/blob/refactor/mvvm/docs/developer/how-viewmodels-are-turned-into-views.md).

## Views and View Models Moved

Previously all views and view models were all in the same folder/namespace, which was starting to get unwieldy. I've moved them under folders based on which part of the UI they appear in: `Dialog`, `GitHubPane` and `TeamExplorer`.

## Developer Documentation Added

I've added a few pieces of developer documentation explaining some of the new features:

https://github.com/github/VisualStudio/blob/refactor/mvvm/docs/developer/readme.md

Fixes #1105 
Fixes #1341 
Fixes #1357


src/GitHub.App/Authentication/TwoFactorRequiredUserError.cs
```diff
@@ -8,11 +8,18 @@ namespace GitHub.Authentication
...
+
+        public TwoFactorRequiredUserError(
+            TwoFactorAuthorizationException exception,
+            TwoFactorType twoFactorType)
             : base(exception.Message, innerException: exception)
         {
             Guard.ArgumentNotNull(exception, nameof(exception));
 
-            TwoFactorType = exception.TwoFactorType;
+            TwoFactorType = twoFactorType;
```
grokys: Added this extra constructor because the detection of the two-factor type is deep within Octokit's header parsing code. This allows us to easily override that in unit tests.

src/GitHub.UI.Reactive/Assets/Controls/Validation/UserErrorMessages.xaml
```diff
@@ -29,20 +29,14 @@
...
-                                        ToolTip="{Binding Path=ErrorCauseOrResolution}"
-                                        HorizontalAlignment="Stretch"
-                                        VerticalAlignment="Center"
-                                        TextWrapping="Wrap"
-                                        FontWeight="{Binding RelativeSource={RelativeSource FindAncestor, AncestorType={x:Type local:UserErrorMessages}},Path=ErrorMessageFontWeight}"
-                                        Margin="{Binding RelativeSource={RelativeSource FindAncestor, AncestorType={x:Type local:UserErrorMessages}},Path=MessageMargin}"/>
-                                </DataTemplate>
-                            </ContentControl.Resources>
-                        </ContentControl>
+                        <TextBlock Grid.Column="1"
```
grokys: I had to make this change because for some reason the old way of doing it was causing a `StackOverflowException` in WPF's layouting code. This change should produce equivalent results and is in fact simpler.

src/GitHub.App/Factories/ViewViewModelFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        readonly IGitHubServiceProvider serviceProvider;
+
+        [ImportingConstructor]
+        public ViewViewModelFactory(
+            IGitHubServiceProvider serviceProvider,
+            ICompositionService cc)
+        {
+            this.serviceProvider = serviceProvider;
+            cc.SatisfyImportsOnce(this);
```
jcansdale: I don't think importing `ICompositionService` or calling `SatisfyImportsOnce` is necessary. This should happen anyway after `ViewViewModelFactory` has been created.

I'm not sure `AllowRecomposition = true` is supported in Visual Studio's MEF implementation. I'm guessing this class some from somewhere else?

jcansdale:
When I click on the margin to add inline comment, visual Studio crashes with the following:

image
image

meaghanlewis:
I noticed that when not signed into GitHub and trying to view a repository's pull requests that the view just spins forever:

**expected**
prompt login pr

**actual**
forever spinning pr

meaghanlewis:
When checking out a PR branch, I can do so successfully but the view doesn't update from saying `checkout test-pr-creation` to `local branch updated`

**expected**
local branch up to date

**actual**
should say local branch up to date

grokys:
@meaghanlewis the spinner should be fixed - it was caused by a late change. The checkout not refreshing is #1345 - it's also happening on master and will be fixed in a separate PR.

grokys:
@jcansdale should now be fixed!

src/GitHub.App/Factories/ViewViewModelFactory.cs
grokys: Not sure - this class used to be [ExportFactoryProvider](https://github.com/github/VisualStudio/blob/e13678c2ea2bbc1cbefa3ed171b6bbf025569040/src/GitHub.Exports/Services/ExportFactoryProvider.cs#L15) and the code comes from there.

meaghanlewis:
thanks @grokys! ill give this one more pass this morning then

docs/developer/how-viewmodels-are-turned-into-views.md
```diff
@@ -0,0 +1,86 @@
...
+}
+```
+
+```xml
+<Window x:Class="MyApp.MainWindow"
+        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:local="clr-namespace:MyApp"
+        Content="{Binding}">
+</pfui:DialogWindow>
```
jcansdale: Should this be `</Window>`?

docs/developer/how-viewmodels-are-turned-into-views.md
```diff
@@ -0,0 +1,86 @@
...
+    <DataTemplate DataType="{x:Type local:ViewModel}">
+      
+      <!-- Display ViewModel.Greeting in a red border with rounded corners -->
+	  <Border Background="Red" CornerRadius="8">
+        <TextBlock Binding="{Binding Greeting}"
+      </Border>
+        
+    </DataTemplate>
+  </Window.Resources>
+</pfui:DialogWindow>
```
jcansdale: Should this be `</Window>`?

docs/developer/how-viewmodels-are-turned-into-views.md
```diff
@@ -0,0 +1,86 @@
...
+        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:local="clr-namespace:MyApp"
+        Content="{Binding}">
+  <Window.Resources>
+    <DataTemplate DataType="{x:Type local:ViewModel}">
+      
+      <!-- Display ViewModel.Greeting in a red border with rounded corners -->
+	  <Border Background="Red" CornerRadius="8">
+        <TextBlock Binding="{Binding Greeting}"
```
jcansdale: Missing a `/>`.

jcansdale:
image

Should we `#pragma warning disable CA1004` to keep the build clean?

src/GitHub.App/Factories/ViewViewModelFactory.cs
jcansdale: I put a breakpoint on the `set` method of `Views` and this property is indeed being set twice (with the same number of elements).

docs/developer/how-viewmodels-are-turned-into-views.md
```diff
@@ -0,0 +1,86 @@
...
+</DataTemplate>
+```
+
+The `DataTemplate.DataType` here applies the template to all classes inherited from [`GitHub.ViewModels.ViewModelBase`](../src/GitHub.Exports.Reactive/ViewModels/ViewModelBase.cs) [1]. The template defines a single `ContentControl` whose contents are created by a `ViewLocator`.
+
+The [`ViewLocator`](../src/GitHub.VisualStudio/Views/ViewLocator.cs) class is an `IValueConverter` which then creates an instance of the appropriate view for the view model using MEF.
+
+And thus a view model becomes a view.
+
+[1]: it would be nice to make it apply to all classes that inherit `IViewModel` but unfortunately WPF's `DataTemplate`s don't work with interfaces.
```
jcansdale: I'm sure I'm missing something, but is this not a `DataTemplate` that uses an interface?
https://stackoverflow.com/a/25642988/121348

src/GitHub.App/Factories/ViewViewModelFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        [ImportMany(AllowRecomposition = true)]
```
jcansdale: Visual Studio's MEF implementation is static (the graph is created once and cached). New exports/assemblies can't be added to the catalog on the fly. Having `AllowRecomposition = true` is confusing because recomposition is impossible.

src/GitHub.App/Factories/ViewViewModelFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+        IEnumerable<ExportFactory<FrameworkElement, IViewModelMetadata>> Views { get; set; }
+
+        /// <inheritdoc/>
+        public TViewModel CreateViewModel<TViewModel>() where TViewModel : IViewModel
+        {
+            return serviceProvider.ExportProvider.GetExport<TViewModel>().Value;
```
jcansdale: We could do the same as with `CreateView` and avoid any dependency on `IGitHubServiceProvider`.

docs/developer/how-viewmodels-are-turned-into-views.md
grokys: Hmm, interesting, I didn't know this was possible! Worth a try!

src/GitHub.App/Factories/ViewViewModelFactory.cs
grokys: Ok, well I'm fine with removing it! We should probably do it in a new PR though in case it causes problems.

src/GitHub.App/Factories/ViewViewModelFactory.cs
grokys: Yep! Lets try that in a separate PR too.

src/GitHub.App/ViewModels/GitHubPane/NavigationViewModel.cs
```diff
@@ -0,0 +1,197 @@
...
+        int index = -1;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="NavigationViewModel"/> class.
+        /// </summary>
+        public NavigationViewModel()
+        {
+            history = new ReactiveList<IPanePageViewModel>();
+            history.BeforeItemsAdded.Subscribe(BeforeItemAdded);
+            history.CollectionChanged += CollectionChanged;
```
sguthals: Where are BeforeItemAdded and CollectionChanged coming from?

sguthals: NVM - I see them below.

src/GitHub.App/GitHub.App.csproj
```diff
@@ -131,19 +132,40 @@
     <Reference Include="WindowsBase" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="Factories\ViewViewModelFactory.cs" />
```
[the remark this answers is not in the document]
grokys: Yeah, it's used for creating both views and view models. I did think it looked strange until I thought that MVVM stands for "model, _view, view model_" and so I decided to keep it.

src/GitHub.App/Factories/ViewViewModelFactory.cs
jcansdale: @sguthals Done, see #1368.

Issue #1317, opened 2017-11-14
Use Mads Kristensen's VSIX Gallery for GHfVS Beta feed


jcansdale:
Mads Kristensen (probably the most prolific creator of VS extensions) has created the VSIX Gallery for automatically publishing extension builds.

All a consumer of extensions needs to do is install the VSIX Gallery extension (yes, it's all very meta 😉).
https://marketplace.visualstudio.com/items?itemName=MadsKristensen.VSIXGallery-nightlybuilds

They then have access to automatic updates of extensions published to the site.

Here is how extensions get submitted to the VSIX Gallery:
http://vsixgallery.com/guide/dev/

meaghanlewis:
Closing out this PR because we don't want every one of our AppVeyor builds to be uploaded to the VSIX Gallery

PR #1354, opened 2017-11-28
[WIP] Get latest version of GHfVS build nightly using VSIX Gallery

meaghanlewis:
Use Mad Kristensen's [VSIX Gallery](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.VSIXGallery-nightlybuilds) to get nightly builds of GHfVS.

Closes https://github.com/github/VisualStudio/issues/1317


meaghanlewis:
👋  @jcansdale do you think you could help me out with this? 🙏 

I followed the instructions in: http://vsixgallery.com/guide/dev/ and first tried this from powershell by executing the VSIX script and then calling a method to upload the extension to the [VSIX 
 Gallery](http://vsixgallery.com/). That part was a breeze.

I added the same two steps in the `appveyor.yml` but the builds are not uploading from Appveyor. I looked through the [script code](https://github.com/madskristensen/ExtensionScripts/blob/master/AppVeyor/vsix.ps1) to see if im missing something obvious, but can't tell.

Mind taking a 👀 ?

shana:
@meaghanlewis It looks like by default the script looks for a vsix in the current directory https://github.com/madskristensen/ExtensionScripts/blob/master/AppVeyor/vsix.ps1#L12

You likely need to pass in the path to the vsix in order to get it working.

Also, this is hooking up into every build we do, not just beta builds, which means we'd be publishing random builds from incomplete PR branches, too... not sure that's what we really want to do?

meaghanlewis:
Thanks for taking a look on this @shana, this PR was pretty stale.

I did previously specify the path to the vsix as `c:\projects\visualstudio\release\**\*.vsix` - which didn't work although im pretty sure that path is correct.

And no, it wouldn't be a good idea to push that many builds to the VSIX Gallery, especially when some builds might be broken and users can end up trying those builds. We thought this could be a good way to easily grab new and install new builds to test, but it's just as easy to download and install builds from Appveyor instead. I also don't see a way we could pick and choose what builds get sent to the gallery.

I think it might be best to decide not to finish implementing this and close the issue.

meaghanlewis:
Closing out this PR because we don't want every one of our AppVeyor builds to be uploaded to the VSIX Gallery

PR #1366, opened 2017-12-05
Add lifecycle management for GitHub Pane pages

### LINKED ISSUE #1360: Make GitHub pane dispose of its view models
At the moment it isn't obvious how to make view models dispose of their resources.

In PR #1359 I've used a workaround that assumes that only one instance of `PullRequestDetailViewModel` will exist at a time:
https://github.com/github/VisualStudio/pull/1359/commits/6c4ff1b7eb97a84d927594ccd666c291f1f13450#diff-347b0f42676901bf61fcbabb3c74ec3eR88

If the the GitHub pane starts disposing of its view models, this can be removed.

grokys:
This PR makes `IPanePageViewModel`s get disposed when they are removed from a `NavigationViewModel`'s history. It also adds `Activated` and `Deactivated` methods to `IPanePageViewModel` which are called when the page is moved to/away from in the navigator. `NavigationViewModel.RegisterDispose` has been removed as disposing of the view models should suffice.

It modifies `PullRequestDetailViewModel` to only refresh when the page is active. When not active, it simply sets a flag to refresh when the page is activated. This should prevent all pages in the GitHub pane history refreshing at once, which is an expensive operation.

Depends on #1346 
Fixes #1360

### DIFF
--- a/src/GitHub.App/ViewModels/GitHubPane/GitHubPaneViewModel.cs
+++ b/src/GitHub.App/ViewModels/GitHubPane/GitHubPaneViewModel.cs
@@ -46,6 +46,7 @@ namespace GitHub.ViewModels.GitHubPane
         readonly ReactiveCommand<Unit> refresh;
         readonly ReactiveCommand<Unit> showPullRequests;
         readonly ReactiveCommand<object> openInBrowser;
+        bool initialized;
         IViewModel content;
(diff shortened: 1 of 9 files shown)


meaghanlewis:
LGTM. Tested this using VS2017 using the github/github repo.

PR #1369, opened 2017-12-06
Task-ize ConnectionManagerExtensions

grokys:
`ConnectionManagerExtensions` methods were returning `IObservables` but only returning a single value. Convert them to return `Task`s so they can be used in Team Explorer etc.

Also made the naming consistent and made sure we don't to try to create an `IModelService` for a non-logged-in connection.

### DIFF
--- a/src/GitHub.Exports.Reactive/Extensions/ConnectionManagerExtensions.cs
+++ b/src/GitHub.Exports.Reactive/Extensions/ConnectionManagerExtensions.cs
@@ -14,7 +14,7 @@ namespace GitHub.Extensions
             IModelServiceFactory factory)
         {
             var connection = await cm.GetConnection(repository);
-            return connection != null ? await factory.CreateAsync(connection) : null;
+            return connection?.IsLoggedIn == true ? await factory.CreateAsync(connection) : null;
         }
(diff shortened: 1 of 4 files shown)


src/GitHub.Exports.Reactive/Extensions/ConnectionManagerExtensions.cs
```diff
@@ -1,80 +1,20 @@
...
-        public static IObservable<IConnection> GetLoggedInConnections(this IConnectionManager cm)
-        {
-            Guard.ArgumentNotNull(cm, nameof(cm));
-
-            return cm.GetLoadedConnections()
-                .ToObservable()
-                .Select(x => x.FirstOrDefault(y => y.IsLoggedIn));
-        }
-
         public static async Task<IModelService> GetModelService(
```
jcansdale: Should this method be in `GitHub.Exports.Reactive`?  I didn't notice any Rx stuff that it depends on.

grokys: `IModelService` is reactive so that's why it's still here.

PR #1370, opened 2017-12-06
Simplify ViewViewModelFactory

### LINKED ISSUE #1361: Creating pull request fails with: "Request failed with status code: 403"
- Visual Studio version: 2015

1. Clicked on `Pull Requests`, `Create New`:

image

2. Entered title and description:

image

3. Request fails with `Request failed with status code: 403`.

Same happened after branch was created and pushed.

### LINKED ISSUE #1368: Simplify the way ViewViewModelFactory uses MEF

The `ViewViewModelFactory` class currently has dependencies on `ICompositionService` and `IGitHubServiceProvider`, which I'm not sure are necessary. It also uses `AllowRecomposition = true`, which I don't think is supported in Visual Studio's flavor of MEF.

I'd like to see if this can be simplified, so we don't confuse people looking at this class in future.

jcansdale:
This PR:

- Removes unnecessary call to `SatisfyImportsOnce`
- Removed unsupported setting of `AllowRecomposition`
- Stores `ViewModelType` as a `string` in metadata to avoid pre-loading assemblies

Storing the `ViewModelType` as a `string` mirrors the way `Export` metadata is stored. For example, the export: `Export(typeof(Foo.IBar))` will create a metadata property `ExportTypeIdentity` with the string "Foo.IBar".

Removing the dependency on `IGitHubServiceProvider` is more complicated and might be tackled in a separate PR, see: 
https://github.com/github/VisualStudio/pull/1370#discussion_r156337188

Fixes #1361

### DIFF
--- a/src/GitHub.Exports/Exports/ExportMetadata.cs
+++ b/src/GitHub.Exports/Exports/ExportMetadata.cs
@@ -37,10 +37,10 @@ namespace GitHub.Exports
         public ExportViewForAttribute(Type viewModelType)
             : base(typeof(FrameworkElement))
         {
-            ViewModelType = viewModelType;
+            ViewModelType = viewModelType.FullName;
         }
 
-        public Type ViewModelType { get; }
+        public string ViewModelType { get; }
     }
 
     /// <summary>
@@ -69,7 +69,7 @@ namespace GitHub.Exports
     public interface IViewModelMetadata
     {
         [SuppressMessage("Microsoft.Performance", "CA1819:PropertiesShouldNotReturnArrays")]
-        Type[] ViewModelType { get; }
+        string[] ViewModelType { get; }
     }
 
     /// <summary>
--- a/src/GitHub.Exports/Exports/ExportMetadata.cs
+++ b/src/GitHub.Exports/Exports/ExportMetadata.cs
@@ -32,6 +32,8 @@ namespace GitHub.Exports
     /// </summary>
     [MetadataAttribute]
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+    [SuppressMessage("Microsoft.Design", "CA1019:DefineAccessorsForAttributeArguments",
+        Justification = "Store string rather than Type as metadata")]
     public sealed class ExportViewForAttribute : ExportAttribute
     {
         public ExportViewForAttribute(Type viewModelType)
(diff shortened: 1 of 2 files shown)


src/GitHub.Exports/Exports/ExportMetadata.cs
```diff
@@ -69,7 +69,7 @@ public ExportMenuAttribute() : base(typeof(IMenuHandler))
     public interface IViewModelMetadata
     {
         [SuppressMessage("Microsoft.Performance", "CA1819:PropertiesShouldNotReturnArrays")]
-        Type[] ViewModelType { get; }
+        string[] ViewModelType { get; }
```
grokys: Why did you change the type of this member? It's not mentioned in the PR description.

jcansdale: My aim was to remove `ViewViewModelFactory`s dependency on `IGitHubServiceProvider`.

Instead of calling `IGitHubServiceProvider.ExportProvider.GetExport<TViewModel>()`, I was hoping to use a second `ExportFactory`, like so:

```
        [ImportMany]
        IEnumerable<ExportFactory<FrameworkElement, IViewModelMetadata>> Views { get; set; }

        [ImportMany("ViewModel")]
        IEnumerable<ExportFactory<object, IViewModelMetadata>> ViewModels { get; set; }
```

We would then enumerate through the `IViewModelMetadata` when finding a factory to create a View or ViewModel.  If the `IViewModelMetadata` contained a `Type` rather than a `string`, it would force *all* assemblies to load the first time a View or ViewModel was created (hence the change to using `string`).

Unfortunately there appear to be some strange limitations with `ExportFactory` which prevent this from working. 😢

jcansdale: I think the change from `Type` to `string` is still valuable because it should prevent all assemblies that contain a view from being pre-loaded when the first view is created.  Does that make sense?

jcansdale:
Note to self, might need to use something like this:

```
        [Export]
        public class RootPart
        {
            [Import, SharingBoundary("SomeBoundary")]
            public ExportFactory<BoundaryPart> Factory { get; set; }
        }

      [Export, Shared("SomeBoundary")]
```

src/GitHub.Exports/Exports/ExportMetadata.cs
grokys: Yes, that makes sense, need to check that this is actually happening!

PR #1373, opened 2017-12-11
Single Sign On Support.

### LINKED ISSUE #977: Support authentication via webflow
We only support authentication via oauth currently, but a lot of enterprise signon system require webflow (okta, etc). We should probably support webflow.

This requires us to have a browser window to handle the login, and the webbrowser in VS2015 is horribly out of date, so we should probably look into alternatives.

### LINKED ISSUE #1323: User images not appearing next to viewed Pull Requests
- GitHub Extension for Visual Studio  version: 2.3.5.277
- Visual Studio version: Visual Studio Pro 2017 15.4.4

__What happened__ (with steps, logs and screenshots, if possible):

I noticed that the "profile" images appearing next to all the pull requests I see from our repo only ever shows the default image and not what I can see when on GitHub and viewing pull requests in our browser. The attached screen shot is that I see in github in Visual studio for all my PR's for example, yet on GitHub I do have a profile image that does appear...

githubextensionimages

grokys:
This PR implements SSO/OAuth support for both GitHub.com and Enterprise instances.

## DotCom

When the user goes to login, the GitHub tab now shows an option to log in with a browser:

image

When clicked the user's system browser will be opened, and the user logged in.

## Enterprise

Note: Enterprise support will not be available until version 2.12.1. To enable support for SSO on enterprise, we first need to check the entered URL to see if OAuth/Username/Password is supported. This results in a slightly different flow for enterprise:

Here you can see that because 2.12.1 isn't yet out, we're not enabling SSO, but we _do_ detect that username/password login isn't supported and instead prompt the user to enter a token. We'll only be able to test SSO support on Enterprise when 2.12.1 is out.

Fixes #1323 
Fixes #977


meaghanlewis:
This is working great @grokys 🎉 just a couple notes:

1. When I try to sign in with an enterprise account using the server address and a token I see: **You need to sign out and back in.** I tried to remove any existing ghe credentials in the credential manager, but I continue to get this error.
screen shot 2017-12-12 at 8 26 35 pm


2. After I sign in with the browser, the browser window stays open. I think this is tangential to Jamie's point, but is it possible to close this after connecting?

grokys:
> When I try to sign in with an enterprise account using the server address and a token I see: You need to sign out and back in. 

Ah, what this is trying to tell you is that the scopes on the token are wrong. When using SSO or a password, we're responsible for requesting scopes so that error makes sense. When using a token, it doesn't make sense. I think we need to tweak the error message there.

> After I sign in with the browser, the browser window stays open. I think this is tangential to Jamie's point, but is it possible to close this after connecting?

I don't think it's possible for a [browser tab to close itself if it wasn't opened with js](https://stackoverflow.com/questions/2076299/how-to-close-current-tab-in-a-browser-window). Desktop just leaves the blank tab open, so I assumed that'd be good enough for us too, however I think we could render a simple page there if we decide that's a good idea. Should probably be a separate PR though.

grokys:
@meaghanlewis I've changed the message that is displayed when the token has incorrect scopes to be more useful:

```
Incorrect API scopes. Required: user,repo,gist,write:public_key
```

This message should actually only ever displayed when logging in with a token so I think this will be more useful.

meaghanlewis:
thanks @grokys for updating that message displayed during Enterprise sign in and for the explanation about not being able to close a browser tab. I agree it should be a separate PR.

meaghanlewis:
@grokys im getting the updated message, and am able to sign in with Enterprise using my token with updates scopes 👍 
screen shot 2017-12-18 at 10 50 50 am


One thing to note- I modified my token to have more scopes than I needed. When I selected **admin:public_key**, rather than the write and read options separately I still got the scopes error which seemed strange to me. 
screen shot 2017-12-18 at 10 58 46 am
And saw this in the log: `Incorrect API scopes: require ["user","repo","gist","write:public_key"] but got ["admin:public_key","gist","notifications","repo","user"]`.

Can that top-level scope **admin:public_key** be accepted as well?

grokys:
@meaghanlewis good find! `admin:public_key` should now be accepted.

grokys:
@jcansdale yeah that would probably be a good idea, it's been something I've been thinking as well. We could have an "information" icon that shows the information, or takes you to a page describing the needed scopes.

However that's not the focus of this PR, so we should probably address it separately.

meaghanlewis:
In the latest build I'm having problems signing in with enterprise. After entering the server address, the token field doesn't appear.

Clicking `Sign in with your browser` opens up visualstudio.github.com

screen shot 2017-12-20 at 1 25 10 pm

grokys:
@meaghanlewis it looks like ghe.io has been updated to 2.12.1 but my PR to set the callback URL hasn't been included. I'll get in touch with the enterprise folks.

grokys:
The "or" text probably shouldn't appear when there's only one choice:

image

Or should we still allow the user to enter a token as well?

meaghanlewis:
Good point @grokys. Perhaps allow the user to enter the token as well and not force them to use SSO similarly to how sign in with GitHub offer both options.

donokuda:
It looks like I have to specify the protocol being used (e.g., `https://`) in order to kick off this flow. If I don't, nothing happens as a result, and it feels like something is broken even though it's waiting for me to input a valid URL.

Would it makes sense to turn this into a multi-step process where the user has to explicitly submit information for each step? That way, if the protocol is omitted, then we could still try to use what's supplied.

> The "or" text probably shouldn't appear when there's only one choice:

👍 Agree. We could consider replacing the "Sign in" button with a "Sign in with your browser" submit button.

sso

For 2-1, we might want to consider guiding the user on where to find the token to log in.

donokuda:
Just a heads up, there's another login modal when you try to push to a repo on GitHub and you're not logged in:

grokys:
@donokuda:

> It looks like I have to specify the protocol being used (e.g., https://) in order to kick off this flow. 

I've tweaked the logic to not need `https://`, hopefully that's better.

> We could consider replacing the "Sign in" button with a "Sign in with your browser" submit button.

There won't actually ever only be one choice now: there's always going to be either user/pass or token sign in available for enterprise.

> Just a heads up, there's another login modal when you try to push to a repo on GitHub and you're not logged in:

Yeah, this isn't actually us... Not sure what we can do about this. Might be worth a separate issue.

meaghanlewis:
@grokys nice job getting this working with enterprise it's working great! 🎉 

Just one thing to mention. When I sign in incorrectly using `https://ghe.io`, I see bad credentials displayed in the dialog, and in the log I get the error:
`(2018-01-11 16:48:14.951 INFO [01] LoginTabViewModel         Error logging into '"https://ghe.io/"' as ''
Octokit.AuthorizationException: Bad credentials)`

Signing in to enterprise incorrectly using just `ghe.io`doesn't give an error in the dialog or an error in the log. This is unexpected.

grokys:
@meaghanlewis good catch! That should now be fixed.

meaghanlewis:
thanks @grokys this looks great to me 👍

shana:
This needs to be tested against a GitHub for Business team with different authentication settings - GitHub for Business is hosted by us but allows enterprises to do their own authentication, including SSO, but I'm not sure that requires them to have a separate domain for the org/repos - which means that our assumption that github.com logins are always oauth may be wrong and we may be locking out GitHub for Business users.

shana:
`LoginManager` is currently being created as a service, but it doesn't have to be, since it's not required during package initialization. The 2fa handler setup that's being done in https://github.com/github/VisualStudio/pull/1373/files#diff-0cd7c470cbea5722f66d871d121f1d3bR221 should be happening in the `LoginManager` constructor so that instances initialized via MEF are valid (as it stands currently, they're not).

I don't think `ITwoFactorChallengeHandler` needs to be created lazily, it's a MEF shared instance and the login code sets the viewmodel accordingly when it needs to. We can just import it in the constructor like we do every other dependency so MEF creates it, and we don't have to worry about threading nor have special initialization code for this.

PR #1375, opened 2017-12-12
Make deleted file visible on Dark theme.

### LINKED ISSUE #1349: Highlight/selected color in Changed Files list makes entry completely unreadable
Screenshot says it best:

image

grokys:
Only change the color of a deleted file's icon and text foreground to a subdued gray when the item is not selected, otherwise it becomes unreadable in the Dark theme.

Before:

image

After:

image

Fixes #1349


meaghanlewis:
LGTM 👍

PR #1379, opened 2017-12-12
Fix Diff Display for Moved Files.

### LINKED ISSUE #1364: Left side of PR changes list not visible
- GitHub Extension for Visual Studio  version:
- Visual Studio version: 2015

When there are added files, the left side of the changes list isn't visible. This truncates important context information (where the files are).

- Open PR 1346 in detail view
- Scroll down to see the following:
image

Because there is no horizontal scroll bar, there is no easy way to see the hidden parts of the tree. This makes the PR quite cumbersome to review.

### LINKED ISSUE #1365: Renamed PR files aren't compared with previous version
- Visual Studio version: 2015

1. Opened a PR that contains renamed files (PR 1346)
2. Selected `View Changes` on renamed file
3. The LHS file in diff view is blank:
image

- Expected renamed file to be compared with original (like on github.com):
image
- Expected renamed file to be prefixed with `[rename ...]` (to be consistent with `[add]`):
image

This has made reviewing PR 1346 tricky because I can't see what's actually new in the renamed files.

### LINKED ISSUE #1383: Paths are the same when diffing moved PR file

1. Double-clicked on moved file in PR details view
2. The base and head paths appear to be the same

image

You can repro this with PR 1346:

image

grokys:
- Store the old path in `PullRequestFileNode`
- Use this path in `ExtractFile` for moved files

Fixes #1365 
Fixes #1383

### DIFF
--- a/src/GitHub.App/SampleData/PullRequestDetailViewModelDesigner.cs
+++ b/src/GitHub.App/SampleData/PullRequestDetailViewModelDesigner.cs
@@ -61,7 +61,7 @@ This requires that errors be propagated from the viewmodel to the view and from 
             var repositoriesDir = new PullRequestDirectoryNode("Repositories");
             var itrackingBranch = new PullRequestFileNode(repoPath, @"GitHub\Models\ITrackingBranch.cs", "abc", PullRequestFileStatus.Modified, null);
             var oldBranchModel = new PullRequestFileNode(repoPath, @"GitHub\Models\OldBranchModel.cs", "abc", PullRequestFileStatus.Removed, null);
-            var concurrentRepositoryConnection = new PullRequestFileNode(repoPath, @"GitHub\Repositories\ConcurrentRepositoryConnection.cs", "abc", PullRequestFileStatus.Added, "add");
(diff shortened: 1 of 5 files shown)


grokys:
@jcansdale  are you saying that moved files are now displayed differently? This PR shouldn't have changed the way moved files are displayed.

jcansdale:
@grokys,

> are you saying that moved files are now displayed differently?

No, just that it doesn't actually fix the issue I was having in #1364. The problem is that it automatically pans right because the tree items are so wide. This makes reviewing a PR with moved files a pain (I ended up reviewing on GitHub instead).

An easy fix would be to replace the previous path with `[moved]` I don't think having the path visible on the tree item is really useful. If the user really wants this info, it is surfaced on the diff view tooltip:

image

(it looks like the one of the paths might be wrong though?)

grokys:
I misread the issue and thought that the problem was that the left hand side of the _diff_ of these files was empty (as it was). Looks like this PR fixes a different issue.

grokys:
Ok, this PR fixes #1365, not #1364 - updated the description. With that in mind, could you review again? :)

PR #1388, opened 2017-12-13
Make sure our assemblies are lazily loaded

### LINKED ISSUE #1387: All assemblies being loaded on startup.
On current `master`, all of our assemblies are being loaded on startup. This isn't good - they should be loaded only when needed.

grokys:
This fixes our lazy loading of assemblies. Previously all assemblies were loaded on VS startup (with Team Explorer and GitHub pane not shown):

image

Following this PR the list now looks like:

image

Note that this only fixes things to the state they were before. ReactiveUI is still used by `GitHubConnectSection` which in VS2017 is the default view in team explorer, meaning that RxUI gets loaded as soon as Team Explorer loads in VS2017. We should fix this, but it's a different PR I think.

Fixes #1387

### DIFF
--- a/src/GitHub.Api/LoginManager.cs
+++ b/src/GitHub.Api/LoginManager.cs
@@ -14,7 +14,7 @@ namespace GitHub.Api
     {
         readonly string[] scopes = { "user", "repo", "gist", "write:public_key" };
         readonly IKeychain keychain;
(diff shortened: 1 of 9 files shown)


grokys:
This has broken logging in because we're now trying to get the `ITwoFactorChallengeHandler` service on a background thread.

meaghanlewis:
LGTM too

Issue #848, opened 2017-02-10
What git commands should we execute when checking out a PR


jcansdale:
## This issue is to discuss what git commands should be executed by default when checking out a PR.

At the moment when we checkout a PR, we don't update any submodules that might be in the repo.

### Issues with the current situation

- If the SHA of a contained submodule has changed the user won't be able to checkout another PR.
- If any user files are created in a submodule that aren't in `.gitignore` the user won't be able to checkout another PR (this happens a lot with NCrunch).

### What commands might we want to run

- git submodule update
- git submodule init
- git submodule sync
- git clean (in each submodule)

`git submodule update`

This would fix the most common issue when checking out a PR where there have been commits to a contained submodule.

`git submodule init`

If a submodule has been added?

`git submodule sync`

If a submodule has been moved?

`git clean` (in each submodule)

If NCrunch or a build has left user config or build artifacts in a submodule.

`git checkout`

Maybe we should allow the user to ignore untracked files (if `git clean` was deemed too dangerous).
(this has been implemented in #1302)

Hopefully the resident Git gurus will chime in with pros/cons and other ideas! 😄

See also #826

grokys:
From what I understand, GHfM does this, but GHfW doesn't. @joshaber @shiftkey @Haacked, do you remember why you decided not to do this in GHfW? Was it because of problems in GHfM or something else?

shiftkey:
@grokys the overhead of time spent updating submodules in the background is one of the reasons why you should be conservative when updating things.

 - `git submodule init`

In theory, you can skip this step by passing `--init` to `git submodule update` and save yourself a command. HOWEVER, I had a look at some of our CI jobs but they do `git submodule init` first, because, and I quote the [docs](https://git-scm.com/docs/git-submodule):

>  This command does not alter existing information in .git/config. 

Why does this matter?

 - `git submodule sync` 

Git stores the submodule remote information in two places - `.gitmodules` and `.git/config`. As soon as these settings change, you'll need to `sync` to ensure they're up to date. Maybe you can compare the `.gitmodules` blob SHA when switching between commits to see if this call is necessary, or maybe it's easier to just let Git do this.

And so, once you have the correct remote information setup for each submodule, you can now update each submodule in the working tree.

 - `git submodule update`

 :thumbsup: definitely needed, but don't forget to include a `--recursive` here in case your submodules also have submodules, dawg.

So yes, I think all three steps are essential, I'd just do them in this order after a checkout:

 - `git submodule init`
[...]

PR #1392, opened 2017-12-14
[Feature] Reviewing PR with submodule changes

### LINKED ISSUE #826: Submodules cause all kinds of issues when checking out a PR

jcansdale:
### Functionality

- Option to sync submodules after checking out a PR with submodule changes
- Allows user to check out a different PR branch when there are submodule change
- Usage tracking using `NumberOfSyncSubmodules` counter
- When when the `Sync` button is pressed, the following GIt commands will be executed (see github/VisualStudio#848):

```
git submodule init
git submodule sync --recursive
git submodule update --recursive
```

image

### To do

- Write test plan for added, removed and changed submodules (paging @meaghanlewis)

### Merged PRs

- Reviewing PR with submodule changes MVP: #1371 

#### Syncing submodules

1. Check out a PR
2. Check out another PR with submodule changes (I've been using PR 614 for testing)
3. Click the the `[SyncSubmodules]` button
4. Once sync submodules completes, the branch should appear `up to date`

#### Ignoring submodules

1. Check out a PR
2. Check out another PR with submodule changes (I've been using PR 614 for testing)
3. Ignore the `[SyncSubmodules]` button and check out original PR
4. Original branch should appear `up to date` (previously this wouldn't have worked)

#### Git.exe not on PATH

1. Remove Git.exe from `PATH`.
2. Check out a PR
3. Check out another PR with submodule changes (I've been using PR 614 for testing)
4. Click the the `[SyncSubmodules]` button
5. Something like this should appear:
image

### Related issues

- Submodules cause all kinds of issues when checking out a PR: github/VisualStudio#826
- What git commands should we execute when checking out a PR: github/VisualStudio#848
- Workflow: Reviewing PRs that contain submodule changes: github/editor-tools#227
- Zoom meeting - Reviewing a PR branch workflow: github/editor-tools#230

Fixes #826


jcansdale:
Deciding what to do when updating submodules and `git.exe` isn't on the users path.

Visual Studio 2018 will show the following if `Git for Windows` isn't installed:

image

image

The `Install` buttons simply opens the following in the users default browser:
https://git-scm.com/download/win

The current commit will show the following (it will show `git` not `gitx`):

image

I'm looking to make the message a little friendlier. The user will probably already have Git installed, so there is no need to go overboard with a solution!

Something along the lines of:
```
Couldn't find Git.exe on PATH.

Please install Git for Windows from:
https://git-scm.com/download/win
```

donokuda:
I pushed a couple of commits that clean up the sync submodules number. Here's what it looks like (with some stubbed info):

screen shot 2018-01-21 at 9 28 19 am

For the sake of space, I kept the action to "Sync." Could we use a tooltip that says something along the lines of "Sync [number] submodules?" I tried adding one myself, but couldn't figure out how to do it in the time I allotted for myself. (The GitHubActionLink component is based of a regular xaml button component if that's any help.) 

> I'm looking to make the message a little friendlier. The user will probably already have Git installed, so there is no need to go overboard with a solution!

@jcansdale I forgot if you mentioned this earlier, but we should move any errors from an action into an infobar.

jcansdale:
@donokuda, Looking sweet! 😄 

I've added a tooltip like this:

image

donokuda:
I checked this out myself and it looks good to me ✨ 

Thanks a bunch, @jcansdale! 🙇

grokys:
Just tried this on #1415 and I'm getting:

image

Which is strange because `submodules/Octokit.GraphQL` doesn't exist on this branch. Doing a `git submodule update` from the command line works fine however.

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -154,6 +186,68 @@ public IObservable<Unit> Push(ILocalRepositoryModel repository)
...
+            {
+                var output = new StringWriter(CultureInfo.InvariantCulture);
+                var exitCode = await SyncSubmodules(repository.LocalPath, line => output.WriteLine(line));
+                if (exitCode != 0)
+                {
+                    // Replace with friendly message if Git.exe isn't on path.
+                    // If culture isn't English, user will see the local equivalent of:
+                    // "'git' is not recognized as an internal or external command"                    
+                    var message = output.ToString();
+                    if (exitCode == 1 && message.StartsWith("'git' is not recognized as an internal or external command,", StringComparison.Ordinal))
```
grokys: This strikes me as a bit of a hack. Might there be a way to check whether `git.exe` is on the path and display an error rather than parsing the (localizable) error message?

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -154,6 +186,68 @@ public IObservable<Unit> Push(ILocalRepositoryModel repository)
...
+                    {
+                        message =
```
grokys: This should be in a resources file.

jcansdale:
@grokys,

> Just tried this on #1415 and I'm getting...

That was a bit of an edge case, but I'm glad you found it. I've tweaked it so that it won't give up quite so easily and will always do `git submodule update`.

src/GitHub.App/Services/PullRequestService.cs
jcansdale: I can change it to use WHERE to find out where `Git.exe` is on the path.

grokys:
A few other things I've noticed 

- When you click the "Sync" button it can take quite a while to update the submodule, in which time there's no indication that anything is happening except for the fact you can't click on anything. This is also a problem for the "Push" and "Pull" buttons but as these operations are usually quicker you don't tend to notice. We should probably address this for all 3 operations in a separate PR
- When there are submodules to sync, the "Push" and "Pull" buttons are always visible (though disabled) even if there is nothing to push and pull:

image

Is this correct or should we just be showing the "Sync" button in this state?

jcansdale:
> When you click the "Sync" button it can take quite a while to update the submodule, in which time there's no indication that anything is happening except for the fact you can't click on anything.

Yes, I've noticed this as well. I'm hoping to add some incremental progress to give the user some idea of what's going on.

> When there are submodules to sync, the "Push" and "Pull" buttons are always visible (though disabled) even if there is nothing to push and pull. Is this correct or should we just be showing the "Sync" button in this state?

This was intentional. The idea is that "Sync" is a peer of "Push" and "Pull", but only appears when there are submodules to sync (no point cluttering the interface for no reason).

jcansdale:
@grokys,

> This is also a problem for the "Push" and "Pull" buttons but as these operations are usually quicker you don't tend to notice. We should probably address this for all 3 operations in a separate PR

I've made some changes should should make reporting incremental progress easier. I agree we should probably address it in a separate PR and keep this one focused.

The issue that occurred when checking out #1415 has now been addressed. I checked it before updating and it successfully completed a `git submodule update`.

It now checks that Git.exe exists using WHERE and displays an error message (form a resource string) and aborts if it doesn't.

I think that's everything. 🤞

jcansdale:
I intended to hide all sync related UI when there are no submodules to sync. It appears this isn't the case. Oops!

image

jcansdale:
Here we go. Less clutter when there are no submodule changes:

image

meaghanlewis:
This is looking good to me 🎉 

I also added some test scenarios here: https://github.com/github/editor-tools/pull/268 can you take a 👀  @jcansdale?

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -154,6 +186,67 @@ public IObservable<Unit> Push(ILocalRepositoryModel repository)
...
+            var cmdArguments = "/C WHERE /Q " + fileName;
+            var startInfo = new ProcessStartInfo("cmd", cmdArguments)
+            {
+                UseShellExecute = false,
+                CreateNoWindow = true
+            };
+
+            using (var process = Process.Start(startInfo))
+            {
+                process.WaitForExit();
```
grokys: This will block the UI thread waiting for the process to exit, probably not a good idea...

src/GitHub.App/ViewModels/GitHubPane/PullRequestDetailViewModel.cs
```diff
@@ -659,6 +676,26 @@ IObservable<Unit> DoPush(object unused)
...
+        {
+            usageTracker.IncrementCounter(x => x.NumberOfSyncSubmodules).Forget();
+
+            var progress = new CaptureProgress();
+            var complete = await pullRequestsService.SyncSubmodules(LocalRepository, progress);
+            if (!complete)
+            {
+                throw new ApplicationException(progress.ToString());
+            }
+        }
```
grokys: Actually a low-impact way for now to show that something is happening during this method would be to do:

```
        async Task DoSyncSubmodules(object unused)
        {
            try
            {
                IsBusy = true;
                usageTracker.IncrementCounter(x => x.NumberOfSyncSubmodules).Forget();

                var progress = new CaptureProgress();
                var complete = await pullRequestsService.SyncSubmodules(LocalRepository, progress);
                if (!complete)
                {
                    throw new ApplicationException(progress.ToString());
                }
            }
            finally
            {
                IsBusy = false;
            }
        }
```

What do you think?

jcansdale:
I've stopped it from blocking the main thread and changed it to use the busy spinner.

I did a spike where it where it shows progress on the status bar, which works very well but isn't ready for prime time. I'll address this in a separate PR so as not to block this one!

What do you reckon?

PR #1106, opened 2017-08-03
Highlight checked out PR in list.

grokys:
This PR highlights the currently checked out PR in the PR list.

Blue:

devenv_2018-02-21_16-00-37

Dark:

Light:

## Not yet implemented

To link a branch to a PR we use a property set in the git `.config` file - currently this is only set when a branch is a) checked out in the PR details view b) the PR details are opened for the current branch's PR.

We can actually do better here: when we load the PR list we have enough information to mark all PR branches with their related PR but there are a couple of things preventing us from doing this:

1. ~~`TrackingCollection.OriginalCompleted` is fired in the PR list when the list starts loading for the first time, not when the load completes. This works correctly on subsequent refreshes - appears to be a bug in `TrackingCollection`~~ - now fixed
2. `TrackingCollection.original` would need to be exposed so we can get an unfiltered list of all PRs
3. ~~Navigating to the PR list causes a reload of all PRs (#1105) which means we'd be doing this way more often than we should; as the operation involves opening the repository and reading/writing config values this may not be a good idea. This appears to be caused by the fact that in the navigation controller there is no distinction between loading a page for the first time and moving to an already loaded page.~~ - now fixed

I think this should probably be implemented in another PR, so bear in mind that the current PR may not be highlighted *until* you open its details.

Issue #43, opened 2015-08-11
Verify that the app behaves nicely when using an unverified email


shana:
[Soon](https://github.com/github/github/pull/44599#issuecomment-129597888), some API actions will require that the user have a verified email, and error when they don't: https://github.com/github/github/pull/44098

We need to ensure the app behaves nicely in those cases.

shana:
Behaviour is appropriate, for now

PR #1396, opened 2017-12-18
[Feature] Show current PR on status bar

jcansdale:
The following PRs target this one:

- Expose IVSGitExt as async VS service: #1510 (extends closed PR #1506)
- Use custom UIContext to load GitHubPackage / PullRequestStatusBarPackage in repo context: #1512
- Simplify the VSGitExt service: #1513

### Functionality

- Show PR associated with current branch on status bar
- Click on PR to navigate to PR details

image

- Tooltip shows PR number and title

image

### To do

- [x] Make UI consistent with other VS status bar buttons (paging @donokuda)
- [x] Fix issue with PR number/title not updating
- [x] Add metrics for navigate to PR command
- [x] Make sure this feature isn't slowing startup
  - [x] Async load package
  - [x] Load package on background thread
  - [x] Load on `[ProvideAutoLoad(Guids.GitSccProviderId)]`?
  - [ ] ~~Use PR number from branch metadata?~~
  - [x] Lazy load `IPullRequestSessionManager`, see  https://github.com/github/VisualStudio/pull/1388/files#diff-8f8b4cd719001e4a8dd4b96e211243a8R53

### Related issues

- Show PR branch on status bar and allow navigation to PR details: #1102
- Highlight checked out PR in list: #1106


jcansdale:
@donokuda I've updated it to surface the number of submodules to sync. It currently looks like this:

image

meaghanlewis:
I've been dogfooding this for a couple days now. It's working well and allows me to remember what PR I'm looking at and to easily navigate back to that PR window.

One thing I find interesting is that when I checkout master, the PR I was just looking at still shows on the status bar and I didn't expect to see it there any longer.

shana:
> One thing I find interesting is that when I checkout master, the PR I was just looking at still shows on the status bar and I didn't expect to see it there any longer.

@jcansdale This feels like a bug?

jcansdale:
@shana,

> This feels like a bug?

Yes this is a bug. It's being tracked by #1440.

jcansdale:
@meaghanlewis You should find this (leaving a PR branch) fixed now. I've merged in #1442 to this PR.

meaghanlewis:
@jcansdale this is looking good to me now 👍

src/GitHub.InlineReviews/PullRequestStatusPackage.cs
```diff
@@ -0,0 +1,23 @@
...
+using Microsoft.VisualStudio.ComponentModelHost;
+using GitHub.VisualStudio;
+using GitHub.InlineReviews.Services;
+
+namespace GitHub.InlineReviews
+{
+    [PackageRegistration(UseManagedResourcesOnly = true)]
+    [Guid(Guids.PullRequestStatusPackageId)]
+    [ProvideAutoLoad(Guids.GitSccProviderId)]
+    public class PullRequestStatusPackage : Package
```
jcansdale: This shold be an aync package.

grokys:
The tooltip as it is could be slightly confusing:

image

This could be taken to mean "clicking this will rename the project dir". If it showed "#43 - Rename project dir" it might be clearer?

jcansdale:
@grokys Yup, I think that works.

image

What do you reckon?

donokuda:
Just driving by to say that I'm 👍 for adding the pull request number. I'm excited for this to land too! 😄

src/GitHub.InlineReviews/PullRequestStatusBarPackage.cs
```diff
@@ -0,0 +1,29 @@
...
+    public class PullRequestStatusBarPackage : AsyncPackage
+    {
+        /// <summary>
+        /// Initialize the PR status UI on Visual Studio's status bar.
+        /// </summary>
+        protected override async Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
+        {
+            var componentModel = (IComponentModel)await GetServiceAsync(typeof(SComponentModel));
+            var exportProvider = componentModel.DefaultExportProvider;
+            var pullRequestStatusManager = exportProvider.GetExportedValue<IPullRequestStatusBarManager>();
```
shana: MEF calls need to happen on the main thread, and there's no guarantee that this method is going to be called on the main thread, especially once the new auto-load changes come (https://docs.microsoft.com/en-us/visualstudio/extensibility/how-to-use-asyncpackage-to-load-vspackages-in-the-background). For things that are required during package loading, they should be registered as services instead to avoid having to hit MEF on startup (see GitHubPackage.cs)

src/GitHub.InlineReviews/PullRequestStatusBarPackage.cs
```diff
@@ -0,0 +1,29 @@
...
+using GitHub.VisualStudio;
+using GitHub.InlineReviews.Services;
+using Task = System.Threading.Tasks.Task;
+
+namespace GitHub.InlineReviews
+{
+    [PackageRegistration(UseManagedResourcesOnly = true)]
+    [Guid(Guids.PullRequestStatusPackageId)]
+    [ProvideAutoLoad(Guids.GitSccProviderId)]
+    public class PullRequestStatusBarPackage : AsyncPackage
```
grokys: Does this need to be in a separate package? For context, I've been thinking recently that we should be merging `GitHub.InlineReviews` into our main package if possible, as it's completely dependent on it. Could this go in the main package? Would there be problems with that?

grokys:
Small thing, but I've just noticed that the color for this PR icon is different to the colors for the rest of the git status bar icons:

image

donokuda:
> Small thing, but I've just noticed that the color for this PR icon is different to the colors for the rest of the git status bar icons:

mmm, I think I'm gonna have to snoop around for the correct visual studio color for that icon. 🤔 🔍

jcansdale:
@shana,

> MEF calls need to happen on the main thread, and there's no guarantee that this method is going to be called on the main thread, especially once the new auto-load changes come (https://docs.microsoft.com/en-us/visualstudio/extensibility/how-to-use-asyncpackage-to-load-vspackages-in-the-background). For things that are required during package loading, they should be registered as services instead to avoid having to hit MEF on startup (see GitHubPackage.cs)

I've converted the package to load on a background thread. It now uses the following attributes.

```
    [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
    [ProvideAutoLoad(Guids.GitSccProviderId, PackageAutoLoadFlags.BackgroundLoad)]
```

I've fixed the `VSGitExt` and `PullRequestStatusBarManager` services so that they can be loaded on the background thread. The other services that is depends on will be loaded just in time on the Main thread.

I started converting them to be VS services, but `VSGitExt` was problematic because it's one of those `GitHub.TeamFoundation` MEF services with two implementations. I've updated `VSGitExt` so that it can be loaded and used from a background thread. Visual Studio's `SComponentModel` is also now an async service.

src/GitHub.InlineReviews/PullRequestStatusBarPackage.cs
jcansdale: When I first wrote this functionality, I made it piggyback a ride with one of the other packages. Keeping it separate can be useful to control exactly when and how it loads.

For example, I wanted this package load when the `GitSccProvider` is available and with the option `AllowsBackgroundLoading = true`. The package it was being loaded with before was using different settings (being loaded when a solution was opened).

jcansdale:
Still needs more work before `AllowsBackgroundLoading = true` works 100%. 😢

src/GitHub.InlineReviews/PullRequestStatusBarPackage.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        /// <summary>
+        /// Initialize the PR status UI on Visual Studio's status bar.
+        /// </summary>
+        protected override async Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
+        {
+            var usageTracker = (IUsageTracker)await GetServiceAsync(typeof(IUsageTracker));
+            var serviceProvider = (IGitHubServiceProvider)await GetServiceAsync(typeof(IGitHubServiceProvider));
+            var componentModel = (IComponentModel)await GetServiceAsync(typeof(SComponentModel));
+            var pullRequestSessionManager = componentModel.DefaultExportProvider.GetExport<IPullRequestSessionManager>();
```
shana: MEF calls need to happen on the main thread.

Also, it would be better if everything goes through `GitHubServiceProvider`. This way we can mock all these types (and even bypass MEF entirely if we want to) without the caller having to change any calls. `GitHubServiceProvider` already does look ups via ComponentModel, and the MEF lookups ensure that the results are loaded from our DLLS and not any other type that might be overriding our stuff, so that's a plus.

Ideally, though, we don't use MEF in package initialization, because loading the catalog for the first time is really expensive. Since you need to grab `IVSGitExt` here, I suggest you expose it as a service in the `ServiceProviderPackage` class like we do the other services like metrics, that way you'll get an instance directly with no need for MEF scanning everything.

shana: Of course, if `IPullRequestSessionManager` has too many dependencies to be exposed as a service, we can go through MEF, but should always do it via the `GitHubServiceProvider` anyway.

src/GitHub.InlineReviews/PullRequestStatusBarPackage.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        /// <summary>
+        /// Initialize the PR status UI on Visual Studio's status bar.
+        /// </summary>
+        protected override async Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
+        {
+            var usageTracker = (IUsageTracker)await GetServiceAsync(typeof(IUsageTracker));
+            var serviceProvider = (IGitHubServiceProvider)await GetServiceAsync(typeof(IGitHubServiceProvider));
+
+            await ThreadingHelper.SwitchToMainThreadAsync();
```
jcansdale: There seems to be an issue where this await doesn't return until a solution is loaded. This means that when Visual Studio is opened without a solution, the PR list might be visible but the current PR doesn't appear on the status bar.

This isn't necessarily a problem, but I'd like to understand what's going on. I wonder if this will impact any of our other async loading packages? We wouldn't notice the issue on packages that use `[ProvideAutoLoad(UIContextGuids80.SolutionExists)]` (because they only load when a solution exists).

src/GitHub.InlineReviews/PullRequestStatusBarPackage.cs
jcansdale: I've fixed it so that everything goes through `GitHubServiceProvider`. No more explicit dependency on `IComponentModel`.

shana:
> I agree it would be best to expose IVSGitExt as a service. I'm inclined to do this as a separate PR though. What do you reckon?

I'd rather not introduce MEF loading code in the package if we're just throwing it away later. Maybe do a separate PR targeting this branch with that change?

src/GitHub.InlineReviews/PullRequestStatusBarPackage.cs
shana: The `GitHubPackage` already loads at the correct time in the correct way, btw.

src/GitHub.InlineReviews/PullRequestStatusBarPackage.cs
shana: Yeah, this is yet another reason why we shouldn't be switching to the main thread for initialization whenever we can avoid it. If all we need are non-MEF services, we shouldn't need to switch to the main thread for any of this.

StanleyGoldman:
image

Playing with this locally, looks good.

src/GitHub.InlineReviews/Services/PullRequestStatusBarManager.cs
```diff
@@ -0,0 +1,169 @@
...
+        IPullRequestSessionManager pullRequestSessionManager;
+
+        [ImportingConstructor]
+        public PullRequestStatusBarManager(IVSGitExt gitExt, IUsageTracker usageTracker, IGitHubServiceProvider serviceProvider)
+        {
+            this.gitExt = gitExt;
+            this.usageTracker = usageTracker;
+            this.serviceProvider = serviceProvider;
+
+            OnActiveRepositoriesChanged();
```
jcansdale: This might be called twice!

src/GitHub.InlineReviews/PullRequestStatusBarPackage.cs
jcansdale: I've changed `IVSGitExt` to be a VS services with no MEF dependencies.

src/GitHub.InlineReviews/PullRequestStatusBarPackage.cs
jcansdale: BTW, the `GitHubPackage` package uses `AllowsBackgroundLoading = true`, but we'd forgotten to use the `PackageAutoLoadFlags.BackgroundLoad` option on `[ProvideAutoLoad(...)]`. This means it uses a background thread when an executing command triggers the package load but the Main thread when the package is auto-loaded. Subtle and confusing! 😕 I've addresses this issue in #1512.

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
```diff
@@ -8,91 +9,67 @@
...
         readonly object refreshLock = new object();
 
         IGitExt gitService;
         IReadOnlyList<ILocalRepositoryModel> activeRepositories;
 
         [ImportingConstructor]
-        public VSGitExt(IGitHubServiceProvider serviceProvider)
-            : this(serviceProvider, new VSUIContextFactory(), new LocalRepositoryModelFactory())
+        public VSGitExt(Func<Type, Task<object>> getServiceAsync)
+            : this(getServiceAsync, new VSUIContextFactory(), new LocalRepositoryModelFactory())
```
shana: Any reason to send a callback instead of sending in `IAsyncServiceProvider` (which provides `GetServiceAsync` here?

jcansdale: The issue was that `AsyncPackage` implements 3 interfaces called `IAsyncServiceProvider`, 2 of which exist in the same namespace. Passing a delegate way a way to avoid taking a dependency on `Microsoft.VisualStudio.Shell.Immutable.14.0` and needing to disambiguate using `extern alias` (which works but is a bit messy).

I've pushed a commit that uses the interface from `Microsoft.VisualStudio.Shell.Immutable.14.0`. If we're likely to be using `IAsyncServiceProvider` elsewhere, having the correct assembly to hand is maybe a good thing (even if we need to use `extern alias`).

An alternative approach would be to use the lower level Interop `IAsyncServiceProvider` interface from here:

```
namespace Microsoft.VisualStudio.Shell.Interop
{
    public interface IAsyncServiceProvider
    {
        IVsTask QueryServiceAsync([ComAliasName("OLE.REFGUID")] ref Guid guidService);
    }
}
```

The fact that is takes a `Guid` and returns a `IVsTask` again makes it a bit messy.

My instinct is to go with this last commit. What do you think?

shana:
> The issue was that AsyncPackage implements 3 interfaces called IAsyncServiceProvider, 2 of which exist in the same namespace. Passing a delegate was a way to avoid taking a new dependency on Microsoft.VisualStudio.Shell.Immutable.14.0 and needing to disambiguate using extern alias (which works but is a bit messy).

Which two assemblies provide the same type?

jcansdale:
@shana,

It's a collision between `Microsoft.VisualStudio.Shell.Immutable.14.0` and `Microsoft.VisualStudio.Shell.Framework, Version=15`.

```
1>------ Build started: Project: GitHub.TeamFoundation.15, Configuration: Debug Any CPU ------
1>C:\Source\github.com\github\VisualStudio\src\GitHub.TeamFoundation.14\Services\VSGitExt.cs(15,60,15,81): error CS0433: The type 'IAsyncServiceProvider' exists in both
'Microsoft.VisualStudio.Shell.Framework, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' and 
'Microsoft.VisualStudio.Shell.Immutable.14.0, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
========== Build: 0 succeeded, 1 failed, 20 up-to-date, 0 skipped ==========
```

I'll see what happens if I remove the `Microsoft.VisualStudio.Shell.Framework, Version=15` dependency from `GitHub.TeamFoundation.15`...

jcansdale:
Ah, we're forced to reference `Microsoft.VisualStudio.Shell.Framework, Version=15` because of this line. 😭 

image

This would however be an easy reflection call.

jcansdale:
I've changed it so that rather than giving an alias to the shared `Microsoft.VisualStudio.Shell.Immutable.14` assembly that contains the `IAsyncServiceProvider` implementation that we want, we give an alias to `Microsoft.VisualStudio.Shell.Framework, v15` that we only use once in `VSGitServices`.

This will make finding the correct `IAsyncServiceProvider` a lot easier in future!

src/GitHub.InlineReviews/Services/PullRequestStatusBarManager.cs
jcansdale: This will be fixed in #1512.

PR #1399, opened 2017-12-20
Wrap text on error message in clone dialog

### LINKED ISSUE #1303: Wrap text on error message in clone dialog
__Version__
- GitHub Extension for Visual Studio  version: issue reported in v2.3.4.54
- Visual Studio version: VS2015, VS2017

__What happened__ 
When a user's repositories have not loaded, an error message appears in the clone dialog. The full error message is cut off in the clone dialog. It would be good to wrap the text and make it fully visible.
screen shot 2017-11-06 at 10 20 30 am

__How to reproduce__
This can be reproduced most easily by turning off the wifi and then opening the clone dialog.

meaghanlewis:
Fixes #1303 

Before:
32522126-eee1cd80-c3ca-11e7-91fa-7b5d461c03f2

After:
screen shot 2017-12-20 at 2 52 15 pm

### DIFF
--- a/src/GitHub.VisualStudio/Views/Dialog/RepositoryCloneView.xaml
+++ b/src/GitHub.VisualStudio/Views/Dialog/RepositoryCloneView.xaml
@@ -166,14 +166,14 @@
                 HorizontalAlignment="Center"
                 VerticalAlignment="Center"
                 DockPanel.Dock="Top">
-        <uirx:ErrorMessageDisplay x:Name="loadingFailedMessage"
-                                  Margin="8,0"
-                                  Icon="stop"
-                                  Message="{x:Static prop:Resources.loadingFailedMessageMessage}"
-                                  Visibility="{Binding LoadingFailed, Mode=OneWay, Converter={ui:BooleanToVisibilityConverter}}">
-          <TextBlock Text="{x:Static prop:Resources.loadingFailedMessageContent}" TextWrapping="Wrap" />
-        </uirx:ErrorMessageDisplay>  
-        </StackPanel>
+      <uirx:ErrorMessageDisplay x:Name="loadingFailedMessage"
+                                Margin="8,0"
+                                Icon="stop"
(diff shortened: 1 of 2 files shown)


donokuda:
ooo good catch. I'm unfamiliar with the component being used, but is there a way to vertically align the icon so that it's aligned to the top? Something similar to `VerticalAlignment="Top"`

src/GitHub.VisualStudio/Views/Dialog/RepositoryCloneView.xaml
```diff
@@ -166,13 +166,14 @@
...
-                                Visibility="{Binding LoadingFailed, Mode=OneWay, Converter={ui:BooleanToVisibilityConverter}}" />
-    </StackPanel>
+        <uirx:ErrorMessageDisplay x:Name="loadingFailedMessage"
+                                  Margin="8,0"
+                                  Icon="stop"
+                                  Message="{x:Static prop:Resources.loadingFailedMessageMessage}"
+                                  Visibility="{Binding LoadingFailed, Mode=OneWay, Converter={ui:BooleanToVisibilityConverter}}">
+          <TextBlock Text="{x:Static prop:Resources.loadingFailedMessageContent}" TextWrapping="Wrap" />
+        </uirx:ErrorMessageDisplay>  
+        </StackPanel>
```
grokys: Nit: Looks like the indentation got messed up here - you need to be careful when editing XAML files - some of our XAML files use a tab spacing of 4 and some 2 (and some are mixed, arrgh!) and VS will go ahead and reformat things for you if you have the wrong setting! It's very annoying, we should fix it, but in the meantime probably best to manually fix this indentation.

meaghanlewis: thanks @grokys will update 👍

meaghanlewis:
@grokys is the formatting better now? Also, do you know if that stop icon can be top aligned with the text? I tried a few things, but didn't have success with it and am still a xaml newbie 😄

PR #1359, opened 2017-12-01
Lightweight fix for PR detail view not refreshing

jcansdale:
This PR listens to the `IVSGitExt` service for `ActiveRepositoriesChanged` events and refreshes the model.

I considered extending `ITeamExplorerServiceHolder`, but this code is already quite confusing and I didn't want to complicate it further.

At the moment there is no obvious way to dispose of view models. ~~As a workaround, this code assumes there is only one active `PullRequestDetailViewModel` object and automatically disposes of any previous instances.~~

I spoke with @grokys about how this might be handled in future. A future implementation of the `GitHub` pane might dispose of view models, making the `AutoDispose` workaround redundant.

Fixes #1345.

PR #1409, opened 2018-01-05
Notify PullRequestSessionManager when branch changes

### LINKED ISSUE #1408: PullRequestSessionManager.CurrentSession not updated when changing branches

When I change to a new PR branch, the `PullRequestSessionManager.CurrentSession` property is no longer being updated.

I believe this is due to the change in the behavior of`ITeamExplorerService.Subscribe`. This now only fires when changing to a new repo, not when simply changing branches.

I've tried listening for property changed events on the `localRepositoryModel` object here:
https://github.com/github/VisualStudio/blob/562bac52557fdc4102bff47b286e1480b046b897/src/GitHub.InlineReviews/Services/PullRequestSessionManager.cs#L181

But no properties appear to change.

This is related to: 
- Making TeamExplorerServiceHolder testable: #1292
- Checked out status of PR doesn't update until refresh: https://github.com/github/VisualStudio/issues/1345#issuecomment-347791859
- Lightweight fix for PR detail view not refreshing: #1359

jcansdale:
- The `PullRequestSessionManager` needs to be notified when a branch changes.
- `ITeamExplorerServiceHolder` only fires when moving to a new repo.
- `IVSGitExt.ActiveRepositoriesChanged` also fires when a branch changes.

I've changed `PullRequestSessionManager` to use a new `ITeamExplorerContext` service rather than `ITeamExplorerServiceHolder`. `TeamExplorerContext` simple facade on top of `IVSGitExt` and `ITeamExplorerServiceHolder`. 

It notifies of any change to the active repository using the `StatusChanged` event and exposes an up to date `ILocalRepositoryModel` via `GetActiveRepository`.

Ultimately it would be nice if `ActiveRepository` was a property and the `LocalRepositoryModel` was updated rather than recreated when its status changes.

This PR is simply intended to unblock #1396. I'm reluctant to attempt any more radical refactoring without @shana's input.

- We should be able to revert: https://github.com/github/VisualStudio/pull/1359

Fixes #1408


jcansdale:
Came across this one while testing:

image

`repository.CloneUrl` can be `null`.  Strange. 😕 

https://github.com/github/VisualStudio/blob/71b80458a7662c16f2817e6ad65f52d820889211/src/GitHub.App/Services/PullRequestService.cs#L290

jcansdale:
I seem to be breaking more than I'm fixing with this PR. I'm going to bail and bring in team members how understand `ITeamExplorerServiceHolder` and `PullRequestSessionManager` better!

//cc @shana

PR #1410, opened 2018-01-08
Fix PullRequestDetailView in designer.

grokys:
`PullRequestDetailView` was not displaying in the Visual Studio designer - it was showing an error related to `OcticonImage`.

 Not sure why this works when the previous didn't, but seems to fix it.


grokys:
Actually, this doesn't seem to have solved it: it was working and now it's not :/

PR #1411, opened 2018-01-09
Move GitHub.InlineReviews.UnitTests to nunit

meaghanlewis:
This PR:
- Updates all the Inline Reviews unit tests to use nunit instead of xunit.
- Uses the latest version of nunit -> `3.9.0`
- Changes assertions from classic model (`Assert.Equal`) to constraint based model (`Assert.That(result, Is.EqualTo(x))`

### DIFF
--- a/test/GitHub.InlineReviews.UnitTests/Services/PullRequestSessionManagerTests.cs
+++ b/test/GitHub.InlineReviews.UnitTests/Services/PullRequestSessionManagerTests.cs
@@ -17,7 +17,7 @@ using Microsoft.VisualStudio.Text;
 using Microsoft.VisualStudio.Text.Editor;
 using Microsoft.VisualStudio.Utilities;
 using NSubstitute;
-using Xunit;
+using NUnit.Framework;
 using System.Reactive.Disposables;
 
 namespace GitHub.InlineReviews.UnitTests.Services
@@ -35,7 +35,7 @@ namespace GitHub.InlineReviews.UnitTests.Services
 
         public class TheConstructor : PullRequestSessionManagerTests
         {
-            [Fact]
+            [Test]
             public void ReadsPullRequestFromCorrectFork()
             {
                 var service = CreatePullRequestService();
@@ -59,7 +59,7 @@ namespace GitHub.InlineReviews.UnitTests.Services
 
         public class TheCurrentSessionProperty : PullRequestSessionManagerTests
         {
-            [Fact]
+            [Test]
             public void CreatesSessionForCurrentBranch()
             {
                 var target = new PullRequestSessionManager(
@@ -73,7 +73,7 @@ namespace GitHub.InlineReviews.UnitTests.Services
                 Assert.True(target.CurrentSession.IsCheckedOut);
             }
 
-            [Fact]
+            [Test]
             public void CurrentSessionIsNullIfNoPullRequestForCurrentBranch()
             {
                 var service = CreatePullRequestService();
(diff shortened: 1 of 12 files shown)


grokys:
@meaghanlewis i think the nunit equivalent of e.g.

```
Assert.Single(result);
```

would be:

```
Assert.AreEqual(1, result.Count);
```

meaghanlewis:
Thanks @grokys I'll try this out today 👍

meaghanlewis:
`Assert.AreEqual(1, result.Count);` didn't work for me because result was usually a type that couldn't return a count.
`Assert.That(result, Has.One.Items);` seemed to do the trick.

test/GitHub.InlineReviews.UnitTests/Services/PullRequestSessionManagerTests.cs
```diff
@@ -106,10 +106,10 @@ public void CurrentSessionChangesWhenBranchChanges()
                 service.GetPullRequestForCurrentBranch(null).ReturnsForAnyArgs(Observable.Return(Tuple.Create("foo", 22)));
                 teService.NotifyActiveRepoChanged();
 
-                Assert.NotSame(session, target.CurrentSession);
+                Assert.That(session, Is.Not.SameAs(target.CurrentSession));
```
grokys: I think you can use `Assert.AreNotSame` here, it's slighly less verbose.

test/GitHub.InlineReviews.UnitTests/Services/PullRequestSessionManagerTests.cs
```diff
@@ -252,10 +252,10 @@ public class TheGetLiveFileMethod : PullRequestSessionManagerTests
                     new FakeTeamExplorerServiceHolder(CreateRepositoryModel()));
                 var file = await target.GetLiveFile(FilePath, textView, textView.TextBuffer);
 
-                Assert.Same(diff, file.Diff);
+                Assert.That(diff, Is.SameAs(file.Diff));
```
grokys: And `Assert.AreSame` here.

test/GitHub.InlineReviews.UnitTests/Tags/InlineCommentTaggerTests.cs
```diff
@@ -44,11 +44,11 @@ public void ShouldReturnShowCommentTagForRhs()
                 var firstPass = target.GetTags(span);
                 var result = target.GetTags(span).ToList();
 
-                Assert.Single(result);
-                Assert.IsType<ShowInlineCommentTag>(result[0].Tag);
+                Assert.That(result, Has.One.Items);
+                Assert.That(result[0].Tag, Is.InstanceOf(typeof(ShowInlineCommentTag)));
```
grokys: I think it would be better to use `Is.InstanceOf<ShowInlineCommentTag>()` here.

meaghanlewis:
Thanks for reviewing @grokys! In nunit3, it seems they prefer assertions to use the constraint-based model where `assertions are written primarily using the 'Assert.That' method, which takes constraint objects as an argument.` Versus the classic assert model used for nunit2.

https://github.com/nunit/docs/wiki/Assertions#two-models

I should probably change all assertions to be constraint-based, because right now I think it's a mixed bag.

meaghanlewis:
Alright @grokys updates all assertions to use the new model...even though its more verbose (sorry)😄

grokys:
That's ok @meaghanlewis, if that's the recommendation let's follow it!

meaghanlewis:
@grokys can you give me a ✅  or a ❌  ?

PR #1415, opened 2018-01-11
WIP: Spike of Pull Request Reviews feature.

grokys:
The PR adds an initial spike of the Pull Request Reviews feature.

Existing reviews are displayed in the PR details pane, filtered to the most recent (non-dismissed, non-pending) review for each user:

image

Clicking on the review opens a new pane:

image

Clicking on "Start a review" similarly opens a new pane for authoring a new review:

image

## Known Issues

This is an initial spike of the feature, and as such there are limitations and known issues:

- ~~The GitHub REST API does not support modifying pending reviews. This means:~~
  - ~~If you've already started a review of the PR on GitHub, don't try to do a review on the same PR here. It will break~~
  - ~~We have to keep a record of the line comments for the current review locally. These aren't serialized currently, so closing the solution/VS will mean you lose your comments~~ Fixed: now using GraphQL
  - The "Approve" and "Request Changes" actions are available when reviewing your own PRs. These will fail in this case, you can only comment on your own PRs
- It's not possible to tell which inline comments are part of the review being looked at/authored and which aren't

## Moving forward

- ~~We need to decide whether to continue storing the state of a pending review locally, or move to GraphQL which supports updating a pending review~~ I've gone with GraphQL
- We need to decide how to display inline comments glyphs for comments that are outside the current review

Depends on #1445


jcansdale:
I've been dogfooding this PR the last couple of days.

When returning to a PR, the first thing I want to see is whether there are any new reviews/approvals. With the current layout, you need to scroll down past the PR description to find this information. The concise `Reviewers` section can easily be missed, particularly when there is a long description and many PR file changes (this PR is a case in point).

I think it would work better if the `Reviewers` section was at the top:

image

That way is would be easy to see the status of a PR, navigate to the commands and find the `Start a review` button.

In a similar vein, maybe the `View conversation on GitHub` should be moved below the `Changes` (it can also be a pain to find). This would mirror the other view options that appear at the bottom of a PR on .com:

image

meaghanlewis:
Hey @grokys this is looking really smooth! It's so nice to be able to do PR reviews within Visual Studio.

Here are a few notes of some things I found strange.

1. I don't think I should be given the option to approve or request changes to my own PR. Perhaps **Approve** and **Request changes** should be disabled to avoid seeing this error:
screen shot 2018-02-20 at 10 20 30 am

2. I experience an issue trying to approve a PR after previously requesting changes. (I wasn't approving my own PR in this case, but using a tester account). I saw the same thing Jamie did here. Refreshing and coming back to try to approve again fixed the issue.
>Is seems there are some issues when you come to approve your review.
>image

3. When I review a PR and I am taken back to the PR details pane, I see under reviewers it says "pending review". After refreshing the status updates correctly.
screen shot 2018-02-20 at 10 40 29 am

4. When requesting changes, I had previously left a comment in one of the changes files, but the comment doesn't seem to be associated with the review to request changes as it says "0 file comments" when there should be 1 comment.
screen shot 2018-02-20 at 10 42 12 am

donokuda:
Hey @grokys! I tested the pull request flow and have some ideas on how to polish the experience up a little more. I'm going to start with the flow of creating a review, then follow up with another comment on reading code reviews.

_Note: I wanted to articulate and document my thoughts before jumping into a polish branch so that we could get consensus on what to change as well as the best way to approach these ideas._

### Creating a review

There are currently a couple of ways to start a review:

1. Explicitly clicking on "Add your review" in the pull request details pane.
2. Clicking on "Start a review" via adding a comment in the diff.

image

In the situation for starting a review via the diff, clicking "Start a review" didn't feel like I was was starting a review. Instead it felt more like I was leaving a single comment. This is probably because I expected something to guide me through reviewing a pull request.

We could address this in a couple of ways:

**We have the GitHub panel navigate automatically to the "Create a Review" view.** That way, the panel reflects that I'm currently in a review mode. Additionally, because the Files Changed section exists in that panel, I still have all the information needed to continue my review.

**We show some sort of notification in the created comment.** Because my eyes are already focused on that specific area, we have an opportunity to put something in front of the reviewer and point them to the right place without having them find the next steps themselves.

image

(Of course, we could always do a combination of both :smile:)

In terms of the content in the "Submit your review" section. I'm a huge :100: on what you have. The only big suggestion I have is how we group and show file comments in the view:

image

I'm hoping by grouping comments by files, we would make it easier for reviewers and pull request authors to understand the scope of work need to address. Additionally, because we group by files, I feel we gain the benefit of being able to navigate to a file or a specific line comment.

The downside to this approach is that this is a custom UI component. I wasn't able to find examples around Visual Studio that I felt was intuitive for this type of information.

That said, I'm open to other ideas or examples that could help.

### Navigating to pending reviews

One last scenario I thought about was around having a review in progress. While we do list a review in progress, I wonder if there's an opportunity to try to call more attention to it. I'm thinking of the situation where someone started a review, left to do something else, and then totally forgot about it.

image

By calling more attention to the pending review, we could avoid situations where someone was about to give critical feedback, but didn't realize that it wasn't published yet.

Here are a few variations that I tried out:

image

I'm personally in favor of 2 and 3 because they feel a more natural in the list. The first idea hides the pending review and the last one feels a little too cluttered and busy.  

Those are the ideas that I have so far in regards to creating a review (and reviews in progress.) As mentioned before, I'll follow up with some stuff around viewing a review (including other things like past reviews.)

grokys:
> We show some sort of notification in the created comment. Because my eyes are already focused on that specific area, we have an opportunity to put something in front of the reviewer and point them to the right place without having them find the next steps themselves.

I like this. However, might we also want to link to the review for _all_ comments that come from reviews? We could maybe place that banner on all comments saying something like "This part of a review. Open in GitHubPane".

grokys:
Another thing to consider is that currently when I request changes in a review and then add a comment review, we just show the latest review, i.e. the comment review. This isn't entirely correct though, because the previous request changes review is still in action.

In this case dotcom shows the `x` requested changes icon next to the username, ignoring the subsequent comment review. Do we want to do this? The problem with this is that it's unclear that the reviewer has added comments to the PR after they requested changes. Maybe we should display buth an `x` and  a comment icon next to the username in this case? Not sure.

donokuda:
> Another thing to consider is that currently when I request changes in a review and then add a comment review, we just show the latest review, i.e. the comment review. This isn't entirely correct though, because the previous request changes review is still in action.
> 
> In this case dotcom shows the x requested changes icon next to the username, ignoring the subsequent comment review (though it is of course visible in the conversation view). Do we want to do this? The problem with this is that it's unclear that the reviewer has added comments to the PR after they requested changes.

I think we should follow dotcom's lead on this for a couple of reasons:

1. As you stated above, unless a reviewer has specifically dismissed their review, the  `❌Requested Changes` is still valid feedback that needs to be address regardless of the subsequent comment.
2. Matching behavior with dotcom isn't necessarily an issue until we figure out our users' workflows differ than what dotcom does.

> Maybe we should display buth an x and a comment icon next to the username in this case? Not sure.

Regarding adding a review comment after someone has already submitted a review, I think this is something that a conversation timeline would solve. I'm against adding another icon next to the review because it's not exactly clear what would show if I clicked into that review. Does that mean we combine the comment and the previous review that requests changes? Would there be a possible that someone would interpret that as "oh I need to address this as well in order to move forward" even though that wasn't the intention?

donokuda:
> I like this. However, might we also want to link to the review for all comments that come from reviews? We could maybe place that banner on all comments saying something like "This part of a review. Open in GitHubPane".

ooo, I'm in favor of that because that visually reminds me that I'm in the reviewing mode if we keep it shown until the review is complete. I kind of want to word-smith it closer to something like "You're currently reviewing a pull request" Any thoughts around that? 💭

donokuda:
Here's some ideas around addressing code review. I tried to approach this perspective as an author of a pull request who needs to address review feedback in order to get their code approved and merged.

The main goals I wanted to accomplish were:

1. Pull request authors should be able to understand which feedback needs to be address in order to get their code merged in.
2. In the middle of addressing feedback, authors should know which comments have been addressed so that they can continue where they left off.

The general flow in my mind was like this:

1. I finish working on a feature branch and open a pull request to discuss my changes
2. My colleagues review my code and leave feedback.
3. I select a review to address, read through the feedback.
4. I click on a filename or on the line to view the comment inline.
5. I implement changes based on the feedback
6. After committing changes, comments that I've addressed are marked `OUTDATED`
7. I go through each of the reviews until all the feedback is addressed.
8. I leave an indication somewhere that I'm ready for another pass.

Given the above, here's what I'm at:

## Show all reviews based on reviewer

Previously, I tried separating each review into its own view. However, I felt this approach was too limiting and fragmented. Navigating between outdated reviews felt clunky, and there wasn't necessarily a straightforward way to get back to the most recent review.

So instead, let's show all the reviews done by an individual in its own view? Here's a basic breakdown of how that would work:

From the reviews section of the pull request details panel, you'll click a reviewer which will bring you to a view with all reviews left by the reviewer:

screen shot 2018-03-05 at 4 09 07 pm

(👆leads to 👇) 

image

## Quick breakdown of this view

**Header**

image

**File Comment (and different states)**

image

Previous versions of the file comments had the line number in its own column but I was concerned about how flexible the layout was. Additionally, I feel like this layout helps with outdated comments. The compromise is that I feel like the line number isn't as scannable.

(Here's a comparison of the two versions that I did:)

image

## Long reviews are long!

I used https://github.com/github/VisualStudio/pull/1422 as an example of a pull request that could have a lot of reviews and comments, and I definitely think that it came become very unwieldy. 😱 

long-review

Here are a couple of things ideas to make sure that the list doesn't get too overwhelming:

1. We could collapse previous reviews to save vertical space. A big assumption here (!) is that the most recent review is the most relevant thing to address. We might want to do something in the case where the most recent is a neutral review and the previous one approves or requests changes but let's follow what happens with dotcom for now:

screen shot 2018-03-05 at 4 11 08 pm

2. I only exposed the top comment for each file. For comments with replies, we'll add a little note that replies exist, but we won't render them. I figured, we should let the peek view handle all the conversation and have the review view be more like a giant TODO list. This should make our use of vertical space a little more manageable:

image

That's what I got! Open to any thoughts and feedback. I'm specifically interested in which ideas are a big unknown or expensive to implement technically. I can adjust some of the designs around that (for example, showing the line number of outdated commits.)

sguthals:
Still up for discussion, should we do a flow more like this:
https://files.slack.com/files-pri/T0CA8C346-F9KUQQWPN/image.png

Or like what @donokuda proposed in the above comment.

sguthals:
Just poking around the PR a bit and here are some comments:

# LONG descriptions
Even if they aren't long, some of the PR descriptions are really really long. I know @jcansdale mentioned this above, but could we have an option to collapse them and then remember your choice per PR? That way if I'm really working through a specific PR I don't have to scroll through it every time/collapse it every time.

# Multiple Reviews by One Person
I think I prefer the "Collapsible Reviews":
37007072-59cb145e-2090-11e8-8199-32eaf55a3181
I like this one better because it's clear which one is the current and there isn't switching panes again. Though the third image in the "Other Reviews By" version has "current" on the one that is current, I think it would be easier for folks to not realize they aren't on the current one. 

Now, this could still be a potential problem in the "Collapsible Reviews" one that @donokuda proposed depending on how the scrolling works. Specifically, the "long reviews are long" comment that @donokuda posed above brings up the point that the most recent review might be hidden _above_ the long review. Should we still have some kind of indication of staleness? I like the "This review is outdated" in @grokys design, but it might be too cluttered mixed with @donokuda design with the collapsible - is there something else we might be able to do...a simple "Outdated" next to "Approved" or "Requires Changes"? Is there space for that? Too cluttered?

donokuda:
> Even if they aren't long, some of the PR descriptions are really really long. I know @jcansdale mentioned this above, but could we have an option to collapse them and then remember your choice per PR? That way if I'm really working through a specific PR I don't have to scroll through it every time/collapse it every time.

This is something I tried to address in the issue https://github.com/github/VisualStudio/issues/1452. I like the idea of remembering to collapse. 

image

> Now, this could still be a potential problem in the "Collapsible Reviews" one that @donokuda proposed depending on how the scrolling works. Specifically, the "long reviews are long" comment that @donokuda posed above brings up the point that the most recent review might be hidden above the long review. Should we still have some kind of indication of staleness?
> ...
> is there something else we might be able to do...a simple "Outdated" next to "Approved" or "Requires Changes"? Is there space for that? Too cluttered?

My assumption is that anything other than the most recent review would be considered stale. So for example, if Steven approved my pull request and then requested changes 3 days later, the approved review would be considered stale. However, I agree that it might not be clear enough in the UI. I like your idea of adding an "Outdated" badge:

image

sguthals:
I like your "outdated" badge mockup @donokuda.

sguthals:
Notes from Zoom Call which was recorded:
Some features @grokys pointed out:
1. Start with all older comments minimized
2. Taking a long time to load because the API does not give us line numbers
3. Having it in the side panel is useful because you don't have to look away (maybe re-visit when we have the conversation view in the main area. 
4. Question: All reviews by one person, or separate views?
5. Question: Losing the tree view

@jcansdale - Can we show the most recent one at the top
Yes - that is just a bug

Once we have the conversation view, the newer comments would most likely be at the bottom - need to revisit this when that happens. That might be a jarring experience for users to have the side panel be top->bottom and the conversation view be bottom->top. But it's worth revisiting, not trying to fix now.

@donokuda:
- The line numbers are a design constraint, so let's just push an API change
- New features (e.g. line numbers) can come in when they are implemented on dotcom
- Click on the file name to open the file

@StanleyGoldman:
- Giving users the view in the side panel when they have the bigger view they may not use it
- @sguthals suggests having the bigger conversation view open up from the side panel with a button that says "minimize to side panel" and it would show our original view. This would allow us to introduce the side panel now, still have the conversation view later, but have them be a workflow instead of competing views.

@meaghanlewis 
- There is a lot of overwhelming amount of information
- Maybe 1 line instead of 4-5.
- @StanleyGoldman suggested maybe strip the code block and put [CODE BLOCK] instead

@jcansdale 
- Mouse-over preview/show more context

Conclusions:
- Start with most recent review opened at the top
- Have the comments be just single lines
- Remove line numbers
- Have the outdated comments in a separate section

Future Things:
- Conversation view in the main viewing area that meshes well with our current side-panel
- Mouse-Over features

src/GitHub.App/Factories/ModelServiceFactory.cs
```diff
@@ -3,6 +3,7 @@
 using System.ComponentModel.Composition;
 using System.Threading;
 using System.Threading.Tasks;
+using GitHub.Api;
```
sguthals: I'm glad we added this!

grokys:
Superseded by #1491.

PR #1418, opened 2018-01-11
Move Unit.Tests to nunit

meaghanlewis:
This PR:
- Updates the `Unit.Test` project to use nunit instead of xunit.
- Uses the latest version of nunit -> `3.9.0`
- Uses the constraint-based model type assertions (`Assert.That`)


meaghanlewis:
There are currently 18 tests failing with 2 different errors:

1. `Expected: assignable from <GitHub.ViewModels.GitHubPane.ILoggedOutViewModel>   But was:  <Castle.Proxies.ILoggedOutViewModelProxy>`
2. `The calling thread must be STA, because many UI components require this`

meaghanlewis:
@grokys @jcansdale i pushed a commit to remove the `Run-Xunit` script and it dismissed your reviews, sorry  😞  - would one of you mind approving again?

jcansdale:
I'm happy that you converted them to using the `Assert.That` syntax. I find this more readable because it's obvious what the expect value is and what it's being compared to. 👍

PR #1420, opened 2018-01-17
Navigate to correct repository for inline comment GitHub link where fork repository is open.

### LINKED ISSUE #1413: Comment link navigates to contributor repo not PR repo
- GitHub Extension for Visual Studio version: v2.3.7

1. Opened "Possible missing comment # 34" from `grokys/PullRequest` in PR detail view
2. Clicked on comments for `Program.cs`
3. Clicked on `5 days ago`

image

The following URL opened with 404:
`https://github.com/jcansdale/PullRequestSandbox/pull/34#discussion_r159834411`

Expected:
`https://github.com/grokys/PullRequestSandbox/pull/34#discussion_r159834411`

grokys:
When a fork repository was open in VS, but the PRs for the parent repository were being viewed, clicking on the "date" link in an inline comment to go to the comment online was navigating to the fork repository instead of the parent repository.

To fix this, use the correct repository owner for GitHub link in `InlineCommentThreadViewModel.GetCommentUrl`. An optional `owner` parameter was added to `UriString.ToRepositoryUrl`. which will change the owner of the produced repository URL.

Fixes #1413


src/GitHub.Exports/Primitives/UriString.cs
```diff
@@ -141,11 +143,15 @@ public Uri ToRepositoryUrl()
                 ? url.Scheme
                 : Uri.UriSchemeHttps;
 
+            var nameWithOwner = owner == null || NameWithOwner == null ?
+                NameWithOwner :
+                string.Format(CultureInfo.InvariantCulture, "{0}/{1}", owner, RepositoryName);
+
```
shana: Nit: Given `owner = null` and `NameWithOwner` != null, won't this mean that `nameWithOwner` will be `/bar` instead of `bar`? I know it probably doesn't make a difference if the `Path` in UriBuilder` starts with a `/`, it will probably get ignored, but wouldn't this be more straightforward as:

```
var nameWithOwner = owner != null ? string.Format(CultureInfo.InvariantCulture, "{0}/{1}", owner, RepositoryName) : NameWithOwner;
```

Also, if `RepositoryName` is null (it can be), should this probably return null even if `owner` is passed in, for consistency?

grokys:
@shana I've updated the code with a few more tests and a fix.

> Nit: Given owner = null and NameWithOwner != null, won't this mean that nameWithOwner will be /bar instead of bar?

Did you mean the parameter `owner` here or the property `Owner`? If you mean the parameter then it would work, unless I'm misunderstanding. If you meant the `Owner` property then that can't be null if `NameWithOwner != null`.

> Also, if RepositoryName is null (it can be), should this probably return null even if owner is passed in, for consistency?

`RepositoryName` gets taken before owner, so if there's an owner there's a repository name.

Issue #23, opened 2015-07-31
Visual Studio crashes when navigating to Manage Connections page in Team Explorer


xperiandri:
When I click on "plug" icon in Team Explorer then Visual Studio immediately crashes.

Stack trace:
```
Unhandled exception:
System.NullReferenceException
Object reference not set to an instance of an object.
   at GitHub.VisualStudio.Base.TeamExplorerServiceHolder.<UIContextChanged>b__0() in Y:\jenkins\workspace\GitHub for Visual Studio speakeasy jenkinskeep\src\GitHub.VisualStudio\Base\TeamExplorerServiceHolder.cs:line 128
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at GitHub.VisualStudio.Base.TeamExplorerServiceHolder.<UIContextChanged>d__2.MoveNext() in Y:\jenkins\workspace\GitHub for Visual Studio speakeasy jenkinskeep\src\GitHub.VisualStudio\Base\TeamExplorerServiceHolder.cs:line 128
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.<>c__DisplayClass2.<ThrowAsync>b__3(Object state)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
   at System.Windows.Threading.DispatcherOperation.InvokeImpl()
   at System.Windows.Threading.DispatcherOperation.InvokeInSecurityContext(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
[...]

PR #1422, opened 2018-01-17
Refresh PullRequestSessionManager.CurrentSession when PR branch changes

### LINKED ISSUE #1408: PullRequestSessionManager.CurrentSession not updated when changing branches

### LINKED ISSUE #1421: ActiveRepository is sometimes set to null when changing PR branches
- GitHub Extension for Visual Studio version: v2.3.7.0
- Visual Studio version: Visual Studio 2015 (14.0.25420.1)

You can see what's going on from the log:

```
2018-01-17 12:13:00.060 INFO [38] TeamExplorerContext       Refresh ActiveRepository: RepositoryPath=C:\Source\github.com\grokys\PullRequestSandbox, BranchName=comments-test, HeadSha=5affff80a92fcb44367df4a1e4244882105ecafa
2018-01-17 12:13:00.060 INFO [38] TeamExplorerContext       Fire PropertyChanged event for ActiveRepository
2018-01-17 12:14:44.513 INFO [09] TeamExplorerContext       Refresh ActiveRepository: RepositoryPath=C:\Source\github.com\grokys\PullRequestSandbox, BranchName=pr/34-possible-missing-comment, HeadSha=a16aa3a7b1288d4c02b7859f1d2de83c10f65b2e
2018-01-17 12:14:44.513 INFO [09] TeamExplorerContext       Fire StatusChanged event for ActiveRepository
2018-01-17 12:14:49.277 INFO [154] TeamExplorerContext       Refresh ActiveRepository: RepositoryPath=null, BranchName=null, HeadSha=null
2018-01-17 12:14:49.277 INFO [154] TeamExplorerContext       Fire PropertyChanged event for ActiveRepository
2018-01-17 12:14:50.306 EROR [01] PullRequestDetailViewModel Error loading pull request grokys/PullRequestSandbox/34 from GitHub
System.NullReferenceException: Object reference not set to an instance of an object.
   at GitHub.Services.PullRequestService.<>c__DisplayClass22_0.<<EnsureLocalBranchesAreMarkedAsPullRequests>b__0>d.MoveNext() in C:\source\github.com\github\VisualStudio\src\GitHub.App\Services\PullRequestService.cs:line 267
--- End of stack trace from previous location where exception was thrown ---
   at System.Reactive.PlatformServices.ExceptionServicesImpl.Rethrow(Exception exception)
   at System.Reactive.ExceptionHelpers.ThrowIfNotNull(Exception exception)
   at System.Reactive.Subjects.AsyncSubject`1.GetResult()
   at GitHub.InlineReviews.Services.PullRequestSessionManager.<GetSession>d__14.MoveNext() in C:\source\github.com\github\VisualStudio\src\GitHub.InlineReviews\Services\PullRequestSessionManager.cs:line 148
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at GitHub.ViewModels.GitHubPane.PullRequestDetailViewModel.<Load>d__100.MoveNext() in C:\source\github.com\github\VisualStudio\src\GitHub.App\ViewModels\GitHubPane\PullRequestDetailViewModel.cs:line 344
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at GitHub.ViewModels.GitHubPane.PullRequestDetailViewModel.<Refresh>d__101.MoveNext() in C:\source\github.com\github\VisualStudio\src\GitHub.App\ViewModels\GitHubPane\PullRequestDetailViewModel.cs:line 448
2018-01-17 12:14:50.400 INFO [125] TeamExplorerContext       Refresh ActiveRepository: RepositoryPath=C:\Source\github.com\grokys\PullRequestSandbox, BranchName=pr/34-possible-missing-comment, HeadSha=a16aa3a7b1288d4c02b7859f1d2de83c10f65b2e
2018-01-17 12:14:50.400 INFO [125] TeamExplorerContext       Fire PropertyChanged event for ActiveRepository
```

jcansdale:
#### What's in this PR

- Add `TeamExplorerContext` service for repo and branch change events 
- Change `PullRequestSessionManager` to use the `ITeamExplorerContext` service
- Change `GitHubPaneViewModel` to use `ITeamExplorerContext`
  - This prevents GitHub pane from changing to PR list when PR checkout touches .sln file
- Revert `IVSGitExt` hack from `PullRequestDetailViewModel`
  - This stops it from receiving multiple events when nothing of interest has changed
- Fix null exception in `IsPullRequestFromRepository` when repo doesn't have a remote
- Fix `PullRequestSessionManager` to refresh when PR `Owner` or `Number` changes
- Add NUnit tests project `GitHub.App.UnitTests` for `TeamExplorerContext` tests
  - This is to avoid stepping on @meaghanlewis's unit test refactorings

#### The TeamExplorerContext service

```c#
namespace GitHub.Services
{
    public interface ITeamExplorerContext : INotifyPropertyChanged
    {
        ILocalRepositoryModel ActiveRepository { get; }
        event EventHandler StatusChanged;
    }
}
```

- The `PropertyChanged` event is fired when the repo changes
- The `StatusChanged` event is fired when the branch name, head SHA or tracking SHA changes
  - Duplicate events from `IVSGitExt` are ignored
- When a PR touches the .sln file only the `StatusChanged` event is fired
  - Previously `ActiveRepository` would be set to null and then again to the original repo
- When a solution that isn't in Git is loaded `ActiveRepository` will be set to null

### Related

- [Feature] Show current PR on status bar: #1396
- Visual Studio crashes when navigating to Manage Connections page in Team Explorer #23
- Make sure we don't crash when VS's git service returns bogus data: https://github.com/github/VisualStudio/commit/bc3d148cb35687b2d50015dc6454b41e23bac507

### Notes

Re: issue #23, it looks like the current implementation of Team Explorer won't return null.

image

The original report was from a user running VS 2015 `14.0.23107.0` (which is pre-Update 1). The above code is from VS 2015 `14.0.25431.01` Update 3. This implementation of `TeamExplorerContext` wouldn't crash if `ActiveRepositories` did come back null.

@meaghanlewis Did some tests using VS 2015 `14.0.23107.0` RTM. It appears that the `IGitExt` service is available as a global service even on this early version. I'm confused about how the service could come up null. I wonder if the `IServiceProvider` used by the Team Explorer UI might have had a bug that made it sometimes erroneously return null? 😕 

Fixes #1408
Fixes #1421 

Also fixes issue when `IsPullRequestFromRepository` is called on a local repository with no remote URL.


test/UnitTests/GitHub.App/ViewModels/GitHubPane/GitHubPaneViewModelTests.cs
```diff
@@ -23,9 +23,9 @@ public class TheInitializeMethod
         [Fact]
         public async Task NotAGitRepositoryShownWhenNoRepository()
         {
-            var te = Substitute.For<ITeamExplorerServiceHolder>();
-            te.ActiveRepo.Returns(default(ILocalRepositoryModel));
-            var target = CreateTarget(teServiceHolder: te);
+            var te = Substitute.For<ITeamExplorerContext>();
+            te.ActiveRepository.Returns(default(ILocalRepositoryModel));
+            var target = CreateTarget(teamExplorerContext: te);
```
grokys: Now that we're using a properly mockable class for the team explorer stuff, we can probably simplify these tests...

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -0,0 +1,176 @@
...
+        {
+        }
+
+        public TeamExplorerContext([Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider, ILogger log,
+            Type gitExtType, bool testing)
+        {
+            this.log = log;
+            this.testing = testing;
+
+            gitExtType = gitExtType ?? FindGitExtType();
```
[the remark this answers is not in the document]
jcansdale: The problem is there are two versions of this assembly with no binding redirections between them. I wanted to avoid dependencies on GitHub.TeamFoundation.14/15.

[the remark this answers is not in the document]
grokys: Ahhh, ok yeah that makes more sense then. Maybe add a comment to this effect to the source.

src/GitHub.InlineReviews/Services/PullRequestSessionManager.cs
```diff
@@ -53,19 +53,31 @@ public class PullRequestSessionManager : ReactiveObject, IPullRequestSessionMana
...
             this.sessionService = sessionService;
             this.connectionManager = connectionManager;
             this.modelServiceFactory = modelServiceFactory;
-            teamExplorerService.Subscribe(this, x => RepoChanged(x).Forget());
+
+            RepoChanged(teamExplorerContext.ActiveRepository).Forget();
+            teamExplorerContext.StatusChanged += (s, e) =>
+            {
+                RepoChanged(teamExplorerContext.ActiveRepository).Forget();
+            };
```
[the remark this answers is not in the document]
jcansdale: It's good to know what's possible. :-)

I'm going with Rx for the property change and regular event for this one.

test/GitHub.App.UnitTests/GitHub.App.UnitTests.csproj
```diff
@@ -0,0 +1,94 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
jcansdale: I wanted to avoid treading on the work @meaghanlewis has been doing when we merge. Maybe we could merge the unit test refactoring first and then move this test into the new NUnit `UnitTests` project?

test/UnitTests/GitHub.App/ViewModels/GitHubPane/GitHubPaneViewModelTests.cs
jcansdale: What did you have in mind?

grokys: A lot of tests are creating a mock `ITeamExplorerContext` that returns an `ActiveRepository` of `null`. Could this could be made the default in `CreateTarget` if no mock is passed in?

test/UnitTests/GitHub.App/ViewModels/GitHubPane/GitHubPaneViewModelTests.cs
```diff
@@ -23,150 +23,150 @@ public class TheInitializeMethod
         [Test]
         public async Task NotAGitRepositoryShownWhenNoRepository()
         {
-            var te = Substitute.For<ITeamExplorerServiceHolder>();
-            te.ActiveRepo.Returns(default(ILocalRepositoryModel));
-            var target = CreateTarget(teServiceHolder: te);
+            var te = CreateTeamExplorerContext();
+            te.ActiveRepository.Returns(default(ILocalRepositoryModel));
+            var target = CreateTarget(teamExplorerContext: te);
```
grokys: Hmm this just seems to replace `Substitute.For<ITeamExplorerServiceHolder>()` with `CreateTeamExplorerContext()`, which isn't much different. my suggestion was to make `CreateTarget` accept a null `ITeamExplorerContext` by default and create the mock itself when the default mock is required.

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -0,0 +1,221 @@
...
+                return;
+            }
+
+            notifyPropertyChanged.PropertyChanged += (s, e) => Refresh(gitExt);
+        }
+
+        /// <summary>
+        /// Used for unit testing.
+        /// </summary>
+        public interface IService
```
shana: What is this service interface for?

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -0,0 +1,221 @@
...
+                    trackedSha = newTrackedSha;
+                }
+            }
+            catch (Exception e)
+            {
+                log.Error(e, "Refreshing active repository");
+            }
+        }
+
+        static RepositoryInfo FindActiveRepository(object gitExt)
```
shana: How would this be unit tested?

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -0,0 +1,221 @@
...
+        {
+            var activeRepositoriesProperty = gitExt.GetType().GetProperty("ActiveRepositories");
```
shana: Instead of passing `gitExt` as an object and doing all this manual work here, how about wrapping it in an interface of our own and exposing a type that this code can safely use?

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -0,0 +1,221 @@
...
+            var headShaProperty = currentBranch?.GetType().GetProperty("HeadSha");
+            var headSha = (string)headShaProperty?.GetValue(currentBranch);
+
+            var nameProperty = currentBranch?.GetType().GetProperty("Name");
+            var branchName = (string)nameProperty?.GetValue(currentBranch);
+
+            return new RepositoryInfo(repositoryPath, branchName, headSha);
+        }
+
+        class RepositoryInfo
```
shana: Don't we already have types that represent a repository somewhere that we can use here? Yet another representation of a repository feels very duplicated.

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -0,0 +1,221 @@
...
+                    var newRepositoryPath = repo?.RepositoryPath;
+                    var newBranchName = repo?.BranchName;
+                    var newHeadSha = repo?.HeadSha;
+                    var newTrackedSha = newRepositoryPath != null ? service.FindTrackedSha(newRepositoryPath) : null;
+
+                    if (newRepositoryPath != repositoryPath)
+                    {
+                        log.Information("Fire PropertyChanged event for ActiveRepository");
+                        ActiveRepository = repo != null ? service.CreateRepository(repo.RepositoryPath) : null;
+                        PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(ActiveRepository)));
```
shana: This should probably happen in the property setter directly, so that it's always fired regardless of where the property is set from (probably with some checks so it's not fired if the value is unchanged maybe?)

src/GitHub.App/Services/TeamExplorerContext.cs
shana: There is already a `IVSGitExt` (https://github.com/github/VisualStudio/blob/master/src/GitHub.Exports/Services/IVSGitExt.cs https://github.com/github/VisualStudio/blob/master/src/GitHub.TeamFoundation.14/Services/VSGitExt.cs) type that represents (and wraps) `IGitExt`, so you can use that, there's no need for this code. If you prefer to take it out of the TeamFoundation assemblies and use reflection in it, you should be able to do that easily in it as well.

src/GitHub.App/Services/TeamExplorerContext.cs
jcansdale: To make it unit testable. I didn't want to create an external, public service that wouldn't be used by anything else. It isn't pretty, but I couldn't think of a less hacky way.

src/GitHub.App/Services/TeamExplorerContext.cs
jcansdale: Makes sense. Stick to a consistent pattern.

src/GitHub.App/Services/TeamExplorerContext.cs
jcansdale: It does have lots of tests. I pass in an object with a similar "shape". See `FakeGitExt`.

src/GitHub.App/Services/TeamExplorerContext.cs
jcansdale: The fact that `TeamExplorerContext` uses `IGitExt` is a detail that I didn't want to expose in a public API. `ITeamExplorerContext` is intended to surface the pattern of events that we need and hide the gory details.

I'm hoping that `IVSGitExt` and the parts of `ITeamExplorerServiceHolder` that use it will no longer be required.

I'm about to push a version that wraps `gitExt` in a private class and encapsulates the reflection code. I want to avoid exposing this as a public API.

test/UnitTests/GitHub.App/ViewModels/GitHubPane/GitHubPaneViewModelTests.cs
jcansdale: By the default mock do you mean `CreateTeamExplorerContext(ValidEnterpriseRepo)` rather than `CreateTeamExplorerContext(null)`?

jcansdale:
> I think you misunderstood my comment about the unit tests ;)

Yup, completely. 😉  Is this more what you had in mind?https://github.com/github/VisualStudio/compare/687db8827df8...f2678a1adb32

src/GitHub.App/Services/TeamExplorerContext.cs
grokys: Actually something that just came to me: could/should we use `dynamic` here instead of reflection?

jcansdale:
@shana I've addressed / replied to the inline comments. Could you take another look? 😄

test/UnitTests/GitHub.TeamFoundation/VSGitExtTests.cs
```diff
@@ -0,0 +1,151 @@
...
+            var sp = CreateGitHubServiceProvider(context, gitExt);
+            var target = new VSGitExt(sp);
+
+            var activeRepositories = target.ActiveRepositories;
+
+            Assert.That(activeRepositories.Count(), Is.EqualTo(0));
+        }
+
+        // TODO: We can't currently test returning a non-empty list because it constructs a live LocalRepositoryModel object.
+        // Move could move the responsibility for constructing a LocalRepositoryModel object outside of VSGitExt.
```
grokys: "Move could move"?

grokys: Given that `VSGitExt` is simply a wrapper for `IGitExt` it might be worth exposing the `IGitRepositoryInfo`s directly instead of wrapping them in `ILocalRepositoryModel`s?

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
```diff
@@ -13,24 +14,55 @@ namespace GitHub.VisualStudio.Base
...
+        [ImportingConstructor]
+        public VSGitExt(IGitHubServiceProvider serviceProvider)
         {
-            if (gitService != null)
-                gitService.PropertyChanged -= CheckAndUpdate;
-            gitService = serviceProvider.GetServiceSafe<IGitExt>();
-            if (gitService != null)
-                gitService.PropertyChanged += CheckAndUpdate;
+            this.serviceProvider = serviceProvider;
+            var factory = serviceProvider.GetService<IVSUIContextFactory>();
```
grokys: Is there a reason not to import this via a constructor parameter?

test/UnitTests/GitHub.TeamFoundation/VSGitExtTests.cs
jcansdale: I don't think we should be `ILocalRepositoryModels` here, but the problem with `IGitRepositoryInfo` is there's two versions. Maybe create another wrapper.

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
jcansdale: Not any more! Previously two MEF services were being exported. Fixed! 😄

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            Refresh();
+            gitExt.ActiveRepositoriesChanged += Refresh;
+        }
+
+        void Refresh()
+        {
+            try
+            {
+                var repo = gitExt.ActiveRepositories?.FirstOrDefault();
```
shana: This needs to be done on a background thread, see `TeamExplorerServiceHolder` for how it does this.

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
```diff
@@ -1,39 +1,90 @@
...
+            if (!context.IsActive || !TryInitialize())
+            {
+                context.UIContextChanged += ContextChanged;
+                log.Information("VSGitExt will be initialized later");
+            }
         }
 
+        void ContextChanged(object sender, VSUIContextChangedEventArgs e)
+        {
+            // If we're in the GitSccProvider context and TryInitialize succeeds, we can stop listening for events.
```
shana: We also need to listen for the context to be deactivated, and throw away the instance of GitExt if that happens, and get it back when it activates again.

src/GitHub.TeamFoundation.14/Base/TeamExplorerServiceHolder.cs
```diff
@@ -1,52 +1,44 @@
...
-        /// <param name="gitService"></param>
+        /// <param name="gitService">Used for monitoring the active repository.</param>
         [ImportingConstructor]
         public TeamExplorerServiceHolder(IVSGitExt gitService)
         {
-            this.GitService = gitService;
+            this.gitService = gitService;
             syncContext = SynchronizationContext.Current;
+
+            UpdateActiveRepo();
```
shana: This cannot happen here, it will call `ActiveRepositories` which is a potentially blocking call. It needs to be shuffled to a background thread.

jcansdale:
Should make sure this can't happen:

image

There's no reason for `ActiveRepositories` to return `null` rather then an empty collection.

src/GitHub.App/Services/TeamExplorerContext.cs
jcansdale: I've fixed it so that `ActiveRepositories` is initialized by `VSGitExt` on a thread pool thread.

src/GitHub.TeamFoundation.14/Base/TeamExplorerServiceHolder.cs
jcansdale: I've shuffled the initial fetching of `ActiveRepositories` to a thread pool thread so this is now safe to call.

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
jcansdale: I've done a bunch of experiments changing to and from solutions in TFS repos.  Here are some findings:

- When changing to a TFS repo there can be two `UIContext` events, first with `IsActive=false` then with `IsActive=true`
- When changing back to a Git repo there are (often?) no events fired
- The `IGitExt` service will only be null when directly opening a TFS solution
- Once set the returned `IGitExt` service object will remain constant not return

This is consistent with the `Microsoft.TeamFoundation.Git.Provider.SccExtensibilityService` being delay loaded but never unloaded.

Given that the context is deactivated and then activated again when opening a TFS based solution, I'm reluctant to do anything to the `IGitExt` object when this happens. We'd still be listening when in a TFS solution and continue to use the same object when returning to a Git based solution.

Does that make sense? I should update the comments to reflect this.

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
```diff
@@ -1,39 +1,114 @@
...
+            {
+                context.UIContextChanged += ContextChanged;
+                log.Information("VSGitExt will be initialized later");
+                InitializeTask = Task.CompletedTask;
+            }
+        }
+
+        void ContextChanged(object sender, VSUIContextChangedEventArgs e)
+        {
+            // If we're in the GitSccProvider context and TryInitialize succeeds, we can stop listening for events.
```
shana: We need to continually listen for events so we stop using GitExt if the scc provider becomes deactivated.

grokys: Could you explain why this is? Isn't @jcansdale's analysis correct? Looking at the code in ILSpy I don't see how `IGitExt` can become null once we have an instance unless the package is unloaded (which I don't think can happen?).

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
```diff
@@ -1,39 +1,109 @@
...
+        {
+            try
+            {
+                ActiveRepositories = gitService?.ActiveRepositories.Select(x => repositoryFactory.Create(x.RepositoryPath)).ToList();
+                ActiveRepositoriesChanged?.Invoke();
+            }
+            catch (Exception e)
+            {
+                log.Error(e, "Error refreshing repositories");
+                ActiveRepositories = new ILocalRepositoryModel[0];
```
shana: Shouldn't this be raising the `ActiveRepositoriesChanged` event as well? And if so, shouldn't this event be raised in the setter of the property directly?

jcansdale: Oops! Good catch. This is why I should have done it on the property itself!

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
```diff
@@ -25,7 +25,7 @@ public class VSGitExt : IVSGitExt
         readonly ILocalRepositoryModelFactory repositoryFactory;
 
         IGitExt gitService;
-        IEnumerable<ILocalRepositoryModel> activeRepositories;
+        IReadOnlyList<ILocalRepositoryModel> activeRepositories;
```
shana: nice! was going to mention that and then forgot 👍

PR #1425, opened 2018-01-22
Auto detect tests in Appveyor

### LINKED ISSUE #1289: Automatically detect and run all test dll's
**Background:**
In the past, we have not updated our tests to execute as new directories of tests were added. Because of this, some tests were not being run regularly.

**Goal:**
As a way to remedy forgetting which tests to run, it would benefit to have test dll's follow a similar convention so that they can be detected and run automatically without having to update the test commands and scripts in the future.

This task will require updating:
- test dll's to end in something like `\*.tests.dll` so that they can be recursively searched
- scripts that run the tests to tell them how to locate these test dll's
  - `scripts/test.ps1`
  - `appveyor.yml`

meaghanlewis:
This PR removes:
- The **test_script** block from `appveyor.yml` and uses Appveyor's auto testing mode instead
- The Appveyor flag from `scripts\test.ps1`
- The package `NUnit.Extension.AppVeyor.NUnit3ResultWriter`and all references to it

Closes #1289


grokys:
> The latest build on master ran 806 total tests, but this PR is running 804 total tests (Seems to be running 4 less tests in UnitTests and 2 more tests in TrackingCollectionTests) 

In the case of `UnitTests`, this seems to be caused by ignored tests like `GraphsNavigationItemTests.TheExecuteMethod`. This test was appearing in the list of tests before, but the status was gray indicating that it was ignored, whereas now the test isn't included in the list. The test has 4 `TestCase`s so this explains that.

In the case of `TrackingCollectionTests`, `MultipleSortingAndFiltering` and `OrderByDoesntMatchOriginalOrderTimings` are now being run where they weren't before. These two tests are marked with `Category("Timings")`, which was marked as [being ignored](https://github.com/github/VisualStudio/blob/master/scripts/Run-NUnit.ps1#L45) before. You can use the appveyor.yml to ignore these files.

shana:
> Should I update scripts\test.ps1 and remove the Appveyor flag since it seems we'll only be using this script locally now?

Yup, and you can also remove the `NUnit.Extension.AppVeyor.NUnit3ResultWriter` package from our nuget solution (in VS, `Tools -> Nuget Package Manager -> Manage Nuget Packages for Solution` and then in the `Installed` tab find the package and uninstall it from all projects that have it installed). After that, you can delete the NUnit.Extension.AppVeyor.NUnit3ResultWriter.0.1.0.nupkg package from `lib/` 😄

PR #1426, opened 2018-01-23
Ensure that the GitHubPaneViewModel is initalized.

grokys:
- Renamed `IGitHubToolWindowManager.ShowHomePane` to `ShowGitHubPane` because the latter is what we call it these days
- Make `IGitHubToolWindowManager.ShowGitHubPane` async to make sure that `GitHubPaneViewModel.InitializeAsync` has finished before passing back a refrence to the `GitHubPaneViewModel`
- Make the methods that call `IGitHubToolWindowManager.ShowGitHubPane` `void async` with logging when they fail


jcansdale:
@grokys,

- I'm finding the `GitHub` pane spins forever when I open VS with the pane visible:

image

- The `GitHub.ShowCurrentPullRequest` command does seem to work when the `GitHub` pane is visible.  

- If VS is opened with the `GitHub` pane closed, the `GitHub.ShowCurrentPullRequest` command doesn't open the pane.

grokys:
@jcansdale could you take another look at this? It's hopefully working now.

src/GitHub.Exports.Reactive/Services/IPullRequestSessionManager.cs
```diff
@@ -31,6 +31,18 @@ public interface IPullRequestSessionManager : INotifyPropertyChanged
         /// </returns>
         IPullRequestSession CurrentSession { get; }
 
+        /// <summary>
+        /// Ensures that the service is initialized.
+        /// </summary>
+        /// <returns>A task that when completed indicates that the service is initialized.</returns>
+        /// <remarks>
```
jcansdale: Simplying? ;-)

jcansdale: Hmm, is seems I missed the line with `simplying` and it isn't showing up for me as an in-line comment. What's going on? 😕

meaghanlewis: I don't see the above as inline comments either

PR #1429, opened 2018-01-24
Correctly detecting VisualStudio version on 2017 and forward

StanleyGoldman:
Recent metrics reporting showed that we are getting `15.0.0` consistently for Visual Studio.

https://developercommunity.visualstudio.com/content/problem/2813/cant-find-registry-entries-for-visual-studio-2017.html

We chose to grab the first `productDisplayVersion` to the first glob match of `C:\ProgramData\Microsoft\VisualStudio\Packages\_Instances\**\state.json`

That value returns a value that resembles what is represented on the About screen in Visual Studio.


StanleyGoldman:
I have to update the body of this pull request, the approach has changed since it was first written up.

StanleyGoldman:
Updated the body

StanleyGoldman:
Fixes: #1434

PR #1430, opened 2018-01-25
Functionality to enable/disable trace logging

StanleyGoldman:
- Added a setting to enable/disable trace logging
- Added functionality to control `Verbose` output in `Serilog`
- Configured `Serilog` file output to allow writing from multiple processes
- Configured the log format to show the process id

Depends on
- [x] #1429 
- [x] #1428


src/GitHub.Exports/Settings/generated/IPackageSettings.cs
```diff
@@ -31,6 +31,7 @@ public interface IPackageSettings : INotifyPropertyChanged
         void Save();
         bool CollectMetrics { get; set; }
         bool EditorComments { get; set; }
+        bool EnableTraceLogging { get; set; }
```
grokys: As the header suggests, this is a generated file so you should be editing `src/Common/settings.json` to modify this.

I don't really like this, but that's how it is for the moment :(

src/GitHub.VisualStudio/Services/GitHubServiceProvider.cs
```diff
@@ -97,16 +97,23 @@ public GitHubServiceProvider(IServiceProviderPackage asyncServiceProvider, IServ
...
-            instance = sp.GetService(serviceType);
-            if (instance != null)
-                return instance;
+            try
+            {
+                instance = sp.GetService(serviceType);
+                if (instance != null)
+                    return instance;
+            }
+            catch (Exception ex)
```
grokys: This seems unrelated to trace logging. No objection to including it here, but could you explain how it got here? :D

src/GitHub.VisualStudio/Settings/generated/PackageSettingsGen.cs
```diff
@@ -43,6 +43,13 @@ public bool CollectMetrics
             set { collectMetrics  = value; this.RaisePropertyChange(); }
         }
 
+        bool enableTraceLogging;
```
grokys: Again, this is a generated file so shouldn't be editing it directly.

src/GitHub.VisualStudio/Settings/PackageSettings.cs
```diff
@@ -10,13 +11,61 @@
 namespace GitHub.VisualStudio.Settings
 {
     [Export(typeof(IPackageSettings))]
-    [PartCreationPolicy(CreationPolicy.Shared)]
+    [PartCreationPolicy(CreationPolicy.NonShared)]
+    public class PackgeSettingsDispatcher : IPackageSettings
```
grokys: I think we can use the technique from #1518 for this too.

src/GitHub.VisualStudio/UI/Settings/OptionsControl.xaml
```diff
@@ -81,6 +81,22 @@
       </TextBlock>
       </DockPanel>
     </GroupBox>
+    <GroupBox
+            Margin="8,0,8,0"
+            VerticalAlignment="Top"
+            BorderBrush="LightGray"
```
grokys: Not sure why this `<WrapPanel>` is necessary seeing as there is only one child (a `WrapPanel` with one child won't do anything as far as I understand) but I see the other sections are the same so /shrug.

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -39,6 +41,19 @@ public class GitHubPackage : AsyncPackage
         {
             LogVersionInformation();
             await base.InitializeAsync(cancellationToken, progress);
+            
+            var packageSettings = await GetServiceAsync(typeof(IPackageSettings)) as IPackageSettings;
```
grokys: Could we move this into an `InitializeLogging` method or something?

src/GitHub.Exports/Settings/generated/IPackageSettings.cs
StanleyGoldman: I can't believe i missed that..

src/GitHub.VisualStudio/Services/GitHubServiceProvider.cs
StanleyGoldman: I'm unable to reproduce the error @shana and I saw that caused us to add this catch.

src/GitHub.VisualStudio/Settings/generated/PackageSettingsGen.cs
StanleyGoldman: Cool

src/GitHub.VisualStudio/Settings/PackageSettings.cs
StanleyGoldman: 👍

src/GitHub.VisualStudio/UI/Settings/OptionsControl.xaml
StanleyGoldman: It does not seem to have an effect on this element.

src/GitHub.VisualStudio/GitHubPackage.cs
StanleyGoldman: 👍

src/GitHub.VisualStudio/Services/GitHubServiceProvider.cs
StanleyGoldman: Should I remove the try/catch?

StanleyGoldman:
@grokys I got to most of the fixes you mentioned.

src/GitHub.VisualStudio/Services/GitHubServiceProvider.cs
grokys: No, leave it in, it makes sense to log exceptions here. I just wondered what caused you to add it.

PR #1437, opened 2018-01-30
Fix clearing navigation history with >10 pages.

### LINKED ISSUE #1436: Resetting NavigationViewModel.History throws.
`NavigationViewModel.History ` is a `ReactiveList` with a [`CollectionChanged` handler](https://github.com/github/VisualStudio/blob/9eb8dfb6ccb7f322bddde133808d801e36a80490/src/GitHub.App/ViewModels/GitHubPane/NavigationViewModel.cs#L142) that doesn't support `NotifyCollectionChangedAction.Reset`.

This is a problem because when the history is cleared (by e.g. moving to a new repository), if the history has >10 items a reset will occur, causing the GitHub pane to hang with a spinner until a new project is loaded,

grokys:
#1436 is a bug whereby clearing the GitHub pane navigation history (by moving to a new project) when there are more than 10 items in the history caused an exception.

We need to handle `Reset` collection events on the `Changing` handler so that we can get to the items before they're removed from the collection and `Dispose` them.

Fixes #1436.

### DIFF
--- a/src/GitHub.App/ViewModels/GitHubPane/NavigationViewModel.cs
+++ b/src/GitHub.App/ViewModels/GitHubPane/NavigationViewModel.cs
@@ -25,8 +25,8 @@ namespace GitHub.ViewModels.GitHubPane
         public NavigationViewModel()
         {
             history = new ReactiveList<IPanePageViewModel>();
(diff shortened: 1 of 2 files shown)


meaghanlewis:
This fix LGTM 👍

PR #1438, opened 2018-01-30
Adding a version field to UsageModel

StanleyGoldman:
- Adding a version field to UsageModel and setting it to `1.1`


grokys:
What is the version field used for?

PR #1439, opened 2018-01-30
Ignore inline comment if it contains no diff line context

### LINKED ISSUE #1435: Sometimes inline comments have no associated diff lines
- GitHub Extension for Visual Studio version: 2.4.0 (dev)
- Visual Studio version: 2015

An exception was thrown on this line:
https://github.com/github/VisualStudio/blob/975816ed6520e74774cf456e749cd435a4c830ef/src/GitHub.InlineReviews/Models/InlineCommentThreadModel.cs#L54

This is the inline comment that has no diff line context: https://github.com/grokys/PullRequestSandbox/pull/24#pullrequestreview-41516276

```
    Body: "comment on main"
    CommitId: "080d67f93efeb557f8bb24a009effb85bde353aa"
    CreatedAt: {01/06/2017 13:55:49 +00:00}
    DiffHunk: "@@ -10,7 +10,6 @@ class Program"
    Id: 119620146
    OriginalCommitId: "b328e39e53c85ea15208a60c7af74698d3e146d2"
    OriginalPosition: 6
    Path: "PullRequestSandbox/Program.cs"
    Position: 0
    User: Account: Login: jcansdale IsUser: True
```

jcansdale:
It's possible (if unlikely) that an inline comment contains no diff line context. For example, the following comment has no context:

grokys/PullRequestSandbox#24

```
    Body: "comment on main"
    CommitId: "080d67f93efeb557f8bb24a009effb85bde353aa"
    CreatedAt: {01/06/2017 13:55:49 +00:00}
    DiffHunk: "@@ -10,7 +10,6 @@ class Program"
    Id: 119620146
    OriginalCommitId: "b328e39e53c85ea15208a60c7af74698d3e146d2"
    OriginalPosition: 6
    Path: "PullRequestSandbox/Program.cs"
    Position: 0
    User: Account: Login: jcansdale IsUser: True
```

This was caused a comment with a `Position` of `0` (it should be 1 or above). Diff file line numbers are `1` based, so this comment is outside the bounds of the diff. Although this wouldn't normally happen, there is a comment like this in `PullRequestSandbox` that would be hit from time to time when testing.

This PR simply ignores and logs comment like this.

Fixes #1435

### DIFF
--- a/src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
+++ b/src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
@@ -93,14 +93,13 @@ namespace GitHub.InlineReviews.Services
                 var chunk = chunks.Last();
                 var diffLines = chunk.Lines.Reverse().Take(5).ToList();
                 var firstLine = diffLines.FirstOrDefault();
-                var diffLineType = firstLine != null ? firstLine.Type : DiffChangeType.None;
-
                 if (firstLine == null)
                 {
-                    log.Warning("Couldn't find diff lines for inline comment. RelativePath={RelativePath}, OriginalCommitSha={Sha}, OriginalPosition={Position}, Chunks={Chunks}, DiffHunk={DiffHunk}",
-                        relativePath, comments.Key.Item1, comments.Key.Item2, chunks.Count(), hunk);
+                    log.Warning("Ignoring in-line comment in {RelativePath} with no diff line context", relativePath);
+                    continue;
                 }
 
+                var diffLineType = firstLine.Type;
                 var thread = new InlineCommentThreadModel(
                     relativePath,
                     comments.Key.Item1,
--- a/src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
+++ b/src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
@@ -99,13 +99,11 @@ namespace GitHub.InlineReviews.Services
                     continue;
                 }
 
-                var diffLineType = firstLine.Type;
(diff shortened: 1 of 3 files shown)


src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
```diff
@@ -88,11 +92,20 @@ public class PullRequestSessionService : IPullRequestSessionService
...
+                    continue;
+                }
+
+                var diffLineType = firstLine.Type;
                 var thread = new InlineCommentThreadModel(
                     relativePath,
                     comments.Key.Item1,
                     comments.Key.Item2,
                     diffLines,
+                    diffLineType,
```
grokys: Do we need to pass `diffLineType` into the ctor separately, considering that at this point `firstLine != null`?

jcansdale: I thought since we need to check for this situation anyway and have the data, we may as well pass it in. Keeps it a little simpler. What are the cons?

grokys: Just duplicated information really - it makes the API less fault tolerant, because you can pass in a `diffLineType != diffLines[0]` and doing a `First()` on `diffLines` isn't going to be expensive. That's my feeling at least. Feel free to disagree though and I'll 👍.

jcansdale: Makes sense. I didn't like `DiffLineType` failing when it came to be used, but there's no reason not to populate this in the constructor.

PR #1441, opened 2018-01-30
Adding a user guid to to UsageModel

StanleyGoldman:
This is a cherry pick of relevant changes to code using e6a2f1e8042fbda24a68fb31ab1468472ffde4c0 as a reference. Currently found in #1267.

We are cherry picking the changes that creates and stores a guid per user and records it in our user metrics.

**Todo**:
- [x] Adding user guid to usage model and storing it in ApplicationData path
  - Adding it to the model
https://github.com/github/VisualStudio/blob/c05fc897f672cefc316fac0a8d042cb00c4350d9/src/GitHub.Exports/Models/UsageModel.cs#L8
  - Adding functionality to load and store the user guid
https://github.com/github/VisualStudio/blob/c05fc897f672cefc316fac0a8d042cb00c4350d9/src/GitHub.VisualStudio/Services/UsageService.cs#L38-L76
  - Adding functionality to set the user guid before submitting the data
https://github.com/github/VisualStudio/blob/c05fc897f672cefc316fac0a8d042cb00c4350d9/src/GitHub.VisualStudio/Services/UsageTracker.cs#L151-L153

- [x] Added `Rothko.IEnvironment` to `UsageService` in order to write some tests for it
https://github.com/github/VisualStudio/blob/c05fc897f672cefc316fac0a8d042cb00c4350d9/src/GitHub.VisualStudio/Services/UsageService.cs#L32
https://github.com/github/VisualStudio/blob/c05fc897f672cefc316fac0a8d042cb00c4350d9/src/GitHub.VisualStudio/GitHubPackage.cs#L253-L254
https://github.com/github/VisualStudio/blob/c05fc897f672cefc316fac0a8d042cb00c4350d9/test/UnitTests/GitHub.VisualStudio/Services/MetricsTests.cs#L526-L607

- [x] Discovered and fixed an issue where the loading usage file no file present will result in submitting usage with a date of the default minimum
https://github.com/github/VisualStudio/blob/c05fc897f672cefc316fac0a8d042cb00c4350d9/src/GitHub.VisualStudio/Services/UsageService.cs#L116-L120

- [x] Fixing tests that were too sensitive to the current date/time to succeed
https://github.com/github/VisualStudio/blob/c05fc897f672cefc316fac0a8d042cb00c4350d9/test/UnitTests/GitHub.VisualStudio/Services/MetricsTests.cs#L609-L618


src/GitHub.VisualStudio/Services/UsageService.cs
```diff
@@ -6,28 +6,71 @@
...
+
+                try
+                {
+                    var data = new UserData { UserGuid = userGuid.Value };
+                    var json = SimpleJson.SerializeObject(data);
+                    await WriteAllTextAsync(userStorePath, json);
+                }
+                catch (Exception ex)
+                {
+                    log.Error(ex, "Failed reading writing metrics GUID");
```
grokys: Failed reading writing?

src/GitHub.VisualStudio/Services/UsageService.cs
```diff
@@ -6,28 +6,71 @@
...
 
 namespace GitHub.Services
 {
     [Export(typeof(IUsageService))]
     public class UsageService : IUsageService
     {
-        static readonly ILogger log = LogManager.ForContext<UsageService>();
         const string StoreFileName = "ghfvs.usage";
+        const string UserStoreFileName = "user.json";
+        private static readonly ILogger log = LogManager.ForContext<UsageService>();
```
shana: The `private` shouldn't be here.

test/UnitTests/GitHub.VisualStudio/Services/MetricsTests.cs
```diff
@@ -419,6 +423,13 @@ static IGitHubServiceProvider CreateServiceProvider(bool hasMetricsService = tru
 
     public class UsageServiceTests : TestBaseClass
     {
+        [Test]
+        public async Task GetUserGuidWorks()
+        {
+            var usageService = new UsageService(Substitute.For<IGitHubServiceProvider>());
+            var guid = await usageService.GetUserGuid();
```
StanleyGoldman: So I started writing this unit test...

src/GitHub.VisualStudio/Services/UsageService.cs
```diff
@@ -100,6 +143,10 @@ public IDisposable StartTimer(Func<Task> callback, TimeSpan dueTime, TimeSpan pe
                     Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
                     program.ApplicationName,
                     StoreFileName);
+                userStorePath = Path.Combine(
+                    Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData),
```
StanleyGoldman: And I realize a lot of this is driven here.
Running this test locally pulls up my current user's `ApplicationData`.
How would I go about abstracting the file...

shana: If this class used Rothko, it would be simple to abstract. Rothko provides some environment abstractions so at least you can return a different path for this even if you go through regular IO. That would be a very minimal change that would make this easier to test I think?

StanleyGoldman:
So I added a dependency on `Rothko.IFileFacade` to `UsageService`

https://github.com/github/VisualStudio/blob/754e9371b2610459a3f3900aee45bae694f33c65/src/GitHub.VisualStudio/Services/UsageService.cs#L32

And I'm making use of it here.

https://github.com/github/VisualStudio/blob/754e9371b2610459a3f3900aee45bae694f33c65/src/GitHub.VisualStudio/Services/UsageService.cs#L157-L175

**First question**: It seems like i'm the only person trying to use `IFileFacade`, am I going about this the correct way?

**Second question**: How do I get an actual `IFileFacade` instance from within `GitHubPackage`?

https://github.com/github/VisualStudio/blob/754e9371b2610459a3f3900aee45bae694f33c65/src/GitHub.VisualStudio/GitHubPackage.cs#L249-L253

StanleyGoldman:
- Added Tests
- Found a bug when there is no usage file present
- Fixed said bug
- Updated the pull request body above

PR #1445, opened 2018-01-31
Use a factory method in PullRequestSessionManagerTests

### LINKED ISSUE #1444: PullRequestSessionManagerTests are hard to update.
Updating `PullRequestSessionManagerTests` when `PullRequestSessionManager`'s dependencies change is tedious because one needs to alter the body of every test. We should use a factory method which will use default values for services that aren't directly used by the test.

grokys:
This way when the constructor dependencies of `PullRequestSessionManager` are changed we don't need to update each test individually by hand.

Fixes #1444 
Depends on #1442


chulito1:
Pull all google kits a Jason kits please

Issue #1319, opened 2017-11-15
Crash when loading repositories


shana:
Got a crash when loading repositories in the Clone dialog. I *think* this is something related to the database having null or invalid data in it.

## Possible repro:
- Login
- Open the clone dialog
- Crash?

![exceptionwhenfilteringclonerepos](https://user-images.githubusercontent.com/310137/32840743-d7823d74-ca18-11e7-9785-39fcfa46ca28.png)
![indekeyisnull](https://user-images.githubusercontent.com/310137/32840746-d9d82368-ca18-11e7-9526-0a5121f0d5a0.png)

The `IndexKey` property of the CacheIndex object being null means that in line https://github.com/github/VisualStudio/blob/ee3341fe09cb5e7a76b465d0ea6d757199496949/src/GitHub.App/Extensions/AkavacheExtensions.cs#L203 the object was returned from the database instead of being newly created by the `CacheIndex.Create(key)`, and the database returned an empty/invalid object.

## Possible fix:

After a `This.GetOrCreateObject` call, before the `Replay` (inside the `Defer`, basically), we should check whether the object is actually valid (has the key filled out, basically), and if it isn't we just go ahead and create one with the same create method that we pass into the `GetOrCreateObject`, to guarantee that even if the database is screwed up, we always have valid objects (and saving them afterwards as normal should fix any corruption in the database, so that's a nice plus).

shana:
To repro: install 2.2.0.10, open the clone dialog so it populates with data, then update to the latest version.

PR #1450, opened 2018-02-01
Fix a ton of potential crashers, odd code and redundant calls in ModelService

shana:
Ugh, what a mess. While fixing #1319, I came across a bunch of repository tests that were not actually using the code that the app is using.

Moved those tests to use the real API and deleted the old deprecated ModelService methods, and in the process fixed an old `TrackingCollection` bug where the `OriginalCompleted` observable would finish before all items have been processed.

Added exception catching to various places so we don't crash if keys aren't found or the network is down while accessing data.

This needs #1449 for tests to pass.


test/UnitTests/GitHub.App/Models/ModelServiceTests.cs
```diff
@@ -157,20 +155,20 @@ public class TheGetAccountsMethod : TestBaseClass
...
             var fetched = await modelService.GetAccounts();
 
-            Assert.That(3, Is.EqualTo(fetched.Count));
-            Assert.That("snoopy", Is.EqualTo(fetched[0].Login));
-            Assert.That("github", Is.EqualTo(fetched[1].Login));
-            Assert.That("fake", Is.EqualTo(fetched[2].Login));
+            Assert.AreEqual(3, fetched.Count);
+            Assert.AreEqual("snoopy", fetched[0].Login);
+            Assert.AreEqual("github", fetched[1].Login);
+            Assert.AreEqual("fake", fetched[2].Login);
```
grokys: The `Assert.That` constraint model is preferred for nunit 3 apparently? See https://github.com/github/VisualStudio/pull/1411#issuecomment-358045651 and the surrounding comments for context on this.

grokys:
Also, I think the fix for `TrackingCollection` should probably be in a separate PR...

test/UnitTests/GitHub.App/Models/ModelServiceTests.cs
shana: Ah, ok. It's just that it inverts the order of the parameters, makes it even more confusing than usual. I'll fix these then.

grokys: Yeah, I think we need to switch the order of parameters in quite a few tests if we want to be technically correct. However, it doesn't actually impact the testing and it's tedious work...

PR #1465, opened 2018-02-07
[wip] Prevent hangs when checking out a PR branch

### LINKED ISSUE #1464: Visual Studio hangs when checking out a PR
- GitHub Extension for Visual Studio version: v2.3.6.391, v2.4.1.1219 (likely others)
- Visual Studio version: 2015

1. Open `GitHubVS.sln` inside VS 2015
2. Open GitHub / Pull Requests pane
3. Checkout PR # 1396
4. Checkout PR # 1392
5. Visual Studio hangs

image

image

image

jcansdale:
This is an experiment to see if it's possible to prevent hangs when checking out a PR branch. These can happen when changing branches touches a project. When the project is reloaded Visual Studio can hang. I noticed this on VS 2015 but have heard this can also happen on VS 2017.

The hang happens pretty consistently when changing branches using out PR details view, but not when using Team Explorer. I believe Microsoft do some magic when Team Explorer is asked to change branches.

This PR unloads the solution before changing branches and reloads it when done. Doing this seems to prevent the hang and also avoids the automatic change to the Solution Explorer pane after checking out using PR details view.

*Update: is appears it only hangs on VS 2015 when there is a shared/linked document open (e.g. `SolutionInfo.cs`).*

Fixes #1464


jcansdale:
This works but is a bit heavy handed. A better way to fix this would be to search for shared/linked files and close them before changing the PR branch. We'd only need to do this on VS 2015.

PR #1467, opened 2018-02-07
Fix metrics, take 2135413

### LINKED ISSUE #1466: Missing information in metrics
All the non-int fields are missing in the metrics that are being sent, caused by https://github.com/github/VisualStudio/blob/master/src/GitHub.Exports/Models/UsageModel.cs#L51. This means we're missing `appVersion`, `vsVersion`, `lang`, `isEnterpriseUser` and `isGitHubUser`.

shana:
Fixes #1466

https://github.com/github/VisualStudio/blob/master/src/GitHub.Exports/Models/UsageModel.cs#L51 made it so only properties with int types get included in the metrics, which is not ideal. 

This PR:
- Fixes the bug
- Adds test to validate that the UsageService writes all the data
- Adds test to validate that the MetricsService sends all the data
- Moves the usage tracker tests into a decent namespace and one single fixture


test/UnitTests/GitHub.VisualStudio/Services/UsageTrackerTests.cs
```diff
@@ -8,209 +7,268 @@
 using GitHub.Settings;
 using NSubstitute;
 using NUnit.Framework;
+using System.Reflection;
+using System.Linq;
 
-
-namespace UnitTests.GitHub.VisualStudio.Services
+namespace MetricsTests
```
grokys: I don't have a problem with this, but for reference many other unit tests have no namespace, e.g. https://github.com/github/VisualStudio/blob/master/test/UnitTests/GitHub.Exports/GitServiceTests.cs. However we're not consistent about this it seems.

shana:
@grokys That was just the failing test to start with. Everything is in there now.

src/GitHub.VisualStudio/Services/UsageService.cs
```diff
@@ -32,7 +32,7 @@ public bool IsSameDay(DateTimeOffset lastUpdated)
 
         public bool IsSameWeek(DateTimeOffset lastUpdated)
         {
-            return GetIso8601WeekOfYear(lastUpdated) == GetIso8601WeekOfYear(DateTimeOffset.Now);
+            return GetIso8601WeekOfYear(lastUpdated) == GetIso8601WeekOfYear(DateTimeOffset.Now) && lastUpdated.Year != DateTimeOffset.Now.Year;
```
grokys: Is this right? The method is `IsSameWeek` not `IsDifferentWeek`.

src/GitHub.VisualStudio/Services/UsageService.cs
```diff
@@ -32,7 +32,7 @@ public bool IsSameDay(DateTimeOffset lastUpdated)
...
         }
 
         public bool IsSameMonth(DateTimeOffset lastUpdated)
```
grokys: I think the year check needs to be made in `IsSameMonth` too.

test/UnitTests/GitHub.VisualStudio/Services/UsageTrackerTests.cs
```diff
@@ -8,209 +7,346 @@
...
+            var metricsService = serviceProvider.TryGetService<IMetricsService>();
+
+            var expected = model;
+            expected.NumberOfStartups++;
+            expected.NumberOfStartupsWeek++;
+            expected.NumberOfStartupsMonth++;
+
+            UsageData result = usageService.ReceivedCalls().First(x => x.GetMethodInfo().Name == "WriteLocalData").GetArguments()[0] as UsageData;
+            CollectionAssert.AreEquivalent(
+                ReflectionUtils.GetFields(expected.GetType()).Select(x => new { x.Name, Value = x.GetValue(expected) }),
```
grokys: Was `SimpleJson.cs` added just for this utility class? Wouldn't `expected.GetType().GetRuntimeFields()` suffice here? I notice you use that method in `CreateUsageModel`.

test/UnitTests/GitHub.VisualStudio/Services/UsageTrackerTests.cs
```diff
@@ -8,209 +7,346 @@
...
+                ReflectionUtils.GetFields(result.Model.GetType()).Select(x => new { x.Name, Value = x.GetValue(result.Model) }));
+        }
+
+        [Test]
+        public async Task MetricserviceSendsDailyData()
```
grokys: Typo: should be `MetricsService` I think.

test/UnitTests/GitHub.VisualStudio/Services/UsageTrackerTests.cs
```diff
@@ -8,209 +7,346 @@
...
+            expected.NumberOfStartups++;
 
-                await service.Received(1).WriteLocalData(data);
+            CollectionAssert.AreEquivalent(
+                ReflectionUtils.GetFields(expected.GetType()).Select(x => new { x.Name, Value = x.GetValue(expected) }),
+                ReflectionUtils.GetFields(result.GetType()).Select(x => new { x.Name, Value = x.GetValue(result) }));
+        }
+
+        [Test]
+        public async Task MetricserviceSendsWeeklyData()
```
grokys: And here?

src/GitHub.Exports/Models/UsageModel.cs
```diff
@@ -3,65 +3,67 @@
...
-        public int NumberOfWelcomeTrainingClicks { get; set; }
-        public int NumberOfGitHubPaneHelpClicks { get; set; }
-        public int NumberOfPRDetailsViewChanges { get; set; }
-        public int NumberOfPRDetailsViewFile { get; set; }
-        public int NumberOfPRDetailsCompareWithSolution { get; set; }
-        public int NumberOfPRDetailsOpenFileInSolution { get; set; }
-        public int NumberOfPRDetailsNavigateToEditor { get; set; }
-        public int NumberOfPRReviewDiffViewInlineCommentOpen { get; set; }
-        public int NumberOfPRReviewDiffViewInlineCommentPost { get; set; }
+        public bool IsGitHubUser;
```
grokys: Why make these fields instead of auto properties? Auto properties should work just as well here, afaik. It's possibly over-dogmatic, but exposing fields, even on structs doesn't seem like good practise...

src/GitHub.VisualStudio/Services/UsageService.cs
shana: We're using it as different week everywhere (i.e., is it the same week? no? then send week data). Of course if you run VS on the same week of a different year, it's going to think it's the same week...

test/UnitTests/GitHub.VisualStudio/Services/UsageTrackerTests.cs
shana: Oh, I must have missed one. Yes, it's for these calls, I don't want to manually type the reflection code, it's really hard to read and easy to miss if you call the wrong one.

src/GitHub.VisualStudio/Services/UsageService.cs
grokys: But with this change, `IsSameWeek(DateTimeOffset.Now) == false` - that's not correct surely? Today is the same week as today.

src/GitHub.Exports/Models/UsageModel.cs
shana: There was an alternate strategy I was pursuing but I changed my mind, so nevermind this, I've gone back to properties.

test/UnitTests/GitHub.VisualStudio/Services/UsageTrackerTests.cs
grokys: Sorry I don't understand what you mean by "manually type the reflection code", The two syntaxes are:

`ReflectionUtils.GetFields(expected.GetType())`

or

`expected.GetType().GetRuntimeFields()`

Is the second more "manually typed" than the first? Both are simply calling a method on `expected.GetType()` but the second is less chars as it's an extension method and doesn't require copying `SimpleJson.cs`.

shana: I took out all the code that isn't needed in SimpleJson and just left the ReflectionUtils stuff so it doesn't complain about conflicting types.

src/GitHub.VisualStudio/Services/UsageService.cs
shana: Oh, you mean the wrong comparison in the year because fat fingers? Funny how there's no tests that caught me doing that backwards...

src/GitHub.VisualStudio/Services/UsageService.cs
shana: Yup.

test/UnitTests/GitHub.VisualStudio/Services/MetricsTests.cs
```diff
@@ -0,0 +1,506 @@
...
+            var metricsService = serviceProvider.TryGetService<IMetricsService>();
+
+            var expected = model;
+            expected.NumberOfStartups++;
+            expected.NumberOfStartupsWeek++;
+            expected.NumberOfStartupsMonth++;
+
+            var result = (usageService.ReceivedCalls().First(x => x.GetMethodInfo().Name == "WriteLocalData").GetArguments()[0] as UsageData).Model;
+            CollectionAssert.AreEquivalent(
+                ReflectionUtils.GetProperties(expected.GetType()).Select(x => new { x.Name, Value = x.GetValue(expected) }),
```
grokys: One last thought about this. While reading https://msdn.microsoft.com/en-us/library/ms182276.aspx I noticed:

> For value types, the inherited implementation of Equals uses the Reflection library, and compares the contents of all fields.

So we could actually just do `Assert.That(actual, Is.EqualTo(expected))` here.

shana: Equals won't say which fields are mismatched, which makes it really hard to figure out what's wrong from the test.

shana:
Do *NOT* click on `Update branch` for this PR. It's based on the 2.4.2 version bump commit and it'll be going out in a release based on that, so it should not be up to date with master.

PR #1470, opened 2018-02-08
Match a maximum of 4 lines when locating an inline comment

### LINKED ISSUE #1469: Inline comment appears on PR page but not in GitHub extension
- GitHub Extension for Visual Studio version: 2.4.1
- Visual Studio version: All

1. Added inline comment using GitHub Extension: https://github.com/github/VisualStudio/pull/1426#pullrequestreview-94823864
2. Comment appears on PR page (possibly the line above where I added it, not sure)
3. Open `IPullRequestSessionManager.cs` from PR details view
4. Inline comment doesn't appear on diff view or PR file tree

image

## Diagnosis

It appears the issue caused a dependency in the use of the `git diff --indent-heuristic` option. GitHub.com uses this but command line and libgit2 doesn't use it by default.

You can see the difference in the output from the following commands (on the `github/VisualStudio` repo):
```
git diff a39d0bb8..87058c77
git diff a39d0bb8..87058c77 --indent-heuristic
```

See here for when it was introduced:
https://github.com/github/experience-engineering-code/issues/483

## Fix

The `--indent-heuristic` option is available in libgit:
https://github.com/libgit2/libgit2/blob/da8138b01217824cf211fa491608a7b067cf8e43/include/git2/diff.h#L134

But isn't currently available in libgit2sharp.

https://github.com/libgit2/libgit2sharp/blob/master/LibGit2Sharp/Core/GitDiff.cs#L191
Needs to learn about the new flag.

https://github.com/libgit2/libgit2sharp/blob/3f9b415fa1edfc31ce1ec2b4b3d18441c34adfff/LibGit2Sharp/Diff.cs#L21
Needs to know to set it.

We would need a `DiffModifiers.IndentHeuristic` bool

If you have a `GitDiffOptions` then you can turn on bit `(1 << 31)` of the `Flags` (which unfortunately we don't have access to). See https://github.com/libgit2/libgit2/commit/7e3faf58cbafd1115d0f4bf6e5a2b422f9ef78f1

@ethomson said he might have a chance to tackle this on on a long flight tomorrow. It would be very much appreciated if he was able to! 😃 

_Update: it looks like this has changed to `GIT_DIFF_INDENT_HEURISTIC = (1u << 18)` in recent versions._
https://github.com/libgit2/libgit2/commit/2d9d246409d8131dfbbded2f880b784f205e93a5

jcansdale:
It appears that github.com might match a maximum of 4 lines when deciding whether an in-line comment outdated or not.

For example the following inline comment only shows 4 added lines:

image

Does that mean that only 4 lines are being matched?

It does allow the inline comment to appear from the #1469 issue, so maybe this is an improvement (if we can't get the diff to match 100%);

It appears the issue is actually to do with us not using the `--indent-heuristic` option.
Here is a repro:
```
git diff a39d0bb8..87058c77
git diff a39d0bb8..87058c77 --indent-heuristic
```

See: https://github.com/github/experience-engineering-code/issues/483

Fixes #1469

### DIFF
--- a/src/GitHub.Exports/Models/DiffUtilities.cs
+++ b/src/GitHub.Exports/Models/DiffUtilities.cs
@@ -88,9 +88,15 @@ namespace GitHub.Models
 
         public static DiffLine Match(IEnumerable<DiffChunk> diff, IList<DiffLine> target)
         {
-            if (target.Count == 0)
+            var matchLines = target.Count;
+            if (matchLines == 0)
             {
                 return null; // no lines to match
+            }
+
+            if (matchLines > 4)
+            {
+                matchLines = 4; // match a maximum of 4 lines
             }
 
             foreach (var source in diff)
@@ -101,7 +107,7 @@ namespace GitHub.Models
                     if (source.Lines[i].Content == target[matches].Content)
                     {
                         matches++;
-                        if (matches == target.Count || i == 0)
+                        if (matches == matchLines || i == 0)
                         {
                             return source.Lines[i + matches - 1];
                         }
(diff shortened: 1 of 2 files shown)


jcansdale:
This issue will be fixed via a new option in LibGit2Sharp (equivalent to `--indent-heuristic`).

PR #1490, opened 2018-02-15
Make AssemblyResolverPackage an AsyncPackage.

### LINKED ISSUE #1489: Make Packages AsyncPackages

grokys:
Fixes #1489.


shana:
Damnit, this is missing a thing and I was too eager with the merge button. According to https://docs.microsoft.com/en-us/visualstudio/extensibility/how-to-use-asyncpackage-to-load-vspackages-in-the-background, we need to set `PackageAutoLoadFlags.BackgroundLoad`, at least.

PR #1492, opened 2018-02-15
Moved changed files tree into its own view.

grokys:
Ported from #1415. Moves the PR details changed files tree into its own view so that it can be shared by the PR review authoring view.

Part of #1491

### DIFF
--- a/src/GitHub.App/Services/PullRequestEditorService.cs
+++ b/src/GitHub.App/Services/PullRequestEditorService.cs
@@ -9,12 +9,14 @@ using System.Threading.Tasks;
 using GitHub.Commands;
 using GitHub.Extensions;
 using GitHub.Models;
+using GitHub.ViewModels.GitHubPane;
 using GitHub.VisualStudio;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Editor;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Editor;
 using Microsoft.VisualStudio.Text.Projection;
 using Microsoft.VisualStudio.TextManager.Interop;
 using Task = System.Threading.Tasks.Task;
@@ -151,6 +153,9 @@ namespace GitHub.Services
                 if (!workingDirectory)
                 {
                     AddBufferTag(diffViewer.RightView.TextBuffer, session, rightPath, DiffSide.Right);
+                    EnableNavigateToEditor(diffViewer.LeftView, session, file);
+                    EnableNavigateToEditor(diffViewer.RightView, session, file);
+                    EnableNavigateToEditor(diffViewer.InlineView, session, file);
                 }
 
                 if (workingDirectory)
@@ -350,6 +355,11 @@ namespace GitHub.Services
             return view;
         }
 
+        void ShowErrorInStatusBar(string message)
+        {
+            statusBar.ShowMessage(message);
+        }
+
         void ShowErrorInStatusBar(string message, Exception e)
         {
             statusBar.ShowMessage(message + ": " + e.Message);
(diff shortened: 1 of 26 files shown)


src/GitHub.App/Services/PullRequestEditorService.cs
```diff
@@ -1,25 +1,195 @@
...
     [Export(typeof(IPullRequestEditorService))]
+    [PartCreationPolicy(CreationPolicy.Shared)]
     public class PullRequestEditorService : IPullRequestEditorService
     {
-        readonly IGitHubServiceProvider serviceProvider;
-
         // If the target line doesn't have a unique match, search this number of lines above looking for a match.
         public const int MatchLinesAboveTarget = 4;
 
+        readonly IGitHubServiceProvider serviceProvider;
```
jcansdale: It looks like this is no longer being used.

src/GitHub.VisualStudio/Views/GitHubPane/PullRequestDetailView.xaml
```diff
@@ -2,47 +2,52 @@
                                     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                                     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
                                     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-                                    xmlns:ghfvs="https://github.com/github/VisualStudio"
```
jcansdale: Should we be replacing this? Isn't it the new style of XAML reference that was recently merged?

src/GitHub.VisualStudio/Views/GitHubPane/PullRequestDetailView.xaml.cs
```diff
@@ -47,38 +32,12 @@ public PullRequestDetailView()
...
             });
-
-            bodyGrid.RequestBringIntoView += BodyFocusHack;
         }
 
-        [Import]
-        ITeamExplorerServiceHolder TeamExplorerServiceHolder { get; set; }
-
         [Import]
         IVisualStudioBrowser VisualStudioBrowser { get; set; }
```
jcansdale: I'm curious why we're not using an `ImportingConstructor` for this. Is this something specific to views?

src/GitHub.VisualStudio/Views/GitHubPane/PullRequestFilesView.xaml
```diff
@@ -0,0 +1,138 @@
+﻿<UserControl x:Class="GitHub.VisualStudio.Views.GitHubPane.PullRequestFilesView"
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
+             xmlns:cache="clr-namespace:GitHub.UI.Helpers;assembly=GitHub.UI"
```
jcansdale: Could we use `xmlns:ghfvs="https://github.com/github/VisualStudio"` for some of these?

src/GitHub.VisualStudio/Views/GitHubPane/PullRequestDetailView.xaml
grokys: Oops, merge/rebase error!

src/GitHub.VisualStudio/Views/GitHubPane/PullRequestDetailView.xaml.cs
grokys: If you don't have a default constructor, the designer can't create an instance. We could have a default constructor as well as the importing constructor, but at the time this was written I guess we decided to do it like this.

src/GitHub.VisualStudio/Views/GitHubPane/PullRequestFilesView.xaml
grokys: Yep, we could!

src/GitHub.App/Services/PullRequestEditorService.cs
grokys: It is still being used!

grokys:
@jcansdale fixed those issues. Turns out #1407 wasn't ported to this branch, which was why `TextViewCommandDispatcher` wasn't being used. Ported that.

PR #1495, opened 2018-02-15
Tweak highlight colors

donokuda:
Targets branch `feature/highlight-current-pr`

----

Adjusts highlighted colors to be match Visual Studio highlighted colors:

image

/cc @grokys since the target branch is yours


grokys:
Thanks @donokuda !

PR #1497, opened 2018-02-16
Make InlineReviewsPackage async.

### LINKED ISSUE #1489: Make Packages AsyncPackages

grokys:
I used the [VS2017 project system package](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/Packaging/ManagedProjectSystemPackage.cs) as a template for the required attributes/procedure for this.

This required a bit of jiggling of the command infrastructure we have in the `GitHub.InlineReviews` package. This infrastructure was intended to be temporary, pending a command infrastructure that merges that in `GitHub.VisualStudio`. I'm looking into doing that right now, so don't dwell _too_ much on that part of it.

Fixes #1489.


src/GitHub.InlineReviews/InlineReviewsPackage.cs
```diff
@@ -1,24 +1,32 @@
...
+        protected override async Task InitializeAsync(
+            CancellationToken cancellationToken,
+            IProgress<ServiceProgressData> progress)
         {
-            base.Initialize();
-            PackageResources.Register(this);
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+            var menuService = (IMenuCommandService)(await GetServiceAsync(typeof(IMenuCommandService)));
+            var componentModel = (IComponentModel)(await GetServiceAsync(typeof(SComponentModel)));
+            PackageResources.Register(this, componentModel.DefaultExportProvider, menuService);
```
shana: It would be better to go through `GitHubServiceProvider` for this. `GitHubServiceProvider` already loads up `SComponentModel` when it gets initialized and uses it for lookups, guaranteeing that only types that are ours get returned in MEF queries.

Going through `GitHubServiceProvider` also means that we can bypass the `ExportProvider` look up at any time by precaching instances in it ahead of time (which is handy for tests and also if we want to avoid a MEF query for some reason).

grokys: The problem with `GitHubServiceProvider` is that it doesn't support async lookups, and given that we're in `InitializeAsync` I figured that we really should be doing these async. Or have I missed something?

grokys:
@shana made the change we discussed.

src/GitHub.InlineReviews/InlineReviewsPackage.cs
```diff
@@ -1,24 +1,33 @@
...
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
+using Task = System.Threading.Tasks.Task;
 
 namespace GitHub.InlineReviews
 {
-    [PackageRegistration(UseManagedResourcesOnly = true)]
+    [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [Guid(Guids.InlineReviewsPackageId)]
     [ProvideAutoLoad(UIContextGuids80.SolutionExists)]
```
jcansdale: This will cause the package to auto-load on the Main thread. I think we should be using this:
```
    [ProvideAutoLoad(UIContextGuids80.SolutionExists, PackageAutoLoadFlags.BackgroundLoad)]
```

In future we might want to only auto-load when there is an active Git repository (see #1512)?
```
    [ProvideAutoLoad(Guids.UIContext_Git, PackageAutoLoadFlags.BackgroundLoad)]
```

jcansdale:
Superseded by #1502.

PR #1498, opened 2018-02-16
Save the PR list filter settings.

### LINKED ISSUE #1494: PR filter view defaults to Closed
Steps to reproduce:

- Open repository
- Go to pull requests list
- Change filter to different options (all, closed, open)
- close VS
- reopen VS and go back to pull requests list

Actual behavior: Pull request filter defaults to Closed
Expected behavior: Pull request filter should default to All

Gif of issue:
filter-states-not-saved-bug

grokys:
Somewhere along the line, the call to `SaveSettings` in `UpdateFilter` got removed, causing the filter settings not to get persisted. This reinstates that call.

Fixes #1494


shana:
Looks like it disappeared on commit https://github.com/github/VisualStudio/commit/b929d55403dfceec0e61aebf777d8cf315221ff8, which is a merge commit (PR #1312)

For reference, a good way to find stuff like this is to do `git log -G [regex/string to search] -m --patch`, it'll output merges that have the regex or string.

PR #1499, opened 2018-02-20
Ensure correct ordering of VSGitExt.ActiveRepositoriesChanged.

### LINKED ISSUE #1493: Team Explorer and GitHub pane show the wrong repository
Steps to repro:

- Pin VS to the windows taskbar
- Open VS
- Open a solution in repository A
- Open a solution in repository B
- Close VS
- Right click on the VS icon in the taskbar and select solution in Repository A

Expected: GHfVS displays the information of Repository A in Team Explorer Home and GitHub pane 
Actual: GHfVS displays the information of Repository B in Team Explorer Home and GitHub pane 

This is likely a regression after 2.3.6.391

Screenshots of issue:
image 1

image

Gif of issue:
switching-repos-bug

grokys:
Wait on a lock in `VSGitExt.RefreshActiveRepositories` so that `ActiveRepositoriesChanged` notifications are raised in the correct order.

Also added a unit test that demonstrated the failure before the fix.

Fixes #1493

### DIFF
--- a/src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
+++ b/src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
@@ -127,7 +127,7 @@ namespace GitHub.VisualStudio.Base
             {
                 if (value != activeRepositories)
                 {
-                    log.Debug("ActiveRepositories changed to {Repositories}", value.Select(x => x.CloneUrl));
+                    log.Debug("ActiveRepositories changed to {Repositories}", value?.Select(x => x.CloneUrl));
                     activeRepositories = value;
                     ActiveRepositoriesChanged?.Invoke();
                 }
--- a/src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
+++ b/src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
@@ -24,6 +24,7 @@ namespace GitHub.VisualStudio.Base
         readonly IGitHubServiceProvider serviceProvider;
         readonly IVSUIContext context;
         readonly ILocalRepositoryModelFactory repositoryFactory;
+        readonly object refreshLock = new object();
 
         IGitExt gitService;
         IReadOnlyList<ILocalRepositoryModel> activeRepositories;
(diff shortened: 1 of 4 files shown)


jcansdale:
@grokys, some good detective work here! 😄

I think updating `ActiveRepositories` might be potentially problematic even now that the ordering is correct. We don't want to quickly change from one repository to another as VS opens (for performance reasons).

It would be easy to only automatically refresh the repositories after a short delay if the repositories haven't been explicitly updated (via a property change event). That way the previously loaded repository would only appear when VS is opened without specifying a solution.

For example we could replace `InitializeTask = Task.Run(() => RefreshActiveRepositories())` in the constructor to something like this:

```
                // Call RefreshActiveRepositories after short delay if ActiveRepositories hasn't changed.
                InitializeTask = Task.Run(() =>
                {
                    Thread.Sleep(100);
                    if (ActiveRepositories.Count == 0)
                    {
                        RefreshActiveRepositories();
                    }
                });
```

This would ensure that only one property changed event is fired as VS opens. Does that make sense?

grokys:
@jcansdale yeah I see your point but I feel like `VSGitExt` would be the wrong place to do this - it's a wrapper around a VS service and shouldn't be concerning itself with throttling. Throttling should be done at the view model layer in my opinion (and indeed is made simple at this level by `Observable.Throttle`.

Thoughts @jcansdale @shana ?

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
```diff
@@ -118,6 +127,7 @@ private set
             {
                 if (value != activeRepositories)
                 {
+                    log.Debug("ActiveRepositories changed to {Repositories}", value.Select(x => x.CloneUrl));
```
jcansdale: I'm curious what happens when `value` is `null`. I know `Select` is an extension method. Does it simply return `null`?

jcansdale:
>  I feel like VSGitExt would be the wrong place to do this - it's a simple wrapper around a VS service and shouldn't be concerning itself with throttling.

Yup, I did wonder about that. I agree this shouldn't be the responsibility of `VSGitExt`.

👍 from me.

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
jcansdale: It looks like it would throw an `ArgumentNullException`.

grokys: Ah yeah, I was thinking it wouldn't be null but it can when `gitService` is null. Fixed.

meaghanlewis:
This LGTM 👍

PR #1501, opened 2018-02-21
Use GraphQL in ModelService to read PR reviews/comments.

grokys:
Ported from #1415.

The REST API filters out pending reviews when reading PR reviews/review comments, making it unusable for the PR reviews feature. This PR updates `ModelService` to take a GraphQL `IConnection` and uses that to read PR review comments from the GraphQL API, and in addition reads the related PR reviews too.

This is integrated a bit hackily into `ModelService`; `ModelService` doesn't really mesh well with GraphQL, and our existing models are a bit of a strange fit for the (more logical) way that GraphQL uses nested models to represent relationships. However, without a lot of refactoring this was the best way to get things up and running, and more importantly I don't really have a good feel yet as to _how_ we should structure things with GraphQL. As we hopefully move more things to GraphQL in the coming months we should start to get a feel for that.

(As a side note, take a look at how much code is required to do nested paging in GraphQL - I think making this work by default in Ocotokit.GraphQL will be a big deal)

Part of #1491

### DIFF
--- a/src/GitHub.App/Services/ModelService.cs
+++ b/src/GitHub.App/Services/ModelService.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Linq;
 using System.Reactive;
 using System.Reactive.Linq;
+using System.Reactive.Threading.Tasks;
 using System.Threading.Tasks;
 using Akavache;
 using GitHub.Api;
@@ -17,7 +18,9 @@ using GitHub.Logging;
 using GitHub.Models;
 using GitHub.Primitives;
 using Octokit;
+using Octokit.GraphQL;
 using Serilog;
+using static Octokit.GraphQL.Variable;
 
 namespace GitHub.Services
 {
@@ -33,13 +36,16 @@ namespace GitHub.Services
 
         readonly IBlobCache hostCache;
         readonly IAvatarProvider avatarProvider;
+        readonly Octokit.GraphQL.IConnection graphql;
 
         public ModelService(
             IApiClient apiClient,
+            Octokit.GraphQL.IConnection graphql,
             IBlobCache hostCache,
             IAvatarProvider avatarProvider)
         {
             this.ApiClient = apiClient;
+            this.graphql = graphql;
             this.hostCache = hostCache;
             this.avatarProvider = avatarProvider;
         }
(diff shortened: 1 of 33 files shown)


StanleyGoldman:
The `ghfvs` branch is branch being used here?

src/GitHub.App/Services/ModelService.cs
```diff
@@ -366,6 +375,154 @@ IObservable<IReadOnlyList<IRemoteRepositoryModel>> GetOrganizationRepositories(s
...
+            {
+                var reviewPage = await graphql.Run(query, vars);
+
+                foreach (var review in reviewPage.Reviews)
+                {
+                    result.AddRange(review.CommentPage.Items);
+
+                    if (review.CommentPage.HasNextPage)
+                    {
+                        result.AddRange(await GetPullRequestReviewComments(review.Id, review.CommentPage.EndCursor));
```
StanleyGoldman: We should throw in some comments that explain this GraphQL a bit, because the pagination is nuanced.
Let me take a stab at it...

- The above `query` gets a single page or reviews and for each review a single page of comments.
- `GetPullRequestReviewComments` will continue to query the comments per review.
- The containing `for` loop will continue this process for each page in this page of reviews
- The containing `while` loop will continue this process to obtain all review pages

grokys:
> The ghfvs branch is branch of Octokit.GraphQL is being used here?

No, that branch can be removed now I think - the stuff in there that were needed by GHfVS has been added to master (strong name signing).

src/GitHub.App/Services/ModelService.cs
grokys: Yep, good point - added some comments, does that make it clearer?

src/GitHub.App/Services/ModelService.cs
```diff
@@ -366,6 +375,158 @@ IObservable<IReadOnlyList<IRemoteRepositoryModel>> GetOrganizationRepositories(s
                     });
         }
 
+#pragma warning disable CS0618 // DatabaseId is marked obsolete by GraphQL but we need it
```
jcansdale: 🚲 You could create a `GetDatabaseId` method surrounded by `#pragma warning disable CS0618` to avoid turning off obsolete warnings for a large block.

test/UnitTests/GitHub.App/Models/AccountModelTests.cs
```diff
@@ -119,13 +119,26 @@ public static bool BitmapSourcesAreEqual(BitmapSource image1, BitmapSource image
...
+                    var encoder = new BmpBitmapEncoder();
+                    encoder.Frames.Add(BitmapFrame.Create(image));
+                    using (MemoryStream ms = new MemoryStream())
+                    {
+                        encoder.Save(ms);
+                        data = ms.ToArray();
+                    }
+                    return data;
+                }
+                catch (Exception ex)
```
jcansdale: I'm guessing this will show a build warning? Replace with `catch { }`. Trying to keep the build clean. 😉

src/GitHub.App/Services/ModelService.cs
grokys: That won't work here as these expressions are converted at runtime to GraphQL queries, and the expression rewriter can't introspect such a method, so building the query will fail.

PR #1502, opened 2018-02-22
Consolidate VS Command Architecture

### LINKED ISSUE #1489: Make Packages AsyncPackages

grokys:
# Previously in Commands

We previously had two separate command architectures

## GitHub.VisualStudio.Menus

These classes had some problems:

1. Defined in `GitHub.VisualStudio` so couldn't be used in lower-level assemblies
2. No support for command parameters
3. The `serviceProvider.AddCommandHandler` extension method did some sync service requests making it non-ideal for use in `AsyncPackage`s
4. Can't be bound to UI elements

## GitHub.InlineReviews.Commands

These classes were written because of the problems with the prior framework (most notably point 1) but had their own problems:

1. Were defined in `GitHub.InlineReviews` so couldn't be used in lower-level assemblies (I'd done that with the intention of doing this consolidation at some point)
2. Made calls to the `Package` for services
3. Used a "magic" MEF registration that could cause all MEF assemblies to be loaded when not needed

# This PR

This PR consolidates the two frameworks and _attempts_ to solve the problems of the two. There are the same `VsCommand` and `VsCommand<T>` classes as in the `GitHub.InlineReviews.Commands` namespace, but:

1. They don't use "magic" package registration
2. Instead I've added some extension methods to `IMenuCommandService` to register them and removed the `AddCommandHandler` extension methods that used sync `GetService` calls
3. They implement `ICommand` so they can be bound in the UI
4. They inherit directly from `OleCommand` rather than wrapping an `OleCommand` so they can be used wherever an `OleCommand` is accepted
5. I also added an extension method `IMenuCommandService.BindCommand` to bind an existing `ICommand` (such as a `ReactiveCommand`) to a VS command. This is used in the `GitHubPaneViewModel`

In addition, I've added a new assembly: `GitHub.Services.Vssdk` for the implementations. We've been talking about this for a while, and it's really needed here. I originally called it `GitHub.Services.VisualStudio` but decided to change it to `Vssdk` because we already use `VisualStudio` in `GitHub.VisualStudio` and thought it was clearer that it means "abstractions over the VSSDK"

Fixes #1489


src/GitHub.App/ViewModels/GitHubPane/GitHubPaneViewModel.cs
```diff
@@ -213,18 +219,13 @@ public void Dispose()
...
-                BindNavigatorCommand(paneServiceProvider, PkgCmdIDList.githubCommand, openInBrowser);
-
-                paneServiceProvider.AddCommandHandler(Guids.guidGitHubToolbarCmdSet, PkgCmdIDList.helpCommand,
-                     (_, __) =>
-                     {
-                         browser.OpenUrl(new Uri(GitHubUrls.Documentation));
-                         usageTracker.IncrementCounter(x => x.NumberOfGitHubPaneHelpClicks).Forget();
-                     });
+                var menuService = (IMenuCommandService)paneServiceProvider.GetService(typeof(IMenuCommandService));
+                BindNavigatorCommand(menuService, PkgCmdIDList.pullRequestCommand, showPullRequests);
```
jcansdale: Is there a reason you're not using the `BindCommand` extension method?

src/GitHub.VisualStudio/Commands/CreateGistCommand.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        public override Task Execute()
+        {
+            return dialogService.Value.ShowCreateGist();
+        }
+
+        protected override void QueryStatus()
+        {
+            Log.Assert(SelectedTextProvider != null, "Could not get an instance of ISelectedTextProvider");
+            Visible = !string.IsNullOrWhiteSpace(SelectedTextProvider?.GetSelectedText());
```
jcansdale: Unrelated to this PR, but just while I remember. We should default to selecting all text. I bet most people don't even know this command exists! ;-)

src/GitHub.VisualStudio/Commands/CreateGistCommand.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class CreateGistCommand : VsCommand, ICreateGistCommand
+    {
+        readonly Lazy<IDialogService> dialogService;
+        readonly Lazy<ISelectedTextProvider> selectedTextProvider;
+
+        [ImportingConstructor]
+        protected CreateGistCommand(IGitHubServiceProvider serviceProvider)
+            : base(CommandSet, CommandId)
+        {
+            dialogService = new Lazy<IDialogService>(() => serviceProvider.TryGetService<IDialogService>());
```
jcansdale: I'm curious why we're not passing `Lazy<IDialogService>` (and others) into the importing constructor?

src/GitHub.App/ViewModels/GitHubPane/GitHubPaneViewModel.cs
grokys: `BindNavigatorCommand` does use `BindCommand`. It's just so we don't have to specify `(new CommandID(Guids.guidGitHubToolbarCmdSet, commandId)` for every command.

src/GitHub.VisualStudio/Commands/CreateGistCommand.cs
grokys: Good question: basically this is how we were doing it before. I'll try using `Lazy<>`.

src/GitHub.InlineReviews/InlineReviewsPackage.cs
```diff
@@ -1,24 +1,38 @@
...
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
+using Task = System.Threading.Tasks.Task;
 
 namespace GitHub.InlineReviews
 {
-    [PackageRegistration(UseManagedResourcesOnly = true)]
+    [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [Guid(Guids.InlineReviewsPackageId)]
     [ProvideAutoLoad(UIContextGuids80.SolutionExists)]
```
jcansdale: This will cause the package to auto-load on the Main thread. I think we should be using this:
```
    [ProvideAutoLoad(UIContextGuids80.SolutionExists, PackageAutoLoadFlags.BackgroundLoad)]
```

In future we might want to only auto-load when there is an active Git repository (see #1512)?
```
    [ProvideAutoLoad(Guids.UIContext_Git, PackageAutoLoadFlags.BackgroundLoad)]
```

src/GitHub.InlineReviews/InlineReviewsPackage.cs
```diff
@@ -1,24 +1,38 @@
...
+    public class InlineReviewsPackage : AsyncPackage
     {
-        protected override void Initialize()
+        protected override async Task InitializeAsync(
+            CancellationToken cancellationToken,
+            IProgress<ServiceProgressData> progress)
         {
-            base.Initialize();
-            PackageResources.Register(this);
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
jcansdale: I don't think we should be on the Main thread.

src/GitHub.VisualStudio/Commands/CreateGistCommand.cs
jcansdale: Added an issue here: #1534

jcansdale:
I've made some auto-load related tweaks to the packages that install commands (`InlineReviewsPackage` and `GitHubPackage`).

Rather than using `await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync()` (which can hang in certain circumstances), change package registration to `AllowsBackgroundLoading = false`. This is the default and will cause the package to initialize on the Main thread. It is then okay to install commands without calling `SwitchToMainThreadAsync`.

I've changed `InlineReviewsPackage` to use: `[ProvideAutoLoad(Guids.UIContext_Git)]`. This will cause it to only load when we're in the context of a Git repo.

There is also some merge related cleanup.

jcansdale:
Forget what I said above about using `AllowsBackgroundLoading = false`.  I've changed it back to using the following:

```cs
[PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
[ProvideAutoLoad(Guids.UIContext_Git, PackageAutoLoadFlags.BackgroundLoad)]
```

I'm using the same workaround as in `PullRequestStatusBarPackage` to avoid using `SwitchToMainThreadAsync`.  By using `await ThreadingHelper.MainThreadDispatcher.InvokeAsync` we avoid the potential hang.

PR #1508, opened 2018-02-26
Use XmlnsDefinitions.

grokys:
Add `XmlnsDefinitionAttribute` to assemblies so that we can use a single `xmlns:` namespace.

Previously, we had to add a new `xmlns:` to each xaml file for each namespace that we wanted to reference. This change means that a single `xmlns:ghfvs="https://github.com/github/VisualStudio"` can be used in many places to reference all our namespaces in all our assemblies. There are 2 exceptions to this:

- `XmlnsDefinitionAttribute` doesn't work for the current assembly (that is, the assembly containing the current xaml file). This is why there are still a number of `xmlns:local` lines
- We have numerous `Resources` classes, so these need to be disambiguated with a `clr-namespace`

I'm not sure if you think this is a useful change, but it's been something in the back of my mind for a while. Fee free to 👎 if you don't think it's worth it.


src/GitHub.VisualStudio/Views/Dialog/GitHubDialogWindow.xaml
```diff
@@ -25,9 +24,9 @@
                    WindowStartupLocation="CenterOwner"
                    Content="{Binding Content}">
   <Window.Resources>
-    <views:ViewLocator x:Key="viewLocator"/>
-    <DataTemplate DataType="{x:Type vm:ViewModelBase}">
+    <!--<ghfvs:ViewLocator x:Key="viewLocator"/>-->
```
jcansdale: Why is this now commented out?

src/GitHub.VisualStudio/Views/Dialog/RepositoryCloneView.xaml
```diff
@@ -2,33 +2,28 @@
                                   xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                                   xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
                                   xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-                                  xmlns:cache="clr-namespace:GitHub.UI.Helpers;assembly=GitHub.UI"
                                   xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-                                  xmlns:helpers="clr-namespace:GitHub.Helpers;assembly=GitHub.UI"
+                                  xmlns:ghfvs="https://github.com/github/VisualStudio"
                                   xmlns:local="clr-namespace:GitHub.VisualStudio.Views.Dialog"
                                   xmlns:prop="clr-namespace:GitHub.VisualStudio.UI;assembly=GitHub.VisualStudio.UI"
```
jcansdale: Is this one still required?

jcansdale: Is this one still required?

jcansdale: Ah or is it one of the `Resources` disambiguations you mentioned are still needed?

jcansdale:
This change does seem to make it look cleaner. You you find it's more maintainable like this?

src/GitHub.VisualStudio/Views/Dialog/GitHubDialogWindow.xaml
grokys: Oops, oversight! That shouldn't be commented out (and should use `local:`).

src/GitHub.VisualStudio/Views/Dialog/RepositoryCloneView.xaml
grokys: Yep, that's a `Resource` disambiguation.

grokys:
> Do you find it's more maintainable like this?

I think it cuts down noise on the root tag, but more importantly when I'm writing XAML I don't have to go look up the control namespace and assembly, I just need to put in the GHfVS URL, so in this way it's a lot more convenient when _writing_ code rather than maintaining (and I don't think it negatively affects maintainability either).

PR #1510, opened 2018-02-27
Expose IVSGitExt as async VS service

jcansdale:
This PR does the following

- Initializes `PullRequestStatusBarPackage` completely on a background thread
- Expose `IVSGitExt` as an async VS service
- Expose `IVSGitExt` as a MEF service
- Initialize in background so as not to delay MEF loading
- ~~Extend `IGitHubServiceProvider` with `GetServiceAsync<T>()` and `<T, Ret>`~~
- Use a delegate for GetServiceAsync
  - This avoids choosing one of the numerous and incompatible `IAsyncServiceProvider` implementations
- Explicitly construct the correct implementation of `IVSGitExt` for the running DTE version
  - With the bonus of avoiding MEF errors for this service

It takes over from where #1506 left off, doing initialization asynchronously using `GetServiceAsync` to retrieve `IGitExt`. This means that we don't need to explicitly switch to the Main thread when retrieving `IGitExt`.


src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
```diff
@@ -41,52 +42,75 @@ public VSGitExt(IGitHubServiceProvider serviceProvider, IVSUIContextFactory fact
...
 
-            // Start with empty array until we have a change to initialize.
+            // Start with empty array until we have a chance to initialize.
             ActiveRepositories = Array.Empty<ILocalRepositoryModel>();
 
-            if (context.IsActive && TryInitialize())
+            PendingTasks = InitializeAsync();
+        }
+
+        async Task InitializeAsync()
```
jcansdale: We could explicitly call this after constructing the service.

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -107,12 +112,28 @@ public GHClient(IProgram program)
         }
     }
 
+    [PartCreationPolicy(CreationPolicy.Shared)]
+    public class VSGitExtPart
```
jcansdale: I put this next to the export for `IGitHubClient`. Maybe we should collect all of the service exports together into a single file?

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -269,5 +296,18 @@ static ToolWindowPane ShowToolWindow(Guid windowGuid)
                 return sp.TryGetService(serviceType);
             }
         }
+
+        IVSGitExt CreateVSGitExt(IGitHubServiceProvider sp)
+        {
+            switch (VSVersion.Major)
```
jcansdale: In the past I've always used `DTE.Version` to differentiate versions of Visual Studio. This always returns a string with the major version number with a `.0` (even for different minor versions). I used `VSVersion` because it was there, but I think it might hit the file system when reading the file version number.

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
jcansdale: Since `IVSGitExt` is also consumed as a MEF service, we probably want the VS service to return as soon as possible (since there is no async loading for MEF services). Awaiting `InitializeAsync` before the service returns is probably a bad idea. Better to continue initializing in the background.

src/GitHub.VisualStudio/GitHubPackage.cs
shana: This shouldn't be here (and GitHubClient shouldn't be either so ignore that example :P). Ideally it should be right next to the implementation, but since we don't want to have multiple exports for both teamfoundation assemblies, and we don't have a separate assembly yet for mef implementations, this should probably go into GitHub.Exports or into the Services folder here in GitHub.VisualStudio.

It should also follow the convention of the other mef dispatchers and be called VSGitExtDispatcher. For it to actually work as a MEF dispatcher, it needs to implement IVSGitExt and have a mef export for that interface.

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -269,5 +296,20 @@ static ToolWindowPane ShowToolWindow(Guid windowGuid)
...
+        IVSGitExt CreateVSGitExt(string dteVersion)
+        {
+            // DTE.Version always ends with ".0" even for later minor versions.
+            switch (dteVersion)
+            {
+                case "14.0":
+                    return new Lazy<IVSGitExt>(() => new VSGitExt14(GetServiceAsync)).Value;
+                case "15.0":
+                    return new Lazy<IVSGitExt>(() => new VSGitExt15(GetServiceAsync)).Value;
+                default:
```
shana: The logic to construct the correct type should be in the VSGitExtDispatcher, so callers don't have to worry about it.
Also, since it needs to call GetServiceAsync on an IAsyncServiceProvider, it should take IAsyncServiceProvider in its constructor, so we know that that's a dependency that this type has. Later, when GitHubServiceProvider supports async calls, we can implement IAsyncServiceProvider on it and pass it in without major changes to the code.

src/GitHub.VisualStudio/Services/VSGitExtFactory.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public async static Task<IVSGitExt> Create(IAsyncServiceProvider sp)
+        {
+            var dte = await sp.GetServiceAsync(typeof(DTE)) as DTE;
+
+            // DTE.Version always ends with ".0" even for later minor versions.
+            switch (dte.Version)
+            {
+                case "14.0":
+                    return new Lazy<IVSGitExt>(() => new VSGitExt14(sp.GetServiceAsync)).Value;
```
shana: Why not just pass the `IAsyncServiceProvider` into `VSGitExt14/15` instead? 
Also, any specific reason why you need a `Lazy` here? You're evaluating it immediately after creating, which feels like it defeats the purpose of the Lazy in the first place.

src/GitHub.VisualStudio/Services/VSGitExtFactory.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            this.serviceProvider = serviceProvider;
+        }
+
+        [Export(typeof(IVSGitExt))]
+        public IVSGitExt VSGitExt => serviceProvider.GetService<IVSGitExt>();
+
+        public async static Task<IVSGitExt> Create(IAsyncServiceProvider sp)
+        {
+            var dte = await sp.GetServiceAsync(typeof(DTE)) as DTE;
```
shana: `ApplicationInfo.GetVersion` will give you the running VS version as a proper type, so you don't have to grab a separate service for that here

src/GitHub.VisualStudio/Services/VSGitExtFactory.cs
jcansdale: > Why not just pass the IAsyncServiceProvider into VSGitExt14/15 instead?

The multiple implementations of `IAsyncServiceProvider` wreck havoc. The `GitHub.TeamFoundation` assemblies pick up more dependencies and I end up having to use `extern alias`. Passing `GetServiceAsync` as a delegate ends up being much simpler.

> You're evaluating it immediately after creating, which feels like it defeats the purpose of the Lazy in the first place.

I've added some comments and change it to use `Func<IVSGitExt>` instead (using `Lazy` was a bit of a hack). By referencing `VSGitExt14` and `VSGitExt15` inside a lambda expression we can avoid loading the incompatible version into VS (e.g. we mustn't load `VSGitExt15` into VS 2015).

PR #1515, opened 2018-03-05
Cleanup GitHubPackage.cs

jcansdale:
Clean up a few things I noticed when working on another PR

- Don't `SwitchToMainThreadAsync` when returning `UsageTracker` service
- Remove a bunch of dead code
- Removed unused export for `GitHubHClient`
   - No errors in MEF log when removed
   - Checked for dynamic loading using `typeof(IGitHubClient)` or `<IGitHubClient>`

### DIFF
--- a/src/GitHub.VisualStudio/GitHubPackage.cs
+++ b/src/GitHub.VisualStudio/GitHubPackage.cs
@@ -253,7 +253,6 @@ namespace GitHub.VisualStudio
             }
             else if (serviceType == typeof(IUsageTracker))
             {
-                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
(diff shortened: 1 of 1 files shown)


src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -94,16 +89,6 @@ void LogVersionInformation()
         }
     }
 
-    [Export(typeof(IGitHubClient))]
-    public class GHClient : GitHubClient
-    {
-        [ImportingConstructor]
-        public GHClient(IProgram program)
-            : base(program.ProductHeader)
```
shana: Just FYI, the reason this was here is so that MEF imports requesting the client will pick it up with the correct ProductHeader value, which sets the user agent for the application on api requests. Since we don't have anything importing it anymore via MEF, it's fine to kill it.

PR #1536, opened 2018-03-14
Fix deadlock when VS is opended with GitHub pane visible

jcansdale:
As a workaround for potential delays before `await ThreadingHelper.SwitchToMainThreadAsync()` returns, we changed to using `await ThreadingHelper.MainThreadDispatcher.InvokeAsync(() => ...)`. Unfortunately rather than simply delays, this workaround can cause deadlock.

This PR:

- Revert back to using `await ThreadingHelper.SwitchToMainThreadAsync()`

These issues happen when Visual Studio is opened with the `GitHub` pane visible. It seems `SwitchToMainThreadAsync` doesn't return until the PR list has finished loading. 😨 

We do need to find out what's causing this, but in a separate issue/PR. For the moment, let's get rid of the deadlock!

### DIFF
--- a/src/GitHub.InlineReviews/InlineReviewsPackage.cs
+++ b/src/GitHub.InlineReviews/InlineReviewsPackage.cs
@@ -28,12 +28,10 @@ namespace GitHub.InlineReviews
             var componentModel = (IComponentModel)(await GetServiceAsync(typeof(SComponentModel)));
             var exports = componentModel.DefaultExportProvider;
 
-            // await ThreadingHelper.SwitchToMainThreadAsync() won't return until after a solution
-            // has been loaded. We're using the following instead as a workaround.
-            await ThreadingHelper.MainThreadDispatcher.InvokeAsync(() =>
-                menuService.AddCommands(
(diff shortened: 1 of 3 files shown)


shana:
If it's deadlocking, that means that the await for the InvokeAsync is happening on the main thread, that's the only thing I can think of that would cause a deadlock. Maybe it's possible that the `InitializeAsync` method is called on the main thread in some conditions. Can we repro the deadlock reliably? If we can, then maybe only doing the `InvokeAsync` if we're not on the main thread can solve this.

jcansdale:
@shana @grokys,

It seems we need to use the following:

```cs
// `JoinableTaskFactory` is a property on `AsyncPackage`
await JoinableTaskFactory
   .WithPriority(VsTaskRunContext.UIThreadNormalPriority)
   .SwitchToMainThreadAsync();
```

I guess the default priority for extensions marked with `AllowsBackgroundLoading = true` must be `UIThreadBackgroundPriority`. This is too low for setting up UI elements like commands and the PR status bar element.

I'll follow up with another issue/PR.

PR #1543, opened 2018-03-19
Fix a few theming issues

donokuda:
There's still a handful of things to address still, but I wanted to get these in sooner than later.

I think one of the last things to fix here is making the dropdown triangle the same color as `VsBrush.WindowKey`. However, I want to use the same type of triangle seen in other places inside of Visual Studio:

----

Screenshots

screen shot 2018-03-19 at 11 21 51 am
screen shot 2018-03-19 at 11 21 36 am
screen shot 2018-03-19 at 11 21 07 am


donokuda:
Pushed some changes to the "Create a review" view:

- Tighten up spacing between header + PR title
- Added a bit more spacing before the summary text field
- Fixed theming
- Change "Review summary" label to "Your review summary" (to be closer to the title of the view)


**Before**

screen shot 2018-03-19 at 2 28 41 pm
screen shot 2018-03-19 at 2 28 24 pm

**After**

screen shot 2018-03-19 at 2 23 59 pm
screen shot 2018-03-19 at 2 23 44 pm
screen shot 2018-03-19 at 2 23 28 pm

donokuda:
ok, I'm going to take a step back from this pull request for real 😛 

One thing I'm stuck on and could use some guidance is how to change the expander arrow/triangle to the default version that (I assume) we get from Visual Studio:

I believe the ones we use here are the [triangle down octicons](https://octicons.github.com/icon/triangle-down/) 

screen shot 2018-03-19 at 5 00 15 pm

But I'd rather use the ones we see in the headers for each section (next to "Description," "Reviewers," etc) 

screen shot 2018-03-19 at 4 59 14 pm
 
By using the default expander triangle, we'll fit better with the Visual Studio UI brand. Also hoping it's less code to maintain in the long run 🤞

PR #1545, opened 2018-03-20
Display Pull Request Reviews for a User.

grokys:
When the user clicks on a "Reviewer" item in the PR details view, navigate to a new view to show the PR reviews by that user.

image

This PR also introduces:

- A new global style for `Expander`: the default WPF expander style doesn't fit in well with Visual Studio
- A `TrimmedPathTextBlock` control to intelligently trim paths (see screenshot above)

Depends on #1523
Part of #1491

### DIFF
--- a/src/GitHub.Exports.Reactive/ViewModels/GitHubPane/IPullRequestReviewFileCommentViewModel.cs
+++ b/src/GitHub.Exports.Reactive/ViewModels/GitHubPane/IPullRequestReviewFileCommentViewModel.cs
@@ -1,0 +1,27 @@
+﻿using System;
+using System.Reactive;
+using ReactiveUI;
+
+namespace GitHub.ViewModels.GitHubPane
+{
+    /// <summary>
+    /// Represents a view model for a file comment in an <see cref="IPullRequestReviewViewModel"/>.
+    /// </summary>
+    public interface IPullRequestReviewFileCommentViewModel
+    {
+        /// <summary>
+        /// Gets the body of the comment.
+        /// </summary>
+        string Body { get; }
+
+        /// <summary>
+        /// Gets the path to the file, relative to the root of the repository.
+        /// </summary>
+        string RelativePath { get; }
+
+        /// <summary>
+        /// Gets a comment which opens the comment in a diff view.
+        /// </summary>
+        ReactiveCommand<Unit> Open { get; }
+    }
+}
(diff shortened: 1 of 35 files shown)


meaghanlewis:
This LGTM 👍

src/GitHub.Exports.Reactive/ViewModels/GitHubPane/IPullRequestReviewFileCommentViewModel.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// </summary>
+        string Body { get; }
+
+        /// <summary>
+        /// Gets the path to the file, relative to the root of the repository.
+        /// </summary>
+        string RelativePath { get; }
+
+        /// <summary>
+        /// Gets a comment which opens the comment in a diff view.
```
jcansdale: Should this be simply, "Opens the comment in a diff view"?

grokys: Ah, should be "Gets a command...". If you look around at C# documentation, properties always start with "Gets a" (for read-only properties) or "Gets or sets a" (for read/write properties). See https://softwareengineering.stackexchange.com/questions/193244/is-the-gets-or-sets-necessary-in-xml-documentation-of-properties and http://stylecop.soyuz5.com/SA1623.html.

src/GitHub.UI/Converters/TrimNewlinesConverter.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// An <see cref="IValueConverter"/> that trims newlines from a string and replaces them
+    /// with spaces.
+    /// </summary>
+    public class TrimNewlinesConverter : ValueConverterMarkupExtension<TrimNewlinesConverter>
+    {
+        public override object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            var text = value as string;
+            if (String.IsNullOrEmpty(text)) return null;
+            return Regex.Replace(text, @"\t|\n|\r", " ");
```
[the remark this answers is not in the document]
grokys: If we're displaying a comment on a single line, it would make sense to me to convert tabs to spaces, because otherwise we'd be wasting space. Tabs are generally used for alignment between lines, so if we've got no lines it makes sense to strip them. Do you have an example where this wouldn't be a good idea?

src/GitHub.Exports.Reactive/ViewModels/GitHubPane/IPullRequestReviewFileCommentViewModel.cs
jcansdale: Thanks, I hadn't noticed that convention.

src/GitHub.UI/Converters/TrimNewlinesConverter.cs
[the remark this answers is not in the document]
jcansdale: Oh, I was just comparing it to the comment: "trims newlines from a string and replaces them with spaces".

grokys:
@jcansdale addressed your feedback.

PR #1549, opened 2018-03-21
Display PR review comments inline.

grokys:
When the user clicks a review comment represented by a `PullRequestReviewFileCommentViewModel`, open the comment inline in a diff view:

This also handles showing outdated comments in a diff representing the commit at which the comment was left. Doing this required a few changes to the inline comment/PR session manager classes to make them accept a commit SHA instead of just assuming that the PR head will be needed.

Depends on #1545 
Part of #1491


meaghanlewis:
@grokys one thing I found interesting is that I couldn't get a diff view (the 3rd one) to open while reviewing comments.

one diff doesnt open

meaghanlewis:
I also found it interesting that for comment reviews, the comment is displayed as a **Description** versus comments or outdated comments.

description vs comment

PR #1552, opened 2018-03-21
Make Navigate to Editor more discoverable

### LINKED ISSUE #1546: Make navigation from diff view to editor more discoverable
- GitHub Extension for Visual Studio version: 2.4.4

At the moment navigating from the diff view to the editor isn't very discoverable. The user needs to have checked out out a PR and already know they can navigate by hitting `Enter` or right-clicking and selecting the the `Open File in Solution` button.

Educating users using an info banner is one option, but getting the balance between attracting attention without being intrusive can be a challenge. It would be better if we could give the user a non-intrusive reminder at the moment they might want to use it.

We could display, `Press Enter to navigate to solution file` if/when the user clicks on the diff view. This would be an action the user could take immediately that wouldn't have worked before they clicked. The status message would be a response to a user action and immediately actionable.

jcansdale:
### What this PR does

This PR gives users a hint that the Navigate to Editor functionality exists.

It adds the following messages:

1. When a PR file diff gets focus
image

2. When a PR file diff gets focus from a PR that isn't checked out
image

3. When user attempts to navigate to a file that isn't checked out
image

4. When the diff file focus is lost the status is cleared (showing the default `Ready` message)
image

- This also works when using `View File` to view a (read-only) PR file

### Reviewers

- What do you think of the wording? Is my alternative suggestion above an improvement?
- Do you think the status bar message is noticeable enough without being overly distracting?

### Todo

- ~~Finalize the wording. Would something like this be better?~~
  - ~~`Press Enter to navigate to Editor (PR branch must be checked out)`~~
- ~~It would be nice if the dialog gave `Checkout` and `Cancel` options~~~
  - ~~Maybe do this in~~ a separate PR

Fixes #1546

### DIFF
--- a/src/GitHub.App/Resources.Designer.cs
+++ b/src/GitHub.App/Resources.Designer.cs
@@ -280,7 +280,7 @@ namespace GitHub.App {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Checkout PR branch and press Enter to navigate to Editor.
+        ///   Looks up a localized string similar to Press Enter to navigate to Editor (PR branch must be checked out).
         /// </summary>
         internal static string NavigateToEditorNotCheckedOutStatusMessage {
             get {
--- a/src/GitHub.App/Resources.Designer.cs
+++ b/src/GitHub.App/Resources.Designer.cs
@@ -271,6 +271,33 @@ namespace GitHub.App {
         }
         
         /// <summary>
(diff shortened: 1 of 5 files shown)


meaghanlewis:
Have been dogfooding this and I think the message is helpful. It's not at all distracting, because the bar is always there- I'm just used of it only saying `Ready`.

PR #1556, opened 2018-03-22
[WIP] Show comments in editor when navigated to from PR

jcansdale:
##### Some Challenges

- The comment margin has an impact on editor performance
  - See if we can improve performace
  - Only show the comment margin when navigated to from a PR
- When there are no comments user might confuse comment margin with the breakpoint margin
  - Showing the add comment / `+` differentiates it from the default breakpoint margin
  - When navigated to from PR diff view the comment margin context is maintained
- Comments vanish when the code they were on changes
  - Let users open the last inline-comment they looked at using a code context menu item
  - Let users view the last comment they looked at using an icon on the status bar
  - Give the users an option to convert to source comments (e.g. `// TODO: #discussion_r171815350`).
- Showing comments inline can be disruptive when viewed inside the editor (code jumping around)
  - Allow users to preview comments without actually opening the inline comment

### DIFF
--- a/src/GitHub.VisualStudio/Views/GitHubPane/PullRequestDetailView.xaml.cs
+++ b/src/GitHub.VisualStudio/Views/GitHubPane/PullRequestDetailView.xaml.cs
@@ -8,6 +8,7 @@ using System.Windows.Controls;
 using System.Windows.Documents;
 using System.Windows.Input;
 using System.Windows.Media;
+using GitHub.Settings;
 using GitHub.Commands;
 using GitHub.Exports;
 using GitHub.Extensions;
@@ -74,6 +75,9 @@ namespace GitHub.VisualStudio.Views.GitHubPane
         [Import]
         IVsEditorAdaptersFactoryService EditorAdaptersFactoryService { get; set; }
 
+        [Import]
+        IPackageSettings PackageSettings { get; set; }
+
         protected override void OnVisualParentChanged(DependencyObject oldParent)
         {
             base.OnVisualParentChanged(oldParent);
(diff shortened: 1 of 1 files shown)


grokys:
I don't really see why we'd only display comments in editor when navigated to from a PR? I'd personally expect to be able to leave comments on any file no matter how the file was opened (assuming I'm on a PR branch). Having to open the file from the PR detail view has 2 disadvantages to me:

1. It's inconvenient, I'm used to opening files from the solution explorer or using Ctrl+,
2. It introduces hidden state: "I could leave comments a second ago, but then I closed and reopened the file and now I can't! What's wrong?"

sguthals:
I agree with @grokys 

@jcansdale is there different functionality that you are breaking up into separate PRs for when you are navigating from diff view vs opening the file outright?

jcansdale:
@grokys I've copied some notes from https://github.com/github/editor-tools/issues/246#issuecomment-374592694 into the description.

The main issues for me are:
- The comment margin has an impact on editor performance
- When there are no comments there margin is easily confused with breakpoint/glyph margin

We definitely need a way to let the user know when there are fresh comments associated with the current file. I'm not convinced that having the comment margin always on is the way to go.

There is however this option if the user wants it:
image

Do you remember #1088 and the surrounding discussions? I was reluctant to disable the comment margin by default. Now despite having the option, I tend to work with it turned off.

grokys:
Yes, I remember that discussion ;) My memory of it was that we were removing it temporarily and then would reinstate it when we were happy with it. The problems that I remember were the problems you listed, plus the fact that comments disappear when you change the line.

From what I remember we'd decided that once we'd addressed these problems (which is what we're investigating now!) we could reinstate the margin. I too tend to work with it turned off, but mainly because it's kinda useless because comments disappear when you edit the line ;)

jcansdale:
@sguthals,

> is there different functionality that you are breaking up into separate PRs for when you are navigating from diff view vs opening the file outright?

Yes, I'm breaking it into separate PRs. I'm trying to avoid an all or nothing, big bang PR. 😉

jcansdale:
@grokys,

I want to focus on discoverability and make it obvious to users when there are comments they might be interested in. Having the comment margin permanently visible is one way to make a certain kind of comment discoverable.

When I've worked with the comment margin permanently on, I've found it gets in the way sometimes (particularly when debugging). I think there are more subtle ways we could do this and surface more information to the user.

For example we might highlight a comment icon/menu on the margin when there are comments on the current document.

At it simplest it might look like this:
comment_menu_mockup

This would allow the user to toggle the comment margin or navigate to a comment (implicitly enabling the margin).

I'd like to go further than this and also surface:
- When a new comment arrives for the current PR
- Navigate to a comment in the current document
- Navigate to a comment on the current PR
- Navigate to the most recently opened comment
  - If the user had started a reply, this should be restored
- Navigate to the PR reviews page
  - Maybe for a particular reviewer

There are a bunch of possibilities and we'd need to think about what's most relevant to use user given the context.

As an analogy, it would be similar to the Git branch menu (which shows branches common branch operations) except for comments:
image

The comment icon/UI would be in the corner of the document (under where the comment margin appears) and so would be more prominent than the status bar controls.

jcansdale:
Obsolete.

PR #1559, opened 2018-03-22
[spike] Feature inline comments in editor

jcansdale:
This is a spike to experiment with the responding to PR feedback workflow.

It brings together the following PR branches:
- Make Navigate to Editor more discoverable: #1552
- Show comments in editor when navigated to from PR: #1556
- Preview inline comments as a tooltip: #1091 
- Allow user to toggle inline comment margin: #1568

Here's a [screen-cast](https://files.slack.com/files-pri/T0CA8C346-F9UCVA31R/2018-03-22_16-42-10.gif) showing a possible workflow:

The screen-cast goes smoothly until the end when I want to respond to the user who's comment I made changes based on. Because I modified the line the comment is on, it loses its context and vanishes. 😢 

#### Related

- WIP: Improve tracking of lines with inline-comments #1032


jcansdale:
We could find somewhere to surface comments that are stale in the editor but not stale on the PR/diff view. For example we might put them at the bottom of the margin.

image

The user could then easily return to the comments they've addressed and describe the changes they've made.

jcansdale:
Obsolete.

PR #1562, opened 2018-03-23
Start PR Reviews from Inline Comments

grokys:
This PR adds a "Start a review" button to the inline comment peek view. When clicked it will start a pending review, and from this point all inline comments will be added to the pending review until the review is submitted:

Depends on #1549 
Part of #1491

### DIFF
--- a/src/GitHub.App/Api/ApiClient.cs
+++ b/src/GitHub.App/Api/ApiClient.cs
@@ -48,6 +48,27 @@ namespace GitHub.Api
             return (isUser ? client.Create(repository) : client.Create(login, repository));
         }
 
+        public IObservable<PullRequestReview> PostPullRequestReview(
+            string owner,
+            string name,
+            int number,
+            string commitId,
+            string body,
+            PullRequestReviewEvent e)
+        {
+            Guard.ArgumentNotEmptyString(owner, nameof(owner));
+            Guard.ArgumentNotEmptyString(name, nameof(name));
+
+            var review = new PullRequestReviewCreate
+            {
+                Body = body,
+                CommitId = commitId,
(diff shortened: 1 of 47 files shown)


meaghanlewis:
This PR looks good to me 👍 

I mainly followed scenarios that I outlined [here](https://github.com/github/editor-tools/pull/289/files#diff-a3aac46ce495c858d9ad5fb102ffe63bR92)

Just a couple things to note:

1. The error that I get when trying to request changes when there are no comments is too long and seems out of place.

screen shot 2018-03-27 at 9 24 56 am


For other errors don't we usually have a banner at the top (like 👇 ) ?
error banner

cc @grokys @donokuda 

2. The cancel button doesn't seem to do anything
cancel doesn t do anything

donokuda:
> For other errors don't we usually have a banner at the top (like 👇 ) ?

Yup, that's correct. I think the banner up there makes the most sense to show the error.

sguthals:
I agree with @meaghanlewis regarding the error message. 

I'm also noticing on VS2017 the font color is too dark for the changes list:
screen shot 2018-03-28 at 6 28 30 pm

I also noticed that if you have "" in your comments you get an error:
screen shot 2018-03-28 at 6 31 55 pm

screen shot 2018-03-28 at 6 32 10 pm

I took out the quotes and it worked fine. 

The interesting thing is that I started the inline comment, clicked "Start Your Review" and it started it, because I was able to see the "Continue your review", but the comment didn't actually get submitted until I took out the quotes.

And I confirmed only one version of the inline comment was saved:
screen shot 2018-03-28 at 6 34 12 pm

Is that a known issue?

PR #1567, opened 2018-03-27
Add PR reviews metrics.

grokys:
Adds two new metrics:

- `NumberOfPRReviewDiffViewInlineCommentStartReview`: incremented when the "Start Review" button is clicked from an inline comment
- `NumberOfPRReviewPosts`: incremented when a PR review is submitted

Depends on #1562 
Part of #1491

### DIFF
--- a/src/GitHub.Exports/Models/UsageModel.cs
+++ b/src/GitHub.Exports/Models/UsageModel.cs
@@ -42,6 +42,8 @@ namespace GitHub.Models
         public int NumberOfPRDetailsNavigateToEditor { get; set; }
         public int NumberOfPRReviewDiffViewInlineCommentOpen { get; set; }
         public int NumberOfPRReviewDiffViewInlineCommentPost { get; set; }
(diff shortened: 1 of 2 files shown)


meaghanlewis:
These metrics look good to me! Tested on VS2015 & VS2017

sguthals:
The first time I started a review from a diff and submitted that review it did not get recorded. I've tried twice since then and it worked. I'm going to try re-starting my experimental instance to see if the first one works when I re-start.

sguthals:
Tests:
2015 Dogfood:
- [x] responding to a comment to start a review did not cause the comment to remain open
- [x] all metrics were updated

2015 Experimental:
- [x] responding to a comment to start a review did not cause the comment to remain open
- [x] all metrics were updated
- [x] This time I submitted as comment only and it increased `numberOfPRReviewPosts` - which I believe is intended behavior, I just wanted to make sure 👍 

2017 Experimental:
- [x] responding to a comment to start a review did not cause the comment to remain open
- [x] all metrics were updated
- [x] This time I submitted as approved and it still worked 👍

PR #1568, opened 2018-03-28
Allow user to toggle inline comment margin

### LINKED ISSUE #1565: Let user know when there are comments on currently edited document
- GitHub Extension for Visual Studio version: 2.4.4

We need a way to let the user know when there are comments on the source file they're currently editing.

There are different ways to achieve this.

- Have the comment margin permanently enabled

- Show how many comments there are in current document and allow user to navigate to comments/toggle the margin

image

Some ideas here https://github.com/github/VisualStudio/pull/1556#issuecomment-376120684.

### Related

- Show comments in editor when navigated to from PR: #1556
- Disable PR comment margin in editor by default: #1088

jcansdale:
### What this PR does

- New horizontal editor margin is hidden behind feature flag
  - image

- Adds a margin to the editor window
  - image

- Clicking on checkbox (or filename/comment bubble) toggles the comment margin
  - image

- Clicking on `view x comments` navigates to PR file changes
  - image

- If there are no PR file comments it will show `view changes`
  - image

- If there are no PR file changes the margin won't appear
  - image

- Right clicking on solution file with active PR branch shows `View Changes in # PR` button 
  - image

- Right clicking a diff view opened using `View Changes in Solution` slows the option `Open File in Code View`. This will open the diff pane in a new code view.
image
  - This also works with the diff view opened using the default `Team Explorer - Changes` action `Compare with Unmodified...`

- Clicking `View Changes in # PR` navigates to equivalent line on PR diff view
  - image

- Clicking `View Changes in # PR` on file without changes in PR shows warning on status bar
  - image

### Metrics

- `NumberOfPullRequestFileMarginToggleInlineCommentMargin` - When the user toggles the checkbox to show/hide comment margin
image

- `NumberOfPullRequestFileMarginViewChanges` - When user clicks on view changes/view comments link
image

- `NumberOfNavigateToPullRequestFileDiff` - When user navigates from editor to PR file diff via the `GitHub.GoToSolutionOrPRFile` command
image

- `NumberOfPRDetailsNavigateToEditor` - When user navigates from PR file diff to editor via the `GitHub.GoToSolutionOrPRFile` command (NOTE: this counter existed before)

image

- `ExecuteToggleInlineCommentMarginCommand` - When the `GitHub.ToggleInlineCommentMargin` command is executed (via the checkbox or executing command directly)
image

### Technical notes

- When opening a comparison window, the `DifferenceViewerOptions.ScrollToFirstDiffId` global option will be temporarily set to `false` (see https://github.com/github/VisualStudio/issues/1631). This lets me immediately set the equivalent caret position when navigating to the the diff view.

### Todo

- [x] Add metrics
  - [x] Add counter for `Open File in Code View` actions
  - [x] Add counter for executing `GoToSolutionOrPullRequestFileCommand`
- [x] Make `View File` show the comment margin
- [x] Fix issue where diff view moves to first change. see:
image

### Issues

- [x] The comments margin doesn't always appear when the comment bubble icon is clicked
  - Same issue when comments margin is always on
- [x] The UI appears on all text view types (including diff view)

### Initial thoughts

My initial impression is that this UI element would be a great opportunity to surface other important information about the current PR. For example:

- The PR number could be shown there
- How many (non-inline) comments comments there are and quick navigation to them
- Show the files that are part of the PR and allow navigation to them
- Navigate to any of the inline comments on the current PR file
- Show if there are any unread comments on the PR and notify when they arrive

Lots of opportunities to show things there would be useful to the user and make this UI element more discoverable (e.g. showing the PR number would be handy for navigation and give context to the comment bubble).

image

### Related UI

People will rarely see the following UI.  They need to be using an Enterprise/Professional version of VS and be editing a non-source file. This is when editing our `GitHub.VisualStudio.vsct` file:

image

When the user clicks on one of the sections, a bubble pops up with a scrollable list of sub-items with an automatic preview of each.

image

We could do a similar thing, except with a list of inline comments in the file and a preview of each comment. When the user clicks on the comment they're interested it, it would navigate to the inline comment in the source.

Similarly we could have a list of PR review comments, with a preview and navigation to the `GitHub` PR review UI.

Lastly when the user clicks on the PR number, there could be a list of changed files in the PR and the number of inline comments they have displayed. Clicking in the file would navigate to the PR file diff or editor.

This would give context for the comment margin toggle switch (which I think is too minimal/subtle at the moment).

### Tackle in another PR (#1671)

- Make comments appear on `View File` (margin appears but no comments 😕)
- Make comments appear on RHS `View Changes in Solution` 

Fixes #1565


src/GitHub.InlineReviews/Views/CommentView.xaml
```diff
@@ -118,7 +118,8 @@
                               Text="{Binding Body, UpdateSourceTrigger=PropertyChanged}"
                               TextWrapping="Wrap"
                               VerticalAlignment="Center"
-                              GotFocus="ReplyPlaceholder_GotFocus">
+                              GotFocus="ReplyPlaceholder_GotFocus"
+                              SpellCheck.IsEnabled="true">
```
jcansdale: Oops, this crept in by mistake. I think we want it, but not as part of this PR!

src/GitHub.InlineReviews/Commands/EnableInlineCommentsCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.TextManager.Interop;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Editor;
+using GitHub.Services;
+
+namespace GitHub.InlineReviews.Commands
+{
+    [Export(typeof(IEnableInlineCommentsCommand))]
+    public class EnableInlineCommentsCommand : VsCommand, IEnableInlineCommentsCommand
```
jcansdale: This should be `ToggleInlineCommentsCommand`.

src/GitHub.InlineReviews/Tags/InlineCommentTagger.cs
```diff
@@ -163,9 +167,16 @@ public IEnumerable<ITagSpan<InlineCommentTag>> GetTags(NormalizedSnapshotSpanCol
...
             if (relativePath != null)
             {
-                var liveFile = await sessionManager.GetLiveFile(relativePath, view, buffer);
-                fileSubscription = liveFile.LinesChanged.Subscribe(LinesChanged);
-                file = liveFile;
+                file = await sessionManager.GetLiveFile(relativePath, view, buffer);
+
+                var options = view.Options;
+                visibleSubscription =
+                    Observable.FromEventPattern<EditorOptionChangedEventArgs>(options, nameof(options.OptionChanged))
```
jcansdale: This creates a subscription for when the `InlineCommentMarginVisible` editor option changes. There might be multiple events fired where the visibility hasn't changed. I was want one initial event initializing the visibility. Is this the cleanest/simplest way of doing that?

src/GitHub.InlineReviews/CommentsMargin.cs
```diff
@@ -0,0 +1,122 @@
...
+using GitHub.InlineReviews.Views;
+using GitHub.InlineReviews.ViewModels;
+using GitHub.InlineReviews.Commands;
+using Microsoft.VisualStudio.Text.Editor;
+using ReactiveUI;
+using Task = System.Threading.Tasks.Task;
+
+namespace GitHub.InlineReviews
+{
+    internal class CommentsMargin : IWpfTextViewMargin
```
jcansdale: I need to find a better name of this margin. It's the horizontal margin that currently appears when editing a PR file in the working directory.

Ultimately it could give more information about the PR the user is working on and include commands to jump to the next PR file.

Maybe `PullRequestMargin` would be a better name? It will only be visible when working on a checked out pull request.

src/GitHub.VisualStudio/Commands/OpenFileInSolutionCommand.cs
```diff
@@ -0,0 +1,157 @@
...
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Editor;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.TextManager.Interop;
+using Task = System.Threading.Tasks.Task;
+
+namespace GitHub.Commands
+{
+    [Export(typeof(IOpenFileInSolutionCommand))]
+    public class OpenFileInSolutionCommand : VsCommand, IOpenFileInSolutionCommand
```
jcansdale: Need to come up with a better name for this command. It's responsible for navigating between a pull request file and the equivalent working directory file (and back again). Something like, `OpenPullRequestOrWorkingFile`?

src/GitHub.VisualStudio/Commands/OpenFileInSolutionCommand.cs
```diff
@@ -0,0 +1,157 @@
...
+                    {
+                        ShowInfoMessage(App.Resources.NavigateToEditorNotCheckedOutInfoMessage);
+                        return;
+                    }
+
+                    // Navigate from PR file to solution
+                    await usageTracker.Value.IncrementCounter(x => x.NumberOfPRDetailsNavigateToEditor);
+
+                    await pullRequestEditorService.Value.OpenFile(info.Session, info.RelativePath, true);
+                    var fileView = FindActiveView();
```
jcansdale: I think we need a delay here. Sometimes it doesn't pick up the destination file. :-(

meaghanlewis:
>My initial impression is that this UI element would be a great opportunity to surface other important information about the current PR. For example:

>- The PR number could be shown there
>- How many (non-inline) comments comments there are and quick navigation to them
>- Show the files that are part of the PR and allow navigation to them
>- Navigate to any of the inline comments on the current PR file
>- Show if there are any unread comments on the PR and notify when they arrive

☝️ these are great ideas. If they are not already captured somewhere, it would be good to move them to an issue.

meaghanlewis:
Hey @jcansdale here are a few things I noticed today:

1. The command **GitHub.ToggleInlineCommentMargin** is not actually toggling the comment margin for me. Although every time I execute the command the metric: `ExecuteToggleInlineCommentMarginCommand` increments just fine.

2. Some of the metrics don't seem to be incrementing correctly. 

`NumberOfNavigateToCodeView` is always 0 even after I select the option
image 2

`NumberOfPullRequestFileMarginViewChanges` is always 0 even after I click on the view comments link.
39930471-b713f17c-5532-11e8-8d3b-8f186d52b073

This is a bit more vague, but I've noticed that some metrics aren't incrementing regularly for: `NumberOfPRDetailsNavigateToEditor`, `NumberOfNavigateToPullRequestFileDiff`, or `NumberOfPullRequestFileMarginToggleInlineCommentMargin`. Instead when I perform those actions it appears to increment `ExecuteGoToSolutionOrPullRequestFileCommand` instead.

metrics for viewing inline comments

StanleyGoldman:
Should it locate me to the comment when i click the "view comments" in the task bar?

src/GitHub.App/Services/PullRequestEditorService.cs
```diff
@@ -229,33 +247,61 @@ public class PullRequestEditorService : IPullRequestEditorService
...
 
-            VisualStudio.Services.Dte.Commands.Raise(
-                Guids.CommandSetString,
-                PkgCmdIDList.NextInlineCommentId,
-                ref param,
-                null);
+            // HACK: We need to wait here for the inline comment tags to initialize so we can find the next inline comment.
+            // There must be a better way of doing this.
+            await Task.Delay(1500);
+            RaiseWhenAvailable(Guids.CommandSetString, PkgCmdIDList.NextInlineCommentId, param);
```
grokys: I thought this hack was no longer necessary with `DifferenceViewerOptions.ScrollToFirstDiffId`?

src/GitHub.InlineReviews/Margins/InlineCommentMarginEnabled.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.Utilities;
+using Microsoft.VisualStudio.Text.Editor;
+
+namespace GitHub.InlineReviews.Margins
+{
+    [Export(typeof(EditorOptionDefinition))]
+    [Name(OptionName)]
+    public class InlineCommentMarginEnabled : ViewOptionDefinition<bool>
```
grokys: I've never used this API, so I'm missing something here, but the only thing that seems to be used from this class is `OptionName`. Do we need to create a subclass of `ViewOptionDefinition` if we only use the name? Why not use an instance of this class as they key when getting options?

src/GitHub.InlineReviews/Views/CommentView.xaml
```diff
@@ -118,7 +118,8 @@
                               Text="{Binding Body, UpdateSourceTrigger=PropertyChanged}"
                               TextWrapping="Wrap"
                               VerticalAlignment="Center"
-                              GotFocus="ReplyPlaceholder_GotFocus">
+                              GotFocus="ReplyPlaceholder_GotFocus"
+                              SpellCheck.IsEnabled="True">
```
grokys: Wasn't there another PR with this feature?

src/GitHub.VisualStudio/Commands/GoToSolutionOrPullRequestFileCommand.cs
```diff
@@ -0,0 +1,228 @@
...
+using Microsoft.VisualStudio.Editor;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Text.Editor;
+using Microsoft.VisualStudio.Text.Differencing;
+using Task = System.Threading.Tasks.Task;
+
+namespace GitHub.Commands
+{
+    [Export(typeof(IGoToSolutionOrPullRequestFileCommand))]
+    public class GoToSolutionOrPullRequestFileCommand : VsCommand, IGoToSolutionOrPullRequestFileCommand
```
grokys: Could we have a comment here (and on the interface) as it's not totally clear from the class name what this command will do.

jcansdale:
@StanleyGoldman,

> Should it locate me to the comment when i click the "view comments" in the task bar?

Good point. It doesn't locate you on the comment, it simply navigates to show changes in the PR (that happens to have x number of comments). The use can then view an already open comment or navigate to the next/previous comment.

Maybe, `view changes in PR (x comments)` would be clearer?

src/GitHub.App/Services/PullRequestEditorService.cs
jcansdale: So did I! 😢  It seems there's another issue where the `NextInlineCommentId` depends on the comment margin tags having been updated in order to navigate to the next comment. Before the tags have been refreshed, this command is unavailable and it would crash when attempting to execute the command.

We should attempt to fix this in another PR.

src/GitHub.InlineReviews/Margins/InlineCommentMarginEnabled.cs
jcansdale: By exporting `ViewOptionDefinition` we declare than a new options exists on the text views (and give it a default value). I probably shouldn't be using `OptionName` directly and should be importing the option and using the `Key` property. This gives us type information for the property.

src/GitHub.InlineReviews/Views/CommentView.xaml
jcansdale: Well, there will be.  I didn't try to sneak this one in, seriusly! ;-)

src/GitHub.VisualStudio/Commands/GoToSolutionOrPullRequestFileCommand.cs
jcansdale: Done. I hope it make sense. Let me know if it isn't clear.

jcansdale:
@meaghanlewis,

> Some of the metrics don't seem to be incrementing correctly.

Thanks for the heads up! I don't understand why this is happening. 😕 I'll investigate more tomorrow.

jcansdale:
@meaghanlewis,

> The command `GitHub.ToggleInlineCommentMargin` is not actually toggling the comment margin for me. Although every time I execute the command the metric: ExecuteToggleInlineCommentMarginCommand increments just fine.

The `GitHub.ToggleInlineCommentMargin` command toggles an option on the current editor that determines whether the comment margin is visible in certain circumstances (the user is on a PR branch and editing a file that has changed in the PR).

When you execute `GitHub.ToggleInlineCommentMargin` from the `Command Window`, it will toggle this option on the command window itself.  If you were to bind the command to a keyboard shortcut, you should find it works as expected (I've bound it to `Alt+G, Alt+T`).

meaghanlewis:
@jcansdale it's still hard for me to understand what `GitHub.ToggleInlineCommentMargin`. I don't see it have any affect on whether the comment margin is visible or not when viewing a changed file on a PR (with the branch checked out). Could you show me perhaps?

meaghanlewis:
Also, @sguthals and I were noticing that when enabling or disabling the feature flag, that a file (or files) need to be closed and reopened to actually see the options to toggle the inline comment margin. This is probably to be expected, and I'm sure users won't just be toggling the flag often, but just wanted to bring it up.

jcansdale:
@meaghanlewis,

> it's still hard for me to understand what `GitHub.ToggleInlineCommentMargin`. I don't see it have any affect on whether the comment margin is visible or not when viewing a changed file on a PR (with the branch checked out).

I think the confusion might be that it's a per-editor view setting, not a global show comments in the margin for all file setting. It is triggered when you click on the check box (when the horizontal margin is visible).

image

I made it that way because the horizontal margin is constructed per-editor view. It might make more sense for it to be a global show the editor margin setting. The checkbox would remain set when moving between PR files and executing `GitHub.ToggleInlineCommentMargin` would toggle the checkbox on and off. I think that would be a lot less confusing!

> Also, @sguthals and I were noticing that when enabling or disabling the feature flag, that a file (or files) need to be closed and reopened to actually see the options to toggle the inline comment margin. This is probably to be expected, and I'm sure users won't just be toggling the flag often, but just wanted to bring it up.

Yes, that is by design. To make it dynamic, the margin would need to be loaded but hidden. I've made it so the switch completely disables this functionality. I got bitten by this once before when hidden functionality was causing issues.

PR #1569, opened 2018-03-28
Initialize GitHubPane and menus asynchronously

### LINKED ISSUE #1550: GitHub pane causes performance warning banner

jcansdale:
As suggested by @AArnott  and @BertanAygun in #1560.

### This PR

- Makes `GitHubPane` initialize MEF on a background thread
- Uses dedicated package for `GitHubPane` without MEF dependency
- Creates commands using MEF on a background thread
- Expose `IGitHubPaneViewModel` via `GetViewModelAsync`
- Logs initialization exceptions in a `TextBox`

This PR would make #1560 obsolete.

### Reviewers

- Should we show anything while MEF is initializing (it doesn't look too bad if left blank)
- If an exception is thrown, I'm simply showing it in a `TextBox` (to allow the user to copy/paste). Is this enough?
- ~~Please sanity check [AsyncMenuPackage](https://github.com/github/VisualStudio/blob/f66afaf8e99e9b3f7a7756e90295d73682f4ece4/src/GitHub.Services.Vssdk/AsyncMenuPackage.cs)~~ (it was crazy and is now gone 😉)

### Todo

- Decide what to display when MEF is loading/refreshing
- Decide how to present exceptions

Fixes #1550


AArnott:
> Do we want to display anything while MEF is loading/refreshing?

You could do what we do with 15.6 async tool windows and display a caption and a spinner control indicating that you're initializing.

> What should happen if MEF cache is corrupt/out-of-date and tool window throws?

If querying MEF throws an exception to you indicating the cache is out of date, you could display an error on your tool window advising the user of this state so they can restart.

src/GitHub.InlineReviews/InlineReviewsPackage.cs
```diff
@@ -18,28 +16,13 @@ namespace GitHub.InlineReviews
...
-            await JoinableTaskFactory.RunAsync(VsTaskRunContext.UIThreadNormalPriority, InitializeMenus);
-        }
-
-        async Task InitializeMenus()
-        {
-            var menuService = (IMenuCommandService)(await GetServiceAsync(typeof(IMenuCommandService)));
-            var componentModel = (IComponentModel)(await GetServiceAsync(typeof(SComponentModel)));
-            var exports = componentModel.DefaultExportProvider;
-
-            await JoinableTaskFactory.SwitchToMainThreadAsync();
```
AArnott: I think you should bring back the `SwitchToMainThreadAsync` call that you removed. I expect the `menuService` is an STA COM object that requires the UI thread.

src/GitHub.Services.Vssdk/AsyncMenuPackage.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace GitHub.Services.Vssdk
+{
+    public abstract class AsyncMenuPackage : AsyncPackage
+    {
+        IVsUIShell vsUIShell;
+
+        sealed protected async override Task InitializeAsync(CancellationToken cancellationToken, IProgress<ServiceProgressData> progress)
+        {
+            vsUIShell = await GetServiceAsync(typeof(SVsUIShell)) as IVsUIShell;
```
AArnott: You should use `await this.JoinableTaskFactory.SwitchToMainThreadAsync(CancellationToken);` as the first line in your method. Casting the SVsUIShell object to an interface itself generally requires the UI thread, and otherwise you're vulnerable to a deadlock by relying on RPC to get you to the UI thread.

src/GitHub.Services.Vssdk/AsyncMenuPackage.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            var menuCommandService = (OleMenuCommandService)(await GetServiceAsync(typeof(IMenuCommandService)));
+            await InitializeMenusAsync(menuCommandService);
```
AArnott: Can you pass `CancellationToken` into this method so that it can cancel async init operations if VS starts to shut down?

src/GitHub.Services.Vssdk/AsyncMenuPackage.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        protected abstract Task InitializeMenusAsync(OleMenuCommandService menuService);
+
+        // The IDesignerHost, ISelectionService and IVsUIShell are requested by the MenuCommandService.
+        // This override allows IMenuCommandService.AddCommands to be called form a background thread.
+        protected override object GetService(Type serviceType)
+        {
+            if (serviceType == typeof(SVsUIShell))
+            {
+                return vsUIShell;
```
AArnott: Your comment here is interesting. Returning this STA COM object object from a free-threaded method doesn't solve the problem that the caller will very likely cast this to an interface while still on the background thread that your comment suggests it will be on. You mustn't cast COM objects to any interface, or invoke those interfaces, while on a background thread unless you're sure they are free-threaded objects (and I don't think this one is one of those few).

src/GitHub.VisualStudio/UI/GitHubPane.cs
```diff
@@ -83,17 +88,26 @@ public void Initialize(IServiceProvider serviceProvider)
         {
             if (!initialized)
             {
-                var provider = VisualStudio.Services.GitHubServiceProvider;
-                var teServiceHolder = provider.GetService<ITeamExplorerServiceHolder>();
-                teServiceHolder.ServiceProvider = serviceProvider;
+                InitializeAsync(serviceProvider).Forget();
```
AArnott: If this async method throws, you'll never know. 
I find it's best to create an extension method on Task (we call ours `FileAndForget()`) that will attach an OnlyOnFaulted continuation (or await) on the Task and file a failure report, or display an error to the user, or whatever... something to draw attention to the failure so it's more diagnosable.

src/GitHub.VisualStudio/UI/GitHubPane.cs
```diff
@@ -83,17 +88,26 @@ public void Initialize(IServiceProvider serviceProvider)
...
+            var provider = VisualStudio.Services.GitHubServiceProvider;
+            var teServiceHolder = provider.GetService<ITeamExplorerServiceHolder>();
+            teServiceHolder.ServiceProvider = serviceProvider;
 
-                View = factory.CreateView<IGitHubPaneViewModel>();
-                View.DataContext = viewModel;
-            }
+            var factory = provider.GetService<IViewViewModelFactory>();
+            viewModel = provider.ExportProvider.GetExportedValue<IGitHubPaneViewModel>();
+            viewModel.InitializeAsync(this).Forget();
```
AArnott: Same here. As I sometimes (flippantly) say around my own team, everywhere you use `Forget()` consider removing the line entirely -- since evidently you don't care whether it succeeds or fails. 😉

src/GitHub.Services.Vssdk/AsyncMenuPackage.cs
BertanAygun: There is an helper method btw that you can use as ServiceProvider.GetGlobalServiceAsync<SVsUIShell, IVsUIShell> that handles this automatically.

jcansdale:
@AArnott @BertanAygun, Thank for the feedback! I'll dig into this properly tomorrow.

Just one quick question.

> You mustn't cast COM objects to any interface, or invoke those interfaces, while on a background thread unless you're sure they are free-threaded objects (and I don't think this one is one of those few).

How do I know what VS services it's okay retrieve using `GetServiceAsync` and call on a background thread?

I know the `menuService` is a managed object, but it retrieves `SVsUIShell` via a call to `GetService`. I thought this was why it was failing when used from an `AsyncPackage`. Are you saying that because depends on `SVsUIShell`, it isn't safe to use from a background thread?

src/GitHub.VisualStudio/UI/GitHubPane.cs
```diff
@@ -83,17 +88,26 @@ public void Initialize(IServiceProvider serviceProvider)
...
-                var provider = VisualStudio.Services.GitHubServiceProvider;
-                var teServiceHolder = provider.GetService<ITeamExplorerServiceHolder>();
-                teServiceHolder.ServiceProvider = serviceProvider;
+                InitializeAsync(serviceProvider).Forget();
+            }
+        }
+
+        async Task InitializeAsync(IServiceProvider serviceProvider)
+        {
+            // Allow MEF to initialize its cache asynchronously
```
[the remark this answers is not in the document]
AArnott: The threading rules state async methods should *not* throw when called on the wrong thread, but rather should simply switch to the thread they require (using `await SwitchToMainThreadAsync` or `await TaskScheduler.Default`.)

src/GitHub.Services.Vssdk/AsyncMenuPackage.cs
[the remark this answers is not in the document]
AArnott: You could... but then as an async method of its own, it *should* be sure to switch within the method itself.

AArnott:
> How do I know what VS services it's okay retrieve using GetServiceAsync and call on a background thread?

We don't have a great answer for that. The best we have is "assume all services you query IServiceProvider/IAsyncServicePRovider for require the UI thread for all access unless documentation explicitly states it's a free-threaded object".

`SComponentModel` *is* a free-threaded object. 

We are formulating a better answer to this with the [vssdk-analyzers](https://github.com/microsoft/vssdk-analyzers) but it's not ready yet.

jcansdale:
> There is an helper method btw that you can use as `ServiceProvider.GetGlobalServiceAsync<SVsUIShell, IVsUIShell>` that handles this automatically.

Thanks for the tip. Is there an equivalent if this given an `IServiceProvider` object? I'd like something that I'd be able to call from my unit tests. For example, say I want to get hold of the `DTE` object from a free-threaded MEF component. 

I'm guessing I could do something like this (not very pretty):
```
        [ImportingConstructor]
        FindDte([Import(typeof(SVsServiceProvider))] IServiceProvider sp)
        {
            var asp = (IAsyncServiceProvider)sp.GetService(
                typeof(Microsoft.VisualStudio.Shell.Interop.SAsyncServiceProvider));
            dte = (DTE)asp.GetServiceAsync(typeof(DTE)).Result;
        }
```

or using the `GetGlobalServiceAsync` method you mentioned:
```
        FindDte()
        {
            dte = await ServiceProvider.GetGlobalServiceAsync<EnvDTE.DTE, EnvDTE.DTE>();
        }
```

Is there a better way that's both concise and unit testable?

jcansdale:
> We don't have a great answer for that. The best we have is "assume all services you query IServiceProvider/IAsyncServicePRovider for require the UI thread for all access unless documentation explicitly states it's a free-threaded object".

Once I have a COM object, is there any way I can interrogate it to see if it's a `STA` component? I'm guessing not and the difference is baked into the code rather than the interfaces/meta-data?

I've noticed that VS SDK sample code tends to use `GetService(typeof(SVsService)) as IVsService` rather than `(IVsService)GetService(typeof(SVsService))`. Is there a reason for this beyond wanting to be sure that the cast doesn't throw? Generally I'd want that extra information to show up in my logs, to I'd cast rather than use `as`. Am I doing it wrong.

Thanks!

AArnott:
> Once I have a COM object, is there any way I can interrogate it to see if it's a STA component? I'm guessing not and the difference is baked into the code rather than the interfaces/meta-data?

There is not. 

> I've noticed that VS SDK sample code tends to use GetService(typeof(SVsService)) as IVsService rather than (IVsService)GetService(typeof(SVsService)).

I have [a blog post on this topic](https://aka.ms/ascast) that might help. In VS, services/packages may fail to load (or simply be skipped such as in safe mode) so we try to code defensively, in expecting nulls from `GetService`. As for cast failures, if we're prepared for a null result, then we'd already be prepared for a cast failure as well so long as it returns null instead of throwing. There are many valid reasons for a cast to fail when considering the code we write may support older versions of VS before the interface was implemented (or was written when an interface was still in development and not guaranteed to be present). Since so many times there is a valid reason for a null result or failed cast, I personally tend to (and advocate for) using `as` cast and failing gracefully on a null result wherever possible, without asking whether it's appropriate in this case since it does no harm if it will never happen, and the pattern is often good.

Similarly, I advocate for using `GetServiceAsync` whenever I'm in an async method even if the service I'm querying for may not (currently) be async. That way when the service becomes available asynchronously in the future, it's not _my_ code that's holding the user back from experiencing the full benefits of not blocking the UI thread any more.

AArnott:
> Is there an equivalent [of GetServiceAsync<TService, TInterface>] given an IServiceProvider object?

I don't think so, and certainly one that makes it safe to call from a background thread. All `GetService` calls should be done from the UI thread -- for unit tests it seems perfectly excusable to call from any thread (if it works), but never in production code.

@jcansdale  You included a snippet where you had an `ImportingConstructor` with a `GetService` call. I advise against this, because MEF is a free-threaded system and your importer (generally speaking) may be on any thread, you shouldn't assume you're on the UI thread as required for calling `GetService`. If you do assume it, then it can deadlock when you're not on that thread due to your caller's thread.

MEF part initialization should be free-threaded. If it has thread affinity, move it to a method that you'll call on the object after importing it.

jcansdale:
@AArnott Thanks for your tips/advice. I've taken note.

I've updated this PR with the changes you requested. Any chance you could take another look?

jcansdale:
@AArnott,

> MEF part initialization should be free-threaded. If it has thread affinity, move it to a method that you'll call on the object after importing it.

The problem is my MEF component isn't tied to a particular package that it would be convenient to initialize it from. It could be used by a number of other MEF components.

Would something like this be okay?

```
        [ImportingConstructor]
        FindDte([Import(typeof(SVsServiceProvider))] IServiceProvider sp)
        {
            InitializeAsync(sp).FileAndForget();
        }

        async Task InitializeAsync(IServiceProvider sp)
        {
            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
            dte = (DTE)sp.GetService(typeof(DTE));
        }
```
(assuming it was able to deal with `dte` being null for a while)

I'd add a shim that makes `SwitchToMainThreadAsync` into a nop when testing.

src/GitHub.VisualStudio/UI/GitHubPane.cs
```diff
@@ -131,6 +144,16 @@ public override void OnToolWindowCreated()
             UpdateSearchHost(pane?.IsSearchEnabled ?? false, pane?.SearchQuery);
         }
 
+        void ShowInitializing()
+        {
+            View = new Label { Content = "Initializing MEF" };
```
grokys: I know you were asking what to display here. We probably shouldn't display a spinner as it will be _really_ jerky while MEF is initializing. I think the following should suffice:

```
View = new TextBlock
{
	Text = "Initializing...",
	HorizontalAlignment = HorizontalAlignment.Center,
	VerticalAlignment = VerticalAlignment.Center,
};
```

I changed the text to just "Initializing..." as MEF might not mean much to users.

src/GitHub.VisualStudio/UI/GitHubPane.cs
```diff
@@ -131,6 +144,16 @@ public override void OnToolWindowCreated()
...
+        }
+
+        void ShowError(Exception e)
+        {
+            View = new TextBox { Text = e.ToString() };
```
grokys: Should make this `TextBox` read-only:

```
View = new TextBox
{
    Text = e.ToString(),
    IsReadOnly = true,
};
```

src/GitHub.InlineReviews/Services/PullRequestStatusBarManager.cs
```diff
@@ -112,42 +121,39 @@ StatusBar FindSccStatusBar(Window mainWindow)
             return contentControl?.Content as StatusBar;
         }
 
-        class RaisePullRequestCommand : ICommand
+        class UsageTrackingCommand : ICommand
```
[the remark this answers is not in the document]
jcansdale: I changed it to use the `IShowCurrentPullRequestCommand` directly rather than using `dte.Commands.Raise`. The problme was I lost the metrics.

I'm thinking it would be a better idea to record the metrics in the command itself? Only issue is, you wouldn't be able to differentiate where the user found the command (not that it matters yet in this case).

src/GitHub.VisualStudio/UI/GitHubPane.cs
```diff
@@ -75,25 +83,30 @@ public GitHubPane() : base(null)
...
+            var provider = VisualStudio.Services.GitHubServiceProvider;
+            var teServiceHolder = provider.GetService<ITeamExplorerServiceHolder>();
+            teServiceHolder.ServiceProvider = this;
 
-                View = factory.CreateView<IGitHubPaneViewModel>();
-                View.DataContext = viewModel;
-            }
+            var factory = provider.GetService<IViewViewModelFactory>();
+            var viewModel = provider.ExportProvider.GetExportedValue<IGitHubPaneViewModel>();
+            viewModel.InitializeAsync(this).Catch(ShowError).Forget();
```
jcansdale: Should `await` this here.

AArnott:
@jcansdale said:

> I'd add a shim that makes SwitchToMainThreadAsync into a nop when testing.

Be careful how you add that shim. Don't avoid using the JTF type directly from all your code because only by using that type directly do the vs-threading analyzers help keep you on the safe threading path.
For testability in the context of a MEF part, you can `[Import] JoinableTaskContext` (I'm not sure what the minimum VS version is for that) and use its JTF instance. Your test can then export an instance of that.

>  InitializeAsync(sp).FileAndForget();

That sounds like your MEF part will be unusable for an undefined length of time, and all callers will have to be prepared to deal with that. What I rather do is make `InitializeAsync` a public/internal method on the type and anyone who imports my MEF part must `await InitializeAsync` before calling any other members on the export. That way deadlocks can be mitigated while ensuring the export will work as expected.

> The problem is my MEF component isn't tied to a particular package that it would be convenient to initialize it from. It could be used by a number of other MEF components.

You shouldn't assume your package is loaded anyway when a MEF part is activated, since MEF can activate parts without the VS package ever having loaded (generally speaking). That's why I advocate for the caller initializing the part by invoking `InitializeAsync()`.

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -59,17 +57,20 @@ void LogVersionInformation()
             var menuService = (IMenuCommandService)(await GetServiceAsync(typeof(IMenuCommandService)));
```
AArnott: Same comment as I made to your other `InitializeMenus()` method.

src/GitHub.VisualStudio/UI/GitHubPane.cs
```diff
@@ -75,25 +83,30 @@ public GitHubPane() : base(null)
 
         protected override void Initialize()
         {
-            base.Initialize();
-            Initialize(this);
+            InitializeAsync().Catch(ShowError).Forget();
         }
 
-        public void Initialize(IServiceProvider serviceProvider)
+        public Task<IGitHubPaneViewModel> GetViewModelAsync() => viewModelSource.Task;
```
AArnott: This looks like it will likely be misused to violate the [3rd threading rule](https://github.com/Microsoft/vs-threading/blob/master/doc/threading_rules.md#rule-3-if-ever-awaiting-work-that-was-started-earlier-that-work-must-be-joined). If someone tries to await this Task within the context of a JTF.Run delegate, you'll quite likely deadlock.
Can you replace your TaskCompletionSource with a JoinableTask instead? Then call `InitializeAsync` within a JTF.RunAsync delegate and store the resulting `JoinableTask` in your field.
Then *this* method should return `joinableTask.JoinAsync()` (or simply `await joinableTask`) to mitigate deadlocks in callers.

src/GitHub.VisualStudio/GitHubPackage.cs
jcansdale: Oops, I remember you said. I'm sorry I missed these. 😊

jcansdale:
@AArnott I'm picking up lots here. Plenty of homework to do!

I've updated it to use `JoinableTask` rather than `TaskCompletionSource` and cast `IMenuCommandService` on Main thread. Hopefully it's getting there.

src/GitHub.InlineReviews/Services/PullRequestStatusBarManager.cs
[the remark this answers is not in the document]
jcansdale: I'll fix this in another PR. It's working for the moment.

jcansdale:
@AArnott,

> For testability in the context of a MEF part, you can [Import] JoinableTaskContext (I'm not sure what the minimum VS version is for that) and use its JTF instance. Your test can then export an instance of that.

That would be handy. Unfortunately it looks like it's a VS 2017 thing.

> That sounds like your MEF part will be unusable for an undefined length of time, and all callers will have to be prepared to deal with that.

This isn't such a problem because it has some extra initialization do do anyway before it starts firing events. All all callers do is subscribe to those events.

src/GitHub.VisualStudio/UI/GitHubPane.cs
```diff
@@ -83,30 +80,38 @@ public GitHubPane() : base(null)
 
         protected override void Initialize()
         {
-            InitializeAsync().Catch(ShowError).Forget();
+            viewModelTask = ThreadHelper.JoinableTaskFactory.RunAsync(InitializeAsync);
```
AArnott: Can you have this class accept an `AsyncPackage` in construction and then use the `AsyncPackage.JoinableTaskFactory` instance instead of `ThreadHelper.JoinableTaskFactory` here?

That way if VS shuts down before this work is done, we won't risk crashing due to arbitrary work going on in background threads.

AArnott:
Your changes look good, @jcansdale.

> I don't completely understand what's going on here (I'm guessing each package has a collection of tasks associated with it). Could you expand a little?

Yes: `AsyncPackage` defines a `JoinableTaskCollection` and an associated `JoinableTaskFactory`. Anytime you use *that* `JoinableTaskFactory` instance, all the work it schedules is added to the `JoinableTaskCollection` while that work is active.
Later when the `AsyncPackage.Dispose` method is called as VS shuts down, that method will block the caller (the UI thread) until the `JoinableTaskCollection` is empty, the idea being to ensure that no threads or async work is ongoing after the package has claimed to have completed shutting down.

When threads are still running at the time VS shuts down the CLR's AppDomain, crashes often result. That's why we curtail the problem this way.

src/GitHub.InlineReviews/Services/PullRequestStatusBarManager.cs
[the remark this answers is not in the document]
jcansdale: I've moved the tracking of `NumberOfShowCurrentPullRequest` into the command itself (where I think it belongs). `RaisePullRequestCommand` is now obsolete.

PR #1580, opened 2018-04-04
Fix navigate to editor.

grokys:
This code was accidentally removed in fa8aab8

- Fix navigate from `View Changes` to editor
- Fix navigate from `View File` to editor

Previously the menu option to navigate from the diff view to the file in the solution was missing from the context menu - this adds it back.

Depends on #1578
Part of #1491 
Fixes #1580


meaghanlewis:
LGTM 👍

PR #1581, opened 2018-04-04
Fix Octokit.GraphQL escaping issue.

grokys:
Updates the Octokit.GraphQL package to 0.0.2-alpha.

This fixes the escaping issue at https://github.com/grokys/Octokit.GraphQL/issues/72.

Now PR review comments that are submitted with a `"` character should work!

Depends on #1580 
Part of #1491


meaghanlewis:
LGTM!

PR #1582, opened 2018-04-04
Validate PR review state before allowing submission.

grokys:
Split the `PullRequestReviewAuthoringViewModel.Submit` command into 2 separate commands and add a `CanExecute` observable for them:

- `Approve` can always be submitted
- `Comment` needs either a review body or >0 review comments
- `RequestChanges` needs either a review body or >0 review comments

## HACK

Disabling the `GitHubActionLink`s that are used for the approve/comment/request changes buttons had no visual effect. Seems that hyperlinks don't respect [dependency property precedence](https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/dependency-property-value-precedence), so the `IsEnabled=False` trigger for `Hyperlink.Foreground` doesn't override the foreground `TemplateBinding`. Not sure what we can do here other than set the control's opacity when disabled, so that's what I did for now.

Depends on #1581 
Part of #1491


meaghanlewis:
This LGTM - except I thought that `RequestChanges` option needs to be either have a review body or > 0 review comments to submit.

grokys:
@meaghanlewis oh, yes you're right! Fixed that.

Issue #1591, opened 2018-04-10
Current PR doesn't appear on status bar when checked out from command line


jcansdale:
- GitHub Extension for Visual Studio version: 
- Visual Studio version: All

#### Repro

1. Clone a new repository
2. Check out PR branch from command line
3. PR doesn't appear on the status bar

Also:

1. Clone a new repository
2. Create a PR using `Create New` on `GitHub` pane
3. PR doesn't appear on the status bar

This means that people who haven't used the extension aren't being guided towards extension features that would be relevant in the current context.

#### Simple mitigation

At the moment the PR icon and the PR number is a single button. We only show this button when we know it's a PR branch (the branch is labeled when opened in the PR detail view). Hence, the button only appears when the has already used the GitHub pane functionality.

![image](https://user-images.githubusercontent.com/11719160/38550498-9b0470fc-3cae-11e8-8f1a-abdf76e45211.png)

We could easily split the button into 2, where the PR icon opens the PR list and the PR number opens the individual PR. This would have 2 advantages.

1. It would keep the command for the PR icon consistent with the `GitHub` pane (opens the PR list)

![image](https://user-images.githubusercontent.com/11719160/38551286-b2919252-3cb0-11e8-8d35-54c168cda0a0.png)

2. It would guide the user towards the PR list when they're using a Git repo (when they can open or create a PR)

#### Proper fix

1. Label the current PR when the PR list is loaded (not when the PR is specifically opened). This would help guide the user towards the current PR after they open the PR list.
2. Label the PR when created using `Create New` on `GitHub` pane

grokys:
Yes, this is a known issue. I'm hoping that we can solve this as we move to GraphQL.

jcansdale:
@grokys what about when creating PRs with the extension? Should they be showing up?

grokys:
Yes, actually we can fix that part already.

PR #1593, opened 2018-04-10
Show `Pull requests` button on status bar when there's no active PR branch

jcansdale:
This PR will show a `Pull requests` button on the status bar when in the context of a GitHub repo but not on a PR branch.

### What it does

- Make the `OpenPullRequests` button visible when in the context of a repo with a remote
- Make the `ShowCurrentPullRequest` button visible when on an PR branch

This means that if we're on a PR branch that hasn't yet been opened in the PR detail view (issue #1591), the `OpenPullRequests` button will still be visible and encourage the user to open the PR by hand. This is also useful when the user simply want to change to a different PR.

The `Pull requests` button has same capitalization as on `github.com` (slightly to screen-cast above)

image

image

### Todo

- [x] Decide what to show when the user is on a GitHub repo, but not on a PR branch (or we don't know)
  - The icon + `Pull requests` would be consistent with .com

- [x] Be more selective about when the `OpenPullRequests` button appears
  - Show `Pull requests` when repo has a remote

At the moment it will appear whenever there is a Git repo with a remote. This means it will also appear on a TFS Git repo. While not the end of the world (clicking on it says `This repository is not on GitHub`), it would be nice if we could be a little more selective.

Do we currently have any code that looks for GitHub-ish URLs?

TFS Git remotes seem to look like this (which would be easy to filter out):
https://user.visualstudio.com/defaultcollection/_git/RepoName

### Reviewers

- Sanity check my use of Rx in `PullRequestStatusBarManager.StartShowingStatus`

- I've noticed the tooltip for the `OpenPullRequests` icon doesn't always appear. Any idea what's going on here?

- The tooltip for our `OpenPullRequests` command on the toolbar is simply `Pull Requests`. In this context I'm using `Open or Create Pull Requests`. What do you prefer out of these:
`Pull Requests`, `Open or Create Pull Requests`, `Open or Create Pull Request` - or something else?

- After splitting the button in two, it the spacing okay? @donokuda

### Rejected

Showing `OpenPullRequests` and `ShowCurrentPullRequest` as separate buttons. Not consistent with other `Team Explorer` status bar buttons/menus.

### Future PRs

- It would be nice if we could point people towards adding a repo to GitHub when the repo doesn't have a remote.

Mitigation for #1591


sguthals:
Just some overall comments from our conversations today:

# High Level Problem
There are three issues that we currently have:
Giving users a way to...
1. open the GitHub pane 
2. get back to the list of PRs
3. open a new PR
That isn't going through levels of menus or through the Team Explorer Pane when the GitHub pane isn't already open.

# Issues with this solution
Though the solution proposed here might resolve the friction of the above three problems, it introduces new friction in terms of design continuity and user experience expectations. Mainly:
1. The task bar sets the tone for icon+text and that should not be broken. 
2. A user might:
  a. Think that the highlighting is broken because the icon or text isn't being highlighted when the other is hovered over
  b. Always click on the text part and get used to that taking them to the details, but then accidentally click on the icon part one day and be taken to a completely different view

# Other Potential Solutions to the High Level Problem
I feel there are two things that need to be addressed:
1. Specifically the PR accessibility (list vs detail vs open) 
2. The GitHub pane accessibility

## Focused on solving the PR accessibility issue
If we are focused on only solving the first, I could imagine a few potential solutions:
1. Add an icon+text that is the PR icon and the number of PRs for this repo, which would take the user to the list of PRs.
2. Add a GitHub icon+text that would take the user to the GitHub pane, which right now only has the list of PRs anyway
3. Add a menu on the current icon+text that comes up and has three options:
  a. Open new PR
  b. Open PR List
  c. Open current PR details
We do not have to also list the PRs in this menu. But it would better follow the UI with the branches menu that Team Explorer has already introduced:
screen shot 2018-04-11 at 7 30 11 pm

_I prefer (3), however I'm wondering what we should do if a user isn't on a PR. What should the text say?_

And I'm sure many other ideas. 

## Focused on the GitHub pane discoverability, workflow, accessibility
I do think, with Forking, Issues, and PR Reviews we have a bigger issue that needs addressing. That is - how are we going to help users be able to discover all of our features when they need to be discovering them.

sguthals:
Agreed @donokuda!

I like this as a first step. Along with metrics, we can get a better workflow/discoverability for the entire extension related to PRs and future features (e.g. Issues)!

meaghanlewis:
This LGTM ✨

src/GitHub.InlineReviews/Views/PullRequestStatusView.xaml
```diff
@@ -35,29 +35,66 @@
...
+                        Pull requests
+                    </TextBlock>
+                </StackPanel>
+            </Button>
+            <Grid.Style>
+                <Style>
+                    <Setter Property="Control.Visibility" Value="Collapsed" />
+                    <Style.Triggers>
+                        <!-- Visible when Number is Null -->
+                        <DataTrigger Binding="{Binding Number}" Value="{x:Null}">
```
grokys: Why not use a binding with `TargetNullValue` here as below?

PR #1597, opened 2018-04-12
Try really hard to avoid deadlocking GetService calls

shana:
`new ShellSettingsManager` in the PackageSettings constructor calls GetService, which might be a potential deadlock situation. Also, to avoid calls to GetService that might deadlock, cache the services we create in GitHubServiceProvider so that it returns cached instances immediately and bypasses GetService.

I also fixed some other calls around the code, particularly direct calls to `ExportProvider` (`GitHubServiceProvider.TryGetService` does that for us)

### DIFF
--- a/src/GitHub.VisualStudio/Commands/CreateGistCommand.cs
+++ b/src/GitHub.VisualStudio/Commands/CreateGistCommand.cs
@@ -5,6 +5,7 @@ using GitHub.Commands;
 using GitHub.Logging;
 using GitHub.Services;
 using GitHub.Services.Vssdk.Commands;
+using Microsoft.VisualStudio.Threading;
 
 namespace GitHub.VisualStudio.Commands
 {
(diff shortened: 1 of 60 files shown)


src/GitHub.VisualStudio/Commands/OpenLinkCommand.cs
```diff
@@ -41,7 +41,7 @@ protected OpenLinkCommand(IGitHubServiceProvider serviceProvider)
             var link = await GenerateLink(LinkType.Blob);
             if (link == null)
                 return;
-            var browser = ServiceProvider.TryGetService<IVisualStudioBrowser>();
+            var browser = ServiceProvider.TryGetMEFComponent<IVisualStudioBrowser>();
```
StanleyGoldman: Try testing the TryGetMEFComponent

src/GitHub.VisualStudio/Commands/CreateGistCommand.cs
```diff
@@ -35,20 +36,17 @@ protected CreateGistCommand(Lazy<IDialogService> dialogService, Lazy<ISelectedTe
...
         {
-            return dialogService.Value.ShowCreateGist();
+            return dialogService.ShowCreateGist();
         }
 
         protected override void QueryStatus()
         {
-            Log.Assert(SelectedTextProvider != null, "Could not get an instance of ISelectedTextProvider");
-            Visible = !string.IsNullOrWhiteSpace(SelectedTextProvider?.GetSelectedText());
+            Visible = !string.IsNullOrWhiteSpace(selectedTextProvider.GetSelectedText());
```
StanleyGoldman: The using TryGetMEFComponent

jcansdale:
Too many conflicts. 😉

PR #1599, opened 2018-04-13
WIP: Fix inline comment peek view focus

### LINKED ISSUE #1588: When adding inline comment, keyboard focus can be split between text box and diff text
- GitHub Extension for Visual Studio version: 2.4.99
- Visual Studio version: 2015

Went to add an inline comment. Typing letters worked but cursor keys control the left hand text of the diff view.

1. Start adding an inline comment on RLS of diff
2. Click on LHS of diff text
3. Click on comment text box (on RHS view)
4. Start typing and using cursor keys

grokys:
Trying to fix #1588 

@olegtk this is the code with the change you suggested, but it seems to have no effect. Any other ideas?

### DIFF
--- a/src/GitHub.InlineReviews/Views/InlineCommentPeekView.xaml.cs
+++ b/src/GitHub.InlineReviews/Views/InlineCommentPeekView.xaml.cs
@@ -2,6 +2,7 @@
 using System.Reactive.Subjects;
 using System.Windows.Controls;
 using GitHub.VisualStudio.UI.Helpers;
+using Microsoft.VisualStudio.Editor;
 
 namespace GitHub.InlineReviews.Views
 {
@@ -17,6 +18,7 @@ namespace GitHub.InlineReviews.Views
             desiredHeight = new Subject<double>();
             threadView.LayoutUpdated += ThreadViewLayoutUpdated;
             threadScroller.PreviewMouseWheel += ScrollViewerUtilities.FixMouseWheelScroll;
+            CommandRouting.SetInterceptsCommandRouting(this, true);
         }
 
         public IObservable<double> DesiredHeight => desiredHeight;


olegtk:
Sorry for the delay. Command handling in Peek is really complicated so I need to debug this case.
I was trying to F5 the solution and while it seems to build and deploy just fine, I cannot get Experimental VS instance to connect to GitHub (normal VS instance is fine). Any ideas what I'm doing wrong?

image

jcansdale:
Hi @olegtk,

I think this might be the issue:
https://github.com/github/VisualStudio/blob/master/src/GitHub.Api/ApiClientConfiguration_User.cs

See here for more info:
https://github.com/github/visualstudio#build

I hope that does the trick. Thanks for investigating this!

jcansdale:
@olegtk we'd like to be more helpful to contributing developers, when they start to work with the extension. If you have any suggestions or observations, please could you comment on #1618. Thanks!

olegtk:
Instructions are fine, I just failed to read them past opening solution :(
It looks like this is a focus issue. When you click first on LHS and then in Peek, LHS still keeps aggregated focus and so diff view routes commands to it (in VS typing alphanumerical chars in general goes directly to the editor, while all commands such as up/down etc are routed through VS command system first).
I'll try to come up with a solution on Monday.

olegtk:
Yeah, it's a bug in Visual Studio, it can be reproduced without your extension. Not sure yet what's a workaround, but please file a bug using "Help/Feedback/Report a Problem" please?

grokys:
Thanks @olegtk - reported the bug here https://developercommunity.visualstudio.com/content/problem/245245/keyboard-focus-in-peek-view-can-be-incorrect.html

olegtk:
We fixed https://developercommunity.visualstudio.com/content/problem/245245/keyboard-focus-in-peek-view-can-be-incorrect.html in VS 2017 15.8 update (starting with Preview4). Thanks!

jcansdale:
@olegtk, thanks for the heads up!

I'm wondering is there's any information you could share from the internal issue about what caused it?

Is this likely to impact VS 2015 users as well?

olegtk:
That was just a focus bug in Peek/Diff . When focused Peek always focuses its containing window frame, which usually fine, except in Diff, whose window frame contains two text views.

Unfortunately VS 2015 users are affected too, but we don't plan to back port the fix to VS 2015 at this point.

grokys:
@olegtk just tried on 15.8 preview 3 and the problem still seems to be there. Any idea what version the fix will be in?

I've also just noticed that Copy/Paste and `TextBox` context menus don't seem to work in the peek view in the diff view. Would this be caused by the same problem?

olegtk:
Probably in Preview 4? We just shipped Preview 4, please try it.

grokys:
@olegtk just tried in 15.8.0. Preview 5.0 and this is still a problem.

Also going to close this PR and move discussion over to #1588.

PR #1606, opened 2018-04-16
Adding metrics tests for the new metrics

shana:
- Fix the `NumberOfStartups` counter, it was never being recorded on disk.
- Revert change to serialize everything to string, it's not needed
- Added a server with nancy to process metrics submissions so we can test our metrics in integration tests
- Added a couple of simple tests


src/GitHub.Exports/Services/MetricsService.cs
```diff
@@ -137,11 +121,25 @@ static StringContent SerializeRequest(UsageModel model)
...
+                {
+                    var value = prop.GetValue(model);
+
+                    if (value == null)
+                    {
+                        dict.Add(ToJsonPropertyName(prop.Name), value);
+                    }
+                    else
+                    {
+                        dict.Add(ToJsonPropertyName(prop.Name), ToModelDictionary(value));
```
StanleyGoldman: Are we expecting nested dictionaries?

src/GitHub.VisualStudio/Services/UsageTracker.cs
```diff
@@ -89,16 +95,8 @@ IDisposable StartTimer()
...
-
-            if (firstTick)
-            {
-                var current = await GetCurrentReport(data);
-                current.Measures.NumberOfStartups++;
-                changed = true;
-                firstTick = false;
-            }
 
+            var changed = false;
```
StanleyGoldman: Because of this change... `WriteLocalData` will not be called for a startup...

src/GitHub.Exports/Services/MetricsService.cs
shana: Not really, I just undid your changes to only output strings, this is the code that was here before, but I think I forgot to revert the other change where you call this method twice - it should be `ToModelDictionary(model)`

src/GitHub.VisualStudio/Services/UsageTracker.cs
shana: It will be called above in `IncrementCounter`

src/GitHub.Exports/Services/MetricsService.cs
StanleyGoldman: No, you caught that change

PR #1607, opened 2018-04-16
Completing GitHub fork functionality

StanleyGoldman:


GitHubVS.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.27130.0
```
StanleyGoldman: I'm not sure about this change.

GitHubVS.sln
```diff
@@ -439,4 +439,7 @@ Global
 		{110B206F-8554-4B51-BF86-94DAA32F5E26} = {8A7DA2E7-262B-4581-807A-1C45CE79CDFD}
 		{17EB676B-BB91-48B5-AA59-C67695C647C2} = {8A7DA2E7-262B-4581-807A-1C45CE79CDFD}
 	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {556014CF-5B35-4CE5-B3EF-6AB0007001AC}
+	EndGlobalSection
```
StanleyGoldman: I'm also unsure of this change.

src/GitHub.App/Services/RepositoryForkService.cs
```diff
@@ -0,0 +1,59 @@
...
+                            await gitClient.SetRemote(repo.LocalRepo, "origin", new Uri(repo.RemoteRepo.CloneUrl));
+                        }
+
+                        if (addUpstream)
+                        {
+                            await gitClient.SetRemote(repo.LocalRepo, "upstream", sourceRepository.CloneUrl.ToUri());
+
+                            if (resetMasterTracking)
+                            {
+                                await gitClient.SetTrackingBranch(repo.LocalRepo, "master", "upstream");
```
jcansdale: I think you need to do the following first:
`await gitClient.Fetch(repo.LocalRepo, "upstream");`

This ensures that the target tracking branch exists.

jcansdale:
@StanleyGoldman when I click on the `Fork` button from a fork, the `Parent` repo appears in the list of organizations that I could create a fork in. I'm guessing if I tried to do that it would fail?

image

Could we add a button that allows the user to switch back to the `Parent` repo (similar to switching to already forked repos)?

StanleyGoldman:
Hey @jcansdale I did some work to make that happen, but I realized I couldn't complete the user story of a person performing that task...

Can you help me fill it in..

#### A user has a GitHub repository open;
- Wants to fork the repository;
	- Does not want to make any changes to local git remotes;
	- Wants to set the fork as their `origin`;
		- Has no need for the original remote;
		- Wants to maintain the original remote as `upstream`;
			- Wants to leave `master` tracking `origin` (the newly fork);
			- Wants `master` set to track `upstream`;
- Wants to switch their `origin` to a parent/fork;

jcansdale:
@StanleyGoldman I'm not very experienced at writing user storys, but I'll give it a go. 😄 

I'd be inclined to start with higher level wants.

#### A user has a GitHub repository open;
- Wants to fix a bug and submit a PR
	- Wants to create a fork and set it as their `origin`
	- Wants to use a fork they already created for their `origin`
- Wants to sync their bug fix branch with the parent repo
	1. Wants to set parent repo as their `origin` (before pulling, rebasing or resetting)
        2. Wants to set `origin` back to their fork

This is hard. Would you be up for chatting about it over zoom?

StanleyGoldman:
Lol, it is hard. Yea we can definitely go over it.
In the mean time i'm going to disable that functionality as it's rather confusing at this point.
I rather 🚢 what we got so far.

src/GitHub.Exports/Models/UsageModel.cs
```diff
@@ -70,6 +70,11 @@ public class MeasuresModel
             public int NumberOfShowCurrentPullRequest { get; set; }
             public int NumberOfStatusBarOpenPullRequestList { get; set; }
             public int NumberOfTeamExplorerHomeOpenPullRequestList { get; set; }
+            public int NumberOfShowRepoForkDialogClicks { get; set; }
+            public int NumberOfReposForked { get; set; }
+            public int NumberOfOriginsUpdatedWhenForkingRepo { get; set; }
+            public int NumberOfUpstreamsAddedWhenForkingRepo { get; set; }
+            public int NumberOfTrackMasterUpstreamWhenForkingRepo { get; set; }
```
StanleyGoldman: I added a good number of metrics

jcansdale:
@StanleyGoldman,

I'm being given the option to fork `PullRequestSandbox` to `jcansdale`.

image

Even though I've already forked this repo.

image

I wonder what's going on there?  Is this a regression?

StanleyGoldman:
@meaghanlewis I'm doing my research on this point.
> For repos where forking is disabled, I am able to select a user/organization to fork to, but on the 2nd step nothing happens when clicking the Fork Repo button.

From what I can tell there is no support in Octokit to determine if a repo "can" be forked...

StanleyGoldman:
@jcansdale I did have a regression here. I also realized that it wasn't quite right to begin with.

```
                      +---------+
                      |         | Users and Orgs
                      |         | where a Repo
               +---------+      | exists
               |XXXXXX|  |      |
               |XXXXXX|  |      |
Places where a |XXXXXX+---------+
User can put a |XXXXXXXXX|
Repo           |XXXXXXXXX|
               +---------+
```

We need the shaded area above, but I'm not a fan of how I got it accomplished.
I have to go find out how to use GraphQL inside of Visual Studio, then I should be able to improve this lookup.

grokys:
> We noted that the prompt: "Reset the master branch to remote/origin" was incorrect when compared to the implemented functionality. I changed the prompt to say "Set the master branch to track upstream/master".

My intention in my spike was that the master branch _would_ be reset to upstream master, as you (pretty much?) always want your fork's `master` to track upstream `master` and if it doesn't you're in for a world of pain unless you know what you're doing.

There's the question of what to do if `master` can't be ffwd-ed to `origin/master` of course. My hand-wavy idea was that it would be renamed as `fork-master` or something.

jcansdale:
> There's the question of what to do if master can't be ffwd-ed to origin/master of course. My hand-wavy idea was that it would be renamed as fork-master or something.

I think in this case let the user rename their branch before recreating master. The chances are it will become a PR branch, so they'll probably want to call it something sensible!

StanleyGoldman:
@grokys 
> My intention in my spike was that the master branch would be reset to upstream master, as you (pretty much?) always want your fork's master to track upstream master and if it doesn't you're in for a world of pain unless you know what you're doing.

Yea, I realized I completely misread the prompt and implemented something different. So the least path of resistance was to change the prompt.

StanleyGoldman:
@meaghanlewis I restored the functionality to display existing forks to the Fork target selection screen...
There is no functionality attached to it yet though...

image

meaghanlewis:
> I'm a bit confused by this one, was the repo in this case public or private?

@StanleyGoldman the fork dialog is blank/empty for all enterprise repositories. 

Another case where the dialog is empty is when I don't have access to the repository. Ex. `@meaghanlewis` cloned repo1 and she has access to the repo and can fork. But then `@evilmeaghanlewis` signs in via Visual Studio and she does not have access to repo1. She can still see and interact with repo1 but some things, like forking, will not work.

meaghanlewis:
@StanleyGoldman thanks for displaying the error messages from the API. This makes it a lot more clear if something goes wrong when trying to fork.

GitHubVS.sln
jcansdale: This change would break me. I have too many versions of VS 2017 installed. ;-) Can you revert it?

src/GitHub.TeamFoundation.14/Home/PulseNavigationItem.cs
```diff
@@ -14,7 +14,7 @@ namespace GitHub.VisualStudio.TeamExplorer.Home
     [PartCreationPolicy(CreationPolicy.NonShared)]
     public class PulseNavigationItem : TeamExplorerNavigationItemBase
     {
-        public const string PulseNavigationItemId = "5245767A-B657-4F8E-BFEE-F04159F1DDA2";
+        public const string PulseNavigationItemId = "5245767A-B657-4F8E-BFEE-F04159F1DDA5";
```
jcansdale: I'm curious why the Guid changed for `PulseNavigationItem`?

src/GitHub.TeamFoundation.14/Home/ForkNavigationItem.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        public override async void Execute()
+        {
+            var connection = await connectionManager.GetConnection(ActiveRepo);
+
+            if (connection?.IsLoggedIn == true)
+            {
+                usageTracker.IncrementCounter(model => model.NumberOfShowRepoForkDialogClicks).Forget();
+                await dialogService.ShowForkDialog(ActiveRepo, connection);
```
jcansdale: Can we expose this as a VS command? I'm trying to make sure we expose all of our commands a VS commands. See:
https://github.com/github/VisualStudio/pull/1638.
https://github.com/github/VisualStudio/commit/437f1619e117f67ff08e17d346836560f5b1f863#diff-6e8b5db6949b54276068dee9aef6d3f9

src/GitHub.VisualStudio/Views/Dialog/ForkRepositoryExecuteView.xaml
```diff
@@ -0,0 +1,86 @@
...
+            Add an <Run Style="{DynamicResource {x:Static markdig:Styles.CodeStyleKey}}">upstream</Run> remote pointing to 
+          <Hyperlink><Run Text="{Binding SourceRepository.CloneUrl, Mode=OneWay}"/></Hyperlink>
+          </TextBlock>
+        </DockPanel>
+      </Border>
+      <Border Style="{StaticResource ItemBorder}">
+        <DockPanel>
+          <CheckBox IsChecked="{Binding ResetMasterTracking}" IsEnabled="{Binding CanResetMasterTracking}" />
+          <TextBlock TextWrapping="Wrap">
+            Set the <Run Style="{DynamicResource {x:Static markdig:Styles.CodeStyleKey}}">master</Run> branch to track
```
jcansdale: It would be nice to have a tooltip saying why this is useful and how it would be used.

src/GitHub.TeamFoundation.14/Home/ForkNavigationItem.cs
StanleyGoldman: I'll add this to the backlog for this feature...

src/GitHub.VisualStudio/Views/Dialog/ForkRepositoryExecuteView.xaml
StanleyGoldman: I have no idea what to write in the tooltip. What do you think I should put?

src/GitHub.TeamFoundation.14/Home/PulseNavigationItem.cs
StanleyGoldman: Reverted

jcansdale:
@StanleyGoldman I've been pulling my hair out trying to understand why the `origin` URL wasn't being refreshed consistently. I think there might be a race condition with Team Explorer monitoring and reading the `.git/config` file. I was getting file locked exceptions when debugging it.

I pushed the following assert as a sanity check that the `origin` URL has been been updated: https://github.com/github/VisualStudio/pull/1607/commits/3b84532a91b6274a3387ae5135bc44bae95a4c40

This commit also moves the call to `RefreshActiveRepositories` out of the `using (activeRepo) { ... }` block. I've been working on a screen-cast demoing the fork repo functionality and it refreshed correctly a bunch of times in a row. I think this might have fixed the most common issue. 🤞

src/GitHub.App/Services/RepositoryForkService.cs
```diff
@@ -101,14 +102,18 @@ public IObservable<object> SwitchRemotes(IRepositoryModel destinationRepository,
...
-                        {
-                            vsGitExt.RefreshActiveRepositories();
-                        }
+                    if (updateOrigin)
+                    {
+                        vsGitExt.RefreshActiveRepositories();
 
-                        return new object();
+                        var updatedRepository = vsGitExt.ActiveRepositories.FirstOrDefault();
+                        log.Assert(updatedRepository.CloneUrl == destinationRepository.CloneUrl,
```
jcansdale: If something went wrong with `RefreshActiveRepositories`, it's possible that `updatedRepository` could be null. Should change this to `updatedRepository?.CloneUrl`. Sorry! 😊

StanleyGoldman: No problem, fixed.

PR #1609, opened 2018-04-17
Fix avatar in inline comments.

grokys:
Don't use the default button width for the button in the avatar view. [The default button width is 76](https://github.com/github/VisualStudio/blob/25fae6b454680894d3300a595fc3bcce65ce84eb/src/GitHub.VisualStudio.UI/Styles/Buttons.xaml#L13); don't use this for the button in the avatar as it will likely be much smaller, and in fact was making the avatar invisible in the inline comment view.

Before:

After:


meaghanlewis:
Tested with GitHub and GitHub Enterprise. LGTM ⚡️

PR #1619, opened 2018-04-20
Simplifying how to login instructions

shana:
Making the images smaller so the text is more readable, adding the other signin options we have, cleaning up the layout a bit and generally simplifying the instructions


docs/getting-started/authenticating-to-github.md
```diff
@@ -1,15 +1,43 @@
...
 - For more information on creating a GitHub account, see "[Signing up for a new GitHub account](https://help.github.com/articles/signing-up-for-a-new-github-account/)".
 - For a GitHub Enterprise account, contact your GitHub Enterprise site administrator.
 
-**Note:** If your organization is on the [Business plan](https://help.github.com/articles/organization-billing-plans) and has not enabled SAML single sign-on or login with username and password, you must create and authorize a personal access token to access protected content. In addition, SAML single sign-on is not available for GitHub enterprise versions less than 2.12.2.
+### Personal access tokens
 
-### Scopes for personal access tokens
+If all signin options above fail, you can manually create a personal access token and use it as your password.
 
 The scopes for the personal access token are: `user`, `repo`, `gist`, and `write:public_key`.
```
jcansdale: I think we should put this information right where the uses needs it.

At the moment we simply say `Token` where the user needs to enter their personal access token. Somehow they need to find this documentation or figure out what scopes are required by themselves.

image

We could be much more helpful and change it to `Token (required scopes: `user`, `repo`, `gist`, and `write:public_key`).

PR #1638, opened 2018-05-02
Consistently name all VS commands

jcansdale:
This PR uses a tip shared by @madskristensen to ensure that all commands have a consistent name.
https://twitter.com/jcansdale/status/991013150932729856
https://twitter.com/mkristensen/status/991018416222818306
> Ahh, there is a way and it is undocumented. Prefix the name in the .vsct file with a period. Eg: <ButtonText>.TestDriven.NET.Foo</BT>

### What this PR does

Previously our commands would appear on the shortcuts search dialog with inconsistent prefixes (some didn't even contain `GitHub` and were difficult to find).

image

- This PR gives them all a consistent prefix of `GitHub`.

image

- Specify a `CanonicalName` for all commands (this is used in the `Command Window`)

- Specify a `LocCanonicalName` for all commands (used in `Customize > Keyboard...` dialog)

- Our commands now appear as follows

```
GitHub.Back
GitHub.Forward
GitHub.Refresh
GitHub.PullRequests
GitHub.ViewOnGitHub
GitHub.Help
GitHub.ConnectToGitHub
GitHub.ShowGitHubPane
GitHub.ShowCurrentPullRequest
GitHub.OpenLink
GitHub.CopyLink
GitHub.OpenFileInSolution
GitHub.Blame
GitHub.CreateGist
GitHub.NextComment
GitHub.PreviousComment
```

Previously they appeared as in the screenshot above + the following:

```
EditorContextMenus.CodeWindow.OpenFileinSolution
Edit.NextComment
Edit.PreviousComment
```

### Reviewers

This PR is deliberately being kept simple. Is there any reason not to do this? 😉 

I hope having all of these commands consistently named will allow for faster and more stable automation testing. //paging @meaghanlewis 

### How to test

These commands names are surfaced in a few places

1. On the `Command Window` if you start typing `GitHub`.

image

2. Right-click on a toolbar, `Customize...`, `Keyboard...` and start typing `GitHub` in `Show commands containing`.

image

3. They can also be accessed programatically via automation.

```
        void ShowGitHubPane(DTE dte)
        {
            dte.ExecuteCommand("GitHub.ShowGitHubPane");
        }
```

### Future PRs

There is a bunch of functionality that is not being exposed as VS commands (e.g. show PR list of sync submodules). We should fix this but in separate PRs.

Fixes #1630

PR #1642, opened 2018-05-02
Factor sync submodules into VS command

jcansdale:
### What this PR does

- Extract the sync submodules action into a VS command called
`GitHub.SyncSubmodules`

### Reviewers

I've extended the command interface to include a method that returns results from the command. Is this a reasonable thing to do if results are sometimes required from a command? It can also be used standalone without returning any results.
https://github.com/github/VisualStudio/blob/36e680e5a0cd936cafbd27af658335727d28841c/src/GitHub.Exports/Commands/ISyncSubmodulesCommand.cs#L16

Related to #1638


src/GitHub.VisualStudio/Commands/SyncSubmodulesCommand.cs
```diff
@@ -0,0 +1,100 @@
...
+            var writer = new StringWriter(CultureInfo.CurrentCulture);
+            var complete = await pullRequestService.SyncSubmodules(repository, line =>
+            {
+                writer.WriteLine(line);
+                statusBarNotificationService.ShowMessage(line);
+            });
+
+            if (!complete)
+            {
+                statusBarNotificationService.ShowMessage("Failed to sync submodules." + Environment.NewLine + writer);
```
grokys: Is it normal for VS commands to write directly to the status bar? Is there no concept of "output" for commands, which e.g. shows in the command pane?

jcansdale: I think the status bar is what commands tend to use as their ambient UI. There isn't really any concept of a native UI or output for a command.

One possibility would be to pass in a progress object that the command could report back to. This could also sidestep the issue of returning results.

I wanted to start off with a simple refactoring and and avoid any extra implementation detail. We can revisit next time this command is touched (if we decide to surface it somewhere else).

PR #1613, opened 2018-04-19
Nest review comments under reviews.

grokys:
Previously we were loading a flat list of PR review comments into the `PullRequestModel.ReviewComments` property because this is how they were exposed in the REST API. 

GraphQL nests review comments under their relevant review, which makes a lot more sense, so this PR removes `PullRequestModel.ReviewComments` and moves them to `PullRequestReviewModel.Comments`.

This is the first step towards GraphQL-ification of Pull Requests; in upcoming PRs I intend to further this process by reading pull request models using purely GraphQL. At this point there are still a few hacks in place, and paging of PR reviews and review comments is not implemented. This will be addressed in upcoming PRs.

PR #1648, opened 2018-05-04
WIP: Read PR details using GraphQL

grokys:
## Description

Use GraphQL to read the PR details in `ModelService.GetPullRequest` instead of using the ugly mix of REST and GraphQL that we were using before.

NOTE: The list of changed files in a PR are not exposed yet via GraphQL so we still need to use REST for these.

## Implementation Notes

The query currently only reads the first 100 comments of the first 100 reviews for each PR. Will this be a problem? Paging with GraphQL is hard currently, we need to add support for automatic paging to Octokit.GraphQL.

Added a default constructor to `PullRequestModel` and made its `Number` property read/write so that all properties can be set from an initializer rather than requiring a mix of ctor parameters and initializers.

## Testing

Nothing user-visible should change, or more precisely nothing should break, when interacting with the Pull Request detail view and PR reviews..

Depends on #1613


grokys:
Marked this as WIP because now Octokit.GraphQL has auto-paging need to update it to use that.

grokys:
Superseded by #1712

PR #1665, opened 2018-05-16
Polish fork repo view

donokuda:
**Note: This branch targets `fixes/github-fork-limit-functionality`**

Couple of polish tasks here for the fork repo feature. This branch isn't fully complete since I need help on a few things.

First, some screenshots of its current state (as of this writing):

screen shot 2018-05-15 at 8 22 18 pm

screen shot 2018-05-16 at 5 25 24 pm

With a dummy error message

screen shot 2018-05-16 at 5 29 26 pm

Here are some things I could use some help on 🙏:

- [x] Hook up the ["Back" button](https://user-images.githubusercontent.com/1174461/40095192-88fe6632-587f-11e8-8bfe-7d65f54501b2.png) in the second screen
- [x] If you have 3 rows of orgs to fork to, they will get clipped but the vertical scrollbar won't show 😭 🤔. I'm not sure why that is and I feel like I spent too much time trying to figure it out 😞 
- [x] Avatars are not the right resolution /cc @grokys since we chatted about this

PR #1672, opened 2018-05-16
Adding functionality to the back button in the fork view

StanleyGoldman:
This pull request ties in the back button for #1665

### DIFF
--- a/src/GitHub.App/ViewModels/Dialog/ForkRepositoryExecuteViewModel.cs
+++ b/src/GitHub.App/ViewModels/Dialog/ForkRepositoryExecuteViewModel.cs
@@ -45,6 +45,7 @@ namespace GitHub.ViewModels.Dialog
                 .Subscribe(tuple => CanResetMasterTracking = tuple.Item1 && tuple.Item2);
 
             CreateFork = ReactiveCommand.CreateAsyncObservable(OnCreateFork);
+            BackCommand = ReactiveCommand.Create();
         }
 
         public IRepositoryModel SourceRepository { get; private set; }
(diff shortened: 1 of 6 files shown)


src/GitHub.VisualStudio/Views/Dialog/ForkRepositoryExecuteView.xaml
```diff
@@ -89,7 +89,7 @@
                                Visibility="{Binding Error, Converter={ui:NullToVisibilityConverter}}" 
                                HorizontalAlignment="Left" />
                     <StackPanel Orientation="Horizontal" HorizontalAlignment="Right">
-                      <Button HorizontalAlignment="Right" Margin="0 0 4 0">Back</Button>
+                      <Button HorizontalAlignment="Right" Margin="0 0 4 0" Click="backButton_OnClick">Back</Button>
```
jcansdale: Could you use `Command="{Binding BackCommand}"` here instead of `Click="backButton_OnClick"`?

meaghanlewis:
This functionality works well for me 👍

src/GitHub.VisualStudio/Views/Dialog/ForkRepositoryExecuteView.xaml
StanleyGoldman: I tried because you mentioned it, but it causes a crash. I'm not sure why.

src/GitHub.App/ViewModels/Dialog/ForkRepositoryExecuteViewModel.cs
```diff
@@ -55,10 +56,14 @@ IRepositoryForkService repositoryForkService
 
         public IReactiveCommand<Repository> CreateFork { get; }
 
+        public IReactiveCommand<object> BackCommand { get; }
+
         public string Title => Resources.ForkRepositoryTitle;
 
         public IObservable<object> Done => CreateFork.Where(repository => repository != null);
 
+        public IObservable<object> Back => BackCommand.AsObservable();
```
grokys: I don't think `Back` and `BackCommand` need to be separate here, you could just subscribe directly to `BackCommand` rather than exposing it in `Back`. The reason it's done like that with `Done` is that I didn't want to tie the `Done` signal to a `ReactiveCommand`: the dialog might be "done" when any number of things happen, some of which might not be easily represented as a command.

src/GitHub.VisualStudio/Views/Dialog/ForkRepositoryExecuteView.xaml
grokys: This should definitely work.

PR #1673, opened 2018-05-16
Make PR# link work in authoring view.

### LINKED ISSUE #1654: Pull Request Review prompt contains a hyperlink that has no effect
This title looks like a hyperlink but has no action.

I'm not sure if we want a link here or one that is more explicit. ala "View on GitHub" in the PR view.

image

grokys:
Add the missing command to navigate back to the Pull Request after clicking the PR# link in the header of the PR authoring view.

image

Fixes #1654.


StanleyGoldman:
My only issue with this is not really with you fixing the hyperlink, but it's more of a design question.
Why is the header for this view so different than the header from the previous view?

image

PR #1681, opened 2018-05-17
Don't only show `Open File in Solution` / `Open Code View` options when on PR branch

### LINKED ISSUE #1679: Open File in Solution / Open Code View menu options not visible
- GitHub Extension for Visual Studio version: `2.5.1`

The `File in Solution` / `Open Code View` menu commands are only visible when a PR branch is checked out.

image

image

These should always be visible when in a PR file diff / changes view (with a hint dialog suggesting to user check out to navigate).

jcansdale:
### What this PR does

There was some code that hid all `GoToSolutionOrPullRequestFileCommand` options when not on a PR branch.

```cs
var session = sessionManager.Value.CurrentSession;
if (session == null)
{
     // No active pull request session
     Visible = false;
     return;
}
```

Only the `View Changes in PR` command option should be hidden when not on a PR branch.

- Make the `Open File in Solution` option visible when viewing a PR file diff
- Make the `Open File in Code View` option visible when viewing a writable diff file

Fixes #1679

### DIFF
--- a/src/GitHub.VisualStudio/Commands/GoToSolutionOrPullRequestFileCommand.cs
+++ b/src/GitHub.VisualStudio/Commands/GoToSolutionOrPullRequestFileCommand.cs
@@ -150,14 +150,6 @@ namespace GitHub.Commands
         {
             try
             {
-                var session = sessionManager.Value.CurrentSession;
-                if (session == null)
-                {
-                    // No active pull request session
-                    Visible = false;
-                    return;
-                }
-
                 var sourceView = pullRequestEditorService.Value.FindActiveView();
                 if (sourceView == null)
                 {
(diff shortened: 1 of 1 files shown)


src/GitHub.VisualStudio/Commands/GoToSolutionOrPullRequestFileCommand.cs
```diff
@@ -185,13 +177,17 @@ void OnBeforeQueryStatus(object sender, EventArgs e)
...
-                var relativePath = sessionManager.Value.GetRelativePath(textView.TextBuffer);
-                if (relativePath != null)
+                var session = sessionManager.Value.CurrentSession;
+                if (session != null)
                 {
-                    // Active text view is part of a repository
-                    Text = "View Changes in #" + session.PullRequest.Number;
-                    Visible = true;
-                    return;
+                    var relativePath = sessionManager.Value.GetRelativePath(textView.TextBuffer);
```
grokys: I don't think `session` is actually needed here. Can't it just be:

```
var relativePath = sessionManager.Value?.GetRelativePath(textView.TextBuffer);

if (relativePath != null) { ... }
```

PR #1689, opened 2018-05-18
[spike] Make MEF constructors free-threaded

jcansdale:
### What this PR does

- Use JoinableTaskFactory to make MEF constructors free-threaded
- Fire TeamExplorerContext events on Main thread

### DIFF
--- a/src/GitHub.App/Services/TeamExplorerContext.cs
+++ b/src/GitHub.App/Services/TeamExplorerContext.cs
@@ -7,6 +7,8 @@ using GitHub.Logging;
 using GitHub.Primitives;
 using Serilog;
 using EnvDTE;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Threading;
 
 namespace GitHub.Services
 {
@@ -29,6 +31,7 @@ namespace GitHub.Services
 
         readonly DTE dte;
         readonly IVSGitExt gitExt;
+        readonly JoinableTaskFactory joinableTaskFactory;
 
         string solutionPath;
         string repositoryPath;
@@ -40,17 +43,18 @@ namespace GitHub.Services
         ILocalRepositoryModel repositoryModel;
 
         [ImportingConstructor]
-        public TeamExplorerContext(IGitHubServiceProvider serviceProvider, IVSGitExt gitExt)
-            : this(gitExt, serviceProvider)
+        TeamExplorerContext(SVsServiceProvider serviceProvider, IVSGitExt gitExt)
+            : this(gitExt, serviceProvider, ThreadHelper.JoinableTaskFactory)
         {
         }
 
-        public TeamExplorerContext(IVSGitExt gitExt, IGitHubServiceProvider serviceProvider)
+        public TeamExplorerContext(IVSGitExt gitExt, IServiceProvider serviceProvider, JoinableTaskFactory joinableTaskFactory = null)
         {
             this.gitExt = gitExt;
+            this.joinableTaskFactory = joinableTaskFactory ?? new JoinableTaskContext().Factory;
 
-            // This is a standard service which should always be available.
-            dte = serviceProvider.GetService<DTE>();
+            // SVsServiceProvider doesn't complain when this isn't called on Main thread
+            dte = (DTE)serviceProvider.GetService(typeof(DTE));
 
             Refresh();
             gitExt.ActiveRepositoriesChanged += Refresh;
@@ -80,27 +84,47 @@ namespace GitHub.Services
                     if (newRepositoryPath != repositoryPath)
                     {
                         log.Debug("ActiveRepository changed to {CloneUrl} @ {Path}", repo?.CloneUrl, newRepositoryPath);
-                        ActiveRepository = repo;
+                        joinableTaskFactory.Run(async () =>
+                        {
+                            await joinableTaskFactory.SwitchToMainThreadAsync();
+                            ActiveRepository = repo;
+                        });
                     }
                     else if (newCloneUrl != cloneUrl)
                     {
                         log.Debug("ActiveRepository changed to {CloneUrl} @ {Path}", repo?.CloneUrl, newRepositoryPath);
-                        ActiveRepository = repo;
+                        joinableTaskFactory.Run(async () =>
+                        {
+                            await joinableTaskFactory.SwitchToMainThreadAsync();
+                            ActiveRepository = repo;
+                        });
                     }
                     else if (newBranchName != branchName)
                     {
                         log.Debug("Fire StatusChanged event when BranchName changes for ActiveRepository");
-                        StatusChanged?.Invoke(this, EventArgs.Empty);
+                        joinableTaskFactory.Run(async () =>
+                        {
+                            await joinableTaskFactory.SwitchToMainThreadAsync();
+                            StatusChanged?.Invoke(this, EventArgs.Empty);
+                        });
                     }
                     else if (newHeadSha != headSha)
                     {
                         log.Debug("Fire StatusChanged event when HeadSha changes for ActiveRepository");
-                        StatusChanged?.Invoke(this, EventArgs.Empty);
+                        joinableTaskFactory.Run(async () =>
+                        {
+                            await joinableTaskFactory.SwitchToMainThreadAsync();
+                            StatusChanged?.Invoke(this, EventArgs.Empty);
+                        });
                     }
                     else if (newTrackedSha != trackedSha)
                     {
                         log.Debug("Fire StatusChanged event when TrackedSha changes for ActiveRepository");
-                        StatusChanged?.Invoke(this, EventArgs.Empty);
+                        joinableTaskFactory.Run(async () =>
+                        {
+                            await joinableTaskFactory.SwitchToMainThreadAsync();
+                            StatusChanged?.Invoke(this, EventArgs.Empty);
+                        });
                     }
 
                     repositoryPath = newRepositoryPath;
--- a/src/GitHub.TeamFoundation.14/Base/TeamExplorerServiceHolder.cs
+++ b/src/GitHub.TeamFoundation.14/Base/TeamExplorerServiceHolder.cs
@@ -2,11 +2,12 @@
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Linq;
-using System.Threading;
 using GitHub.Extensions;
 using GitHub.Models;
 using GitHub.Services;
 using Microsoft.TeamFoundation.Controls;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Threading;
 
 namespace GitHub.VisualStudio.Base
 {
@@ -20,19 +21,26 @@ namespace GitHub.VisualStudio.Base
 
         IServiceProvider serviceProvider;
         readonly IVSGitExt gitService;
-
-        // ActiveRepositories PropertyChanged event comes in on a non-main thread
-        readonly SynchronizationContext syncContext;
+        readonly JoinableTaskFactory joinableTaskFactory;
 
         /// <summary>
         /// This class relies on IVSGitExt that provides information when VS switches repositories.
         /// </summary>
         /// <param name="gitService">Used for monitoring the active repository.</param>
         [ImportingConstructor]
-        public TeamExplorerServiceHolder(IVSGitExt gitService)
+        TeamExplorerServiceHolder(IVSGitExt gitService) : this(gitService, ThreadHelper.JoinableTaskFactory)
+        {
+        }
+
+        /// <summary>
+        /// This constructor can be used for unit testing.
+        /// </summary>
+        /// <param name="gitService">Used for monitoring the active repository.</param>
+        /// <param name="joinableTaskFactory">Used for switching to the Main thread.</param>
+        public TeamExplorerServiceHolder(IVSGitExt gitService, JoinableTaskFactory joinableTaskFactory = null)
         {
             this.gitService = gitService;
-            syncContext = SynchronizationContext.Current;
+            this.joinableTaskFactory = joinableTaskFactory ?? new JoinableTaskContext().Factory;
 
             UpdateActiveRepo();
             if (gitService != null)
@@ -40,7 +48,6 @@ namespace GitHub.VisualStudio.Base
                 gitService.ActiveRepositoriesChanged += UpdateActiveRepo;
             }
         }
-
 
         // set by the sections when they get initialized
         public IServiceProvider ServiceProvider
@@ -142,8 +149,14 @@ namespace GitHub.VisualStudio.Base
             var repo = gitService?.ActiveRepositories.FirstOrDefault();
 
             if (!Equals(repo, ActiveRepo))
-                // so annoying that this is on the wrong thread
-                syncContext.Post(r => ActiveRepo = r as ILocalRepositoryModel, repo);
+            {
+                // Fire property change events on Main thread
+                joinableTaskFactory.Run(async () =>
+                {
+                    await joinableTaskFactory.SwitchToMainThreadAsync();
+                    ActiveRepo = repo;
+                });
+            }
         }
 
         void ActiveRepoPropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
--- a/src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
+++ b/src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
@@ -2,13 +2,13 @@ using System;
 using System.Linq;
 using System.Threading.Tasks;
 using System.Collections.Generic;
-using System.ComponentModel.Composition;
 using GitHub.Models;
 using GitHub.Services;
 using GitHub.Logging;
 using GitHub.TeamFoundation.Services;
 using Serilog;
 using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Threading;
 using Microsoft.VisualStudio.TeamFoundation.Git.Extensibility;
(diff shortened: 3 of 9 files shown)


src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -40,17 +43,18 @@ public class TeamExplorerContext : ITeamExplorerContext
         ILocalRepositoryModel repositoryModel;
 
         [ImportingConstructor]
-        public TeamExplorerContext(IGitHubServiceProvider serviceProvider, IVSGitExt gitExt)
-            : this(gitExt, serviceProvider)
+        TeamExplorerContext(SVsServiceProvider serviceProvider, IVSGitExt gitExt)
```
grokys: If this were imported using `[Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider` then I think these two constructors could be merged?

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -40,17 +43,18 @@ public class TeamExplorerContext : ITeamExplorerContext
...
+            : this(gitExt, serviceProvider, ThreadHelper.JoinableTaskFactory)
         {
         }
 
-        public TeamExplorerContext(IVSGitExt gitExt, IGitHubServiceProvider serviceProvider)
+        public TeamExplorerContext(IVSGitExt gitExt, IServiceProvider serviceProvider, JoinableTaskFactory joinableTaskFactory = null)
         {
             this.gitExt = gitExt;
+            this.joinableTaskFactory = joinableTaskFactory ?? new JoinableTaskContext().Factory;
```
grokys: What does `new JoinableTaskContext().Factory` do?

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -80,27 +84,47 @@ void Refresh()
                     if (newRepositoryPath != repositoryPath)
                     {
                         log.Debug("ActiveRepository changed to {CloneUrl} @ {Path}", repo?.CloneUrl, newRepositoryPath);
-                        ActiveRepository = repo;
+                        joinableTaskFactory.Run(async () =>
```
grokys: Might it be worth adding a `RunOnMainThread()` extension method to do this, or is it better to make it explicit like this?

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
```diff
@@ -49,29 +53,28 @@ public VSGitExt(IAsyncServiceProvider asyncServiceProvider, IVSUIContextFactory
             // The IGitExt service isn't available when a TFS based solution is opened directly.
             // It will become available when moving to a Git based solution (and cause a UIContext event to fire).
             var context = factory.GetUIContext(new Guid(Guids.GitSccProviderId));
-            context.WhenActivated(() => Initialize());
+            context.WhenActivated(() => JoinableTaskFactory.RunAsync(InitializeAsync));
```
grokys: I'm not clear on why this is different from just calling `Initialize` as we were previously? The result of `RunAsync` isn't awaited,.

src/GitHub.App/Services/TeamExplorerContext.cs
jcansdale: We do need to allow a `JoinableTaskFactory` to be specified when unit testing. If we exported it with a `ContractName`, we could combine that with what you suggested, e.g:

```
[Import(typeof(SVsServiceProvider))] IServiceProvider serviceProvider,
[Import("GitHub.VisualStudio")] JoinableTaskFactory joinableTaskFactory
```
(or perhaps `JoinableTaskContext`)

src/GitHub.App/Services/TeamExplorerContext.cs
jcansdale: It constructs a new task factory that consideres the current thread to be the `Main` thread.  This is useful when unit testing because it effectively turns `await.. SwitchToMainThreadAsync` into a nop.

src/GitHub.App/Services/TeamExplorerContext.cs
jcansdale: When `Refresh` becomes `async` (from your other PR), all we'll need is `await joinableTaskFactory.SwitchToMainThreadAsync()` before the `ActiveRepository = repo`.

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
jcansdale: It automatically adds running tasks to `JoinableTaskCollection`, which can be awaited by unit tests. It also allows use to `SwitchToMainThreadAsync` in the `GetServiceAsync<T>()` method below.

It we use a common `JoinableTaskContext` across the extension, I'm hoping it will let us more easily debug our task related code and avoid possible deadlocks.

StanleyGoldman: asdf

StanleyGoldman: test

StanleyGoldman: Hello

src/GitHub.TeamFoundation.14/Base/TeamExplorerServiceHolder.cs
```diff
@@ -142,8 +152,14 @@ void UpdateActiveRepo()
             var repo = gitService?.ActiveRepositories.FirstOrDefault();
 
             if (!Equals(repo, ActiveRepo))
-                // so annoying that this is on the wrong thread
-                syncContext.Post(r => ActiveRepo = r as ILocalRepositoryModel, repo);
+            {
+                // Fire property change events on Main thread
```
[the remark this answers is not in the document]
jcansdale: I'm glad you spotted this. It would have changed the behavior from the previous `SynchronizationContext.Post(...)`! I missed the fact that `Post` is asynchranous.

test/UnitTests/GitHub.TeamFoundation/VSGitExtTests.cs
```diff
@@ -39,22 +39,23 @@ public void GetServiceIGitExt_WhenUIContextChanged(bool activated, int expectCal
 
             context.IsActive = activated;
 
-            target.PendingTasks.Wait();
+            target.JoinableTaskCollection.JoinTillEmptyAsync().Wait();
```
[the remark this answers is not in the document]
jcansdale: > "I won't stop blocking till that collection is empty, but the very JoinableTask that is doing that waiting is also a member of the collection it wants to have empty!"

I fell into exactly this trap, before I reverted back to using `Wait()` which seemed to work. Thanks for the exaplanation and code. I'll use this pattern in future.

jcansdale:
@AArnott,

> Kudos to you for testing this. One thing I couldn't tell from your test snippet is whether you block the main thread while activating all your MEF parts from a background thread.

I had some fun and games with my async method running code, see here:
https://github.com/jcansdale/TestDriven.Net-Issues/issues/110

I ended up using this (where `TryRequiresMainThread` checks the target method for an `[STAThread]` attribute - is doesn't have one in this case):
```
                    var testRunState = JoinableTaskFactory.Run(async () =>
                    {
                        if (TryRequiresMainThread(assembly, cref))
                        {
                            await JoinableTaskFactory.SwitchToMainThreadAsync();
                        }

                        return await memberTestRunner.RunAsync(adaptorTestListener, assembly, cref);
                    });
```

I guess this means the Main thread wasn't being blocked when activating all the MEF parts? I've updated the test project to do the following:
```
        // Right click `Test With > In-Proc (VS SDK)`
        void MainThreadBlockingLoadServices(IComponentModel componentModel)
        {
            // Block the main thread while activating MEF parts from a background thread.
            // https://github.com/github/VisualStudio/pull/1689#issuecomment-391399687
            var jtf = ThreadHelper.JoinableTaskFactory;
            jtf.Run(async () =>
            {
                using (jtf.Context.SuppressRelevance())
                {
                    await Task.Run(() => LoadServices(componentModel));
                }
            });
        }
```

Does that look right?

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -43,28 +47,56 @@ public class TeamExplorerContext : ITeamExplorerContext
...
-            // This is a standard service which should always be available.
-            dte = serviceProvider.GetService<DTE>();
-
-            Refresh();
+            StartRefresh();
             gitExt.ActiveRepositoriesChanged += Refresh;
         }
 
-        async void Refresh()
+        void StartRefresh() => JoinableTaskFactory.Context.Factory.RunAsync(QueueRefreshAsync).Task.FileAndForget(log);
```
jcansdale: @AArnott I want to ensure that there are no overlapping calls to `RefreshAsync`. This seems to work, but I worry about other people looking at the code. Is there a more obvious way to do this?

I've just reminded myself of the following:
> Rule #3. If ever awaiting work that was started earlier, that work must be joined.

I'll see how it might look using `Join`.

jcansdale:
@AArnott,

> The one thing you may be missing from your free-threaded test is that you must be on the UI thread before the jtf.Run command so that you're blocking the UI thread (and not some random background thread) during the rest of the test.

Oops, yes. It looks like I am missing that. I just need to add a `[STAThread]` attribute so the test runner knows to run it on the Main thread.

BTW, does this look like a reasonable way to implement a lock-equivalent using a `JoinableTask`?
```
        async Task QueueRefreshAsync()
        {
            await refreshJoinableTask?.JoinAsync();
            await (refreshJoinableTask = JoinableTaskFactory.RunAsync(RefreshAsync));
        }
```

src/GitHub.App/Services/TeamExplorerContext.cs
jcansdale: This is actually my own helper method to file exceptions in our own logs:
```cs
        public static void FileAndForget(this Task task, ILogger log)
        {
            task.ContinueWith(t =>
            {
                if (t.IsFaulted)
                {
                    log.Error(t.Exception, nameof(FileAndForget));
                }
            });
        }
```

Maybe I should call it `LogAndForget` to disambiguate?

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -43,28 +47,56 @@ public class TeamExplorerContext : ITeamExplorerContext
...
+        void Refresh() => JoinableTaskFactory.Context.Factory.Run(QueueRefreshAsync);
+
+        async Task QueueRefreshAsync()
+        {
+            await JoinableTaskCollection.JoinTillEmptyAsync();
+            await JoinableTaskFactory.RunAsync(RefreshAsync);
```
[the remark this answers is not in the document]
jcansdale: The most important thing is that the data from the most recent call refresh is up to date. I know that just becuase task A starts befoire task B, doesn't mean that it will complete before task B. If task A happened to complete after task B, it might set `ActiveRepository` back to null (which has been a problem in the past).

Since `RefreshAsync` is relatively quick and I'm going to use Rx for throttling, I'm inclined to keep it simple and just queue up refreshes. I could maybe use a `CancellationToken`, but I think this would be optimzing an edge case.

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -80,13 +81,17 @@ public class TeamExplorerContext : ITeamExplorerContext
...
+        void StartRefresh() => JoinableTaskFactory.RunAsync(QueueRefreshAsync).Task.FileAndForget(log);
+        void Refresh() => JoinableTaskFactory.Run(QueueRefreshAsync);
 
         async Task QueueRefreshAsync()
         {
-            await JoinableTaskCollection.JoinTillEmptyAsync();
-            await JoinableTaskFactory.RunAsync(RefreshAsync);
+            if (refreshJoinableTask != null)
+            {
+                await refreshJoinableTask.JoinAsync(); // make sure StartRefresh has completed
```
AArnott: You can drop the `.JoinAsync()` call. It's implied any time you await a `JoinableTask`.

jcansdale:
@AArnott I'm currently using `ThreadHelper.JoinableTaskContext` to construct my `JoinableTaskCollection`. Am I doing this right, or should I be using the `JoinableTaskFactory.Context` from one of our packages?

If I should be using a different context, I'm wondering what the best way to expose it to MEF would be. Maybe something like: `[Export("GitHub.JoinableTaskContext")]`?

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -80,13 +81,17 @@ public class TeamExplorerContext : ITeamExplorerContext
             gitExt.ActiveRepositoriesChanged += Refresh;
         }
 
-        void StartRefresh() => JoinableTaskFactory.Context.Factory.RunAsync(QueueRefreshAsync).Task.FileAndForget(log);
-        void Refresh() => JoinableTaskFactory.Context.Factory.Run(QueueRefreshAsync);
+        void StartRefresh() => JoinableTaskFactory.RunAsync(QueueRefreshAsync).Task.FileAndForget(log);
+        void Refresh() => JoinableTaskFactory.Run(QueueRefreshAsync);
 
         async Task QueueRefreshAsync()
```
AArnott: Does this method need to be thread-safe? It isn't at the moment. If two callers on different threads call it concurrently, you may find yourself with multiple `RefreshAsync` calls executing simultaneously.

AArnott:
> I'm currently using ThreadHelper.JoinableTaskContext to construct my JoinableTaskCollection. Am I doing this right, or should I be using the JoinableTaskFactory.Context from one of our packages?

There is (and must be) only one `JoinableTaskContext` in VS. Where you get that instance from is therefore not important, so long as you get one. Getting it from your package's JTF property is very reasonable. Another option if you're in a MEF world is to `[Import] JoinableTaskContext`. 

> If I should be using a different context, I'm wondering what the best way to expose it to MEF would be. Maybe something like: [Export("GitHub.JoinableTaskContext")]?

Per my earlier comments, you should *not* create your own, and therefore you needn't export your own, since the *one* is already exported.

AArnott:
> Unfortunately await refreshJoinableTask?.JoinAsync() throws. It would need to be...

Correct: `anything?.Something()` will return `null` if `anything` is null. That means the expression you're awaiting on can actually be null. The compiler doesn't emit code to check for null, so you end up throwing NRE if `refreshJoinableTask` is null. Lesson: Never await expressions that can be null.

> Am I correct in thinking SemaphoreSlim should be avoided?

Probably. Mixing semaphores and JTF is possible, but often done wrong. You're essentially creating a private queue that JTF has no insight into, so deadlocks hit you when something in the back of the queue owns the UI thread and the thing at the head of the queue requires the UI thread. There is a pattern that avoids deadlocks for this when you need it, but it's usually easier to avoid semaphore.

AArnott:
Oh! Excellent. Renaming the method is up to you.

jcansdale:
@AArnott,

> Given the Rx framework (which I believe you're using here?) allows you to schedule work, can you arrange for it to be scheduled to the UI thread in the first place? I see you deleted an ObserveOn(RxApp.MainThreadScheduler) line. Why is that? If that doesn't work, can we create something like that that does?

I've reverted to using `RxApp.MainThreadScheduler` because it does appear to work fine.

My concert was that `SynchronizationContext.Current` is different depending on whether a task is started using `RxApp.MainThreadScheduler` or a `JoinableTaskFactory`. Maybe this isn't a problem and I'm worrying about nothing.

Alternatively, maybe having a `JoinableTaskFactory` flavored `IScheduler` would make it easier to track any tasks kicked off by Rx? I'm not familiar enough with Rx and `JoinableTaskFactory` to make a good call on this.

AArnott:
Not every JoinableTaskCollection is created with the purpose of blocking disposal. Some are there just to avoid deadlocks. But any time you trigger async work without awaiting it yourself, you should make sure that *someone* is tracking it and blocking *some* disposal somewhere till it's done.

Your code to block disposal looks good.

AArnott:
Regarding Rx and `JoinableTaskFactory`, if Rx notifications can ever become blocking (i.e. if you can ever block the main thread waiting for a notification to come in), then yes, some careful JTF authoring in a custom `IScheduler` is merited. But if they are always asynchronous, I imagine the default scheduler for the main thread is probably sufficient. That said, if you want to try implementing `IScheduler` with JTF I'm willing to review it.

src/GitHub.Logging/Logging/ILoggerExtensions.cs
```diff
@@ -26,5 +27,21 @@ public static void Assert(this ILogger logger, bool condition, string messageTem
 #pragma warning restore Serilog004
             }
         }
+
+        /// <summary>
+        /// Log any exceptions when a task throws.
+        /// </summary>
+        /// <param name="task">The <see cref="Task"/> to log exceptions from.</param>
+        /// <param name="log">The <see cref="ILogger"/> to use.</param>
+        public static void LogAndForget(this Task task, ILogger log)
```
grokys: Why add this new method rather than add logging to the existing `Forget` method in `TaskExtensions`? Also its name suggests that the main purpose of this method is to log, but that's not true.

jcansdale: The idea was for it to be equivalent to `FileAndForget` which logs to Visual Studio's error log, but that takes an instance of one of our loggers instead.

Having the existing `Forget` method log somewhere as well would be a good idea, but the question is where since it won't have an instance of a logger.

Could we open an issue to do this?

grokys: Ok, that makes sense. IMO this should probably just be called `Forget` as well though, so there would be two versions: one that takes a logger and one that doesn't.

PR #1699, opened 2018-05-21
Fix scrolling in PR details view.

### LINKED ISSUE #1698: Scrolling is broken in PRs with lots of changed files.
When looking at the details for a large PR, the horizontal scroll bar is shown at the bottom of the scrollable area, meaning that one has to scroll right to the bottom in order to scroll horizontally. This makes navigating this panel very difficult:

grokys:
## Description

Attempt to fix the scrolling in the PR details view.

Before:

After:

## Implementation

There is no panel in WPF which would allow the behavior we need, so added a `ScollingVerticalStackPanel` control which implements `IScrollInfo` meaning that it implements its own scrolling behavior.

This panel can be used to mark controls as "fixed" meaning they don't scroll horizontally. All controls in the PR details are marked as fixed except the changes tree.

To make it that that the "Changes" header doesn't scroll with the changes tree, I had to put the tree outside its `SectionControl` and bind it's visibility to the `SectionControl`'s `IsExpanded` property.

Fixes #1698

### DIFF
--- a/src/GitHub.UI/Controls/ScollingVerticalStackPanel.cs
+++ b/src/GitHub.UI/Controls/ScollingVerticalStackPanel.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Controls.Primitives;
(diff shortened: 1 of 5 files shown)


jcansdale:
There appears to be a strange performance issue. Vertical scrolling is snappy when there is either no scroll bar or when there's a scroll bar with space to scroll.

If you restrict the view and expand it again (so there is nothing to scroll), suddenly scrolling becomes very laggy.  There doesn't seem to be a way to get rid of the bar.

image

Tested using the closed `github/VisualStudio` PR 1004.

jcansdale:
There seems to be something funny going on when I I use the mouse wheel.

The vertical scroll bar doesn't get updated.



Using the wheel I can keep on scrolling even after I reach the bottom.

grokys:
@jcansdale pushed changes to hopefully fix those problems.

> Would it be possible to only show the horizontal bar when when the tree view component is visible?

Pushed a change to do this, but `ScrollViewer` doesn't really like you doing that, so lets see what problems that introduces ;)

src/GitHub.UI/Controls/ScollingVerticalStackPanel.cs
```diff
@@ -0,0 +1,206 @@
...
+                offsetY += delta;
+                rect.Y -= delta;
+            }
+
+            if (rect.Y < 0)
+            {
+                offsetY += rect.Y;
+            }
+
+            // We technially should be trying to also show the right-hand side of the rect here
```
jcansdale: This is a good call.

meaghanlewis:
This LGTM, thanks for fixing! ✨

PR #1701, opened 2018-05-21
Adding Edit/Delete functions to inline comments

StanleyGoldman:
Depends on:
- [ ] #1708

image

- Added Edit/Delete comment functionality for inline comment views
  - Even though an administrator can edit/delete any comment. For a first pass we only added the functionality to comments owned by the current viewer.

### DIFF
--- a/src/GitHub.Exports.Reactive/Services/IPullRequestSession.cs
+++ b/src/GitHub.Exports.Reactive/Services/IPullRequestSession.cs
@@ -111,7 +111,7 @@ namespace GitHub.Services
         /// <param name="body">The comment body.</param>
         /// <param name="inReplyTo">The REST ID of the comment to reply to.</param>
         /// <param name="inReplyToNodeId">The GraphQL ID of the comment to reply to.</param>
-        /// <returns></returns>
+        /// <returns>A comment model.</returns>
         Task<IPullRequestReviewCommentModel> PostReviewComment(
             string body,
             int inReplyTo,
@@ -152,5 +152,13 @@ namespace GitHub.Services
         /// <param name="number">The number of the pull request comment to delete</param>
         /// <returns>A task which completes when the session has completed updating.</returns>
         Task DeleteComment(int number);
+
+        /// <summary>
+        /// Edit a PR review comment reply.
+        /// </summary>
+        /// <param name="number">The number of the pull request comment</param>
+        /// <param name="body">The replacement comment body.</param>
+        /// <returns>A comment model.</returns>
+        Task<IPullRequestReviewCommentModel> EditComment(int number, string body);
     }
 }
--- a/src/GitHub.Exports.Reactive/Services/IPullRequestSession.cs
+++ b/src/GitHub.Exports.Reactive/Services/IPullRequestSession.cs
@@ -134,7 +134,7 @@ namespace GitHub.Services
         /// Posts the currently pending review.
         /// </summary>
         /// <param name="body">The review body.</param>
-        /// <param name="state">The review event.</param>
+        /// <param name="e">The review event.</param>
         /// <returns>The review model.</returns>
         Task<IPullRequestReviewModel> PostReview(string body, PullRequestReviewEvent e);
 
@@ -145,5 +145,12 @@ namespace GitHub.Services
         /// <param name="pullRequest">The new pull request model.</param>
         /// <returns>A task which completes when the session has completed updating.</returns>
         Task Update(IPullRequestModel pullRequest);
+
+        /// <summary>
+        /// Deletes a pull request comment.
+        /// </summary>
+        /// <param name="number">The number of the pull request comment to delete</param>
+        /// <returns>A task which completes when the session has completed updating.</returns>
+        Task DeleteComment(int number);
     }
 }
--- a/src/GitHub.Exports.Reactive/Services/IPullRequestSession.cs
+++ b/src/GitHub.Exports.Reactive/Services/IPullRequestSession.cs
@@ -156,9 +156,9 @@ namespace GitHub.Services
         /// <summary>
         /// Edit a PR review comment reply.
         /// </summary>
-        /// <param name="number">The number of the pull request comment</param>
+        /// <param name="commentNodeId">The node id of the pull request comment</param>
         /// <param name="body">The replacement comment body.</param>
         /// <returns>A comment model.</returns>
-        Task<IPullRequestReviewCommentModel> EditComment(int number, string body);
+        Task<IPullRequestReviewCommentModel> EditComment(string commentNodeId, string body);
     }
 }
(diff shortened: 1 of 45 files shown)


src/GitHub.InlineReviews/Views/CommentView.xaml
```diff
@@ -50,31 +50,46 @@
...
                         Visibility="{Binding IsPending, Converter={ui:BooleanToVisibilityConverter}, FallbackValue=Collapsed}">
-                    <TextBlock FontSize="10">Pending</TextBlock>
+                  <TextBlock FontSize="10">Pending</TextBlock>
                 </Border>
-            </StackPanel>
+              </StackPanel>
+
+              <StackPanel Orientation="Horizontal" HorizontalAlignment="Right" DockPanel.Dock="Right"
+                          Visibility="{Binding CanEditOrDelete, Converter={ui:BooleanToVisibilityConverter}}">
+                <ui:GitHubActionLink Content="Edit"
```
jcansdale: image

image

Rather than `Edit`, should this be `Update comment` like on .com? 👆

donokuda: I think this references the Edit / Delete links in the screenshot above?

Screenshot of feature

If so, I think it's okay to leave this the way how it is. However, I'm 👍 for having the submit button text be "Update comment" like the dotcom screenshot you've posted.

src/GitHub.InlineReviews/Views/CommentView.xaml
```diff
@@ -170,18 +192,38 @@
...
+                    <Style TargetType="FrameworkElement">
+                        <Setter Property="Visibility" Value="Collapsed"/>
+                        <Style.Triggers>
+                            <DataTrigger Binding="{Binding EditState}" Value="Editing">
+                                <Setter Property="Visibility" Value="Visible"/>
+                            </DataTrigger>
+                        </Style.Triggers>
+                    </Style>
+                </StackPanel.Style>
+                <Button Command="{Binding CommitEdit}">Edit</Button>
```
donokuda: Per @jcansdale's comment, can we change this to "Update comment" so it's inline with the interaction on dotcom?

src/GitHub.InlineReviews/Views/CommentView.xaml
StanleyGoldman: Fixed up.

src/GitHub.InlineReviews/Views/CommentView.xaml
StanleyGoldman: Fixed up.

src/GitHub.InlineReviews/ViewModels/CommentThreadViewModel.cs
```diff
@@ -44,6 +45,30 @@ public ReactiveCommand<ICommentModel> PostComment
             }
         }
 
+        public ReactiveCommand<ICommentModel> EditComment
+        {
+            get { return editComment; }
+            set
```
grokys: I see you followed `PostComment` and made this setter public - I think that was an oversight, it should be `protected`. When you fix this, could you fix `PostComment` too?

src/GitHub.InlineReviews/ViewModels/CommentThreadViewModel.cs
```diff
@@ -44,6 +45,30 @@ public ReactiveCommand<ICommentModel> PostComment
...
+                editComment = value;
+
+                value.ThrownExceptions.Subscribe(_ => { });
+            }
+        }
+
+        public ReactiveCommand<object> DeleteComment
+        {
+            get { return deleteComment; }
+            set
```
grokys: As with this one.

src/GitHub.InlineReviews/ViewModels/CommentThreadViewModel.cs
StanleyGoldman: I made them private, but they aren't used. Are we just leaving them here in case we ever need them? So we know what actions should be taken on set?

StanleyGoldman: Made them `protected`

src/GitHub.InlineReviews/ViewModels/CommentThreadViewModel.cs
StanleyGoldman: 👍

meaghanlewis:
This functionality works well! Tested in VS2017 with public, private and forked repositories. Also confirmed that the edit/delete options only show for your own comments.

The icons look really snazzy here ⚡️ 
screen shot 2018-05-29 at 12 51 52 pm

One thing I was wondering here is that I thought you changed the name of the button to `Update comment` in [this commit](https://github.com/github/VisualStudio/pull/1701/commits/81f7397c1ef120b62ce0b00a33b0052ffa3d2c97) to address Jamie and Don's feedback, but it seems that was changed and is now `Edit Comment`.
screen shot 2018-05-29 at 12 51 59 pm

src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
```diff
@@ -60,9 +61,17 @@ public class CommentViewModel : ReactiveObject, ICommentViewModel
             User = user;
             UpdatedAt = updatedAt;
 
+            var canDelete = this.WhenAnyValue(
+                x => x.EditState,
+                x => x == CommentEditState.None && user.Login.Equals(currentUser.Login));
+
+            canDelete.Subscribe(b => CanDelete = b);
```
grokys: You could use [`ObservableAsPropertyHelper`](https://reactiveui.net/docs/guidelines/framework/prefer-oaph-over-properties) to make this a bit neater.

grokys:
Oops, clicked "approve" before writing my comment ;)

Looks good and works well - just one small suggestion, but I don't think that should hold up an approval.

src/GitHub.Exports.Reactive/Services/IPullRequestSession.cs
```diff
@@ -111,7 +111,7 @@ public interface IPullRequestSession
         /// <param name="body">The comment body.</param>
         /// <param name="inReplyTo">The REST ID of the comment to reply to.</param>
         /// <param name="inReplyToNodeId">The GraphQL ID of the comment to reply to.</param>
-        /// <returns></returns>
+        /// <returns>A comment model.</returns>
```
sguthals: Adding an inline comment

meaghanlewis:
Nice catch @grokys! I'm able to update comments now when a review is pending.

meaghanlewis:
@StanleyGoldman - on a similar note about pending reviews...i don't seem to be able to delete comments while a review is pending. And then trying to update after clicking delete shows a `Not Found` error even though I still seem to be able to update the comment successfully.

cant delete pending comment

meaghanlewis:
One more thought about this. On dotcom when I go to delete a comment, I get a confirmation popup from the browser:

confirm delete comment

I'm wondering if it would be possible to add a confirmation in Visual Studio as well?

src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
```diff
@@ -24,6 +24,7 @@ public class CommentViewModel : ReactiveObject, ICommentViewModel
         CommentEditState state;
         DateTimeOffset updatedAt;
         string undoBody;
+        private bool canDelete;
```
grokys: Nit: Stray `private`

PR #1706, opened 2018-05-25
Don't block checkout when a file appears to have been deleted

### LINKED ISSUE #1691: "Cannot checkout as your working directory has uncommitted changes" appearing without pending changes
- GitHub Extension for Visual Studio version: 2.5.1.2234 & previous version
- Visual Studio version: 15.7.1 & previous

I am not able to checkout a branch from the pull request.  When I hover over the link, it a help text pops up that reads, "Cannot checkout as your working directory has uncommitted changes."  I do not have any pending changes in my current directory.  I have checked multiple times.  Could it be that it is not looking at the gitIgnore?  This seems to only be an issue with one repository.  It does allow me to checkout the branch on a sample and another small repo I have for another project, but not sure why it doesn't allow me on our main repo.  Any ideas?

image

jcansdale:
### Why this PR

Sometimes files can appear as `DeletedFromWorkdir` that that aren't visible on Team Explorer or when `git status` is executed. When this happens there is no obvious solution.

image

Since restoring a deleted file is non-destructive, there is no reason that I can think of to block a user from checking out another PR when a file has been deleted.

### What this PR does

- Don't block checkout when a file has been deleted (`DeletedFromWorkdir`)
- Don't block checkout when a file has been renamed
  - This appears as a combined `RenamedInWorkingDir` and `DeletedFromWorkdir`

### How to test

This PR fixes an issue a user was seeing where files appeared to be missing after a fresh clone that weren't visible to `git status` or Team Explorer changes. See #1691. This PR allows a checkout when a file is missing, so these phantom missing files are no longer a blocker.

1. Check our a clean branch
2. Delete a file that is part of the repo
3. Checkout a PR branch

Expect the `Checkout ...` link to be visible and for the PR branch to checkout successfully.

User has also confirmed the fix https://github.com/github/VisualStudio/issues/1691#issuecomment-392776846.

### What this PR doesn't do

We don't currently allow a checkout to be attempted in any circumstance. @grokys considers having implemented this check to have been a mistake.

A future PR could always allow checkouts to be attempted, but warn the user when they'd be taking changes with them to another PR branch (this might or might not we what they want - but it's impotent they know what's going on).

### Questions

*I've gone with not blocking checkout when a file has been renamed. We want to move in the direction of being more relaxed about checkouts.*

What should we do if a repository file has been renamed? By default this will appear as a missing (`FileStatus.DeletedFromWorkdir`) and an untracked (`FileStatus.NewInWorkdir`) file. After this change the user will be able to checkout a new PR, leaving the renamed file in the working directory. This isn't necessarily a bad thing since they would be able to do this from the command line anyway.

The alternative is to use `new StatusOptions { DetectRenamesInWorkDir = true }` and continue to block checkouts when it appears that a repository file has been renamed. See [here](https://github.com/jcansdale/VisualStudio/blob/e0896810c2ff1d55607fdf751440a46d6d524c79/test/UnitTests/GitHub.App/Services/PullRequestServiceTests.cs#L227).

### Related

- Don't block checking out a PR if there are untracked files: #1302
- [SPIKE] Don't do "clean" check before attempting to check-out PR: #1304

Fixes #1691

### DIFF
--- a/src/GitHub.App/Services/PullRequestService.cs
+++ b/src/GitHub.App/Services/PullRequestService.cs
@@ -150,23 +150,38 @@ namespace GitHub.Services
             using (var repo = gitService.GetRepository(repository.LocalPath))
             {
                 var statusOptions = new StatusOptions { ExcludeSubmodules = true };
-                var isClean = !IsFilthy(repo.RetrieveStatus(statusOptions));
+                var status = repo.RetrieveStatus(statusOptions);
+                var isClean = !IsCheckoutBlockingDirty(status);
                 return Observable.Return(isClean);
             }
         }
 
-        static bool IsFilthy(RepositoryStatus status)
+        static bool IsCheckoutBlockingDirty(RepositoryStatus status)
         {
             if (status.IsDirty)
             {
-                // This is similar to IsDirty, but also allows NewInWorkdir files
-                return status.Any(entry =>
-                    entry.State != FileStatus.Ignored &&
-                    entry.State != FileStatus.Unaltered &&
-                    entry.State != FileStatus.NewInWorkdir);
+                return status.Any(entry => IsCheckoutBlockingChange(entry));
             }
 
             return false;
+        }
+
+        // This is similar to IsDirty, but also allows NewInWorkdir and DeletedFromWorkdir files
+        static bool IsCheckoutBlockingChange(StatusEntry entry)
+        {
+            switch (entry.State)
+            {
+                case FileStatus.Ignored:
+                    return false;
+                case FileStatus.Unaltered:
+                    return false;
+                case FileStatus.NewInWorkdir:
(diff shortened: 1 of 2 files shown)


grokys:
Maybe we should just not be even trying to do this, and just let the checkout fail if it has to?

jcansdale:
@grokys,

> Maybe we should just not be even trying to do this, and just let the checkout fail if it has to?

I tried doing it this way a while back, but didn't think it worked well with the current UI. The problem is that it invites the user to potentially end up in a messy state (bringing changes to a PR they're about to review). When reviewing PRs, I think a clean checkout is particularly important.

meaghanlewis:
Im able to checkout PR branches successfully after deleting and renaming files now 👍

PR #1712, opened 2018-05-31
Refactoring Pull Request Models.

grokys:
## Description

Previously we were using a mix of REST and GraphQL for Pull Requests and Pull Request Reviews, and our data model didn't fit either very well. This PR:

- Adds a `PullRequestDetailModel` which is modeled after [what we're hoping will be the future shape of the GraphQL pull request/reviews API](https://github.com/github/experience-engineering-code/pull/720)
- Moves reading pull request details out of `ModelService` and into `PullRequestSession` - this is now the only place to get hold of a PR details model. Doing this means that all parts of the extension always agree about the state of a pull request, and prevents some unnecessary API reads that we were carrying out before - our UI should feel a lot snappier now!
- Reads `PullRequestDetailModel` using GraphQL. Because the GraphQL API for reviews and comments isn't complete yet, after reading this we massage the model into what should be the final shape after reading (see [`BuildPullRequestThreads`](https://github.com/github/VisualStudio/compare/refactor/pr-models?expand=1#diff-27e704a8cecf704234bf4fe243bcfc93R764)

The old `PullRequestModel` is still there (although in a cut-down form) and still read using REST. It is now used just for the PR list. This will be moved to GraphQL in a later PR.

## Testing

Basically, everything around pull requests/reviews/comments should work as before!

## TODO

- [x] Make adding comments work - Octokit.GraphQL is currently trying to use the `Mutation` object for auto-paging and failing
  - Worked around this by re-reading the PR details after submitting a comment. This might not be ideal, so might need to be revisited. The Octokit.GraphQL issue (https://github.com/grokys/Octokit.GraphQL/issues/93) also needs to be fixed.
- [x] Update with changes from #1701 once merged
- [x] Add missing XML doc comments


test/GitHub.InlineReviews.UnitTests/ViewModels/NewInlineCommentThreadViewModelTests.cs
```diff
@@ -10,6 +10,11 @@ namespace GitHub.InlineReviews.UnitTests.ViewModels
 {
     public class NewInlineCommentThreadViewModelTests
     {
+        public NewInlineCommentThreadViewModelTests()
+        {
+            Splat.ModeDetector.Current.SetInUnitTestRunner(true);
```
jcansdale: We should probably move this into an assembly wide setup for the whole test assembly. This is possible with NUnit in a way it wasn't with xUnit.

For example we could move it into `AssemblyInfo.cs` like this:
https://github.com/github/VisualStudio/blob/ebb900e8ef58fd25af32d69421624bac5fe52c0b/test/GitHub.App.UnitTests/Properties/AssemblyInfo.cs#L45

Otherwise it's too easy to forget and end up with hanging or not depending on the order they're executed. 😨

src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
```diff
@@ -260,6 +266,108 @@ public ITextDocument GetDocument(ITextBuffer buffer)
             return null;
         }
 
+        public virtual async Task<PullRequestDetailModel> ReadPullRequestDetail(HostAddress address, string owner, string name, int number)
+        {
+            if (readPullRequest == null)
+            {
+                readPullRequest = new Query()
+                    .Repository(Var(nameof(owner)), Var(nameof(name)))
```
jcansdale: I'm curious why you're caching the compiled query here but not other queries? I've found an easy way to automatically handle the named arguments and caching if this is something users of GraphQL should be doing.

src/GitHub.VisualStudio/Views/GitHubPane/PullRequestReviewSummaryView.xaml
```diff
@@ -15,9 +16,9 @@
 
     <d:DesignData.DataContext>
         <vm:PullRequestReviewSummaryViewModel Id="1" State="Pending" FileCommentCount="2">
-            <vm:PullRequestReviewSummaryViewModel.User>
+            <!--<vm:PullRequestReviewSummaryViewModel.User>
```
jcansdale: Is this commented out for a reason?

src/GitHub.App/ViewModels/GitHubPane/PullRequestFilesViewModel.cs
```diff
@@ -105,6 +105,7 @@ public void Dispose()
                 { string.Empty, new PullRequestDirectoryNode(string.Empty) }
             };
 
+            await Task.Delay(0);
```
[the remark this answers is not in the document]
grokys: Oops, that was left in from when I commented out the whole of the method. Putting a `Task.Delay(0)` in was my way of preventing the "method doesn't use await" error. Will remove.

src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
grokys: Basically just because this was written from scratch and the other queries were just modified. We should probably be doing the same with the other queries, not sure if that should be in this PR or another?

src/GitHub.VisualStudio/Views/GitHubPane/PullRequestReviewSummaryView.xaml
grokys: Fixed.

src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
jcansdale: Can I show you my GraphQL prototype before we do the the same with the other queries? I'd like to find out if you think it's a good idea in principle before developing it any further. So maybe push this to another PR.

grokys:
@jcansdale are you sure this was caused by this PR? This doesn't look like anything that this PR should have changed. Can you leave a comment in the same position on master?

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -273,7 +273,7 @@ static Task<int> Where(string fileName)
             }
         }
 
-        public IObservable<Unit> Checkout(ILocalRepositoryModel repository, IPullRequestModel pullRequest, string localBranchName)
+        public IObservable<Unit> Checkout(ILocalRepositoryModel repository, PullRequestDetailModel pullRequest, string localBranchName)
```
sguthals: Leaving a comment!

src/GitHub.InlineReviews/Services/PullRequestSessionService.cs
jcansdale: I wanted to get your thoughts on this:
https://github.com/grokys/Octokit.GraphQL/pull/104

PR #1719, opened 2018-06-07
[spike] Infer PR associated with current branch

### LINKED ISSUE #1591: Current PR doesn't appear on status bar when checked out from command line

jcansdale:
This is a spike to experiment with infering the PR associated with the current branch.

### What this PR does

- Search current and parent repos for a PR with the same HEAD branch name as the current branch

### TODO

Use `associatedPullRequests` property of `ref`.

### Findings

There can be multiple PRs that use the same branch as their HEAD. Can we could up with a conservative strategy that minimizes false positives but that still works in most cases?

For example, say a user creates a PR using their `master` branch. They probably won't want this PR associated with their `master` branch forever!

There are attributes associated with a PR that we could take into account to infer when a PR is likely stale and shouldn't be automatically associated with the current branch. For example:

- If the SHA a branch is different to the PR it might be stale
- If a PR has been closed it is likely to be of less interest to a user to an open one

Fixes #1591

### DIFF
--- a/src/GitHub.App/SampleData/LocalRepositoryModelDesigner.cs
+++ b/src/GitHub.App/SampleData/LocalRepositoryModelDesigner.cs
@@ -11,7 +11,7 @@ namespace GitHub.App.SampleData
     public class LocalRepositoryModelDesigner : ILocalRepositoryModel
     {
         public UriString CloneUrl { get; set; }
-        public IBranch CurrentBranch { get; set; }
+        public ILocalBranch CurrentBranch { get; set; }
         public Octicon Icon { get; set; }
         public string LocalPath { get; set; }
         public string Name { get; set; }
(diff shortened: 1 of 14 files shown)


jcansdale:
We're planning to warn the user with a call-out notification (see #2026) rather than having alternative logic to choose the default remote when "origin" isn't defined. This will make it obvious what's going on and avoid any magic.

PR #1737, opened 2018-06-15
Refactoring PR list

grokys:
This PR refactors the PR list to:

## Use GraphQL for loading

By using GraphQL we can load only the fields we're interested in, which speeds up our queries.

## Use data virtualization so that only visible items will be loaded

Previously, when the GitHub pane was opened, all pull requests were loaded. This could take minutes for repositories with a lot of pull requests, and the whole time they were loading VS would be slow.

With this PR, only the first 100 PRs will be loaded when the GitHub pane is shown, which is a single request. When the user scrolls down, more PRs will be loaded on demand.

## Display a message when no items found

We currently have two messages:

When no filters are active and no PRs are found:

image

When filters are active and no PRs are found:

image

- Improve the author filter

The author filter is now searchable:

image

Unfortunately because we're now only loading the first page of results, we don't have a full list of all users that have opened a PR, instead we use the GraphQL "Assignable Users" endpoint to get a list similar to on .com and allow the user to also select logins that they enter that aren't on that list.

Depends on #1712


meaghanlewis:
@grokys Dogfooding this and am so impressed with all the improvements to the PR list! 💥 

Leaving some feedback and questions. I know this is still WIP so apologies if you're already addressing some of these things.

### General feedback (and screenshots 😄 )
- The new PR list is looking so smooth. It's nice to tell from a glance things like which PR branch I'm on, how many comments a PR has, and the number of open/closed/all PRs.
new pr list

- I love the new PR author filter!
new author filter

- And the view when there are no PR's is great too.
no prs

### Things to fix

- No link to open PR in browser
no link on pr

- Link to open `pulls` in browser doesn't do anything
link to open pulls doesnt work


### Questions

- When the PR list is still loading the count shows 0 for PRs. I think it's fine as is, but wondering if it would be possible to not show this count until it's loaded? 
loading prs

- When there are no PRs the author filter shows but there is nothing to filter so this option doesn't really do anything. Is it possible to not show this filter or the count of PRs here?
filter when no prs

cc @donokuda for thoughts on the above questions ^

jcansdale:
Most people on the list of users haven't submitted any PRs (open or closed). Could we populate the list of users based on the filter? For example could we only show users with open PRs on the open PRs filer and users with closed PRs on the closed PRs filer?

image

This would whittle down the list considerably and make it much quicker to find users!

donokuda:
Jumping in to answer some questions. 

> When the PR list is still loading the count shows 0 for PRs. I think it's fine as is, but wondering if it would be possible to not show this count until it's loaded?

I like the idea of not showing the count and wonder if it makes sense to also hide the filtering options until the pull request list is loaded. We could consider moving the Create New action to the right side again and keep the filtering options closer to each other:



> When there are no PRs the author filter shows but there is nothing to filter so this option doesn't really do anything. Is it possible to not show this filter or the count of PRs here?

I think it's okay to have the count be present even if there are no pull requests in the list. 

>  Could we populate the list of users based on the filter? For example could we only show users with open PRs on the open PRs filer and users with closed PRs on the closed PRs filer?

I'm 👍for  this. We might want to be on the look out to see if any folks think it's a bug if they expected someone to appear in the list but isn't there. We could consider showing a "no results found" message if someone is filtering and we're unable to match anyone.

This is starting to look good! I'll block out some time to explore visual / UX improvements today.

grokys:
> No link to open PR in browser

This was by design. The list is now a selectable list, and selectable lists don't generally have inner clickable items. I was thinking of adding a context menu with "Open on GitHub"?

> When the PR list is still loading the count shows 0 for PRs. I think it's fine as is, but wondering if it would be possible to not show this count until it's loaded?

Yep, definitely! 👍 

> Most people on the list of users haven't submitted any PRs (open or closed). Could we populate the list of users based on the filter? For example could we only show users with open PRs on the open PRs filer and users with closed PRs on the closed PRs filer?

Unfortunately this is a little more difficult than it would first seem. Because the list is now virtualized we don't _have_ a complete list of users with open/closed PRs - we only have potentially the first page. I've considered adding the authors that we _do_ have to the filter, but haven't tried it out. That might work, but it would be an incomplete list.

Because of this, the filter uses the "assignable users" GraphQL endpoint, which is what .com seems to do. If you go to the author filter on .com you will notice something similar to what we have here.

> When there are no PRs the author filter shows but there is nothing to filter so this option doesn't really do anything. Is it possible to not show this filter or the count of PRs here?

Not showing the filter would certainly be possible, though I'm a little wary for hiding things when they're not relevant because from what I understand of UX, it can hurt discoverability. I could be wrong on the latest thinking around that though ;)

As @donokuda said, I think we should still show the count: seeing that there are 0 results is useful info.

donokuda:
> Unfortunately this is a little more difficult than it would first seem. Because the list is now virtualized we don't have a complete list of users with open/closed PRs - we only have potentially the first page. I've considered adding the authors that we do have to the filter, but haven't tried it out. That might work, but it would be an incomplete list.
> 
> Because of this, the filter uses the "assignable users" GraphQL endpoint, which is what .com seems to do. If you go to the author filter on .com you will notice something similar to what we have here.

Good technical info to know about 😄I'm fine with not worrying about filtering and using what's already implemented there. By going with what we get out of the box, that will allow us to move and learn quickly.

donokuda:
FYI: After chatting with @grokys, we're going to move any design discussions to a separate issue so that this pull request can stay focused on refactoring instead of redesigning 😄

donokuda:
Out of curiosity, what is the timestamp of each PR list item referencing? It looks like it's something other than the time it was opened. Would it make sense to use the "Opened at" date instead?

simurai:
> Would it make sense to use the "Opened at" date instead?

👍 

.com allows to sort by "[Least recently updated](https://github.com/github/VisualStudio/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-asc)". It's interesting because it highlights which are the more "active" PRs. There could be "old" PRs that just take a while, but still get updated regularly.

On the other hand, sorted by "Newest" makes the list much more constant/predictable. There is no random re-ordering and new PRs just show up at the top. Maybe good to show that as the default.

donokuda:
> On the other hand, sorted by "Newest" makes the list much more constant/predictable. There is no random re-ordering and new PRs just show up at the top. Maybe good to show that as the default.

Yeah, I think if we're settled on sorting by recently updated, then we should consider reflecting that in the timestamp (like what you mentioned that .com does.)

grokys:
This PR is now ready for review! There's still some discussion around #1768 so I've not merged that in yet. If a decision is made before this PR is merged, then we can merge it in here, otherwise it can be made to master.

src/GitHub.App/Services/RepositoryService.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        [ImportingConstructor]
+        public RepositoryService(IGraphQLClientFactory graphqlFactory)
+        {
+            Guard.ArgumentNotNull(graphqlFactory, nameof(graphqlFactory));
+
+            this.graphqlFactory = graphqlFactory;
+        }
+
+        public async Task<string> ReadParentOwnerLogin(HostAddress address, string owner, string name)
```
jcansdale: I wonder if this would be more useful as a general `FindParent` method that returns a `(string owner, string repositoryName)`? A parent repository might not have the same name as its child and we may want to support this as some point.

src/GitHub.App/Collections/IVirtualizingListSource.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Threading.Tasks;
+
+namespace GitHub.Collections
+{
+    /// <summary>
+    /// A loader for a virtualizing list.
+    /// </summary>
+    /// <typeparam name="T">The item type.</typeparam>
+    /// <remarks>
+    /// This interface is used the the <see cref="VirtualizingList{T}"/> class to load pages of data.
```
jcansdale: "by the" not "the the"?

src/GitHub.App/Services/RepositoryService.cs
grokys: Ah yeah, good point. I will make this change.

grokys: I changed this to return the repository name and owner. We don't actually use the `name` yet,: `UriString` doesn't support getting a new URL with a different name as well as owner so want to do that in a separate PR, this is big enough as it is ;)

PR #1744, opened 2018-06-16
Using builtin file & folder icons

### LINKED ISSUE #1745: PR changes view should use builtin VS icons
image

The icons (folders, files) are custom and not native VS icons. It would be nice if they were the same as those used in the team explorer changes window to
1. make the UI fit in better with the rest of VS
2. show a special icon for each file type just like in the solution/team explorer as opposed to having the same generic icon for all files

Neme12:
Fixes #1745

Before:
image

After:
image

image


src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
```diff
@@ -386,6 +392,8 @@
     <Compile Include="Views\Dialog\RepositoryRecloneView.xaml.cs">
       <DependentUpon>RepositoryRecloneView.xaml</DependentUpon>
     </Compile>
+    <Compile Include="Views\GitHubPane\DirectoryIsExpandedToImageMonikerConverter.cs" />
+    <Compile Include="Views\GitHubPane\FileNameToImageMonikerConverter.cs" />
```
Neme12: Where would be an appropriate place to put these converters? I noticed there is a folder with some converters in GitHub.UI but that project doesn't have a lot of dependencies on VS itself (which these do). Also these aren't really general purpose converters like the other ones seem to be, they're specifically made for one view. But this isn't a good place either. Where should I put them?

src/GitHub.VisualStudio/Views/GitHubPane/PullRequestFilesView.xaml
```diff
@@ -60,27 +66,8 @@
                             Tag="{Binding DataContext, ElementName=root}"
                             KeyboardNavigation.DirectionalNavigation="None">
 
-                    <!-- 
-                      We need to change the color of the file icon when the file is deleted, but applying the style
-                      to OcticonImage directly borks the designer (see #1410). Work around this by applying the color
-                      to a parent control and let the foreground be inherited by the icon.
-                    -->
```
Neme12: I noticed this used to gray out the icon if the item was deleted. This can't really be done with arbitrary icons. I tried replacing the color change with reduced opacity, but it didn't look very good in my opinion. And the built-in team explorer view doesn't do anything like that either (it's only the text that is grayed out). Is it OK to remove this?

src/GitHub.VisualStudio/Views/GitHubPane/PullRequestDetailView.xaml
```diff
@@ -8,8 +8,10 @@
                                     xmlns:prop="clr-namespace:GitHub.VisualStudio.UI;assembly=GitHub.VisualStudio.UI"
                                     xmlns:markdig="clr-namespace:Markdig.Wpf;assembly=Markdig.Wpf"
                                     xmlns:vsui="clr-namespace:Microsoft.VisualStudio.Shell;assembly=Microsoft.VisualStudio.Shell.14.0"
+                                    xmlns:theming="clr-namespace:Microsoft.VisualStudio.PlatformUI;assembly=Microsoft.VisualStudio.Imaging"
                                     Background="{DynamicResource GitHubVsToolWindowBackground}"
                                     Foreground="{DynamicResource GitHubVsWindowText}"
+                                    theming:ImageThemingUtilities.ImageBackgroundColor="{Binding RelativeSource={RelativeSource Self}, Path=Background.Color}"
```
Neme12: This is important to ensure that the icons have the right background in all themes. `Transparent` doesn't work either (it seems to behave almost the same as white).

If this was set to `Transparent`, it would look like this on a dark theme:
image

Correct behavior:
image

Neme12:
Note that this also shows any custom icons correctly. For example after installing this extension:
https://marketplace.visualstudio.com/items?itemName=MadsKristensen.FileIcons

we can see all the new icons:
image
In this case the icon for `.editorconfig` is actually provided by the extension and not part of Visual Studio (the icon for `.gitignore` is wired up by the extension but the image itself comes from VS).

Neme12:
I'm not sure what's failing in CI. Can someone help please? Everything is building for me locally.

Neme12:
@grokys Please consider adding `<LangVersion>6</LangVersion>` to the project file. It will prevent contributors from using newer language features and set a standard for everyone, no matter which version of VS they happen to be using. On top of that, it will also prevent the IDE from offering suggestions or refactorings for features that can't be used.

Doing this is always a good idea even if you move to VS 2017 so that there's never confusion about which language version we're using - whether it be C# 6, 7.0, 7.1, 7.2 or 7.3.

Neme12:
There's another error in CI:
> error CA1812: Microsoft.Performance : 'DirectoryIsExpandedToImageMonikerConverter' is an internal class that is apparently never instantiated.

This is incorrect. The class is used from XAML :-/

Neme12:
I was able to simplify this further and find the proper way to do this when I stumbled upon `IVsImageService2` by looking through APIs in intellisense :smile:

src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
grokys: Personally I think that where you've put them is the best place for them: as you say they're tied to a particular view, so it makes sense to put the converters alongside. Not sure if anyone else things differently though?

src/GitHub.VisualStudio/Views/GitHubPane/PullRequestFilesView.xaml
grokys: Yeah that's fine with me!

src/GitHub.VisualStudio/Views/GitHubPane/FileNameToImageMonikerConverter.cs
```diff
@@ -19,6 +22,10 @@ public FileNameToImageMonikerConverter()
 
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
+            // In design mode, imageService will be null
+            if (DesignerProperties.GetIsInDesignMode(new DependencyObject()))
+                return default(ImageMoniker);
```
Neme12: Alternatively, we could just check if `imageService` is null and return `KnownMonikers.Document` as a fallback, even though it should never be null when running inside VS.

Neme12: What we return doesn't matter because these icons can't be shown in the designer anyway

Neme12:
Sorry for so many merge commits. I didn't notice that you added a merge commit in my branch and just used the regular sync feature in VS, so that created another one. Now I saw this message with "This branch is out-of-date with the base branch" which I've never seen before and was curious about what it does :smile: So that's another merge commit... But interestingly, that message is still there and didn't go away. Could that be a bug?

grokys:
No, it's probably just that another PR was merged in the meantime. Don't worry about the merge commits, they're pretty much a fact of life ;)

grokys:
@meaghanlewis this looks good to me. If you could take a look and it looks good to you I think we should merge asap

meaghanlewis:
This looks great to me! Thanks @Neme12  🎉

Neme12:
Note: When writing code, I was often wondering if there is any preferred code style for this repository that I should use. For example: would you prefer to be explicit about access modifiers (private, internal) or omit them? are braces required? can I use expression bodied methods? are there any naming conventions for const fields?

If there are any standard conventions for these, it would be nice to codify them inside .editorconfige so that contributors can be aware of them, obide by them if necessary and the IDE can help them do the right thing. If there are any concerns about setting a hard rule, these could be set to a "suggestion" (not a warning). The same could be done for naming styles. Even if the severity is set to none (it's not a suggestion, warning or an error if you get this wrong) it can still be helpful because the IDE will generate code based on this preferred style when using refactorings.

But given that there were no comments on the style I chose, I'm wondering if it's maybe not considered that important in this codebase and everyone is free to use the style they prefer? (or did I just get everything right? :smile:) If that's the case, doing this might not bring that much value after all. I don't know the preferences, so that's why I'm asking and suggesting ideas. 🎉 What do you think? Would that be a good fit for this repo?

jcansdale:
@Neme12 This would  be an excellent fit for this repo!  It's just a question of carving out the time. If you're up for driving it in #1750, that would be great. 😄 

Could you add a list of questions that would need answering? I'm not very familiar with `.editorconfig` files. I'm not sure about the rest of the team.

src/GitHub.VisualStudio/Views/GitHubPane/PullRequestDetailView.xaml
grokys: Just ran into this on the PR review authoring view. I've tried moving this property to `PullRequestFilesView` and it seems to work fine. What am I missing?

PR #1746, opened 2018-06-17
Enforce language version for C# 6

Neme12:


Neme12:
> If we do decide to require VS 2017, what language version do you think we should target?

Is that a question for me? It depends on what version of VS 2017 you're happy mandating. Unless there's need for some of the `ref` or `in` features for performance somewhere, it's mostly just minor convenience features, so probably not a huge benefit..

That said, I've gotten quite used to some of them, for example non-trailing named arguments for clarity in C# 7.2:
```c#
M("argument", allowEmpty: true, cancellationToken) // as opposed to
M("argument", allowEmpty: true, cancellationToken: cancellationToken)
```

I think 7.2 would probably be a good compromise, it came out with 15.5 last year. 7.3 came out very recently with 15.7 (though I would definitely recommend everyone to update, the performance improvements are really amazing).

grokys:
I'm not sure why it's even a question to be honest ;) As you say, we should set it at the highest lang version that the version of VS we use supports.

Issue #1727, opened 2018-06-11
Update the messaging in the GitHub panel when a solution is not open


sguthals:
When a solution isn't open, the messaging in the GitHub panel is:
<img width="526" alt="screen shot 2018-06-11 at 12 51 32 pm" src="https://user-images.githubusercontent.com/1314285/41255230-b2ebe07c-6d7a-11e8-86af-898ad581b37b.png">

But there is no such menu item when there is no open solution:
<img width="1591" alt="screen shot 2018-06-11 at 12 57 37 pm" src="https://user-images.githubusercontent.com/1314285/41255243-c2eb3284-6d7a-11e8-95e5-d862a9c03b2d.png">

If there is no open solution, we should change the messaging to "Go to Team Explorer and open a solution" or something along those lines.

meaghanlewis:
I believe this messaging just shows in VS2017. With VS2015, the last opened solution stays open after restarting VS and until a new solution is selected.

meaghanlewis:
The message looks like it is defined here: https://github.com/github/VisualStudio/blob/0bf5d88f1ddbaaab62f9cf1d91f3a9d530e6d3a0/src/GitHub.VisualStudio.UI/Resources.resx#L328

micheleissa:
Hello,
I can take this one. Do you guy just want the verbiage to change?

meaghanlewis:
Hi @micheleissa that would be great if you can update this! Yes, we just want the message to be simplified to say something like `Open a git project or solution to get started.` Feel free to open a PR with the change and we'll review it.

mariombaltazar:
Hi,
Is this issue still open?

GuillemSeCa:
Is this issue still open should I work on it?

PR #1751, opened 2018-06-19
Change Not git repo message

micheleissa:
- Per #1727  change the text of the message.


grokys:
Thanks @micheleissa! I'm wondering if we should maybe have two messages: one for when there's no solution opened and one for when there's a solution opened but it's not a git repository? Not sure really, @sguthals what do you think?

jcansdale:
@grokys,

> I'm wondering if we should maybe have two messages: one for when there's no solution opened and one for when there's a solution opened but it's not a git repository?

I think we should do this with another message for when a repository exists but there is no `origin` remote defined. At the moment users are given no clues as to what's going on.

grokys:
Yeah, possibly, but that sounds like a separate PR to me?

jcansdale:
> Yeah, possibly, but that sounds like a separate PR to me?

I just wanted to mention there were a few different things we might want to check for -- eventually. 😄

micheleissa:
So will this pr be accepted or not?

sguthals:
@grokys yes! I agree with:
> one for when there's no solution opened and one for when there's a solution opened but it's not a git repository?

@micheleissa Are you up for making that distinction as part of this pr?

micheleissa:
@sguthals I can try if it is easy to know there is a an opened non-git solution. I might ask for some directions.

micheleissa:
I'm not familiar at all with Xaml so I don't know where to begin..all I know for now is the view GenericNotAGitRepositoryView is using NotAGitRepositoryMessage resource. Could you guys shed some light on that and where does this panel even displayed?

micheleissa:
@grokys  thanks. Do I need to check in https://github.com/github/VisualStudio/blob/master/src/GitHub.App/ViewModels/GitHubPane/GitHubPaneViewModel.cs#L396-L407 if a solution open or not and also for a msg if not a git repo I think this is handled already.

grokys:
@micheleissa yes that's what you'll need to do. You'll want to abstract the "is a solution open?" check into a service so that it can be unit tested.

StanleyGoldman:
Hey @micheleissa are you still interested in working on this?

micheleissa:
I might take a look at it this weekend. Do you guys have a deadline for it?

vbarcaru:
I'm not working yet

meaghanlewis:
No, there is no deadline @micheleissa. Just wondering if you were still interested in completing this work.

micheleissa:
@meaghanlewis thank you, I'm trying to re-produce the issue but I can't. Have any of you guys managed to produce it and if so could you tell me how? Thanks again.

jcansdale:
Closing some very old PRs.

PR #1761, opened 2018-06-22
Update eligible projects to new SDK format.

grokys:
Now we're using VS2017 we can use the new simplified SDK-style .csproj format. This PR updates the projects that can be updated to this format. It also enables `Microsoft.CodeAnalysis.FxCopAnalyzers` on them, which is the successor to CodeAnalysis. I've not enabled warnings as errors yet though, because these analyzers find a fair few problems that will need to be fixed first.

WPF projects and VSSDK projects can't use the SDK format yet. In addition `GitHub.TeamFoundation.15` uses an alias on a nuget package which can't currently be done, so the `GitHub.TeamFoundation.X` projects haven't been converted.


Neme12:
> VSSDK projects can't use the SDK format yet

I've seen VS extension projects using the new project system in Microsoft repos, for example here:
https://github.com/dotnet/roslyn/blob/master/src/VisualStudio/Setup/VisualStudioSetup.csproj

but haven't found any info or documentation on that. Maybe this is an experimental & undocumented feature?

Specifically this part seems relevant:
```xml
<ImportVSSDKTargets>true</ImportVSSDKTargets>
```

jcansdale:
I appears the `nunit3-vs-adapter` recently went through a similar process https://github.com/nunit/nunit3-vs-adapter/pull/499.

Maybe @jnm2 will have some tips for us. 😄

Neme12:
@jcansdale I think you can use PackageReference even without the new project file format, it should be orthogonal to that.

Neme12:
In fact I thought this would be at least a 2-step transition with multiple PRs to make this huge change easier to review and take iteratively.

jnm2:
Sorry to disappoint, but when I looked at the VSIX project, I didn't see any ROI in getting it working with the new SDK. It's possible Oren has something up his sleeve—I know he has WPF working on the new csproj SDK.

Neme12:
I don't understand what the problem with WPF is? I just tried creating a new WPF project & changing the csproj and everything worked fine.

The only problem is that `.xaml` files are not included properly by default, so you have to list each one of them in the csproj just like today. At least that's what I thought until I found this:
https://github.com/Microsoft/VSProjectSystem/issues/169#issuecomment-277981978

So now my csproj looks like:
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net46</TargetFramework>
    <OutputType>WinExe</OutputType>
  </PropertyGroup>
  <ItemGroup>
    <Reference Include="System" />
    <Reference Include="System.Data" />
    <Reference Include="System.Xml" />
    <Reference Include="Microsoft.CSharp" />
    <Reference Include="System.Core" />
    <Reference Include="System.Xml.Linq" />
    <Reference Include="System.Data.DataSetExtensions" />
    <Reference Include="System.Net.Http" />
    <Reference Include="System.Xaml" />
    <Reference Include="WindowsBase" />
    <Reference Include="PresentationCore" />
    <Reference Include="PresentationFramework" />
  </ItemGroup>
  <ItemGroup>
    <None Remove="**\*.xaml" />
    <ApplicationDefinition Include="App.xaml">
      <Generator>MSBuild:Compile</Generator>
      <SubType>Designer</SubType>
    </ApplicationDefinition>
    <Page Include="**\*.xaml" Exclude="App.xaml">
      <Generator>MSBuild:Compile</Generator>
      <SubType>Designer</SubType>
    </Page>
    <Compile Update="**\*.xaml.cs">
      <DependentUpon>%(Filename)</DependentUpon>
      <SubType>Code</SubType>
    </Compile>
  </ItemGroup>
</Project>
```
everything works perfectly and it is stable when adding/removing xaml inside VS (it doesn't modify the project file).

Directory.Build.Props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <PropertyGroup>
+    <Product>GitHub Extension for Visual Studio</Product>
+    <Version>2.5.4.0</Version>
+    <Copyright>Copyright © GitHub, Inc. 2014-2018</Copyright>
```
jcansdale: We could maybe do something like this to future-proof... 😉 
```
<CopyrightYear>$([System.DateTime]::Now.ToString("yyyy"))</CopyrightYear>
<Copyright>Copyright © GitHub, Inc. 2014-$(CopyrightYear)</Copyright>
```

src/GitHub.Services.Vssdk/GitHub.Services.Vssdk.csproj
```diff
@@ -1,161 +1,16 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
jcansdale: I'm confused where this project gets its VS SDK references from.

grokys: It gets them from GitHub.Exports.

grokys:
@Neme12 yep, i've managed to get WPF projects working now. Thanks for the pointer - I'd tried something similar before but couldn't get it to work. Not sure what I was missing.

src/common/xaml.props
```diff
@@ -0,0 +1,12 @@
+<Project DefaultTargets="Build">
+  <ItemGroup>
+    <Compile Update="**\*.xaml.cs">
+      <DependentUpon>%(Filename)</DependentUpon>
+      <SubType>Code</SubType>
+    </Compile>
+    <Page Include="**\*.xaml">
```
Neme12: The reason why I included `<None Remove="**\*.xaml" />` before this in my example is that without that, when adding a new xaml control inside VS adds a new line here with `<None Remove="AddedFile.xaml" />`

But I have no idea how this behaves and what VS would do when this is in am imported project. Did you check that it works properly and doesn't modify the csproj?

grokys: Hmm yeah you're right, it does modify the .csproj...

grokys: Looks like the is the associated issue: https://github.com/dotnet/project-system/issues/3639

grokys: I think for the moment I'm going to revert the change for converting our WPF projects to the new csproj and try again when 15.8 is available.

Neme12: I didn't think this was a bug. They are included as None by default so VS tries to do its best to undo that and make sure they're only included as Content.

jnm2:
@jcansdale Hey, here's an example of a new-style VSIX project by @sharwell: https://github.com/tunnelvisionlabs/JustMyCodeToggle/blob/master/Tvl.VisualStudio.JustMyCodeToggle/Tvl.VisualStudio.JustMyCodeToggle.csproj

sharwell:
If you want WPF projects to work, I recommend using https://github.com/onovotny/MSBuildSdkExtras by @onovotny.

jnm2:
@sharwell Just FYI, my ballpark guess is that I use your JMC toggle around ten times per the average day.

sharwell:
😆 I don't use it at all. I only built it because you wanted it.

grokys:
@sharwell just got back to this and tried using MSBuild.Sdk.Extras but I'm still getting the:

```
  <ItemGroup>
    <Compile Remove="Controls\UserControl1.xaml.cs" />
  </ItemGroup>
```

added whenever I add a XAML/cs file combination, and they don't appear nested.

onovotny:
Have you added `<ExtrasEnableWpfProjectSetup>true</ExtrasEnableWpfProjectSetup>` to your project? That turns on the WPF defaults.

grokys:
@onovotny yeah, I have. Just pushed the commit with the change in. If you want to try it, try going to GitHub.UI and adding a WPF user control and saving the solution.

grokys:
@onovotny this seems to be fixed with MSBuild.Sdk.Extras 1.6.52 - thanks!

test/GitHub.TeamFoundation.UnitTests/VSGitExtTests.cs
```diff
@@ -17,7 +17,7 @@
 
 public class VSGitExtTests
 {
-    public class TheConstructor : TestBaseClass
+    public class TheConstructor
```
shana: Why is this inheritance changing?

grokys:
> Looks good in general, I'm just wondering where the settings for code analysis are.

Yes it's using the default ruleset for now, and not enabling warnings as errors. The new analyser has a _lot_ more checks that are very useful and that I don't want to disable (such as always using `ConfigureAwait()` on `Task`s) but I didn't want to litter this PR with changes to fix these warnings.

My reasoning was that we should fix these warnings and enable warnings as errors in a separate PR.

test/GitHub.TeamFoundation.UnitTests/VSGitExtTests.cs
grokys: Ummm... I can't remember. I'll put this back.

Directory.Build.Props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <Product>GitHub Extension for Visual Studio</Product>
-    <Version>2.5.5.0</Version>
+    <Version>2.5.6.0</Version>
```
onovotny: Do you need to manually version? There are better ways to do this :)

grokys: Oh really? Please do elucidate! What would be a better way to do this?

jcansdale: I've always hated version stamping. I'd be interested in finding the least-hacky way! 😉

onovotny: I have a project that matches this one very closely -- xUnit for Devices. It has a VSIX for its templates and auto-versions.

https://github.com/xunit/devices.xunit

Here are some key things:

1. Uses Nerdbank.GitVersioning to get a version for each commit. That handles the csproj's.
2. Build scripts update other xml assets to match: https://github.com/xunit/devices.xunit/blob/master/.vsts-shared.yml#L31-L36

PR #1767, opened 2018-06-28
Update to Octokit.GraphQL to 0.1.0

### LINKED ISSUE #1774: Extension ships with a newer Newtonsoft.Json than Visual Studio 2017
The currently published extension ships with Json.NET version `10.0.x` whereas VS ships built-in with `9.0.x`. It's strongly advised that you do not ship newer versions of built-in VS assemblies, since that can cause incompatibilities or breakages in other packages. 

Might be related to https://developercommunity.visualstudio.com/content/problem/287648/xamarin-android-and-ios-projects-will-not-load-aft.html.

StanleyGoldman:
Update Octokit.GraphQL to 0.1.0-beta version from nuget.org

Fixes #1774.


grokys:
I think `lib/Octokit.GraphQL.0.0.6-alpha.nupkg` can be removed, right?

grokys:
Regarding the CI failures,I'm not sure why they're failing now when they weren't failing before, but at one point I had to add a binding redirect to the unit tests project, before it was split apart: 

https://github.com/github/VisualStudio/pull/1501/files#diff-1567700eee9e01c12d5752e9988072a1R7

You might have to add a similar config to all the unit test projects.

It seems strange that it worked without this on 10.0 but not on 9.0 though!

StanleyGoldman:
Nice..

PR #1768, opened 2018-06-29
PR list polish: Move comment count to the right 

donokuda:
_(Targets @grokys's `refactor/pr-list` branch (https://github.com/github/VisualStudio/pull/1737) but wanted to open this separately for discussion)_

The goal of this change is to make the information a little more scannable by having the comment counts of each pull request be vertically aligned. A way to test my attempt is to scan down the list of pull requests, be able to identify which ones have comments, and know which ones have a high level of activity/discussion.

 **Before:**

screen shot 2018-06-29 at 10 09 59 am

 **After:**

screen shot 2018-06-29 at 10 08 48 am

/cc @simurai for 👀 and 💭(if you have any! 😄)

### DIFF
--- a/src/GitHub.VisualStudio/Views/GitHubPane/PullRequestListItemView.xaml
+++ b/src/GitHub.VisualStudio/Views/GitHubPane/PullRequestListItemView.xaml
@@ -77,7 +77,7 @@
                     Opacity="0.5">
                 <Grid.ColumnDefinitions>
                     <ColumnDefinition Width="*" />
-                    <ColumnDefinition Width="Auto" />
+                    <ColumnDefinition MinWidth="35" Width="Auto" />
                 </Grid.ColumnDefinitions>
                 <TextBlock Grid.Column="0">
                     <Run Text="{Binding Number, Mode=OneWay, StringFormat=#{0}}"/>
--- a/src/GitHub.VisualStudio/Views/GitHubPane/PullRequestListItemView.xaml
+++ b/src/GitHub.VisualStudio/Views/GitHubPane/PullRequestListItemView.xaml
@@ -72,21 +72,24 @@
                        Text="{Binding Title}"
                        TextTrimming="CharacterEllipsis"/>
 
-            <StackPanel DockPanel.Dock="Top"
+            <Grid DockPanel.Dock="Top"
                     Margin="0,2,5,0"
-                    Orientation="Horizontal"
                     Opacity="0.5">
-                <TextBlock>
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="*" />
+                    <ColumnDefinition Width="Auto" />
+                </Grid.ColumnDefinitions>
+                <TextBlock Grid.Column="0">
                     <Run Text="{Binding Number, Mode=OneWay, StringFormat=#{0}}"/>
                     <Run Text="{Binding UpdatedAt, Converter={ghfvs:DurationToStringConverter}, Mode=OneWay}"/>
                     by
                     <Run Text="{Binding Author.Login, Mode=OneWay}"/>
                 </TextBlock>
-                <TextBlock Margin="4 0" Visibility="{Binding CommentCount, Converter={ghfvs:CountToVisibilityConverter}}">
-                <ghfvs:OcticonImage Icon="comment" Width="16" Height="16" Margin="0 0 0 -4"/>
-                <Run Text="{Binding CommentCount, Mode=OneWay}" BaselineAlignment="Top"/>
+                <TextBlock Grid.Column="1" Margin="4 0" HorizontalAlignment="Right" Visibility="{Binding CommentCount, Converter={ghfvs:CountToVisibilityConverter}}">
+                    <ghfvs:OcticonImage Icon="comment" Width="16" Height="16" Margin="0 0 0 -4"/>
+                    <Run Text="{Binding CommentCount, Mode=OneWay}" BaselineAlignment="Top" />
                 </TextBlock>
-            </StackPanel>
+            </Grid>
         </DockPanel>
     </Grid>
 </UserControl>
--- a/src/GitHub.VisualStudio/Views/GitHubPane/PullRequestListItemView.xaml
+++ b/src/GitHub.VisualStudio/Views/GitHubPane/PullRequestListItemView.xaml
@@ -58,12 +58,14 @@
                                    Height="30"
                                    VerticalAlignment="Top"/>
 
+            <!--
             <ghfvs:OcticonImage DockPanel.Dock="Right"
                                 Background="Transparent"
                                 Icon="git_pull_request"
                                 ToolTip="This is the current branch"
                                 VerticalAlignment="Stretch"
                                 Visibility="{Binding IsCurrent, Converter={ghfvs:BooleanToVisibilityConverter}}"/>
+                                -->
 
             <TextBlock DockPanel.Dock="Top"
                        Margin="0,-1,5,0"
--- a/src/GitHub.VisualStudio/Views/GitHubPane/PullRequestListItemView.xaml
+++ b/src/GitHub.VisualStudio/Views/GitHubPane/PullRequestListItemView.xaml
@@ -69,9 +69,19 @@
 
             <TextBlock DockPanel.Dock="Top"
                        Margin="0,-1,5,0"
-                       Text="{Binding Title}"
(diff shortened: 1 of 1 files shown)


src/GitHub.VisualStudio/Views/GitHubPane/PullRequestListItemView.xaml
```diff
@@ -58,33 +58,28 @@
                                    Height="30"
                                    VerticalAlignment="Top"/>
 
-            <ghfvs:OcticonImage DockPanel.Dock="Right"
-                                Background="Transparent"
-                                Icon="git_pull_request"
-                                ToolTip="This is the current branch"
-                                VerticalAlignment="Stretch"
-                                Visibility="{Binding IsCurrent, Converter={ghfvs:BooleanToVisibilityConverter}}"/>
```
donokuda: I removed this to keep the number of icons down in the list, but I still think it's useful to call out if you have checked out a branch with a pull request associated with it.

One idea that @grokys and I kicked around is bolding the title. This is similar to a treatment used in the "Connect" page showing which repository that you are currently in:

screen shot 2018-06-29 at 10 44 30 am

simurai:
One concern might be that when the comment counts are on the right, it's a bit harder to see where they belong to? To help that, there could be subtle borders:

with lines

Not super necessary, but something to consider.

sguthals:
I like the subtle border @simurai

donokuda:
Pushed up a couple of changes to try out adding a border and vertically centering the comment count. I think it solves a handful of grouping issues that I was experiencing earlier.

That said, I wonder if having the list item separator feels out of place... Part of that is mostly because I haven't really seen examples of this type of treatment around Visual Studio.

simurai:
>  I wonder if having the list item separator feels out of place...

Hmm.. Is it that in VS borders are more used to separate entire sections and not just "list items"?

Another alternative might be to not have a "column" on the right, but move the comments count next to the PR number. Then you can still scan them by going down a more or less straight line.

screen shot 2018-07-04 at 12 00 21 pm

Also added a bit more space to not have the comment count and month clash with each other.

jcansdale:
Something I'm definitely missing from the new version is clickable hyperlinks to PRs on GitHub:

image

Is this something that could be re-enabled?

jcansdale:
Now this is looking closer to .com, something that is painfully missing is the CS status checks.

image

If all this did was open the same link as it does on .com, having CI status would be so useful! I know this info is available via GraphQL. 😄

grokys:
> Something I'm definitely missing from the new version is clickable hyperlinks to PRs on GitHub

As mentioned on the PR list PR: https://github.com/github/VisualStudio/pull/1737#issuecomment-399184139

> This was by design. The list is now a selectable list, and selectable lists don't generally have inner clickable items. I was thinking of adding a context menu with "Open on GitHub"?

That context menu has been added.

jcansdale:
@grokys sorry I missed your previous comment.

> That context menu has been added.

Oh. I never thought to look for a context menu. I can see this being a discoverability issue.

Could we have a tooltip pointing users to the context menu if they hover over the `#` link? It wouldn't be a clickable item, but it would point them in the right direction.

grokys:
> Could we have a tooltip pointing users to the context menu if they hover over the #nnn label? It wouldn't be a clickable item, but it would point them in the right direction.

Perhaps, yes that might be an option. What would you suggest?

One thing to think about is that you're maybe trained that there's a clickable link there. Usually one is used to trying a right-click in VS to see more options, so it might not be necessary for new users? Of course there will be users though who are used to having a link there.

jcansdale:
@grokys,

> Perhaps, yes that might be an option. What would you suggest?

Maybe something like: `Use context menu to Open on GitHub`. I'm afraid it isn't terribly elegant. 😢 

Re https://github.com/github/VisualStudio/pull/1737#issuecomment-399184139:
> This was by design. The list is now a selectable list, and selectable lists don't generally have inner clickable items. I was thinking of adding a context menu with "Open on GitHub"?

I think my first preference would be to break this convention and allow users to click directly on the `#link`. If we later add support for CI status ✔️ / ❌, we might want to allow users to click directly on the icon (just like they're used to doing on .com).

It isn't like having clickable items on a selectable list is unprecedented. Even on File Explorer there's a checkbox that can be toggled directly without using a context menu.

image

We're in an interesting position with some visual queues and ways of interacting that users will have learnt from the website.

donokuda:
As much as I understand and agree with the concerns around discoverability of the "View On GitHub" action, I'm not convinced that the pull request list is the place where we want to make that action obvious (including using a tooltip). I think a more appropriate place would be in the Pull Request Details view (which we already have in the form of a "View on GitHub" link.

If we find that folks are used to clicking the pull request number in the details list to view the PR on GitHub, then I'd like to understand what drives them to take that action at that moment:

- Is it because they don't know that the pull request detail feature exists (e.g., a different discoverability issue in and of itself)
- Is it because they know the details view exists, but doesn't suit their needs / workflow
- Is it because they know the details view exists, but is hard to use or understand
- _...etc._

I think if we have a better understanding of what drives that behavior, we can improve the extension so that action is not necessary from the pull request list view. (Side note: from my understanding, we link out to the pull request page on GitHub.com because we didn't have a details view implemented when the extension was first shipped. Therefore, viewing on GitHub.com has been more of a stop-gap.)

For a similar reason, I'm also hesitant on making CI statuses actionable from the pull request list. However, I do think it's useful to surface that information in the PR list so users can quickly identify which pull requests have failing build statuses so that they can see what's broken and fix it as quick as possible.

donokuda:
Wanted to show an update with @simurai's suggestions: 

screen shot 2018-07-05 at 12 22 57 pm

This feels a little better than before. Specifically, adding some space between the comment count and timestamp helped makes a more parseable. That said, I think I want to try mocking up a few more ideas for us to review and move forward with.

jcansdale:
@donokuda,

> That said, I wonder if having the list item separator feels out of place... Part of that is mostly because I haven't really seen examples of this type of treatment around Visual Studio.

The `History` in `Simple View` looks somewhat similar to this:

image

I actually liked the scan-ability of the comments when they're lined up on the right.

> If we find that folks are used to clicking the pull request number in the details list to view the PR on GitHub, then I'd like to understand what drives them to take that action at that moment.

I do that mainly because we don't currently support general (non-inline) PR comments. It's the quickest way to see what someone has written. It also means I don't have to wait for the PR to load.

jcansdale:
@donokuda,

> If we're not a huge fan of the smaller avatar treatment, then I think the first one would be a nice iteration over what exists already

I'm leaning towards the first one. I think the smaller avatar is a bit too small.

- There are many avatars that I can't quite work out what they are at the smaller size
- This isn't mitigated by having the user peek view like on dotcom [1]

[1]
image

donokuda:
I ended up circling back to the first iteration to avoid moving around things unnecessarily.

Here's what the list looks like now:

screen shot 2018-07-11 at 2 16 07 pm

I updated the PR so that it targets `master` given that @grokys's PR was merged. This is ready for review.

sguthals:
One more question, which may not be easily done or preferred but I figured I'd throw it out there: 

What about cutting off the title a bit more and putting the comment icon+# in the center of the two lines that we are showing?

meaghanlewis:
This LGTM 🙌 ! 

One thing that I liked was being able to tell which branch you have checked out with an associated PR. I saw you debated the idea of bolding the title a while back, just wondering if this was still on the table.
>I removed this to keep the number of icons down in the list, but I still think it's useful to call out if you have checked out a branch with a pull request associated with it.
One idea that @grokys and I kicked around is bolding the title. This is similar to a treatment used in the "Connect" page showing which repository that you are currently in...

sguthals:
I agree with @meaghanlewis -> that would be cool to see a design of as well

jcansdale:
> One thing that I liked was being able to tell which branch you have checked out with an associated PR. I saw you debated the idea of bolding the title a while back, just wondering if this was still on the table.

Well spotted @meaghanlewis. I would definitely miss this functionality.

donokuda:
> Which is different than Visual Studio. Not saying things have to be aligned, but I like exploring these different types and understanding why we make different choices (e.g. is it because of different workflows for a given editor? Different aesthetics? )

I explored something like that in a [previous comment](https://github.com/github/VisualStudio/pull/1737#issuecomment-399256787) but punted on it to not hold up the PR.  I really liked hoisting the current pull request on top and thought it worked well.

PR #1772, opened 2018-07-10
Fix standalone comment replies

grokys:
#1712 broke a couple of things with inline comments:

- It was using the pull request ID as a review ID, which obviously doesn't work. Added a `CreatePendingReviewCore` method which both `CreatePendingReview` and `PostStandaloneReviewCommentReply` can use, which exposes the info needed by both methods.
- It was breaking out of the loop to create threads after a single reply. D'oh. Not sure how that got there.

## Testing

Previously, leaving standalone comment replies didn't work. Should now work.


meaghanlewis:
This LGTM, thanks for fixing this. ✨

PR #1773, opened 2018-07-10
Show errors in PR list.

grokys:
When an error occurs initializing the PR list (because for example a repository could not be found), handle the error correctly and display an error message. Previously was displaying an empty broken list.

## Testing

Alter the clone URL for a repository to a repository that does not exist. Previously showed a broken PR list, should now show an error.


meaghanlewis:
This is looking good to me. It's nice to have errors instead of an empty broken list, but some of the messages im seeing still are not very descriptive.

I tried a few different scenarios when testing this.

1. No network connectivity
error occurred

2. GitHub repo was deleted
could not resolve repository

3. User does not have access to pull requests (which shows the same error as shown above)

4. I tried to change the clone URL as you suggested, but that always seemed to show me "This repository is not on GitHub" so perhaps I was doing something wrong.
not on github

grokys:
@meaghanlewis yeah, some of those error messages could definitely be improved. I'll merge this PR and take a look at seeing if I can get more information for some of those failures in a separate PR.

PR #1783, opened 2018-07-17
Enable navigation from historical file to solution

jcansdale:
### What this PR does

Extend the `GoToSolutionOrPullRequestFile` command to support navigation from a file opened using the history view to the same file in the users solution/working directory.

### The history of a file

1. Right click on a source file from a repository and `Code context > View History...`
image

2. Double click on a revision in the history
image

3. Right click on the read-only historical file and `Open File in Solution`
image

4. It should navigate to the equivalent position in your solution / working directory

### The commit history

1. Select `View History...` from the status bar
image

2. Double click on a commit in the history
image

3. Double click on a changed file in the commit
image

4. Right click on the commit diff view and `Open File in Solution`
image

4. It should navigate to the equivalent position in your solution / working directory

### DIFF
--- a/src/GitHub.App/Services/GitHubContextService.cs
+++ b/src/GitHub.App/Services/GitHubContextService.cs
@@ -58,6 +58,8 @@ namespace GitHub.Services
         static readonly Regex treeishCommitRegex = new Regex($"(?<commit>[a-z0-9]{{40}})(/(?<tree>.+))?", RegexOptions.Compiled);
         static readonly Regex treeishBranchRegex = new Regex($"(?<branch>master)(/(?<tree>.+))?", RegexOptions.Compiled);
 
+        static readonly Regex tempFileObjectishRegex = new Regex(@"\\TFSTemp\\[^\\]*[.](?<objectish>[a-z0-9]{8})[.][^.\\]*$", RegexOptions.Compiled);
+
         [ImportingConstructor]
         public GitHubContextService(IGitHubServiceProvider serviceProvider, IGitService gitService)
         {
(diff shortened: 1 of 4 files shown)


src/GitHub.App/Services/GitHubContextService.cs
```diff
@@ -305,6 +307,55 @@ public bool TryOpenFile(string repositoryDir, GitHubContext context)
             }
         }
 
+        /// <inheritdoc/>
+        public string FindObjectishForTFSTempFile(string tempFile)
+        {
+            var match = tempFileObjectishRegex.Match(tempFile);
+            if (match.Success)
+            {
```
StanleyGoldman: Why did you name the group `objectish`?

src/GitHub.App/Services/GitHubContextService.cs
```diff
@@ -305,6 +307,55 @@ public bool TryOpenFile(string repositoryDir, GitHubContext context)
...
+                return match.Groups["objectish"].Value;
+            }
+
```
StanleyGoldman: Test single comment

src/GitHub.App/Services/GitHubContextService.cs
StanleyGoldman: Why did you name the group `objectish`?

src/GitHub.App/Services/GitHubContextService.cs
```diff
@@ -305,6 +307,55 @@ public bool TryOpenFile(string repositoryDir, GitHubContext context)
             }
         }
 
+        /// <inheritdoc/>
+        public string FindObjectishForTFSTempFile(string tempFile)
+        {
+            var match = tempFileObjectishRegex.Match(tempFile);
+            if (match.Success)
+            {
+                return match.Groups["objectish"].Value;
```
StanleyGoldman: Why did you name the group `objectish`?

StanleyGoldman: `(first 8 chars of a blob SHA)` ahh

src/GitHub.App/Services/GitHubContextService.cs
```diff
@@ -305,6 +307,55 @@ public bool TryOpenFile(string repositoryDir, GitHubContext context)
...
+                var blob = repo.Lookup<Blob>(objectish);
+                if (blob == null)
+                {
+                    return (null, null);
+                }
+
+                foreach (var commit in repo.Commits)
+                {
+                    var trees = new Stack<Tree>();
+                    trees.Push(commit.Tree);
```
StanleyGoldman: I now know what a Git blob sha is, I've yet to wrap my head around a Commit Tree.

src/GitHub.App/Services/GitHubContextService.cs
jcansdale: It started off being called blobish, but I stole the name from the `Lookup` method you found. 😄

src/GitHub.VisualStudio/Commands/GoToSolutionOrPullRequestFileCommand.cs
```diff
@@ -198,6 +216,81 @@ void OnBeforeQueryStatus(object sender, EventArgs e)
...
+            if (path == null)
+            {
+                return false;
+            }
+
+            var objectish = gitHubContextService.Value.FindObjectishForTFSTempFile(path);
+            if (objectish == null)
+            {
+                return false;
+            }
```
StanleyGoldman: The first three if statements from this function and the function above are meant to be the same.
The only minor minor difference _(took me a moment to spot)_ is you are saving the repository dir in the second implementation.

The intent might be clearer if you combine both to one function that returns `repositoryDir` if the conditions are right or `null` if they fail.

StanleyGoldman: `GetRepositoryDirectoryForSourceViewAndPath` or something..

src/GitHub.App/Services/GitHubContextService.cs
StanleyGoldman: I have to read the tests more

StanleyGoldman:
Everything works here. Just one thing that can be clarified i spotted.
I still feel like I need to read the tests more to understand what a Commit Tree is.

PR #1787, opened 2018-07-19
Refactored Clone Dialog

### LINKED ISSUE #1757: Using a slash in the clone dialog
Rather than typing `nunit`, reaching for the mouse and clicking to expand the org 'nunit,' then finding nunit in the alphabetically sorted list and clicking on it, my instinct was to type a slash like I would at the command line or in the browser:

Do you already support something similar, or would you be interested in such a feature?

grokys:
Refactored the clone dialog, basing it initially on the clone dialog from desktop:

## Features

- Now uses GraphQL to load the list of repositories. This improves the load time of the list.
- Separate GitHub and Enterprise tabs
- A new "Clone from URL" tab

## Screenshots

Particularly the URL clone view needs some input from @donokuda.

image

image

# Scopes

GraphQL requires the `read:org` scope the access the user's repositories. This means that the user will have to log out and back in again to use the clone dialog. The following message will be shown in this case (needs input from @donokuda):

image

# Caveats

- ~~You currently need to log out and back in because the GraphQL query needs the `read:org` scope which your previous login won't have. Need to implement a "You need to log out and back in again" flow in the dialog for this.~~ This is now implemented
- ~~Only showing first 100 org repositories because of octokit/octokit.graphql.net#132~~ now fixed.

Depends on #1761 

Closes #1757


donokuda:
> You currently need to log out and back in because the GraphQL query needs the read:org scope which your previous login won't have. Need to implement a "You need to log out and back in again" flow in the dialog for this.

In case it hasn't been considered yet, a potentially "quick & dirty" flow for this is to pop up a [modal dialog](https://docs.microsoft.com/en-us/visualstudio/extensibility/ux-guidelines/notifications-and-progress-for-visual-studio#BKMK_ModalErrorMessageDialogs) on top of the clone dialog that give the individual the option to log out.

I'd imagine the flow to break down like this:

1. I click on "Clone" from the GitHub section of Team Explorer's connect page
2. A dialog appears explaining that I need to log out in order to proceed with two options: [Cancel] [Log out]
3. I click "Log out," the extension attempts to log me out.
- On success ✅: the login modal appears and I step through that flow. (Ideally: After logging in, I'm presented with the clone dialog since that's the action I've been meaning to do in the first place.) 
- On failure ❌: Another dialog appears with the error and two options: [Cancel] [Try again]
  - Selecting "Try again" attempts logging out again.

meaghanlewis:
This looks great to me @grokys !

One thing that I ran into was when I wanted to enter the username and repo to clone - ex `github/quality`. I wasn't signed into my GitHub account at the time - only GitHub Enterprise. This login dialog appeared:

screen shot 2018-09-06 at 2 29 11 pm

I authenticated with it, but it didn't actually sign me in and I got an error like this `Error encountered while cloning the remote repository: Git failed with a fatal error.
unable to access 'https://github.com/github/quality/': The requested URL returned error: 403`. 

Is it possible to show our Sign in dialog here instead to avoid this from possibly confusing users?

grokys:
@meaghanlewis hmm yes, I'm not quite sure how to deal with that. If the repository is publicly visible then this clone should actually work. We _could_ make a check to see if we have a connection to the server that the clone is being requested from, but again, the clone may actually work.

I notice desktop does the same thing in this situation and shows a login dialog. I'm tempted to say that this is expected behavior.

meaghanlewis:
hey @grokys im okay to leave the behavior the same when trying to clone a private repo while signed out of a GitHub account. This might just be an edge case and there is a workaround anyway to sign into the account first and then clone.

meaghanlewis:
One more question - could you add a message for when someone tries to clone a repo they already have?

Currently, the clone button is disabled when trying to clone an existing repo but it might not be clear why.
screen shot 2018-09-10 at 8 01 53 am

In the old (current) clone dialog we at least showed an icon in the path to indicate it can't be cloned to that location.
screen shot 2018-09-10 at 8 10 06 am

And in desktop, they have an explicit message saying `the destination already exists.` Could we also show a message similar to this?
screen shot 2018-09-10 at 8 01 39 am

Issue #1784, opened 2018-07-17
Unable to delete inline comment


meaghanlewis:
__Version__
- GitHub Extension for Visual Studio version: [2.5.4.3295](https://ci.appveyor.com/project/github-windows/visualstudio/build/2.5.4.3295)

__What happened__ 
When I try to delete an inline comment I see the error `Specified cast is not valid.`
<img width="227" alt="screen shot 2018-07-17 at 10 12 54 am" src="https://user-images.githubusercontent.com/10103121/42833878-496d6930-89aa-11e8-885b-b9d69a13be65.png">

Log snippet:
`2018-07-17 10:12:52.418 [05908] EROR [01] CommentViewModel          Error Deleting comment
System.InvalidCastException: Specified cast is not valid.
   at GitHub.InlineReviews.ViewModels.InlineCommentThreadViewModel.<DoDeleteComment>d__6.MoveNext() in C:\projects\visualstudio\src\GitHub.InlineReviews\ViewModels\InlineCommentThreadViewModel.cs:line 83
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at GitHub.InlineReviews.ViewModels.CommentViewModel.<DoDelete>d__12.MoveNext() in C:\projects\visualstudio\src\GitHub.InlineReviews\ViewModels\CommentViewModel.cs:line 129`

PR #1789, opened 2018-07-20
Fixing functionality to delete a pull request comment

StanleyGoldman:
Fixes: #1784 

#### First Issue:
- **InvalidCastException** in `InlineCommentThreadViewModel.DoDeleteComment(...)`
  https://github.com/github/VisualStudio/blob/49ad951549e3362db0ecfff561c6f8360cc25d41/src/GitHub.InlineReviews/ViewModels/InlineCommentThreadViewModel.cs#L83-L84
- Meanwhile in `CommentViewModel.DoDelete(...)`
   https://github.com/github/VisualStudio/blob/49ad951549e3362db0ecfff561c6f8360cc25d41/src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs#L129-L130
  `Id` which is a `string` value is being provided to the delete comment operation. There is no way to delete a pull request comment by GraphQL ID. We need to use the pull request comment database id.

#### Second Issue:
 - `PullRequestSessionService.DeleteComment(...)`
    https://github.com/github/VisualStudio/blob/49ad951549e3362db0ecfff561c6f8360cc25d41/src/GitHub.InlineReviews/Services/PullRequestSessionService.cs#L650-L664
  1. The variable `number`, which refers to the pull request comment number. Is used to delete the pull request comment by database Id, then is used in an attempt to re-load the current pull request.
  1. In order to call `ReadPullRequestDetail` we need the pull request id.

#### Fixes
1. Added `DatabaseId` and `PullRequestId` to `CommentModel`  and made the changes in `PullRequestSessionService` to populate them.
1. Modified `PullRequestSessionService` to require comment database id and pull request id in order to delete a pull request comment.
1. Modified `InlineCommentThreadViewModel.DoDeleteComment(...)` and `CommentViewModel.DoDelete(...)` to provide an expect the correct values.

**Note**:
I'm noticing a different bug when a user now deletes the last comment in a list. I will log that bug after this.

#### Additional Fixes
- Added functionality to `InlineCommentPeekViewModel` to listen to the `CancelEdit` command of every assigned `NewInlineCommentThreadViewModel`
- Added functionality to the `InlineCommentPeekService` to dismiss the session when that event is fired

### DIFF
--- a/src/GitHub.Exports/Models/CommentModel.cs
+++ b/src/GitHub.Exports/Models/CommentModel.cs
@@ -11,6 +11,16 @@ namespace GitHub.Models
         /// Gets the ID of the comment.
         /// </summary>
         public string Id { get; set; }
+        
+        /// <summary>
+        /// Gets the DatabaseId of the comment.
+        /// </summary>
+        public int DatabaseId { get; set; }
+
+        /// <summary>
+        /// Gets the PullRequestId of the comment
+        /// </summary>
+        public int PullRequestId { get; set; }
 
         /// <summary>
         /// Gets the author of the comment.
--- a/src/GitHub.Exports/Models/CommentModel.cs
+++ b/src/GitHub.Exports/Models/CommentModel.cs
@@ -18,9 +18,12 @@ namespace GitHub.Models
         public int DatabaseId { get; set; }
 
         /// <summary>
-        /// Gets the PullRequestId of the comment
+        /// Gets the PullRequestId of the comment.
         /// </summary>
         public int PullRequestId { get; set; }
+        // The GraphQL Api does not allow for deleting of pull request comments.
+        // REST Api must be used, and PullRequestId is needed to reload the pull request.
+        // This field should be removed with better GraphQL support.
 
         /// <summary>
         /// Gets the author of the comment.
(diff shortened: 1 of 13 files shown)


src/GitHub.Exports/Models/CommentModel.cs
```diff
@@ -11,6 +11,16 @@ public class CommentModel
...
+        
+        /// <summary>
+        /// Gets the DatabaseId of the comment.
+        /// </summary>
+        public int DatabaseId { get; set; }
+
+        /// <summary>
+        /// Gets the PullRequestId of the comment
+        /// </summary>
+        public int PullRequestId { get; set; }
```
grokys: Yeah, it's not ideal having this here, but because the REST API requires it, there's not really a better way. Hopefully we'll be able to delete comments using GQL soon!

grokys: Might want to add a comment saying that this is just needed for deleting comments via the REST API and should be temporary?

StanleyGoldman: Updated

meaghanlewis:
This LGTM. Thanks @StanleyGoldman ❤️

PR #1793, opened 2018-07-20
Polish check suite results in pull request list

donokuda:
image


donokuda:
Updated the original post with screenshots. This is ready for review! 😄

StanleyGoldman:
Looks great

PR #1798, opened 2018-07-24
[WIP] Functionality to support Check Suites API

StanleyGoldman:


src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -96,6 +96,12 @@ public class PullRequestService : IPullRequestService
                             LastCommit = pr.Commits(null, null, 1, null).Nodes.Select(commit =>
                                 new LastCommitSummaryModel
                                 {
+//                                    CheckSuites = commit.Commit.CheckSuites(null, null, null, null, null).AllPages(10)
```
StanleyGoldman: asdf

StanleyGoldman: asdf

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -136,36 +144,63 @@ public class PullRequestService : IPullRequestService
...
             var result = await graphql.Run(readPullRequests, vars);
 
-            foreach (var item in result.Items.Cast<ListItemAdapter>())
+            foreach (ListItemAdapter item in result.Items)
             {
                 item.CommentCount += item.Reviews.Sum(x => x.Count);
                 item.Reviews = null;
 
-                var hasStatuses = item.LastCommit.Statuses != null 
+                var hasCheckSuites = item.LastCommit.CheckSuites != null
```
grokys: Nit: could these be written more simply?

```
var hasCheckSuites = item.LastCommit.CheckSuites?.Any() ?? false;
```

src/GitHub.App/ViewModels/GitHubPane/PullRequestCheckViewModel.cs
```diff
@@ -53,8 +55,55 @@ public static IEnumerable<IPullRequestCheckViewModel> Build(IViewViewModelFactor
...
+                    pullRequestCheckViewModel.CheckType = PullRequestCheckType.ChecksApi;
+                    pullRequestCheckViewModel.Title = model.Name;
+                    pullRequestCheckViewModel.Description = model.Summary;
+                    pullRequestCheckViewModel.Status = checkStatus;
+                    pullRequestCheckViewModel.DetailsUrl = new Uri(model.DetailsUrl);
+
+                    return pullRequestCheckViewModel;
+                }) ?? new PullRequestCheckViewModel[0];
+
+            return statuses.Union(checks).OrderBy(model => model.Title);
```
grokys: Is `.Union` the best operator to use here? It will do equality checks for each item, even though the items will never be equal. Not a big deal but I would have said that `.Concat` would be a better choice here.

src/GitHub.Exports/Models/AnnotationModel.cs
```diff
@@ -0,0 +1,48 @@
+﻿namespace GitHub.Models
+{
+    /// <summary>
+    /// Model for a single check annotation.
+    /// </summary>
+    public class CheckRunAnnotationModel
```
grokys: Nicely documented ;)

src/GitHub.Exports/Models/PullRequestDetailModel.cs
```diff
@@ -93,7 +93,12 @@ public class PullRequestDetailModel
         public IReadOnlyList<PullRequestReviewThreadModel> Threads { get; set; }
 
         /// <summary>
-        /// Gets or sets a collection of pull request Checks & Statuses
+        /// Gets or sets a collection of pull request Checks Suites
+        /// </summary>
+        public List<CheckSuiteModel> CheckSuites { get; set; }
```
grokys: `IReadOnlyList` is used for the lists elsewhere in this model (and oops I didn't notice this with `Statuses`).

src/GitHub.App/ViewModels/GitHubPane/PullRequestCheckViewModel.cs
StanleyGoldman: Oh. I never knew that...

src/GitHub.Exports/Models/AnnotationModel.cs
StanleyGoldman: :+1:

meaghanlewis:
This LGTM! Thanks for your work on this @StanleyGoldman ✨

PR #1805, opened 2018-07-25
Fixing incorrect casing in the clone url owner issues

### LINKED ISSUE #1804: Incorrect casing in the clone url owner causes issues
#### Issue
If you clone a repo using incorrect casing in the owner. Some functionalities just don't work.
- Visual Studio is `https://github.com/github/VisualStudio.git`
- If you instead clone it with `https://github.com/GitHub/VisualStudio.git`

##### The following functionalities are broken
- Any checked out branch will not be correlated to the corresponding pull request.
  https://github.com/github/VisualStudio/blob/4af9d1ca0071f48782412eae5dad82d4f0078702/src/GitHub.App/Services/PullRequestService.cs#L527
- Checking out an internal pull request will result in the branch path resembling that of a fork pull request checkout
  https://github.com/github/VisualStudio/blob/4af9d1ca0071f48782412eae5dad82d4f0078702/src/GitHub.App/Services/PullRequestService.cs#L395
- Multiple pull request sessions can exist
  https://github.com/github/VisualStudio/blob/4af9d1ca0071f48782412eae5dad82d4f0078702/src/GitHub.InlineReviews/Services/PullRequestSessionManager.cs#L228
  - The `OpenFileInWorkingDirectory` functionality does not work.

##### Other suspicious code paths
- https://github.com/github/VisualStudio/blob/4af9d1ca0071f48782412eae5dad82d4f0078702/src/GitHub.VisualStudio.UI/UI/Views/GitHubConnectContent.xaml.cs#L78

<hr>

#### Original Issue

I noticed this while working with another project...

https://github.com/github/VisualStudio/blob/4af9d1ca0071f48782412eae5dad82d4f0078702/src/GitHub.App/Services/PullRequestService.cs#L525-L528

`pullRequest.HeadRepositoryOwner` is "justaprogrammer"
`repository.CloneUrl.Owner` is "JustAProgrammer"

image

StanleyGoldman:
Fixes #1804 

1. Using a case-insensitive comparison when..
   1. attempting to determine if (the current displayed pull request == the current branch)
   1. checking out a pull request
1. Invariant lowercasing the owner name when..
   1. Finding/Creating a `PullRequestSession`

### DIFF
--- a/src/GitHub.App/Services/PullRequestService.cs
+++ b/src/GitHub.App/Services/PullRequestService.cs
@@ -524,7 +524,7 @@ namespace GitHub.Services
 
         public bool IsPullRequestFromRepository(ILocalRepositoryModel repository, PullRequestDetailModel pullRequest)
         {
-            return pullRequest.HeadRepositoryOwner == repository.CloneUrl.Owner;
+            return pullRequest.HeadRepositoryOwner.ToLowerInvariant() == repository.CloneUrl.Owner.ToLowerInvariant();
         }
 
         public IObservable<Unit> SwitchToBranch(ILocalRepositoryModel repository, PullRequestDetailModel pullRequest)
--- a/src/GitHub.App/Services/PullRequestService.cs
+++ b/src/GitHub.App/Services/PullRequestService.cs
@@ -524,7 +524,7 @@ namespace GitHub.Services
 
         public bool IsPullRequestFromRepository(ILocalRepositoryModel repository, PullRequestDetailModel pullRequest)
         {
-            return pullRequest.HeadRepositoryOwner.ToLowerInvariant() == repository.CloneUrl.Owner.ToLowerInvariant();
+            return string.Equals(repository?.CloneUrl.Owner, pullRequest.HeadRepositoryOwner, StringComparison.OrdinalIgnoreCase);
         }
 
         public IObservable<Unit> SwitchToBranch(ILocalRepositoryModel repository, PullRequestDetailModel pullRequest)
--- a/src/GitHub.App/Services/PullRequestService.cs
+++ b/src/GitHub.App/Services/PullRequestService.cs
@@ -392,7 +392,7 @@ namespace GitHub.Services
                     {
                         await gitClient.Checkout(repo, localBranchName);
                     }
-                    else if (repository.CloneUrl.Owner == pullRequest.HeadRepositoryOwner)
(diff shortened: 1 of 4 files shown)


src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -524,7 +524,7 @@ public IObservable<bool> EnsureLocalBranchesAreMarkedAsPullRequests(ILocalReposi
 
         public bool IsPullRequestFromRepository(ILocalRepositoryModel repository, PullRequestDetailModel pullRequest)
         {
-            return pullRequest.HeadRepositoryOwner == repository.CloneUrl.Owner;
+            return pullRequest.HeadRepositoryOwner.ToLowerInvariant() == repository.CloneUrl.Owner.ToLowerInvariant();
```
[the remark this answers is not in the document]
StanleyGoldman: No problem

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -524,7 +524,7 @@ public IObservable<bool> EnsureLocalBranchesAreMarkedAsPullRequests(ILocalReposi
 
         public bool IsPullRequestFromRepository(ILocalRepositoryModel repository, PullRequestDetailModel pullRequest)
         {
-            return pullRequest.HeadRepositoryOwner == repository.CloneUrl.Owner;
+            return string.Equals(repository?.CloneUrl.Owner, pullRequest.HeadRepositoryOwner, StringComparison.OrdinalIgnoreCase);
```
StanleyGoldman: @jcansdale 

> It's possible that repository.CloneUrl.Owner will be null if ILocalRepositoryModel doesn't have a remote.

I put the `?` here because of what you said, but is that really possible?

meaghanlewis:
Hey @StanleyGoldman im happy with this. I tried to find other scenarios where casing of the repository owner could mess things up, but haven't encountered any other negative experiences.

src/GitHub.App/Services/PullRequestService.cs
grokys: It's definitely possible that `CloneUrl` is null but you put the `?` after `repository`. I don't think passing a null `repository` to this method should be allowed, so we probably need a `Guard.ArgumentNotNull(repository, nameof(repository));` at the top of the method instead of the `?`.

Issue #1713, opened 2018-05-31
Deleting an inline comment should show a confirmation dialog


StanleyGoldman:
After discussion with @meaghanlewis in https://github.com/github/VisualStudio/pull/1701 we realized it would be best if deleting of a comment raised a notification as that is what happens on .com.

Issue #1797, opened 2018-07-24
Delete comment icon is very close to close window icon


jcansdale:
- GitHub Extension for Visual Studio version: 2.5.4

The ✖️ icon to delete an inline-comment is very close to the icon used to close the inline-comment window and also appears in the top right hand corner of a box.
![image](https://user-images.githubusercontent.com/11719160/43134910-85f7be28-8f3b-11e8-9c85-79b664d9478d.png)

If a user accidentally clicks on the delete icon, there is no confirmation to correct their mistake.

Maybe we could change it to be a `...` drop down menu with a `Delete` option like .com relatively recently changed to.

![image](https://user-images.githubusercontent.com/11719160/43135252-a91cbe66-8f3c-11e8-9ebd-1e917a55ea52.png)

StanleyGoldman:
Related to #1713

donokuda:
I'm going to take a quick stab at this and https://github.com/github/VisualStudio/issues/1713.

I'll focus on two things:

1. Increasing the spacing between the ✏️ and the ❌ octicon
2. Popup a dialog when a user clicks on the ❌ asking them to confirm if they want to delete the comment. (and clicking "Delete" should delete the comment.)

meaghanlewis:
I think we can close this out now. Please re-open if I'm incorrect 😄

PR #1822, opened 2018-08-01
Show a confirmation when attempting to delete a comment

donokuda:
Fixes https://github.com/github/VisualStudio/issues/1713 and https://github.com/github/VisualStudio/issues/1797

This pull request introduces a confirmation modal when a user tries to delete a comment. It also increases the spacing between the ✏️ and ❌ octicon to decrease the chance of clicking on the wrong icon:

delete-comment

A couple of things that I noticed while working on this:

- I did not provide any designs for an activity indicator. When I was testing the delete action, there wasn't anything to signal that deleting was in progress. I'll open up an issue to track this.
- If deleting a comment was successful, then it would be great if there was a way to update the UI to reflect this. Currently, the deleted comment persists, which makes it difficult to understand if deleting was successful or not. (in the gif above, I toggle the comment thread to check) This feels related to the first observation above and I will track in the same issue.
- ~The hover styles for the octicons need to be tweaked. I'll work on that in a separate branch.~ Done in 53cee1b

### DIFF
--- a/src/GitHub.InlineReviews/Services/ICommentService.cs
+++ b/src/GitHub.InlineReviews/Services/ICommentService.cs
@@ -1,7 +1,14 @@
 ﻿namespace GitHub.InlineReviews.Services
 {
+    /// <summary>
+    /// This service allows for functionality to be injected into the chain of different peek Comment ViewModel types.
+    /// </summary>
     public interface ICommentService
     {
+        /// <summary>
+        /// This function uses MessageBox.Show to display a confirmation if a comment should be deleted.
+        /// </summary>
+        /// <returns></returns>
         bool ConfirmCommentDelete();
     }
 }
--- a/src/GitHub.InlineReviews/Services/ICommentService.cs
+++ b/src/GitHub.InlineReviews/Services/ICommentService.cs
@@ -1,0 +1,7 @@
+﻿namespace GitHub.InlineReviews.Services
+{
+    public interface ICommentService
+    {
+        bool ConfirmCommentDelete();
+    }
+}
--- a/src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
+++ b/src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
@@ -131,13 +131,11 @@ namespace GitHub.InlineReviews.ViewModels
 
         async Task DoDelete(object unused)
         {
-            string deleteCommentMessageBoxText = VisualStudio.UI.Resources.DeleteCommentConfirmation;
-            string deleteCommentCaption = VisualStudio.UI.Resources.DeleteCommentConfirmationCaption;
-            
-            MessageBoxButton deleteCommentButton = MessageBoxButton.YesNo;
-            MessageBoxImage deleteCommentImage = MessageBoxImage.Question;
-
-            MessageBoxResult result = MessageBox.Show(deleteCommentMessageBoxText, deleteCommentCaption, deleteCommentButton, deleteCommentImage);
+            var result = MessageBox.Show(
+                VisualStudio.UI.Resources.DeleteCommentConfirmation, 
+                VisualStudio.UI.Resources.DeleteCommentConfirmationCaption, 
+                MessageBoxButton.YesNo,
+                MessageBoxImage.Question);
 
             switch(result)
             {
--- a/src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
+++ b/src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
@@ -5,6 +5,7 @@ using System.Reactive.Linq;
 using System.Threading.Tasks;
 using System.Windows;
 using GitHub.Extensions;
+using GitHub.InlineReviews.Services;
 using GitHub.Logging;
 using GitHub.Models;
 using GitHub.ViewModels;
@@ -19,6 +20,7 @@ namespace GitHub.InlineReviews.ViewModels
     public class CommentViewModel : ReactiveObject, ICommentViewModel
     {
         static readonly ILogger log = LogManager.ForContext<CommentViewModel>();
+        ICommentService commentService;
         string body;
         string errorMessage;
         bool isReadOnly;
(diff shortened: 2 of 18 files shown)


donokuda:
Pushed 53cee1b which updates the hover + IsPressed styles to use colors provided by VS (for easy theming):

**Blue Theme**
hover-blue

**Dark Theme**
hover-dark

**Light Theme**
hover-light

src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
```diff
@@ -130,22 +131,41 @@ protected void AddErrorHandler<T>(ReactiveCommand<T> command)
...
-                await Thread.DeleteComment.ExecuteAsyncTask(new Tuple<int, int>(PullRequestId, DatabaseId));
-            }
-            catch (Exception e)
-            {
-                var message = e.Message;
-                ErrorMessage = message;
-                log.Error(e, "Error Deleting comment");
-            }
-            finally
+            MessageBoxResult result = MessageBox.Show(deleteCommentMessageBoxText, deleteCommentCaption, deleteCommentButton, deleteCommentImage);
```
StanleyGoldman: I was searching for something like this in the codebase that I could use as an example

StanleyGoldman: Adding another comment.

StanleyGoldman:
I moved strings to resources as well, I'll let someone else merge this.

src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
grokys: Yeah, there's also `VsShellUtilities.ShowMessageBox` - I'm not sure if we should be using that one or this one? @AArnott would you know?

src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
```diff
@@ -130,22 +131,42 @@ protected void AddErrorHandler<T>(ReactiveCommand<T> command)
...
         {
-            try
-            {
-                ErrorMessage = null;
-                IsSubmitting = true;
+            string deleteCommentMessageBoxText = VisualStudio.UI.Resources.DeleteCommentConfirmation;
+            string deleteCommentCaption = VisualStudio.UI.Resources.DeleteCommentConfirmationCaption;
+            
+            MessageBoxButton deleteCommentButton = MessageBoxButton.YesNo;
+            MessageBoxImage deleteCommentImage = MessageBoxImage.Question;
```
grokys: @donokuda interested to know why you create variables for these things? Wouldn't it be easier just to do:

```
var result = MessageBox.Show(
    VisualStudio.UI.Resources.DeleteCommentConfirmation,
    VisualStudio.UI.Resources.DeleteCommentConfirmationCaption,
    MessageBoxButton.YesNo,
    MessageBoxImage.Question);
```

grokys:
> If deleting a comment was successful, then it would be great if there was a way to update the UI to reflect this.

That should definitely be happening! It happens here, though after a delay. Is it the delay that's confusing?

src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
StanleyGoldman: Took care of this one.

donokuda:
> That should definitely be happening! It happens in my testing, though after a delay. Is it the delay that's confusing?

I'm not sure because the UI doesn't update for me even if I wait for a while:

deleting-delay

src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs
```diff
@@ -130,22 +131,31 @@ protected void AddErrorHandler<T>(ReactiveCommand<T> command)
 
         async Task DoDelete(object unused)
         {
-            try
-            {
-                ErrorMessage = null;
-                IsSubmitting = true;
+            var result = MessageBox.Show(
```
grokys: We shouldn't be showing a message box from the view model because this will make message boxes pop up during unit testing, breaking things. This will need to be extracted out to a service.

grokys: I'm also not sure if we should be using `VsShellUtilities.ShowMessageBox` instead of `MessageBox.Show` in VS? @AArnott any guidance?

AArnott: `VsShellUtilities.ShowMessageBox` just calls into [`IVsUIShell::ShowMessageBox`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivsuishell.showmessagebox?view=visualstudiosdk-2017) I'm pretty sure. You're welcome to call that interface directly and mock it up in your unit tests, or to use the `VsShellUtilities` method if you're already mocking up the global service provider.

grokys: Thanks @AArnott - that's definitely a good option. The question was actually regarding `MessageBox.Show` vs the VS-specific message-box methods, but given that we can pass in a `IVsUIShell` instance I'm not sure that's relevant any more ;)

test/GitHub.InlineReviews.UnitTests/ViewModels/InlineCommentThreadViewModelTests.cs
```diff
@@ -15,7 +16,8 @@ public class InlineCommentThreadViewModelTests
         public void CreatesComments()
         {
             var target = new InlineCommentThreadViewModel(
-                CreateSession(),
+                Substitute.For<ICommentService>()
+                , CreateSession(),
```
grokys: Nit: this comment is in a weird position.

src/GitHub.InlineReviews/Services/ICommentService.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace GitHub.InlineReviews.Services
+{
+    public interface ICommentService
```
grokys: Would be good to get some XML docs on this service to indicate where/for what it's used.

test/GitHub.InlineReviews.UnitTests/ViewModels/InlineCommentThreadViewModelTests.cs
StanleyGoldman: comma you mean 😼

Issue #1806, opened 2018-07-25
"Start a Review" button gets enabled when it shouldn't


StanleyGoldman:
If I'm replying to a review comment. My only available option should be to Comment in return.

![image](https://user-images.githubusercontent.com/417571/43228668-359ecdb8-9030-11e8-8bd1-a9e121821281.png)

But when replying to a comment in GitHub for Visual Studio, the buttons are different and "Start a Review" is also enabled when it shouldn't be.

![2018-07-25_17-29-20](https://user-images.githubusercontent.com/417571/43228769-78d93aa0-9030-11e8-8a0b-bd933d201da3.gif)

StanleyGoldman:
Turns out that I compared things to the comment view on github.com and not the diff view.

PR #1827, opened 2018-08-03
Fix when "Start a Review" button is enabled

StanleyGoldman:
Fixes: #1806 

Users cannot start a review from a comment if there are already comments in the thread.

- Correct the observable that is referenced to enable the `StartReview` command.
   https://github.com/github/VisualStudio/blob/8c9f77afaafd26d4bd2a641b46fff3b2c4b9958f/src/GitHub.InlineReviews/ViewModels/PullRequestReviewCommentViewModel.cs#L83-L85

` Added a count of comments in thread of `None` state into the criteria for `CanStartReview`
  - And used that information in determining if a review can be started.
   https://github.com/github/VisualStudio/blob/956d9d6f900e7f4831f88218887c0df3be79a816/src/GitHub.InlineReviews/ViewModels/PullRequestReviewCommentViewModel.cs#L61-L68

- Changed or removed tests that depending on being able to start a review after comments were made

### DIFF
--- a/test/GitHub.InlineReviews.UnitTests/ViewModels/PullRequestReviewCommentViewModelTests.cs
+++ b/test/GitHub.InlineReviews.UnitTests/ViewModels/PullRequestReviewCommentViewModelTests.cs
@@ -191,12 +191,12 @@ namespace GitHub.InlineReviews.UnitTests.ViewModels
             }
 
             [Test]
-            public void IsDisabledWhenSessionHasNoPendingReview()
+            public void IsEnabledWhenSessionHasNoPendingReview()
             {
                 var session = CreateSession(false);
                 var target = CreateTarget(session);
 
-                Assert.That(target.StartReview.CanExecute(null), Is.False);
+                Assert.That(target.StartReview.CanExecute(null), Is.True);
             }
 
             [Test]
--- a/test/GitHub.InlineReviews.UnitTests/ViewModels/PullRequestReviewCommentViewModelTests.cs
+++ b/test/GitHub.InlineReviews.UnitTests/ViewModels/PullRequestReviewCommentViewModelTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.ObjectModel;
 using System.Reactive.Linq;
 using System.Threading.Tasks;
 using GitHub.InlineReviews.ViewModels;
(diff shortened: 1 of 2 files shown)


StanleyGoldman:
@grokys I could use your help understanding what some of these tests are intending to test and why some of them fail.

test/GitHub.InlineReviews.UnitTests/ViewModels/PullRequestReviewCommentViewModelTests.cs
```diff
@@ -172,12 +191,12 @@ public void IsDisabledWhenSessionHasPendingReview()
...
 
             [Test]
-            public void IsDisabledWhenSessionHasNoPendingReview()
+            public void IsEnabledWhenSessionHasNoPendingReview()
             {
                 var session = CreateSession(false);
                 var target = CreateTarget(session);
 
-                Assert.That(target.StartReview.CanExecute(null), Is.False);
+                Assert.That(target.StartReview.CanExecute(null), Is.True);
```
grokys: I don't think this is correct. This was asserting that you can't start a review when `Body` is empty. With this change the "Start review" button is enabled with no body:

image

The test should probably have been named `IsDisabledWhenSessionHasNoPendingReviewAndEmptyBody` to match the following `IsEnabledWhenSessionHasNoPendingReviewAndBodyNotEmpty`.

grokys:
Hmm, actually is this correct?

image

The "Start a review" button appears for me for replies.

test/GitHub.InlineReviews.UnitTests/ViewModels/PullRequestReviewCommentViewModelTests.cs
StanleyGoldman: Okay cool, those tests make more sense now.

StanleyGoldman:
There are two more tests that are failing for me that I don't immediately follow...

https://github.com/github/VisualStudio/blob/f0de5e86188d1694b4464d221616fdea437ef8e8/test/GitHub.InlineReviews.UnitTests/ViewModels/InlineCommentPeekViewModelTests.cs#L202-L236

https://github.com/github/VisualStudio/blob/f0de5e86188d1694b4464d221616fdea437ef8e8/test/GitHub.InlineReviews.UnitTests/ViewModels/InlineCommentPeekViewModelTests.cs#L238-L272

StanleyGoldman:
I was just massively confused

PR #1834, opened 2018-08-07
Only update PR list state on main thread.

### LINKED ISSUE #1828: Searching for Pull Request too soon can freeze Visual Studio
1. Goto pull request view
1. Switch to display Closed Issues
1. Quickly type in an issue number and push enter before list loads

grokys:
`IVirtualizingListSource.IsLoading` can be signalled on a background thread. This was causing a deadlock in #1828. Make sure we call `IssueListViewModelBase.UpdateState` on the UI thread.

Fixes #1828


StanleyGoldman:
I was not able to reproduce the error here...

StanleyGoldman:
The only decent thing to do at this point is review this pull request

PR #1838, opened 2018-08-07
Adding functionality to confirm before cancelling a review

### LINKED ISSUE #1653: Pull Request summary Cancel button needs confirmation
The Cancel button noted here will remove all of a user's in progress review comments.
The user may not be aware of the effect and may simply be trying to leave the screen because they are not ready to complete their review.

The effect of the cancel button should be clearly explained with a dialog/confirmation before allowing the user to continue.

image

StanleyGoldman:
Fixes #1653

Adding a confirmation to cancel pending reviews as well as a test around it.

### DIFF
--- a/src/GitHub.App/Services/PullRequestEditorService.cs
+++ b/src/GitHub.App/Services/PullRequestEditorService.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using System.Reactive.Linq;
 using System.Threading.Tasks;
+using System.Windows.Forms;
 using GitHub.Commands;
 using GitHub.Extensions;
 using GitHub.Models;
@@ -286,6 +287,15 @@ namespace GitHub.Services
             dte.ItemOperations.OpenFile(fullPath, EnvDTE.Constants.vsViewKindCode);
             var codeView = FindActiveView();
             NavigateToEquivalentPosition(sourceView, codeView);
+        }
+
+        public bool ConfirmCancelPendingReview()
+        {
+            return MessageBox.Show(
+                       GitHub.App.Resources.CancelPendingReviewConfirmation,
+                       GitHub.App.Resources.CancelPendingReviewConfirmationCaption,
+                       MessageBoxButtons.YesNo,
+                       MessageBoxIcon.Question) == DialogResult.Yes;
         }
 
         public bool IsEditableDiff(ITextView textView)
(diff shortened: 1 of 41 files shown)


StanleyGoldman:
Shouldn't we stay on the pull request view if the confirmation dialog is cancelled?
I have a branch where I do that, I'm not sure how we feel about the functionality change...

https://github.com/github/VisualStudio/tree/pull-request-dont-close-after-cancel-confirm

@donokuda any feelings?

grokys:
Yeah it does feel to me like nothing should happen if I click "No" in the dialog box, rather than navigating back to the PR.

StanleyGoldman:
I merged that functionality in.

meaghanlewis:
Pressing `Yes` to confirm cancelling a review freezes and restarts Visual Studio for me.

cancel review freezes vs

StanleyGoldman:
Sigh..

src/GitHub.InlineReviews/Services/PullRequestSession.cs
```diff
@@ -223,20 +223,27 @@ public async Task StartReview()
...
-        public async Task CancelReview()
+        public async Task<bool> CancelReview()
         {
             if (!HasPendingReview)
             {
                 throw new InvalidOperationException("There is no pending review to cancel.");
             }
 
-            await service.CancelPendingReview(LocalRepository, PendingReviewId);
+            if (service.ConfirmCancelPendingReview())
```
grokys: Confirming the cancelation should be done in the view model, not the service. The view model should handle user interaction.

src/GitHub.App/Services/PullRequestEditorService.cs
```diff
@@ -288,6 +289,15 @@ public void OpenActiveDocumentInCodeView(IVsTextView sourceView)
             NavigateToEquivalentPosition(sourceView, codeView);
         }
 
+        public bool ConfirmCancelPendingReview()
```
grokys: I'm still not sure this is the right place for this method. The XML doc on the class says:

```
Services for opening views of pull request files in Visual Studio.
```

Which doesn't really fit with this. I'd suggest adding it to `PullRequestSessionService`. I notice you added a `using System.Windows.Forms;` to that file, so I assume you actually started putting it there and then decided not to? Why was that?

StanleyGoldman: - `IPullRequestSessionService` cannot be imported into `PullRequestReviewAuthoringViewModel`
  - `PullRequestReviewAuthoringViewModel` lives in `GitHub.App`
  - `IPullRequestSessionService` lives in `GitHub.InlineReviews`

PR #1840, opened 2018-08-07
Always create GitHub Gists on GitHub

### LINKED ISSUE #1832: Create a GitHub Gist creates an Enterprise Gist without warning
- GitHub Extension for Visual Studio version: `2.5.5`
- Visual Studio version: All

1 . Select some text
2. Right click, `GitHub > Create GitHub Gist`
3. Create a GitHub Gist dialog appears:
image

- With no indication that Enterprise is being targeted, a GitHub Enterprise Gist is created and opened.

Since `Create a GitHub Gist` implies that it will be created on github.com and Enterprise Gists are private, this is most likely not what the user wants or was expecting.  It isn't obvious how to change to creating Gists on github.com.

jcansdale:
Previously Gists were being created on the first available connection. There was no indication when it was being created on GitHub Enterprise rather than github.com. This didn't make much sense when creating and sharing a GitHub Gist.

## What this PR does

Ensure that Gists are created on github.com.

- Allow `ShowWithFirstConnection` to specify a `HostAddress`
- Make `ShowCreateGist` use `HostAddress.GitHubDotComHostAddress`

## How to test

1. Log in to GitHub Enterprise
2. Log out of GitHub
3. Select some text and right-click `GitHub > Create GitHub Gist`
4. Login dialog should appear
5. Log in to GitHub
6. Create a Gist
7. Browser should appear with Gist on GitHub not GitHub Enterprise

Fixes #1832

### DIFF
--- a/src/GitHub.App/Services/DialogService.cs
+++ b/src/GitHub.App/Services/DialogService.cs
@@ -4,6 +4,7 @@ using System.Threading.Tasks;
 using GitHub.Extensions;
 using GitHub.Factories;
 using GitHub.Models;
+using GitHub.Primitives;
 using GitHub.ViewModels.Dialog;
 
 namespace GitHub.Services
@@ -54,7 +55,7 @@ namespace GitHub.Services
         public async Task ShowCreateGist()
         {
             var viewModel = factory.CreateViewModel<IGistCreationViewModel>();
-            await showDialog.ShowWithFirstConnection(viewModel);
+            await showDialog.ShowWithFirstConnection(viewModel, HostAddress.GitHubDotComHostAddress);
         }
 
         public async Task ShowCreateRepositoryDialog(IConnection connection)
--- a/src/GitHub.App/ViewModels/Dialog/GitHubDialogWindowViewModel.cs
+++ b/src/GitHub.App/ViewModels/Dialog/GitHubDialogWindowViewModel.cs
@@ -6,6 +6,7 @@ using System.Reactive.Subjects;
 using System.Threading.Tasks;
 using GitHub.Factories;
 using GitHub.Models;
+using GitHub.Primitives;
 using GitHub.Services;
 using ReactiveUI;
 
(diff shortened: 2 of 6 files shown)


meaghanlewis:
@jcansdale thanks for looking into this!

This works well for the situation you outlined above where previously the gist was being created on GitHub Enterprise instead of the GitHub instance. In the reverse situation, I: 
1. Log in to GitHub
2. Log in to GitHub Enterprise
3. Connect to a GitHub Enterprise repo
4. Open a file in the Enterprise repo and select some text and then choose to create a gist

The gist is created with GitHub, when I expect it to be created on Enterprise instead.

Is there a way where the gist can be created with my current connection ?

grokys:
> Is there a way where the gist can be created with my current connection ?

That would make a lot of sense! Also we have a username avatar in the gist dialog, which doesn't do anything. What if this could be used to select the instance you're posting to?

image

jcansdale:
@meaghanlewis,

> The gist is created with GitHub, when I expect it to be created on Enterprise instead.

The way I'm interpreting it is that `Create a GitHub Gist` is offering to create a Gist on GitHub not whatever GitHub Enterprise server the user happens to be using.

Gists are a convenient way to share public or private code snippets and hosting them on an internal server doesn't make much sense. I accidentally shared a link to our internal server before realizing and deleting the link.  Even if I'm working on an internal project, I might want to share a snippet on Stack Overflow.

What do you think?  Not many people use the Gist functionality anyway and GitHub Enterprise Gists will be even more of an edge case.

meaghanlewis:
@jcansdale I agree that Gists are a convenient way to share public or private code snippets and I think this applies to Gists on both GitHub and GitHub Enterprise. The use case you shared about sharing a snippet of code on an internal project can apply to both GitHub and GHE, right?

For example, I can view Gists on https://gist.github.com/discover and https://gist.ghe.io/discover.

I don't know how much this functionality is used, but I think we should at least offer the option to create both types of Gists. 

I also like @grokys suggestion about using the avatar to select the instance you post the Gist to.

jcansdale:
@meaghanlewis @grokys,

I've submitted an alternative solution #1854  which preserves the existing functionality, but allows the the user to choose between `Create a GitHub Gist` and `Create an Enterprise Gist` when logged into both. The `Create an Enterprise Gist` command also makes it clear when they're about to create a Gist on their Enterprise server.

> That would make a lot of sense! Also we have a username avatar in the gist dialog, which doesn't do anything. What if this could be used to select the instance you're posting to?

I started exploring this, but thought I'd try a less complex approach first. 😉

jcansdale:
Don't want to remove functionality. Killing this PR. 🔥

PR #1842, opened 2018-08-08
Style TreeViewItem

### LINKED ISSUE #1294: Changes list: wrong colors when highlighting selection
The selection highlight colors used in the PR detail changes list feels wrong:

image

Pretty sure it should be gray and not white.

grokys:
This PR adds a new style for `TreeViewItem` to match VS' theme.

- Correctly uses gray as the non-active selection background (see #1294)
- Uses the same expander size as elsewhere
- Styles the expanders to be hollow when not expanded and use correct mouseover colors

Before:

image

After:

image

Fixes #1294

### DIFF
--- a/src/GitHub.VisualStudio.UI/Styles/ItemsControls.xaml
+++ b/src/GitHub.VisualStudio.UI/Styles/ItemsControls.xaml
@@ -74,4 +74,151 @@
             </Setter.Value>
         </Setter>
     </Style>
+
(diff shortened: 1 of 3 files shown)


jcansdale:
This is looking great. 😄 

Being super pedantic, the vertical alignment of the icons isn't quite the same as `Solution Explorer`.

- Solution Explorer
image

- Changes
image

grokys:
I think pedanticism is warranted here @jcansdale - that's pretty much what this PR is about ;) Will fix.

meaghanlewis:
This LGTM ✔️

PR #1852, opened 2018-08-09
Treat negative durations in comments as having just occurred

### LINKED ISSUE #1851: Creating a comment often leaves a negative timestamp
Due to slight differences in local time vs server time, we often display something like `-3 second ago` for comments when they are created.

image

StanleyGoldman:
Fixes #1851 

image


grokys:
Just a question: wouldn't it have been easier to put this logic in the view model? That is, wouldn't just changing [this line](https://github.com/github/VisualStudio/blob/master/src/GitHub.InlineReviews/ViewModels/CommentViewModel.cs#L73) to something like:

```
UpdatedAt = updatedAt >= DateTimeOffset.UtcNow ? updatedAt : DateTimeOffset.UtcNow;
```

Have the same effect?

grokys:
Though I can understand that maybe we want this logic in a single place... In that case, it's worth thinking about whether `DurationToStringConverter` should even be able to say "-1 seconds ago". Will that ever be needed?

StanleyGoldman:
We can definitely add it to the existing converter, I just wasn't sure...

StanleyGoldman:
Ok, I merged the functionality into one converter.

PR #1853, opened 2018-08-09
Better failure modes

### LINKED ISSUE #1305: Provide better failure modes when a connection error occurs
__Problem:__
Currently if VS is started up disconnected from the internet, the connection is simply removed from Team Explorer. 

__Solution ideas:__
We need to display a message with a "Retry" button or similar if a connection error happens, and similarly display a "You need to log in again" button or similar if the users credentials have changed.

grokys:
When a login fails at VS startup, currently we just remove the connection from the TE connect page and pretend it never existed. With this PR, we show a message and allow the user to retry (in the case of non-auth-related) errors or sign out:

General connection error:

image

Auth error:

image

In addition, an error is shown in the GitHub pane with a link to open the Team Explorer Connect section to fix the problem:

image

The UI needs some love from @donokuda , but this is ready to play with!

Fixes #1305


meaghanlewis:
Tested this by changing my GitHub for Visual Studio credentials in the credential manager to be incorrect and by using the proxy server Burp.

@donokuda if you need any help testing your changes locally, let me know.

This LGTM, thanks @grokys! ✨

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
```diff
@@ -249,6 +291,23 @@ void OnPropertyChange(object sender, PropertyChangedEventArgs e)
                 settings.IsExpanded = IsExpanded;
         }
 
+        void InvitationSectionPropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == "IsVisible")
```
StanleyGoldman: `nameof` maybe?

src/GitHub.App/ViewModels/GitHubPane/GitHubPaneViewModel.cs
```diff
@@ -410,26 +417,56 @@ async Task UpdateContent(ILocalRepositoryModel repository)
...
                 }
                 else
                 {
                     log.Debug("Found a a GitHub repository but not logged in: {CloneUrl}", repository?.CloneUrl);
                     Content = loggedOut;
                 }
             }
-            else
+            
+            if (notGitHubRepo)
```
StanleyGoldman: `else if`?

grokys: No, this flag is set in the preceding block.

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
grokys: Yeah I was trying to stick with the style of the previous method which didn't use `nameof`, but you're right. I've converted this method and the previous to use `nameof`.

PR #1860, opened 2018-08-13
localize VS 'Extensions and Update' And  Add Chinese Traditional language

maikebing:
localize the package metadata which is used by VS 'Extensions and Update'

image


meaghanlewis:
For some reason, the extension description doesn't update for me.

screen shot 2018-08-15 at 2 30 17 pm

maikebing:
Sorry, I have not added traditional Chinese, I only added Simplified Chinese.

PR #1714, opened 2018-06-01
Localization

maikebing:
Localization work started
Let's all translate together, click here to join HTTPS://CRWD.IN/GITHUBVS

- [x]   Chinese (Simplified) -
- [x] Chinese (Traditional) -  

- [x] English - 

```
- [ ] Czech
- [ ] French - 
- [ ] German -  @wenz
- [ ] Italian -  
- [ ] Japanese
- [ ] Korean
- [ ] Polish
- [ ] Portuguese (Brazil)
- [ ] Russian - 
- [ ] Spanish - 
- [ ] Turkish
```

PR #1871, opened 2018-08-22
Fix PR icon in Status Bar and GitHub Icon on Start Page

### LINKED ISSUE #1863: PR icon disappeared on the status bar
__Version__
- GitHub Extension for Visual Studio version: 2.5.5 master

__What happened__
Connect to a repo and notice that the PR icon or PR number doesn't show up in the status bar any longer. This seems to have disappeared in the past week.

Expected to see this:
pr icon

pr number

See no PR icon instead:

screen shot 2018-08-15 at 12 41 02 pm

grokys:
#1714 broke a few things that were causing this:

- It removed the `IncludeOutputGroupsInVSIX`/`IncludeOutputGroupsInVSIXLocalOnly` properties from `GitHub.VisualStudio.csproj`. In the case of `GitHub.InlineReviews` and `GitHub.StartPage` this prevented the packages loading which caused #1863 
- Once this was fixed, needed to add `SatelliteDllsProjectOutputGroup` to the relevant references in order to get the satellite assemblies to be shipped along with the extension

This wasn't a bug, but was unnecessary:

- `OcticonPaths.resx` doesn't need to be translated as it consists solely of XAML paths.

Fixes #1863


shana:
There are more changes in #1714 that seem unnecessary or not really clear:

- https://github.com/github/VisualStudio/pull/1714/files#diff-44ee993861822d4f52c2cc353f059cffR482 is odd: if the localization falls back to the neutral resource, why have a `GitHub.VisualStudio.en-US.vsct` and a `GitHub.VisualStudio.vsct` if the second is empty? If the point is to have localized strings in the menus, than anything that doesn't have strings should probably be in the non localized vsct, otherwise adding things is going to be double the pain.

- https://github.com/github/VisualStudio/pull/1714/files#diff-930c4aa2d9b0f5a330b6b5deab76a257L16 is not a good change - every project needs to define the neutral resources language, which is why it was on SolutionInfo.cs in the first place, so we might want to revert all the SolutionInfo.cs and AssemblyInfo.cs changes in #1714 

- `src/GitHub.VisualStudio/VSPackage.zh-Hans.resx` doesn't have any localized strings so it probably doesn't need to be there.

- Why is there a `src/GitHub.VisualStudio.UI/Resources.resx` and a `src/GitHub.VisualStudio.UI/Resources.en-US.resx`? They both contain localizable strings and look to be copies of each other? Both are embedded but only one is getting compiled.

Issue #1603, opened 2018-04-14
UX Improvement: HiDPI Avatars for HiDPI users


NickCraver:
- GitHub Extension for Visual Studio version: Pull Request Reviews Beta 1
- Visual Studio version: 15.7 Preview 3

For high DPI screen,s avatars are currently very fuzzy. Not sure this'll come across in a picture due to dPI being oh so fun, but screenshot:
![image](https://user-images.githubusercontent.com/454813/38770694-12fafc5a-3fe5-11e8-83b9-c3fb1e162d71.png)

cc @sguthals

grokys:
@NickCraver this should be fixed in the latest 2.5.4 release, so I'm going to close this. Please let us know if it's not fixed for you!

Issue #1547, opened 2018-03-20
Avatars being repeatedly downloaded.


grokys:
From halp:

> I manage a github enterprise instance. A lot of the people I support internally use the https://visualstudio.github.com/ open source project that seems to be maintained by GitHub staff https://github.com/github/VisualStudio/commits/master. The problem I am experiencing is that the VisualStudio GitHub extension makes 10's of thousands of API calls to download account avatars when accessing a repo. This spikes CPU usage to 100% over the 10+ minutes it is downloading the avatars. For example below is the accumulated API calls to download avatars made from a single developers system while accessing a few repos over a 1 hour period:

> $ grep -F 'now="2018-03-19T08' unicorn.log > unicorn_8AM.log
> $ grep -F '10.13.102.88' unicorn_8AM.log | grep Favatars | grep -o now=[^\ ]* | cut -c 17-20 | uniq -c | > uniq -c | sort -rn | head -10
> > 1 29643 08:0
> 1 28866 08:3
> 1 18158 08:2
> 1 16378 08:4
> 1 918 08:1

> This behavior is consistent for everyone that uses this extension.

grokys:
I suspect this is a similar issue that desktop were seeing, in that avatars fail to authenticate with a token, and so we're repeatedly trying to download them, willing it to not fail: https://github.com/desktop/desktop/issues/821#issuecomment-296272496

darn:
It should be noted that the github enterprise instance in the OP is running the latest public version 12.2.7.

darn:
I am wondering why avatar downloads are a requirement? Could there be a setting that allows the client to ignore avatars?

jcansdale:
Hi @darn,

Could you try the `GitHub.VisualStudio.vsix` at the following location and let me know if it resolves the issue?
https://ci.appveyor.com/project/github-windows/visualstudio/build/2.4.4.2016/artifacts

The avatar should revert to a default image rather than bombard your server with requests. There might be a few requests but not the number you're currently seeing!

darn:
Confirmed improvement. The following graphic shows (1) instance of opening a repo prior to installing the Fix and (2) instances of opening the same repo after installing the Fix.
![image](https://user-images.githubusercontent.com/2659996/38318404-17473bc4-37e4-11e8-96e3-72cf8b6c5c41.png)

jcansdale:
Thanks @darn for the initial report and confirming this fix! I look forward to getting it into the next release.

[...]

PR #1873, opened 2018-08-22
Display default avatar for avatars from enterprise hosts.

### LINKED ISSUE #1870: No user avatars displayed for enterprise repositories
## Version

- GitHub Extension for Visual Studio version: 2.5.5
- Visual Studio version: VS2017

## What happened

**Steps to Reproduce**

1. Connect to an enterprise repository
2. View the pull request list, individual pull requests, and inline comments

**Expected behavior:** 
Avatars are shown

**Actual behavior:**
No avatars are shown. It's just an empty space

**Screenshot or GIF:**
screen shot 2018-08-21 at 10 09 43 am
screen shot 2018-08-21 at 10 09 35 am
screen shot 2018-08-21 at 10 42 48 am

I know this has been an issue before, but if we couldn't get the avatar we at least would show something like this:
screen shot 2018-08-21 at 10 06 57 am

It seems so empty just to have a blank space where an avatar should be.

grokys:
Image requests to enterprise hosts over https always fail currently, so just display the default avatar (see #1547).

This was already fixed for the REST portion of the application in #1579 by remembering which hosts image downloads have already failed for and making sure not to try again, however the framework of `AvatarProvider`/`ImageCache`/`ImageDownloader` used there assumes a fixed image size of 32x32, causing #1603.

The whole thing needs a major revamp, which is a little tricky while we're still using a combination of REST and GraphQL, so for the moment just don't try to download images from enterprise hosts that are using `https`: this should be OK because this will currently never work anyway.

Fixes #1870.


meaghanlewis:
LGTM. Thanks for fixing this!

src/GitHub.App/ViewModels/ActorViewModel.cs
```diff
@@ -16,10 +17,30 @@ public ActorViewModel()
         public ActorViewModel(ActorModel model)
         {
             Login = model?.Login ?? "[unknown]";
-            Avatar = model?.AvatarUrl != null ?
-                new BitmapImage(new Uri(model.AvatarUrl)) :
-                AvatarProvider.CreateBitmapImage(DefaultAvatar);
-            AvatarUrl = model?.AvatarUrl ?? DefaultAvatar;
+
+            if (model?.AvatarUrl != null)
```
jcansdale: Nit: I find the new `if (model?.AvatarUrl is string avatarUrl)` syntax more readable.

src/GitHub.App/ViewModels/ActorViewModel.cs
```diff
@@ -16,10 +17,30 @@ public ActorViewModel()
...
+                    // Image requests to enterprise hosts over https always fail currently,
+                    // so just display the default avatar. See #1547.
+                    if (uri.Scheme != "https" ||
+                        uri.Host.EndsWith("githubusercontent.com", StringComparison.OrdinalIgnoreCase))
+                    {
+                        AvatarUrl = model.AvatarUrl;
+                        Avatar = new BitmapImage(new Uri(AvatarUrl));
+                    }
+                }
+                catch { }
```
jcansdale: Is the `catch { }` required with the above `url` checks? If I'm reading it correctly, an exception here would be unexpected and we should log it.

grokys: Yes, the catch is in case the URL is somehow an invalid format. And yes we should probably be logging that ;)

src/GitHub.App/ViewModels/ActorViewModel.cs
grokys: Interesting, what do you find more readable about that? I use the new syntax a lot, but only when a cast is involved. `AvatarUrl` is already typed to string so there is no cast, and so it's just a comparison with null...

grokys: Yeah, I dunno... The problem for me with "if model?.AvatarUrl is a string then call it avatarUrl and do stuff with it" is that it's _always_ a string. That pattern should be used for casting IMO, not just checking for null, and it surprises me when there isn't actually a cast.

PR #1901, opened 2018-09-06
Record CurrentUICulture as well as InstalledUICulture

jcansdale:
~~edit (shana):_ DO NOT MERGE MASTER INTO THIS. _This branch is based off of the last release tag for easy backporting.~~ *We didn't do an out of band release in the end so this can be merged into the next normal release.*

### What this PR does

GitHub for Visual Studio will be localized according to what Visual Studio language pack is being used, not what localized version of Windows is installed. The property we're interested in is `CultureInfo.CurrentUICulture` not `CultureInfo.InstalledUICulture`.

- Create a new `CurrentUILang` metric for `CultureInfo.CurrentUICulture`

### Questions

Initially wasn't sure whether to update `Lang` or create a new metric. I've opted to create a new `CurrentUILang` metric.

It's possible that `InstalledUICulture` is actually a reasonable approximation of `CurrentUICulture`. I think the OS language will be used as the default when installing a VS language pack. `Same as Microsoft Windows` is one of the options when selecting language.

image


shana:
Could we base this off of the latest release tag instead of master? That way we can ship a hotfix if we want to without having to backport the changes.

PR #1906, opened 2018-09-10
Include owner in default path when cloning a repository

jcansdale:
### What this PR does

Rather than suggesting a flat folder structure when cloning a repository, suggest a folder structure that includes the repositories owner

- Suggest a default clone location of `DefaultClonePath/Owner/RepoName`
- Allow user to change `DefaultClonePath` portion of path (keep when new repo is selected)

#### Immediately change default clone path
1. Change the `DefaultClonePath` before selecting a repository
2. Select a repository
3. Check that `owner\repo` has been added to path

#### Change directory name after selecting repository
1. Select a repository
2. Change name of repository in path
3. Select another repository
4. Check the `owner\repo` in path has changed

#### Remove repository name from path
1. Select a repository
2. Remove name of repository in path
3. Select another repository
4. Check the `repo` has been added to path

#### Remove owner and repository name from path
1. Select a repository
2. Remove owner and name of repository from path
3. Select another repository
4. Check that `owner\repo` has been added to path

#### Clear the path
1. Clear the path
2. Select a repository
3. Check that default path has been restored and `owner\repo` added to path

#### Remove owner from path
1. Select a repository
2. Remove owner from path
3. Select another repository
4. Check that previous repository name isn't included in path


meaghanlewis:
👋 @jcansdale I just had a look at how this works.

Question: 
What was your motivation for doing this functionality? Do you usually have more structure in the way your organize your repositories? Just curious here 😄 

Feedback:
Thanks for all the test scenarios! Those all work how you describe. A couple other scenarios that stuck out for me were:
- **removing the full path** : select a repository, remove the path, select another repository the path is only populated as `owner\repo` 
- **removing only the owner name** select a repository, remove just the owner and keep the repo name, select another repository and notice the path gets appended with another `owner\repo` so the full path becomes `default clone path\repo\owner\repo`

Both of which are fine as they are since a user will likely select a new local clone path they're happy with if something undesired happened.

I think this works really nicely and will be helpful at providing more structure for those who have lots of repositories from different owners or organizations.

jcansdale:
@meaghanlewis,

Thanks for taking a 👀

> What was your motivation for adding this functionality? Do you usually have more structure in the way your organize your repositories? Just curious here 😄

Yup, I always organize my folders in a way that mirrors GitHub. I'll often clone a few repositories from the same owner. It's the only way I can keep any kind of order!

> removing the full path : select a repository, remove the path, select another repository the path is only populated as owner\repo

It would be easy enough to reset back to the default clone location if the user removes the full path. I think if someone was to do this, that is what they'd be hoping for?

> removing only the owner name select a repository, remove just the owner and keep the repo name, select another repository and notice the path gets appended with another owner\repo so the full path becomes default clone path\repo\owner\repo

I'll see if this can be fixed as well without any undesirable consequences. I think its okay for the reason you mentioned, but maybe it could be better. 😄

jcansdale:
@meaghanlewis,

I've fixed the two scenarios that you mentioned and added them to the list.

I've also fixed an an edge case where the owner and repository had the same name. Previously it would only delete the repository rather than both.

meaghanlewis:
This LGTM! ✨ ✨

test/GitHub.App.UnitTests/ViewModels/Dialog/Clone/RepositoryCloneViewModelTests.cs
```diff
@@ -114,12 +114,10 @@ public async Task Path_Is_Initialized()
         public async Task Repository_Name_Is_Appended_To_Base_Path()
```
grokys: Need to update test name.

PR #1919, opened 2018-09-13
Unified Clone / Open from GitHub UI

jcansdale:
### What this PR does

- Allow user to Open a repository when the target path already exists
- View/open cloned repositories on Team Explorer Home (same as built in clone)
- In Visual Studio 2017 open cloned repositories in Folder view (same as built in clone)
- Fix GitHub button on Start Page
- Add `Add File / Open / Open from GitHub...` button
- Pre-populate Url from clipboard/command args when available
- Only show warning banner when a file is in the way of the target path
- Use Clone/Open dialog for existing `GitHub.OpenFromUrl` command

### TODO

- [ ] Add metrics

### What this PR doesn't do

The `GitHub.OpenFromUrl` command will no longer attempt to navigate the the file and line number from a target Url. This was done to trim the scope of this PR to simply cloning/opening a repository.

The next step will be to incorporate functionality from the `Open from clipboard` command. This will be more complex to test and will be done as a separate PR.

At the moment these two commands can be combined manually:
1. Copy a GitHub URL to the clipboard
2. Use `Add File / Open / Open from GitHub...`
3. Use `GitHub > Open from clipboard`

#### Here is an example workflow

1. Copy a GitHub URL
image

2. Select `File > Open > Open from GitHub...`
image

3. URL is copied into username/repository field and path is populated with a default location
image

4. Click `Open` if local repository already exists or update path and  `Clone` to create a new repository
image

5. Repository folder is loaded into `Solution Explorer` and `Team Explorer - Home` appears
image

6. Select from list of solutions


meaghanlewis:
This LGTM @jcansdale ✨

src/GitHub.App/Services/RepositoryCloneService.cs
```diff
@@ -103,6 +106,24 @@ public async Task<ViewerRepositoriesModel> ReadViewerRepositories(HostAddress ad
...
+            Guard.ArgumentNotEmptyString(url, nameof(url));
+            Guard.ArgumentNotEmptyString(repositoryPath, nameof(repositoryPath));
+
+            if (!DestinationDirectoryExists(repositoryPath) && !DestinationFileExists(repositoryPath))
+            {
+                var cloneUrl = url.ToRepositoryUrl().ToString();
+                await CloneRepository(cloneUrl, repositoryPath, progress);
+            }
+
+            teamExplorerServices.OpenRepository(repositoryPath);
```
jcansdale: We don't need to do this when cloning (it happens as part of the clone).

src/GitHub.App/Services/DialogService.cs
```diff
@@ -29,9 +29,13 @@ public class DialogService : IDialogService
             this.showDialog = showDialog;
         }
 
-        public async Task<CloneDialogResult> ShowCloneDialog(IConnection connection)
+        public async Task<CloneDialogResult> ShowCloneDialog(IConnection connection, string url)
```
grokys: The interface adds a default value of `null` to this parameter. Should probably add it here as well to make the signatures match.

src/GitHub.App/Services/RepositoryCloneService.cs
```diff
@@ -103,6 +106,24 @@ public async Task<ViewerRepositoriesModel> ReadViewerRepositories(HostAddress ad
...
 
+        /// <inheritdoc/>
+        public async Task CloneOrOpenRepository(
+            UriString url,
+            string repositoryPath,
+            object progress = null)
+        {
+            Guard.ArgumentNotEmptyString(url, nameof(url));
+            Guard.ArgumentNotEmptyString(repositoryPath, nameof(repositoryPath));
+
```
grokys: Should we be checking whether the repository at this path has the same clone URL? I'm thinking of the case where e.g. the user wants to clone a fork and they already have the parent repository at this location.

src/GitHub.App/ViewModels/Dialog/Clone/RepositoryCloneViewModel.cs
```diff
@@ -232,7 +239,8 @@ string ValidatePath(IRepositoryModel repository, string path)
         {
             if (repository != null)
             {
-                return service.DestinationExists(path) ?
+                // TODO: Update wording for DestinationAlreadyExists
```
grokys: Does this TODO need addressing before we merge?

src/GitHub.App/Services/RepositoryCloneService.cs
```diff
@@ -103,6 +106,24 @@ public async Task<ViewerRepositoriesModel> ReadViewerRepositories(HostAddress ad
             return result;
         }
 
+        /// <inheritdoc/>
```
grokys: I'm wondering whether we need both `Clone` and `CloneOrOpen`? Wouldn't `Clone` and `Open` make more sense? The decision of which to call would typically be made at a higher level I would have thought as we're checking whether the repository already exists in order to show/hide the `Open` button.

jcansdale: I've updated this as discussed in Slack. There is now a `CloneOrOpenRepository(CloneDialogResult)` method which is indented to be used when the clone dialog returns a non-null `CloneDialogResult`.

While updating this I discovered that the create repository functionality is also calling `CloneRepository`. I this means that for every `NumberOfReposCreated` incremented, the `NumberOfClones` counter will be incremented as well. Oops! //cc @telliott27

src/GitHub.App/Services/RepositoryCloneService.cs
```diff
@@ -121,20 +172,7 @@ public async Task<ViewerRepositoriesModel> ReadViewerRepositories(HostAddress ad
             try
             {
                 await vsGitServices.Clone(cloneUrl, repositoryPath, true, progress);
-
                 await usageTracker.IncrementCounter(x => x.NumberOfClones);
```
jcansdale: This method is being called by `CreateRepository` here:
https://github.com/github/VisualStudio/blob/8067273b31420d5f9d487fcd29d35421d50e68c8/src/GitHub.App/Services/RepositoryCreationService.cs#L40

src/GitHub.App/ViewModels/Dialog/Clone/RepositoryCloneViewModel.cs
jcansdale: Yes. I was holding off because the the internationalization work. I'll update it now.

jcansdale: Changed to "A file exists at the destination path".

src/GitHub.App/Services/RepositoryCloneService.cs
jcansdale: Yes! In fact there are a few scenarios we should be checking for.

1. That a repository exists at the target path
2. That the repository has an `origin` remote
3. That the repository's remote matches the target URL

I'll let you know when these checks land.

jcansdale:
@grokys re: https://github.com/github/VisualStudio/pull/1919#discussion_r220139796

> Should we be checking whether the repository at this path has the same clone URL? I'm thinking of the case where e.g. the user wants to clone a fork and they already have the parent repository at this location.

I've added checks for this and also for when there is no repository at the target path or when the repository doesn't have a remote named `origin`.

I've moved the warning banner to above the `Local path` text box that it's validating. This was for a couple of reasons.

1. Having at appear above the tree view meant that when it appeared, it moved the tree view control that the user might have clicked on. This lead to rapid cycling between two selections.

2. Moving it to above the field that it's validating means the context is obvious and the warnings can be more concise (we're somewhat restricted for space).

Here are the warnings:

When there is no repository at the local path (e.g. an empty folder):
image

When the remote URL of the local repository is different to the URL selected:
image

When there is a file in the way of the local path (in this case there is a file called `PReview`:
image

When the target repository doesn't have a remote named `origin` and won't be recognized by GHfVS:
image

jcansdale:
- It will now always this this when a local clone already exists:
image

jcansdale:
Here are the updated warning messages:

image

image

image

image

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
```diff
@@ -151,16 +151,14 @@ async Task DoClone()
...
-                        result.Path);
+                    await cloneService.CloneOrOpenRepository(result);
 
                     usageTracker.IncrementCounter(x => x.NumberOfGitHubConnectSectionClones).Forget();
                 }
                 catch (Exception e)
                 {
                     var teServices = ServiceProvider.TryGetService<ITeamExplorerServices>();
-                    teServices.ShowError(e.GetUserFriendlyErrorMessage(ErrorType.ClonedFailed, result.Repository.Name));
+                    teServices.ShowError(e.GetUserFriendlyErrorMessage(ErrorType.ClonedFailed, result.Url.RepositoryName));
```
StanleyGoldman: Should the user message here change?

jcansdale: You're right. Commit incoming. 🚀

PR #1920, opened 2018-09-14
Fix log source on IServiceProviderExtensions 

jcansdale:
The incorrect type was being specified as the source.

- Log for `IServiceProviderExtensions` not `VSServices`

image


jcansdale:
@grokys I think this was one of yours. Could you sanity check and merge? 😄

PR #1933, opened 2018-09-19
Internationalize VS menus/commands

grokys:
Use https://github.com/dotnet/xliff-tasks to translate `GitHub.VisualStudio.vsct` and `VSPackage.resx` using `.xlt` files.

This should make all menus installed by the extension appear translated. A good way to test if this is working is to check the "Connect to GitHub" menu in the Team Explorer Connect Pane:

image

Includes a partial translation in `zh-CN`, @maikebing if you could take a look at the current state of translations on crowdin and approve/update the missing translations, that would be great!


src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
```diff
@@ -17,6 +18,7 @@
     <VsixType>v3</VsixType>
     <IsProductComponent>false</IsProductComponent>
     <ExtensionInstallationFolder>GitHub\GitHub</ExtensionInstallationFolder>
+	<XlfLanguages>zh-CN</XlfLanguages>
```
jcansdale: I notice you've specified `zh-CN` here rather than `zh-Hans` (which I think is simplified Chinese). Is the documentation on the https://github.com/dotnet/xliff-tasks README wrong?

> `XlfLanguages` - The set of locales to which the project is localized. Defaults to the thirteen locales supported by Visual Studio: `cs;de;es;fr;it;ja;ko;pl;pt-BR;ru;tr;zh-Hans;zh-Hant`.

Does Visual Studio support `zh-CN`, `zh-Hans` and `zh-Hant`? 😕

grokys: I'm... not sure. The [original PR](https://github.com/github/VisualStudio/pull/1714) had `zh-HANS` translations, but then when it was re-done in #1888, the translations that came from crowdin were `zh-CN` and this seems to work fine.

As for which is correct, I'm afraid I don't know enough about internationalization or Chinese in particular to say. I've tried googling and I'm not the wiser ;)

As I say, it seems to work fine, so ¯\_(ツ)_/¯

nguerrera: > https://twitter.com/jcansdale/status/1042490984892903424
> Is it okay to use zh-CN instead of zh-Hans/zh-Hant in Visual Studio?

The resource manager will probe based on CurrentUICulture from a culture through its parents:

For zh-CN, this goes zh-CN -> zh-CHS -> zh-Hans.

This means, for example that if CurrentUICulture is zh-CN, then it will use zh-Hans resources, but not vice versa. It might be the case that VS always sets CurrentUICulture to zh-CN when using Chinese (Simplified) language pack, in which case it would work to use zh-CN, but I am not certain of that. If that isn't true, then using zh-CN will not work in all cases.

In the end, there are just two possible choices for VS language packs:

image

And their descriptions align to zh-Hans and zh-Hant
```
> using System.Globalization;
> new CultureInfo("zh-Hans").DisplayName
"Chinese (Simplified)"

> new CultureInfo("zh-Hant").DisplayName
"Chinese (Traditional)"
```

So I would suggest lining up with that and just using zh-Hans and zh-Hant.

grokys: @maikebing when you submitted the latest round of translations from crowdin using `zh-CN` rather than `zh-Hans`, was there a reason for that?

grokys:
@sguthals yes, `.vsct` files can't be easily translated, so microsoft created the `xliff-tasks` tool which uses `.xlf` files to apply translations to `.vsct` files, and we're using that. `xliff-tasks` essentially duplicates the `.vsct` file with translations in place and compiles them into to the binary.

src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
jcansdale: I'm afraid it doesn't appear to work when the `Chinese (Traditional)` language pack is installed. 😭 

image

maikebing:
@grokys    https://crowdin.com/project/github-visual-studio was not created by me, nor can I add a language! 
Mainland China using Simplified Chinese zh-Hans
Hong Kong and Taiwan and other regions to use traditional Chinese zh-Hant
 if it is zh-CN when the system language zh-Hans or zh-Hant can be compatible

grokys:
Ah so it was @StanleyGoldman who created the language with `zh-CN`.

@maikebing so I assume we should be using `zh-HANS` rather than `zh-CN`? Problem is, we're really close to a release - will we cause problems shipping with `zh-CN` for this version?

src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
jcansdale: It appears that `CultureInfo.CurrentUICulture` is set to `zh-TW` when using the `Chinese (Traditional)` language pack. The parents of this appear to be `zh-TW -> zh-CHT -> zh-Hant -> zh`.

@maikebing am I correct in thinking that `zh-Hans` and `zh-Hant` are treated like completely separate languages? There is no mechanism for `zh-Hant` to fall back to using `zh-Hans`?

maikebing:
Chinese (Traditional) need to re-translate and revise the grammar, Simplified Chinese (zh-Hans) is not fully applicable to traditional Chinese (zh-Hant). So I propose to change zh_CN to Simplified Chinese (zh-hans), after adding zh-Hant, I will translate and solicit the views of Hong Kong and Taiwan users as soon as possible.

grokys:
@maikebing thanks! Will we be OK for this release shipping with `zh-CN`, or will that cause problems?

maikebing:
I corrected my mistake!
zh_CN on behalf of mainland China,
zh_TW on behalf of the Taiwan Region of China
zh_HK on behalf of Hong Kong region of China

zh_Hans is a notation for Simplified Chinese 
zh_Hant is a notation for traditional Chinese

grokys:
@maikebing excellent! Thanks so much for your help on this, I'm afraid our team's lack of knowledge of internationalization has really been showing ;)

StanleyGoldman:
Yea, I had no idea what I was doing when I was setting up CrowdIn. I was more concerned with the automatic integration and figured we would get the languages correct as we went along. Thanks @maikebing.

PR #1937, opened 2018-09-19
Fix GitHub button on Start Page

### LINKED ISSUE #1927: GitHub button not working on start page
## Version

- GitHub Extension for Visual Studio version: 2.5.6.4513 and later
- Visual Studio version: 2017

## What happened

**Steps to Reproduce**

1. Open up VS2017 and go to start page
2. Click GitHub icon

**Expected behavior:**
Clone dialog or sign in dialog to appear

**Actual behavior:**
Nothing happens

**Screenshot or GIF:**

**More info:**
The icon stopped responding between these builds: 
https://ci.appveyor.com/project/github-windows/visualstudio/build/2.5.6.4503 - https://ci.appveyor.com/project/github-windows/visualstudio/build/2.5.6.4513

jcansdale:
### What this PR does

Default to using the URL tab when no explicit connection is specified by the StartPage button. This is the neutral tab that works with both GitHub and GitHub Enterprise.

- Let StartPage pass `null` to `ShowCloneDialog`
- Updated unit tests
- Fixed exception that was appearing in log
`GetServiceSafe: Could not obtain instance of 'Microsoft.TeamFoundation.Controls.ITeamExplorerPage'`

### How to test

1. Open Visual Studio 2017
2. Click the GitHub button on the Start Page

Expect the `Clone` dialog to appear with the URL tab selected. Check that a GitHub or GitHub Enterprise repository can be cloned.

Fixes #1927

### DIFF
--- a/src/GitHub.App/ViewModels/Dialog/Clone/RepositoryCloneViewModel.cs
+++ b/src/GitHub.App/ViewModels/Dialog/Clone/RepositoryCloneViewModel.cs
@@ -118,6 +118,11 @@ namespace GitHub.ViewModels.Dialog.Clone
                 SelectedTabIndex = 1;
             }
 
+            if (connection == null)
+            {
+                SelectedTabIndex = 2;
+            }
+
             this.WhenAnyValue(x => x.SelectedTabIndex).Subscribe(x => tabs[x].Activate().Forget());
         }
 
--- a/src/GitHub.App/Services/DialogService.cs
+++ b/src/GitHub.App/Services/DialogService.cs
@@ -31,8 +31,6 @@ namespace GitHub.Services
 
         public async Task<CloneDialogResult> ShowCloneDialog(IConnection connection)
         {
-            Guard.ArgumentNotNull(connection, nameof(connection));
-
             var viewModel = factory.CreateViewModel<IRepositoryCloneViewModel>();
 
             return (CloneDialogResult)await showDialog.Show(
--- a/src/GitHub.Exports/Services/IDialogService.cs
+++ b/src/GitHub.Exports/Services/IDialogService.cs
@@ -13,8 +13,7 @@ namespace GitHub.Services
         /// Shows the Clone dialog.
         /// </summary>
         /// <param name="connection">
-        /// The connection to use. If null, the first connection will be used, or the user promted
-        /// to log in if there are no connections.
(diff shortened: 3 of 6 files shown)


src/GitHub.App/ViewModels/Dialog/Clone/RepositoryCloneViewModel.cs
```diff
@@ -118,6 +118,11 @@ public async Task InitializeAsync(IConnection connection)
                 SelectedTabIndex = 1;
             }
 
+            if (connection == null)
+            {
+                SelectedTabIndex = 2;
```
meaghanlewis: When I open the clone dialog from the Start Page it always opens up to the URL tab. Is this why?

meaghanlewis:
LGTM except for the clone dialog always opening up to the URL tab. 👍

src/GitHub.App/ViewModels/Dialog/Clone/RepositoryCloneViewModel.cs
jcansdale: When a GitHub or GitHub Enterprise account isn't explicitly targeted, it will default to using the URL tab.

When using this functionality I've found that more often than not, I'm cloning a repository that I've found on github.com rather than one of my own (or my organization's) repositories. Having to wait while a list of repositories that probably won't included the one I'm interested in isn't very useful (in fact it can be off-putting).

I think it makes more sense for the user to either explicitly target a GitHub/Enterprise account or paste the URL they were just looking at on github.com and Clone.

Maybe @donokuda has some thoughts?

meaghanlewis: I get what you're saying about waiting to wait for a long list of repositories to load and being able to easily paste the URL, but I would expect the dialog on the Start to open to the first tab just like it does from Team Explorer.

grokys: Yes we should definitely open the first tab rather than the URL tab. @jcansdale the reason it's slow is that the `github` org contains 2,461 repositories. Most users won't be members of an org with that many repositories and their list should load in 1 or 2 requests.

jcansdale:
@grokys is working on a PR that restores the previous behavior. Closing this one.

PR #1953, opened 2018-09-25
Including the xlf files in ncrunch's build

StanleyGoldman:
NCrunch build was failing with:

```
NCrunch: This project was built on server '(local)'
..\..\packages\XliffTasks.0.2.0-beta-63125-01\build\XliffTasks.targets (92, 5): 'xlf\VSPackage.zh-CN.xlf' for 'VSPackage.resx' does not exist. Run `msbuild /t:UpdateXlf` to update .xlf files or set UpdateXlfOnBuild=true to update them on every build, but note that it is strongly discouraged to set UpdateXlfOnBuild=true in official/CI build environments as they should not modify source code during the build.
```

### DIFF
--- a/.ncrunch/GitHub.VisualStudio.v3.ncrunchproject
+++ b/.ncrunch/GitHub.VisualStudio.v3.ncrunchproject
@@ -1,8 +1,7 @@
 ﻿<ProjectConfiguration>
   <Settings>
     <AdditionalFilesToIncludeForProject>
-      <Value>xlf\GitHub.VisualStudio.vsct.zh-CN.xlf</Value>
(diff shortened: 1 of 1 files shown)


.ncrunch/GitHub.VisualStudio.v3.ncrunchproject
```diff
@@ -1,5 +1,9 @@
 ﻿<ProjectConfiguration>
   <Settings>
+    <AdditionalFilesToIncludeForProject>
+      <Value>xlf\GitHub.VisualStudio.vsct.zh-CN.xlf</Value>
```
jcansdale: I wonder if wildcards are supported here, e.g. `<Value>xlf\*.xlf</Value>`?

PR #1971, opened 2018-10-04
Fix slowness and non-determinism of GetLatestPushedSha (GitHub > Open on GitHub)

### LINKED ISSUE #1707: Open on GitHub and Blame is extremely slow
When I click GitHub -> Blame or Open on GitHub, it takes more than 10 minutes until the website pops up in my browser.

jcansdale:
The `GetLatestPushedSha` method which is called when creating or opening a link on github.com could be extremely slow, depending on the repository and how many commits ahead the local branch was. This PR resolves a number of issues with that method.

### What this PR does

- Always run `GetLatestPushedSha` on thread pool (use `Task.Run` instead of `Task.Factory.StartNew`)
- Optimize for when a local branch is ahead of its tracking remote branch
- Optimize for when a local branch was forked from another local branch with a remote
- Speed up search when the remote branch isn't obvious
- Add façade for `LibGit2Sharp.Repository` static methods 
- Add `Libgit2Sharp` tests for the `GetLatestPushedSha` method
- Search commits using `SortBy = Topological` (by default `Repository.Commits` order isn't deterministic)

### Latest Pushed Commit Algorithm

The previous implementation had a simple strategy.

1. If the current HEAD and remote branch point at the same commit, return the SHA for that
2. Search back through local commits for a commit that is reachable from any remote HEAD

https://github.com/github/VisualStudio/blob/53a2f892c15ceaddf01673687ba7e80c256591ac/src/GitHub.Exports/Services/GitService.cs#L95-L107

There were a number of issues.

1. The current HEAD and remote branch are only the same when a repository hasn't been touched
2. For large repositories, `ReachableFrom` could take a few seconds per commit searched
3. The commits returned by `Repository.Commits` aren't necessarily most recent first!
4. All of these combined meant that `GetLatestPushedSha` could easily take 20 seconds

The new implementation fixes these issues and implements a new search strategy:

1. `HEAD` is tracking a remote branch then simply return the `CommonAncestor` (this will be the case when a user has added commits to an existing branch)
2. If the branch has no remote, look for a local branch with a remote that it might have been forked from (this is the common case where a user has started working on a new fork nu hasn't pushed yet)
3. If the forked from branch can't be found, search all remote branches for the nearest `CommonAncestor` (this will be the case when the forked from branch has moved ahead or the local branch was created directory from a remote branch)

When searching for the nearest `CommonAncestor`, I've changed to using `ObjectDatabase.CalculateHistoryDivergence` instead of `Refs.ReachableFrom`. This needs to be called once per remote branch rather than once per commit. On the `dotnet/roslyn` repository I've been testing this on, a search that was taking 20 seconds now only takes a second.

https://github.com/github/VisualStudio/blob/7de9572d3b1181f854c2d8218bdf2fe0367904b0/src/GitHub.Exports/Services/GitService.cs#L110-L149

### How to test

Taken from https://github.com/github/VisualStudio/issues/1707#issuecomment-393330543.

1. Install current version of GitHub for Visual Studio
2. Clone https://github.com/dotnet/roslyn, open any file and click `GitHub > Open on GitHub` on any line. It's nice and fast.
3 Run following commands:
```
git checkout -b newBranch
git commit --allow-empty -m "hello 1"
git commit --allow-empty -m "hello 2"
git commit --allow-empty -m "hello 3"
git commit --allow-empty -m "hello 4"
git commit --allow-empty -m "hello 5"
```
4. Open any file and click `GitHub > Open on GitHub` on any line. It's super slow! 😭 
5. Install the version from this PR
4. Open any file and click `GitHub > Open on GitHub` on any line. It's nice and fast! 😄 

This will trigger search strategy 2. above (because `newBranch` doesn't have a remote tracking branch).

- To test the slower strategy 3, fork directly from a remote branch. This will stop it finding a local branch with a remote and force it to perform an exhaustive search. For example:
```
git checkout remotes/origin/Check64bit -b mybranch
```

`GitHub > Open on GitHub` should only take a second or so.

- Finally it should also cope with this repro https://github.com/github/VisualStudio/issues/1707#issuecomment-393682578 which previously was taking 6 minutes! 😱 

Fixes #1707


Neme12:
Thank you!! I just checked this out and the delays are indeed gone. :tada: No matter what I do, it never takes longer than 3 seconds.

src/GitHub.Exports/Services/IRepositoryFacade.cs
```diff
@@ -0,0 +1,13 @@
+﻿using LibGit2Sharp;
+
+namespace GitHub.Services
+{
+    /// <summary>
+    /// Facade for <see cref="LibGit2Sharp.Repository"/> static methods.
+    /// </summary>
+    public interface IRepositoryFacade
```
grokys: Do we need a new interface for these? Would they make sense in `IGitService`?

jcansdale: That was actually the problem I had. The methods I needed to mock were in `IGitService`. I wanted a pure facade for the static methods in `LibGit2Sharp.Repository`. I've hit the problem before, so thought having a `IRepositoryFacade` interface/service would be useful in other places.

grokys: Oh, I understand: you mean the methods you had to _test_ are in `IGitService`. Ok yeah I missed that, LGTM ;)

PR #1982, opened 2018-10-09
[wip] Fix cloning using Open From GitHub on Visual Studio 2015

### LINKED ISSUE #1981: File > Open > Open from GitHub silently fails in Visual Studio 2015
## Version

- GitHub Extension for Visual Studio version: v2.5.7.5090
- Visual Studio version: 2015

## What happened

**Steps to Reproduce**

1. Click on `File > Open > Open from GitHub `
2. Select a repository
3. Click on `Clone`
4. Clone dialog closes and nothing happens

**Expected behavior:** [What you expect to happen]

Expect repository to clone and `Team Explorer - Home` to appear.

jcansdale:
TODO

Fixes #1981


sguthals:
I just confirmed that this worked on VS2015 Community 👍 

Looks like there is still a build failing though.

src/GitHub.StartPage/StartPagePackage.cs
```diff
@@ -59,7 +57,6 @@ async Task<CodeContainer> RunAcquisition(IProgress<ServiceProgressData> download
             try
             {
                 var uiProvider = await Task.Run(() => Package.GetGlobalService(typeof(IGitHubServiceProvider)) as IGitHubServiceProvider);
-                await ShowTeamExplorerPage(uiProvider);
```
sguthals: So this was the piece that was waiting for the Team Explorer to let us know that the clone had completed?

src/GitHub.TeamFoundation.14/Services/VSGitServices.cs
```diff
@@ -39,6 +36,8 @@ public class VSGitServices : IVSGitServices
 
         [SuppressMessage("Microsoft.Performance", "CA1823:AvoidUnusedPrivateFields", Justification = "Used in VS2017")]
         readonly Lazy<IStatusBarNotificationService> statusBar;
+        [SuppressMessage("Microsoft.Performance", "CA1823:AvoidUnusedPrivateFields", Justification = "Used in VS2015")]
+        readonly Lazy<IVSServices> vsServices;
```
sguthals: Can you quickly explain this bit for my edification?

src/GitHub.TeamFoundation.14/Services/VSGitServices.cs
```diff
@@ -81,25 +83,64 @@ public string GetLocalClonePathFromGitProvider()
...
-#if TEAMEXPLORER14
-            var gitExt = serviceProvider.GetService<IGitRepositoriesExt>();
-            gitExt.Clone(cloneUrl, clonePath, recurseSubmodules ? CloneOptions.RecurseSubmodule : CloneOptions.None);
+            var teamExplorer = serviceProvider.TryGetService<ITeamExplorer>();
+            Assumes.Present(teamExplorer);
 
-            // The operation will have completed when CanClone goes false and then true again.
-            await gitExt.WhenAnyValue(x => x.CanClone).Where(x => !x).Take(1);
-            await gitExt.WhenAnyValue(x => x.CanClone).Where(x => x).Take(1);
+#if TEAMEXPLORER14
```
sguthals: So this is the new bit right? So instead of waiting for the response from TE, we're starting the clone and immediately navigating to the TE Home page and then the user will see the TE change as a result of the clone completing but TE is handling that. And we can identify that change (instead of actually waiting for an update from TE)

src/GitHub.TeamFoundation.14/Services/VSGitServices.cs
```diff
@@ -81,25 +83,64 @@ public string GetLocalClonePathFromGitProvider()
...
+            var gitExt = connectPage.GetService<IGitRepositoriesExt>();
+            Assumes.Present(gitExt);
+
+            gitExt.Clone(cloneUrl, clonePath, recurseSubmodules ? CloneOptions.RecurseSubmodule : CloneOptions.None);
+        }
+
+        static async Task WaitForCloneOnHomePageAsync(ITeamExplorer teamExplorer)
+        {
+            // The clone progress bar appears on the GettingStartedSection of the Home page,
+            // so we wait for this to be hidden before continuing.
```
sguthals: ✨

src/GitHub.StartPage/StartPagePackage.cs
jcansdale: This was a hack to make the `IGitRepositoriesExt` service available later on when the `Clone` method attempts to retrieve it from `IGitHubServiceProvider`.

The problem was that it required two pieces to be in place before it wold work.
1. The `Team Explorer - Connext` page must be visible
2. Something must have set the `GitServiceProvider` on `IGitHubServiceProvider`

This was way too fragile and the reason the bug crept in.

src/GitHub.TeamFoundation.14/Services/VSGitServices.cs
jcansdale: There is a `TryOpenRepository` method on `IVSServices` that we only need to call when running on Visual Studio 2015. On Visual Studio 2017, the cloned repository automatically opens in folder view (which is new in Visual Studio 2017).

If we didn't call `TryOpenRepository`, you wouldn't see the URL change on `Team Explorer - Home` and there'd be no option to open a solution. You'd need to manually find and open the cloned solution.

src/GitHub.TeamFoundation.14/Services/VSGitServices.cs
jcansdale: We can infer that the clone operation has completed when the progress bar is hidden and automatically change context to the newly cloned repository. If we didn't change repository context at this point, the user would know the clone had completed, but they'd be left to find and open a solution from the newly cloned folder (not a nice experience).

src/GitHub.TeamFoundation.14/Services/VSGitServices.cs
```diff
@@ -81,25 +83,68 @@ public string GetLocalClonePathFromGitProvider()
...
-            {
-                typedProgress.ProgressChanged += (s, e) => statusBar.Value.ShowMessage(e.ProgressText);
-                await gitExt.CloneAsync(cloneUrl, clonePath, recurseSubmodules, default(CancellationToken), typedProgress);
-            });
+            NavigateToHomePage(teamExplorer); // Show progress on Team Explorer - Home
+            await cloneTask;
 #endif
         }
 
+        static async Task StartClonenOnConnectPageAsync(
```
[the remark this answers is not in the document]
jcansdale: Oh, that's actually by design. Is so that I can test the alternative code path in Visual Studio 2017, before double checking that it works the same in 2015.

I think it will make testing easier in future if we light up new features in 2017, but don't exclude that is designed to run inside 2015. We could have an option to run inside 2017, but with the alternative code path used inside 2015.

PR #1984, opened 2018-10-09
Adding C# preferences to .editorconfig

### LINKED ISSUE #1750: Consider adding C# code style preferences to editorconfig
see https://github.com/github/VisualStudio/pull/1744#issuecomment-398171141

Neme12:
Fixes #1750

More information about each option can be found at https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference?view=vs-2017

This is mostly the default generated .NET editorconfig style in VS 2017.
image

I removed VB and tweaked a few options based on what I saw in the repo. Please let me know if you have any concerns or any option is unclear. I will add comments explaining them and asking whether the selected preference is correct.

Each option can be configured to a severity of one of 5 options as to how it is enforced and what is shown in the IDE if there's a violation:
- **none** - it's as if this feature was disabled (no lightbulb, there may still be an available refactoring though), except that the code VS generates will adhere to this style
- **silent** - there will be no squiggle in the editor, only the lightbulb, but all code that VS generates (for example by using features such as "GenerateMethod") will adhere to this style
- **suggestion** - there will be 3 small grey dots on the first 2 characters (+ a lightbulb), and this goes into the "Message" category in the error list
- **warning**
- **error**

Tagging @jcansdale


.editorconfig
```diff
@@ -1,6 +1,69 @@
...
 
-[*.cs]
+[*.{cs,vb}]
 indent_style = space
-indent_size = 4
\ No newline at end of file
+indent_size = 4
+
+# Organize usings
+dotnet_sort_system_directives_first = true
```
Neme12: When sorting using directives, everything inside `System` should come first. Is that correct?

.editorconfig
```diff
@@ -1,6 +1,69 @@
...
+indent_size = 4
+
+# Organize usings
+dotnet_sort_system_directives_first = true 
+
+# this. preferences
+dotnet_style_qualification_for_field = false:silent
+dotnet_style_qualification_for_property = false:silent
+dotnet_style_qualification_for_method = false:silent
+dotnet_style_qualification_for_event = false:silent
```
Neme12: Never use `this.` unless absolutely necessary (violation of this is silent).

.editorconfig
```diff
@@ -1,6 +1,69 @@
...
+
+# Language keywords vs BCL types preferences
+dotnet_style_predefined_type_for_locals_parameters_members = true:suggestion
+dotnet_style_predefined_type_for_member_access = true:suggestion
```
Neme12: Prefer `int.Parse` over `Int32.Parse`

.editorconfig
```diff
@@ -1,6 +1,69 @@
...
+
+# Parentheses preferences 
+dotnet_style_parentheses_in_arithmetic_binary_operators = always_for_clarity:silent 
+dotnet_style_parentheses_in_relational_binary_operators = always_for_clarity:silent 
+dotnet_style_parentheses_in_other_binary_operators = always_for_clarity:silent 
+dotnet_style_parentheses_in_other_operators = never_if_unnecessary:silent
```
Neme12: This is the default. I'm not completely sure how it works out in all the cases and combinations of operators, but it should never require anything since it is silent. The last option is for simple cases like `var x = (value)`.

.editorconfig
```diff
@@ -1,6 +1,69 @@
...
+
+# Modifier preferences
+dotnet_style_require_accessibility_modifiers = never:silent
```
Neme12: Don't use explicit default access modifiers (`private`, `internal`), prefer omitting them.

.editorconfig
```diff
@@ -1,6 +1,69 @@
...
+# Expression-level preferences
+dotnet_style_object_initializer = true:suggestion
+dotnet_style_collection_initializer = true:suggestion
+dotnet_style_explicit_tuple_names = true:suggestion
+dotnet_style_null_propagation = true:suggestion
+dotnet_style_coalesce_expression = true:suggestion
+dotnet_style_prefer_is_null_check_over_reference_equality_method = true:suggestion
+dotnet_style_prefer_auto_properties = true:silent
+dotnet_style_prefer_conditional_expression_over_assignment = true:silent
+dotnet_style_prefer_conditional_expression_over_return = true:silent
```
Neme12: These are mostly suggestions to use newer language features.
`dotnet_style_explicit_tuple_names` suggests using tuple.CustomName as opposed to tuple.Item1
`dotnet_style_null_propagation` suggests using ?.
`dotnet_style_coalesce_expression` suggests using ??
`dotnet_style_prefer_is_null_check_over_reference_equality_method` suggests using `is null` over `ReferenceEquals`
the last to are for using ?: over if statements

.editorconfig
```diff
@@ -1,6 +1,69 @@
...
+
+[*.cs]
+
+# var preferences
+csharp_style_var_for_built_in_types = true:none
+csharp_style_var_when_type_is_apparent = true:silent
+csharp_style_var_elsewhere = true:silent
```
Neme12: Prefer var everywhere except for builtin types (int, string...). Does this look good? Or use var everywhere?

.editorconfig
```diff
@@ -1,6 +1,69 @@
...
+
+# Expression-bodied members
+csharp_style_expression_bodied_methods = false:none
+csharp_style_expression_bodied_constructors = false:none
+csharp_style_expression_bodied_operators = false:none
+csharp_style_expression_bodied_properties = false:none
+csharp_style_expression_bodied_indexers = false:none
+csharp_style_expression_bodied_accessors = false:none
```
Neme12: I didn't see any expression bodied methods in this repo, so I put these as false.

.editorconfig
```diff
@@ -1,6 +1,69 @@
...
+
+# Pattern matching preferences
+csharp_style_pattern_matching_over_is_with_cast_check = true:suggestion
+csharp_style_pattern_matching_over_as_with_null_check = true:suggestion
```
Neme12: These suggest using pattern matching `(object is string s)`

.editorconfig
```diff
@@ -1,6 +1,69 @@
...
+
+# Null-checking preferences
+csharp_style_throw_expression = true:suggestion
+csharp_style_conditional_delegate_call = true:suggestion
```
Neme12: Prefer throw expressions (`this.field = parameter ?? throw new ArgumentNullException()`) and `eventHandler?.Invoke()` over `if (eventHandler != null) eventHandler();`

.editorconfig
```diff
@@ -1,6 +1,69 @@
...
+
+# Expression-level preferences
+csharp_prefer_braces = false:none
```
Neme12: I saw a lot of if statements without braces, so I'm putting as false

.editorconfig
```diff
@@ -1,6 +1,69 @@
...
+csharp_style_deconstructed_variable_declaration = true:suggestion
+csharp_prefer_simple_default_expression = true:suggestion
+csharp_style_inlined_variable_declaration = true:suggestion
```
Neme12: Prefer `default` over `default(T)` and `out var` over `out` + declaration outside.

.editorconfig
Neme12: There is no option for this, they always come last 😉

Neme12:
> I've loaded this PR branch but am struggling to see where these settings appear. 😕

Unfortunately, they're not shown anywhere in the UI and there isn't any way to find out if an option is invalid or isn't working, so it's hard to diagnose these at the moment 😕 

> For example, I tried changing the `dotnet_style_require_accessibility_modifiers` option to:
> 
> ```
> # Modifier preferences
> dotnet_style_require_accessibility_modifiers = always_for_clarity : warning
> ```

For `dotnet_style_require_accessibility_modifiers`, the value would be `always`, not `always_for_clarify` (that only works for parentheses). Each option can have a different set of values, which you can find here: https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference?view=vs-2017#normalize_modifiers

Also, it may sometimes be hard to see this refresh in Visual Studio unless you edit and save the .editorconfig from inside Visual Studio (otherwise it might require a restart). I recommend adding the .editorconfig to solution items and editing in VS, then you can see them refresh immediately.

Neme12:
Also, I think specifically for `dotnet_style_require_accessibility_modifiers`, a setting for "never" will not actually ask you to remove them (that might not be implemented, or it's an explicit design decision, I'm not sure). It's only "always" that will actually require you to add them.

.editorconfig
grokys: Yep, we prefer `var` everywhere I think.

.editorconfig
grokys: That's because we've only just moved to a version of C# which supports them. Moving forward, for new code (for me at least) the use of expression-bodied members depends on the length of the expression. Probably best to leave these as `false:none` in that case.

.editorconfig
grokys: Personally I would prefer to always use braces, but I'm not sure about anyone else: @StanleyGoldman @jcansdale @shana what do you think?

Neme12:
@grokys Sorry for the delay. Does the change look good?

.editorconfig
jcansdale: I tend to always uses braces, but realize this is maybe a little verbose for `if (foo) return;`.

jcansdale: I'm changing it to suggest using braces.

.editorconfig
```diff
@@ -63,7 +63,7 @@ csharp_style_throw_expression = true:suggestion
 csharp_style_conditional_delegate_call = true:suggestion
 
 # Expression-level preferences
-csharp_prefer_braces = false:none
+csharp_prefer_braces = false:suggestion
```
Neme12: @jcansdale Did you mean to make this `true:suggestion`?

.editorconfig
```diff
@@ -1,6 +1,72 @@
...
+csharp_style_conditional_delegate_call = true:suggestion
+
+# Expression-level preferences
+csharp_prefer_braces = false:suggestion
+csharp_style_deconstructed_variable_declaration = true:suggestion
+csharp_prefer_simple_default_expression = true:suggestion
+csharp_style_inlined_variable_declaration = true:suggestion
+
+# Spacing options
+csharp_space_after_keywords_in_control_flow_statements = true:warning
```
jcansdale: I was hoping this would warn in the case of `if(foo)`  rather than `if (foo)`. Doesn't seem to be working though. 😕

Neme12: These are formatting options. They should apply when you format your code but the code won't be flagged if it violates this rule. Fortunately that is going to change in VS 2019 with the addition of a formatting analyzer: https://github.com/dotnet/roslyn/pull/28873

Neme12: I think this should just be `true`, not `true:warning`. Specifying a severity is not supported here and it won't be supported in VS 2019 either. You will be able to set a severity for formatting violations in general, no matter which option they come from.

.editorconfig
jcansdale: I did, thanks for the catch!

.editorconfig
```diff
@@ -1,6 +1,73 @@
 # top-most EditorConfig file
 root = true
 
-[*.cs]
+[*.{cs,vb}]
```
meaghanlewis: Good addition!

PR #1989, opened 2018-10-11
#1727 messaging when solution is not open

ptoninato:
Updated values at line 574.  References bug #1727


grokys:
Thanks for the PR @ptoninato - unfortunately I think what we were intended is a little more involved than this: we should be showing a different message when no solution is opened to when a solution is opened, but we're not in a repository.

You can find some more discussion of this in https://github.com/github/VisualStudio/pull/1751.

StanleyGoldman:
Hey @ptoninato even though it's been a while, thanks for the contribution.

So i think what @grokys was saying is that #1751 is more in the direction of where we want to go.
I'm going to close this for now and go pick up the conversation there.

PR #1997, opened 2018-10-18
Various unit test fixes.

grokys:
Various unit test fixes:

- Applies https://github.com/editor-tools/ReactiveUI/pull/1 to our ReactiveUI submodule. This makes tests run a lot more stably (at least in ncrunch, should also improve other test runners too)
  - This means we can remove the `TestUtils.WithScheduler(Scheduler.CurrentThread)` hacks
- Allow running tests in parallel in ncrunch (greatly speeds up test time)
- Fix or ignore a load of CA warnings in unit test projects


test/TrackingCollectionTests/TrackingCollectionTests.cs
```diff
@@ -14,6 +14,7 @@
 using System.Threading.Tasks;
 using System.Reactive.Threading.Tasks;
 using GitHub;
+using System.Globalization;
 
 [TestFixture]
```
jcansdale: I'm not sure there is any value in continuing to run these flaky tests.

```suggestion
[TestFixture, Ignore("These tests are flaky and we're planning to move from using TrackingCollection")]
```

Issue #2007, opened 2018-10-23
Refactor LocalRepositoryModel, RemoteRepositoryModel, RepositoryModel and IGitService


jcansdale:
This group of classes and interfaces have a number of issues that make them difficult reason about and test.

### Some issues

-  `RepositoryModel`, `LocalRepositoryModel` and `RemoteRepositoryModel` aren't models at all
- They depend on and hold a reference to `IGitService`
- `LocalRepositoryModel` contains mutating methods like `Refresh`
- `LocalRepositoryModel` contains utility methods like `GenerateUrl`

### Plan of attack

- Move construction of `LocalRepositoryModel` into `IGitService`
- Convert `RepositoryModel` and `LocalRepositoryModel` into pure models
- Remove the interfaces (because interfaces aren't needed for [POCO](https://kent-boogaart.com/blog/view-models-pocos-versus-dependencyobjects) types)
- `GenerateUrl` is only used by `LinkCommand` so move that there as a private method

#### Should the models be mutable or immutable?
@grokys said:
> immutable is easier to reason about, but most of our other models are mutable
note that most of our other models are mutable only because the syntax for initializing mutable instances is a lot more readable than for immutable instances

@jcansdale said:
> i'd be inclined to keep them mutable for the moment, so we're refactoring less code initially

jcansdale:
Closed by #2028

PR #2008, opened 2018-10-23
Refactor repository models (repository refactor part 1)

jcansdale:
This is the first iteration of refactoring `LocalRepositoryModel` and `GitService` (see #2007).

I've focused on moving the code to where is should be, with as few functional changes as possible. There is more to do, but I think this makes sense as a chunk.

- Replace `ILocalRepositoryModel.CurrentBranch` property with `IGitService.CreateCurrentBranchModel(model)` method
- Move construction of `LocalRepositoryModel` to `GitService.CreateLocalRepositoryModel`
- Move `LocalRepositoryModel.GenerateUrl` to `LinkCommandBase`
- Move `LocalReposotoryModel.Refresh` to `GitService.RefreshCloneUrl` (this will be removed in #2025)
- Get rid of `LocalRepositoryModelFactory` (we can use `GitService` instead)
- Remove unused `LocalRepositoryModel.HeadSha` property

The most significant functional change was to replace the `LocalRepositoryModel.CurrentBranch` property with a `GitService.CreateCurrentBranchModel(repositoryModel)` method.

This was necessary because the `IGitExt.ActiveRepositories` property changed event fires before the current branch is actually changed. If we populate `LocalRepositoryModel.CurrentBranch` when the model is created, `CurrentBranch` ends up with information about the previous branch.

By deferring creation of the current branch model, we can populate it when the branch has actually changed and only create the branch model when it's actually required (the property change event can fire a lot).

### Bugs to fix

- [x] The `ActiveRepositoriesChanged` event fires _before_ the branch has changed on the local repository
https://github.com/github/VisualStudio/blob/47430fd92c1a7e17acd5dd04dc34d7628d5c6f0c/src/GitHub.Exports/Services/GitService.cs#L52

- [x] Tuples aren't comparable so this was always coming back `false`
https://github.com/github/VisualStudio/blob/47430fd92c1a7e17acd5dd04dc34d7628d5c6f0c/src/GitHub.App/Services/TeamExplorerContext.cs#L146

### Still to do

Now the junk has been removed from `LocalReposotoryModel`, we can decide how we want the models to look!

- [x] Where possible use `ITeamExploerContext` instead of `ITeamExplorerServiceHolder` (done in #2025)
- [ ] Refactor `BranchModel` to make it an actual model
- [ ] Change `RemoteRepositoryModel.OwnerAccount` to be a `ActorModel` instead of an `IAccount` (@grokys)
- [ ] Remove the model interfaces
- [ ] Remove the `Octicon` from the model https://github.com/github/VisualStudio/pull/2008#discussion_r228090302
- [ ] `BranchModel` is being passed a repository constructed from a URL rather then a local path
https://github.com/github/VisualStudio/blob/e3f811400f73aa722664c388253b76b24cf079c9/src/GitHub.Exports/Models/BranchModel.cs#L36
- [ ] `LocalRepositoryModel.GetHashCode` and `Equals` should use consistent comparisons
https://github.com/github/VisualStudio/blob/e3f811400f73aa722664c388253b76b24cf079c9/src/GitHub.Exports/Models/LocalRepositoryModel.cs#L36
https://github.com/github/VisualStudio/blob/e3f811400f73aa722664c388253b76b24cf079c9/src/GitHub.Exports/Models/LocalRepositoryModel.cs#L55
`OrdinalIgnoreCase` or `InvariantCultureIgnoreCase` maybe?


jcansdale:
I think updating the `Checkout ...` link state might have broken with this PR. Need to investigate....

image

src/GitHub.Exports/Models/RepositoryModel.cs
```diff
@@ -33,38 +31,22 @@ public class RepositoryModel : NotificationAwareObject, IRepositoryModel
...
         /// <summary>
         /// Gets the name of the repository.
         /// </summary>
-        public string Name { get; }
+        public string Name { get; set; }
 
         /// <summary>
         /// Gets the repository clone URL.
         /// </summary>
         public UriString CloneUrl
```
grokys: Future: should these properties be raising change notifications? If so, should other properties also be raising them (`CurrentBranch` for example).

src/GitHub.Exports/Models/RepositoryModel.cs
```diff
@@ -85,7 +67,7 @@ protected set
         public Octicon Icon
```
grokys: Future: move this out of the model: it's a view model-level reponsibility that only really applies to the repository list in Team Explorer.

src/GitHub.Exports/Models/BranchModel.cs
```diff
@@ -32,7 +32,8 @@ public BranchModel(LibGit2Sharp.Branch branch, IRepositoryModel repo, IGitServic
             Extensions.Guard.ArgumentNotNull(repo, nameof(repo));
             Name = DisplayName = branch.FriendlyName;
 #pragma warning disable 0618 // TODO: Replace `Branch.Remote` with `Repository.Network.Remotes[branch.RemoteName]`.
-            Repository = branch.IsRemote ? new LocalRepositoryModel(branch.Remote.Url, gitService) : repo;
+            // NOTE: This method expects a localPath not a URL!
+            Repository = branch.IsRemote ? gitService.CreateLocalRepositoryModel(branch.Remote.Url) : repo;
```
grokys: Should we also refactor `BranchModel` to make it a model here or in another PR?

src/GitHub.TeamFoundation.14/Services/LocalRepositoryModelFactory.cs
```diff
@@ -7,7 +7,7 @@ class LocalRepositoryModelFactory : ILocalRepositoryModelFactory
     {
         public ILocalRepositoryModel Create(string localPath)
         {
-            return new LocalRepositoryModel(localPath, GitService.GitServiceHelper);
+            return GitService.GitServiceHelper.CreateLocalRepositoryModel(localPath);
```
grokys: Future: this is now a part of `IGitService`, we can remove this factory.

src/GitHub.VisualStudio/Commands/LinkCommandBase.cs
```diff
@@ -46,7 +50,7 @@ protected ILocalRepositoryModel GetRepositoryByPath(string path)
                 if (!string.IsNullOrEmpty(path))
                 {
                     var repo = ServiceProvider.TryGetService<IGitService>().GetRepository(path);
-                    return new LocalRepositoryModel(repo.Info.WorkingDirectory.TrimEnd('\\'), GitService.GitServiceHelper);
+                    return GitService.GitServiceHelper.CreateLocalRepositoryModel(repo.Info.WorkingDirectory.TrimEnd('\\'));
```
grokys: Is `TrimEnd` needed here? If so, should that logic be put into `CreateLocalRepositoryModel`?

src/GitHub.Exports/Services/GitService.cs
```diff
@@ -21,6 +23,51 @@ public GitService(IRepositoryFacade repositoryFacade)
...
+        {
+            Guard.ArgumentNotNull(localPath, nameof(localPath));
+
+            var dir = new DirectoryInfo(localPath);
+            if (!dir.Exists)
+            {
+                throw new ArgumentException("Path does not exist", nameof(localPath));
+            }
+
+            var cloneUrl = GetUri(localPath);
```
jcansdale: Should this logic be part of RepositoryViewModel?

jcansdale: Should this logic be part of RepositoryViewModel?

jcansdale: Should this logic be part of RepositoryViewModel?

src/GitHub.Exports/Models/ILocalRepositoryModel.cs
```diff
@@ -18,22 +15,6 @@ public interface ILocalRepositoryModel : IRepositoryModel, INotifyPropertyChange
...
-        /// Generates a http(s) url to the repository in the remote server, optionally
-        /// pointing to a specific file and specific line range in it.
-        /// </summary>
-        /// <param name="linkType">The type of repository link to create</param>
-        /// <param name="path">The file to generate an url to. Optional.</param>
-        /// <param name="startLine">A specific line, or (if specifying the <paramref name="endLine"/> as well) the start of a range</param>
-        /// <param name="endLine">The end of a line range on the specified file.</param>
-        /// <returns>An UriString with the generated url, or null if the repository has no remote server configured or if it can't be found locally</returns>
-        Task<UriString> GenerateUrl(LinkType linkType, string path = null, int startLine = -1, int endLine = -1);
+        IBranch CurrentBranch { get; set; }
```
jcansdale: It looks like we can't construct the `CurrentBranch` as soon as the repository change event fires. I think it might make sense to remove this property and construct the current branch model when needed.

src/GitHub.Exports/Services/GitService.cs
```diff
@@ -21,6 +23,51 @@ public GitService(IRepositoryFacade repositoryFacade)
...
+            var name = cloneUrl?.RepositoryName ?? dir.Name;
+
+            var model = new LocalRepositoryModel
+            {
+                LocalPath = localPath,
+                CloneUrl = cloneUrl,
+                Name = name,
+                Icon = Octicon.repo
+            };
+
```
jcansdale: This is currently being refreshed before the branch has actually changed. :-(

src/GitHub.Exports/Models/BranchModel.cs
jcansdale: I'd prefer to attack `BranchModel` in another PR because it's largely independent of `LocalRepositoryModel`.

src/GitHub.Exports/Models/RepositoryModel.cs
jcansdale: I think we should aim to get rid of the `Refresh` command so this model would be immutable after creation.

jcansdale: I think we should aim to get rid of the `Refresh` command so this model would be immutable after creation.

jcansdale: Sorry about the spam.  I'm having another go....

jcansdale: Any another go.

src/GitHub.Exports/Services/GitService.cs
jcansdale: There is now a `CreateCurrentBranchModel` command instead of the `CurrentBranch` property. We'll still need to be careful about calling it too soon.

jcansdale: Another comment.

src/GitHub.Exports/Services/IGitService.cs
```diff
@@ -7,6 +7,19 @@ namespace GitHub.Services
...
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LocalRepositoryModel"/> class.
+        /// </summary>
+        /// <param name="localPath">The repository's local path.</param>
+        /// <returns>A repository model.</returns>
+        ILocalRepositoryModel CreateLocalRepositoryModel(string localPath);
+
+        /// <summary>
+        /// Updates the ClonePath information based on the local path
```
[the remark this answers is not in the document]
jcansdale: The `CurrentBranch` property no longer exists. I've renamed this method to `RefreshCloneUrl` to be explicit about what it does. I'm hoping it will go away completely in a future refactor.

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -128,33 +122,20 @@ async Task RefreshAsync()
                         log.Debug("ActiveRepository changed to {CloneUrl} @ {Path}", repo?.CloneUrl, newRepositoryPath);
                         ActiveRepository = repo;
                     }
-                    else if (newBranchName != branchName)
-                    {
-                        log.Debug("Fire StatusChanged event when BranchName changes for ActiveRepository");
-                        StatusChanged?.Invoke(this, EventArgs.Empty);
-                    }
-                    else if (newHeadSha != headSha)
```
grokys: So StatusChanged is no longer firing when HEAD changes? Is this because it wasn't working before anyway? (i.e. it was firing before the actual change)

src/GitHub.TeamFoundation.14/Services/LocalRepositoryModelFactory.cs
jcansdale: Done. 👍

src/GitHub.Exports/Models/ILocalRepositoryModel.cs
jcansdale: Creating the current branch model is now the responsibility of `IGitService` and no longer a property on the repository model.

src/GitHub.App/Services/TeamExplorerContext.cs
```diff
@@ -128,33 +122,20 @@ async Task RefreshAsync()
...
                     {
-                        log.Debug("Fire StatusChanged event when TrackedSha changes for ActiveRepository");
+                        log.Debug("Fire StatusChanged event when PullRequest changes for ActiveRepository");
                         StatusChanged?.Invoke(this, EventArgs.Empty);
                     }
-                    else if (newPullRequest != pullRequest)
+                    else if (newRepositoryPath != null)
                     {
-                        log.Debug("Fire StatusChanged event when PullRequest changes for ActiveRepository");
+                        log.Debug("Fire StatusChanged event when on a repository and anything changes");
```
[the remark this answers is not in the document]
jcansdale: I'm not sure it that's a feature? 😉

I'll change it to:
"Fire StatusChanged event if anything about an active repository has changed"

src/GitHub.Exports/Services/IGitService.cs
```diff
@@ -7,6 +7,26 @@ namespace GitHub.Services
...
+        /// </summary>
+        /// <param name="localPath">The repository's local path.</param>
+        /// <returns>A repository model.</returns>
+        ILocalRepositoryModel CreateLocalRepositoryModel(string localPath);
+
+        /// <summary>
+        /// Creates a new branch model for the current branch.
+        /// </summary>
+        /// <param name="model">The <see cref="ILocalRepositoryModel" /> to create a current branch model for.</param>
+        /// <returns>A new branch model.</returns>
```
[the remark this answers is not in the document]
jcansdale: We could change it to `GetBranch(string name = "HEAD")`. Do you think that would make sense? Otherwise we'd probably need to qualify it with `GetHeadBranch` or something?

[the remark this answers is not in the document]
grokys: Yeah I think that would make sense.

jcansdale:
Merged as part of #2028.

Issue #1010, opened 2017-06-13
Checked out PR file diffs sometimes complain about different encodings


jcansdale:
- GitHub Extension for Visual Studio version: 2.2.1.100 (dev)
Visual Studio version: 2015

1. Select a PR
2. Checkout PR
3. Double-click on PR file and open diff view

Sometimes the diff view complains that the files have different encodings (UTF-8 vs Windows).

![image](https://user-images.githubusercontent.com/11719160/27085420-19f04166-5047-11e7-8498-0261a12ce72b.png)

jcansdale:
It appears Visual Studio likes to add UTF-8 headers to files it touches. 😕 This seems to change them from being the default Western European (Windows-1252) to being UTF-8.

When creating a temp file to diff against, we should save it as UTF-8, if the working file has been changed to UTF-8.

jcansdale:
Fixed in #1011

PR #2009, opened 2018-10-23
Don't change file encoding unnecessarily.

### LINKED ISSUE #1908: Unicode error can occur when viewing diff for PR files
1. Checkout a different branch than PR #1888 as an example
2. Right click and "View Changes" of any `zh-CN.resx` file.
3. Observe error

**_Note:_** If you are actually on the branch this error does not occur.

### Images

image

image

grokys:
When comparing files in a PR. we're currently taking a look at the file on disk and converting the encoding of the LHS of the comparison to match the on-disk file. However, if the file doesn't exist locally, we were returning `Encoding.Default` for this value, which was causing the unicode errors seen in #1908.

This PR changes to behavior to return `null` if the file doesn't exist locally, and if a `null` encoding is passed to `PullRequestService.ExtractToTempFile` then just write the contents out without any encoding.

### Related

- #1010: Checked out PR file diffs sometimes complain about different encodings

Fixes #1908.


meaghanlewis:
This LGTM 👍

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -769,7 +769,7 @@ public Encoding GetEncoding(ILocalRepositoryModel repository, string relativePat
                 }
             }
 
-            return Encoding.Default;
+            return null;
```
jcansdale: I believe can we can simply replace this with the following and avoid having to spacial case `null`.

```suggestion
            return new UTF8Encoding(false);
```

This makes the `GetEncoding` responsible for returning `Encoding.UTF8` if the target file has a preamble or `new UTF8Encoding(false)` if the target file doesn't exist or have a preamble. UTF8 without preamble is Git's default text encoding.

It's possible we could always use `new UTF8Encoding(false)` and avoid this issue completely. The question is what a diff looks like between UTF8 files with and without preamble. I didn't know `new UTF8Encoding(false)` was a thing until today.

jcansdale: Re: wondering if we could always use `new UTF8Encoding(false)`. Alas, this doesn't work. I tried the following:

```
File.WriteAllText("foo.txt", "foobar", new UTF8Encoding(true));
File.WriteAllText("bar.txt", "foobar", new UTF8Encoding(false));
```

When I diff them together, I see:
image

StanleyGoldman: This should change

Issue #1859, opened 2018-08-10
Corrupt extension cache and missing MEF exports


jcansdale:
- GitHub Extension for Visual Studio version: 2.5.4.3349 (probably any)
- Visual Studio version: 15.6.6, 15.7.5, 15.7.3, 15.7.5, 15.7.5, 15.7.6, 5.8 Preview 5, 5.8 Preview 4

We have been seeing many issues that involve missing MEF components, for example:

- #1826 GitHub Extension broken with VS 15.7.6 
- #1810 SetSite failed for the package [GitHubPackage], VS 15.8 Preview 4/5 
- #1812 Error On Project Load 

### Symptoms

These tend to manifest themselves as follows on the GitHub pane:
```
System.InvalidOperationException: Could not locate any instances of contract GitHub.Services.ITeamExplorerServiceHolder.
   at GitHub.VisualStudio.GitHubServiceProvider.GetService(Type serviceType) in C:\projects\visualstudio\src\GitHub.VisualStudio\Services\GitHubServiceProvider.cs:line 147
   at GitHub.VisualStudio.GitHubServiceProvider.GetService[T]() in C:\projects\visualstudio\src\GitHub.VisualStudio\Services\GitHubServiceProvider.cs:line 153
   at GitHub.VisualStudio.UI.GitHubPane.<InitializeAsync>d__12.MoveNext() in C:\projects\visualstudio\src\GitHub.VisualStudio\UI\GitHubPane.cs:line 98
```

Or like this when opening a solution that's hosted on GitHub:

![image](https://user-images.githubusercontent.com/11719160/43982975-2e10d5ce-9cf0-11e8-837b-9c9c562aa1d1.png)

With an entry in the ActivityLog:
```xml
<description>
SetSite failed for package [GitHubPackage]
  Source: 'System.ComponentModel.Composition'
  Description: No exports were found that match the constraint:
 ContractName=GitHub.Commands.IOpenFromUrlCommand
 RequiredTypeIdentity=GitHub.Commands.IOpenFromUrlCommand
</description>
```

### What's going on

*Update from Daniel Griffen @dgriffen on [Gitter](https://gitter.im/Microsoft/extendvs?at=5b7739225921214124814450).*

> @jcansdale I was able to reproduce this locally! I've figured out the issue and am working on how to fix it on our end or provide you with a workaround.
> @jcansdale there is a slight issue where the target version is getting the incorrect version of the running VS process and therefore excluding both your teamfoundation mef assets.
[...]

PR #2014, opened 2018-10-26
Fix for Visual Studio 2019 Preview

jcansdale:
This PR does the required changes to make the extension play nicely with Visual Studio 2019 Preview.

- Add Visual Studio 2019 compatible `GitHub.TeamFoundation.16` project (`Team Explorer` extension)
- Allow `IGitActionsExt.CloneAsync` to be called from Visual Studio 2019 (required a little reflection)
- Use `ITeamExplorerServices.OpenRepository` rather than `IVSServices.TryOpenRepository`
- Add `Microsoft.TeamFoundation.*` v16 assemblies to `lib` folder
- Pass `IServiceProvider` rather than `IAsyncServiceProvider` to `VSGitExt` (avoiding type name collisions)
- Commented out `GraphsNavigationItemTests`, which were previously ignored but now not compiling

### How to test

This PR has touched the code responsible for cloning a repository and for opening an already cloned repository. We should check this functionality on Visual Studio 2015, 2017 and 2019.

#### Check Cloning a Repository

1. Click `File > Open > Open from GitHub...`
2. Select (or paste) a repository and `Clone`
3. Check `Team Explorer - Home` appear with clone process
4. Check name of repository appears on banner when clone completed
5. Check `Solution Explorer` pane has target repository folder open

#### Check Opening a Repository

1. Click `Team > Manager Connections...`
2. Make sure you're logged into GitHub or GHE
3. Double click on a repository under `GitHub` / GHE
4. Check `Team Explorer - Home` opens with target repository
5. Check `Solution Explorer` pane
  - Expect this to be empty on Visual Studio 2015
  - Expect this to contain target repository folder on Visual Studio 2017 and 2019

### Possible issues

I believe the issue from #1859 is back in Visual Studio 2019 Preview, but I've heard this should be fixed in the public release. This shouldn't happen every time you install, but it's something to watch out for!

- #1859: Corrupt extension cache and missing MEF exports


src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
```diff
@@ -133,10 +133,9 @@ public void JoinTillEmpty()
         async Task<T> GetServiceAsync<T>()
         {
             await JoinableTaskFactory.SwitchToMainThreadAsync();
-            return (T)await asyncServiceProvider.GetServiceAsync(typeof(T));
+            return (T)serviceProvider.GetService(typeof(T));
```
jcansdale: This will block the UI thread. Should get `IAsyncServiceProvider` and then use `GetServiceAsync`. That way we can pass in `IServiceProvider` but still use `IAsyncServiceProvider`.

src/GitHub.VisualStudio/Services/VSGitExtFactory.cs
```diff
@@ -15,23 +16,24 @@ public class VSGitExtFactory
...
-                    return Create(() => new VSGitExt14(asyncServiceProvider));
+                    return Create(() => new VSGitExt14(serviceProvider));
                 case 15:
+                    return Create(() => new VSGitExt15(serviceProvider));
                 case 16:
-                    return Create(() => new VSGitExt15(asyncServiceProvider));
+                    return Create(() => new VSGitExt16(serviceProvider));
                 default:
                     log.Error("There is no IVSGitExt implementation for DTE version {Version}", vsVersion);
                     return null;
```
jcansdale: Returning `null` for a MEF service can cause unexpected behavior and isn't a good idea. We should throw here instead.

src/GitHub.TeamFoundation.14/Services/VSGitExt.cs
jcansdale: On second thoughts, we're waiting until `IGitExt` is available before retrieving it. Using the non-async `GetService` shouldn't be a problem.

jcansdale:
Double clicking a repository on the connect list brings up the `Getting started...` dialog.

image

It seems closing a solution caused the dialog to appear. This _should_ be easy enough to fix.

meaghanlewis:
This looks great to me! Tested in VS2015, 2017 and 2019

PR #2017, opened 2018-10-30
PR conversation view

grokys:
Work in progress for attn of @donokuda.

~~To show the conversation view, right-click on an item in the PR list and select "Open Conversation" (note: this is just a temporary method to open the view, see Open Questions below)~~

## Progress so far

# Open Questions

- I've added a button to close/reopen the pull request, but closing/reopening doesn't reflect the changes yet in the GitHub pane, and doesn't add/remove the entry from the PR list
  - Syncing with other parts of the UI will take a bit more work
  - Given that we don't have "Merge" yet shall we punt this?
- We need to decide what to do with the PR description in the GitHub pane, and how to navigate to the conversation view from there
  - Remove the description from the GitHub pane? Display a summary with a link to the conversation view (summary being e.g. just the first paragraph?)


jcansdale:
Something I really miss is emoji support. Could we have a button or even just a reminder to click `Windows Key + .` to bring up the Windows 10 emoji picker? 😜 

image

jaredpar:
> Maybe @jaredpar would give us a sanity check/tips on how to create a read-only editor component like this

Not quite sure what you were looking for advice on here. Could you elaborate?

jcansdale:
@jaredpar,

> Not quite sure what you were looking for advice on here. Could you elaborate?

I remembered you're wrote the https://github.com/jaredpar/EditorUtils library for hosting the Visual Studio editor outside of VS. What I'm wondering about here is a potentially simpler scenario (hosting it inside VS).

At the moment we're using Markdig for rendering .md content inside Visual Studio. Unfortunately it doesn't have built in syntax highlighting.

I was wondering how feasible it would be to create a read-only WPF editor control with syntax highlighting, that corresponds to a markdown code block? We could then render any code blocks in pull request/issue in a similar way to Visual Studio. The highlighting and color themes would be the same as Visual Studio so it wouldn't look out of place.

Is this realistic or do you think there are performance/complexity issues that are likely to bite us?

jaredpar:
> What I'm wondering about here is a potentially simpler scenario (hosting it inside VS).

Yes that is feasible. Within VS the editor is just there. You can grab all of the factory interfaces from the MEF container and just create the `IWpfTextView` instances directly.

grokys:
r

donokuda:
I went and did an updated mock on how comments could be represented in the conversation view and wanted to get feedback.

**Before**
before


**After** (This is a static mock to communicate intention and not a screenshot of live code)
after

The biggest goal here is to align, both visually and functionally, the conversation page in Visual Studio with what is on dotcom. My base assumption here is that our users expect a similar experience in Visual Studio as they would on dotcom.

### Updating the styling for the comments

I think the biggest difference here is updating so that a pull request's original post its subsequent comments share the same styling:

main-comment

I think we're also currently using the inline comment view for the regular conversation comments. Would it be possible to share the same template as the original comment?

### Styling the header for added commits

There's a small tweak for the "added some commits" header where the octicon is enclosed in a grey circle (to match the styling on dotcom)

groupd events

Something I noticed on dotcom is that they only add the header if there more than one added commit. Currently we add the header to every "added commits" event, even if it's just one. Would it be possible to introduce some sort of logic where we show the "[user] added commits" header only when there's more than one commit?

### Updating the comment form

It looks like we use the comment form with a collapsed / expanded reply field (and "Start a review" button.) Would it make sense to have a different form template for comments here like this?

comment me

### Supporting reviews and other timeline events

There are a handful of missing [timeline events](https://developer.github.com/v4/union/pullrequesttimelineitems/) that we should consider supporting, but it might make sense to also track and tackle these in separate issues/pull requests.

One type event that is missing is showing reviews in the conversation timeline.

I went ahead and took a pass at designing what that could be like:

review with comment

([Example of a review without a summary](https://user-images.githubusercontent.com/1174461/51656579-85908400-1f56-11e9-9af9-3b5200e0ca39.png))

🤔I'm wondering if it's possible to share the same comment view between here and the inline diff comment, primarily because they share the same layout except for the bar to the left.

comments comments comments

We should consider splitting this discussion off to a different issue since there are a handful of pieces outside of the conversation view pertaining to this.

src/GitHub.App/ViewModels/IssueishViewModel.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        /// <inheritdoc/>
+        public Uri WebUrl
+        {
+            get { return webUrl; }
+            private set { this.RaiseAndSetIfChanged(ref webUrl, value); }
+        }
+
+        /// <inheritdoc/>
+        public ReactiveCommand<Unit, Unit> OpenOnGitHub { get; }
```
StanleyGoldman: The `OpenOnGitHub` `ReactiveCommand` is never assigned to.

meaghanlewis:
Is it possible to linkify references to issues and PR's in the conversation view?

screen shot 2019-02-11 at 12 02 11 pm

grokys:
@meaghanlewis yes, this is definitely a feature we should have! I did look into implementing it a while ago, but never finished it. It would involve writing a markdig extension, which shouldn't be too difficult; would just require the time to be allocated to it.

donokuda:
> One thing I noticed while testing. I'm not sure what the correct functionality should be here, Don says he expected to open the browser, I expected it to open the Detail page.

Floating this idea:

giphy

If I click on a link, I'd mostly expect the page that the action took place to update instead of something on the side. However, I probably expect that since the page itself looks and feels like closer to a webpage.

We could consider doing both: navigate the conversation view to the corresponding pull request _and also_ navigate the GitHub panel to show the details view with relevant information to the PR.

There are a few risks that I anticipate (which we would want to consider validating with some folks): 😬 

1. When I'm focusing on a specific pull request, it's confusing to have the panel navigate without me interacting with it. If I'm not aware that the panel has updated, I might be disoriented or accidentally real irrelevant information.
2. It's not clear how to get back to the previous screen in the pane. (The solution here would be to use the [navigation bar above](https://user-images.githubusercontent.com/1174461/52596203-9c632180-2e04-11e9-95e1-a3e0930f20f5.png), so this would mostly be testing if that functionality is discoverable.) 
3. People don't expect the conversation page to change. 👈I doubt this is a problem, but I also understand that my hunch could be wrong. 

I am leaning towards navigating both the conversation page and the GitHub panel to the pull request, but I'm open to feedback in case I'm missing something. 😄

jcansdale:
@donokuda I've been thinking about how I use the current `View on GitHub` link.  I pretty much always use it after creating a PR to edit the markdown description and to add images. This is something the current conversation view doesn't allow and it would take a lot effort to get feature parity with dotcom. I think this is a pretty important use-case/workflow.

How about rather than replacing or combining the `View on GitHub` / `View Conversation` functionality, we keep the `View on GitHub` link and take inspiration from the comments icon on the dotcom PR list.

image

I think showing the number of comments gives people an incentive and invites them to click. It would also be clear that clicking on the comment link would show those comments.

I think we've always been pretty explicit about when a command will open an external browser (for both menus and links). I'd be hesitant to introduce a command that both opens an external browser and does an action inside Visual Studio.

image

In fact we should have a common command/keyboard shortcut to `Open on GitHub` no matter the context (if we don't already). 🤔

donokuda:
@jcansdale 👍 Sure, let's try that idea out! 😄

My only concern is if the text isn't blue (like the links around it), then folks might assume they cannot interact with it (and that it's only informational.)

Would you be open to making it a blue link?

jcansdale:
> Would you be open to making it a blue link?

Absolutely! The main thing was the comment icon/number of comments. I just copy/pasted from dotcom. 😉

jcansdale:
@StanleyGoldman,

> One thing I noticed while testing. I'm not sure what the correct functionality should be here, Don says he expected to open the browser, I expected it to open the Detail page.

Is this the link you're referring to above?

image

I'd expect the `#` link on the conversation view to open in a browser, just like the link in the the notification does on the left. In fact, could we include the owner/repo in the conversation view link as well to add a little more context?

PR #2019, opened 2018-10-31
Updating to the latest Octokit.GraphQL

StanleyGoldman:


src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -90,7 +90,7 @@ public class PullRequestService : IPullRequestService
                 if (readPullRequests == null)
                 {
                     readPullRequests = new Query()
-                          .Repository(Var(nameof(owner)), Var(nameof(name)))
+                          .Repository(Var(nameof(name)), Var(nameof(owner)))
```
grokys: Rather than simply switching the order of these args, I'd prefer to use named arguments where possible.

StanleyGoldman: Fixed.

PR #2024, opened 2018-11-01
Package Newtonsoft.Json.9.0.1 from OutputPath

jcansdale:
Copy Newtonsoft.Json.9.0.1 to the OutputPath and package from there,
rather than referencing the one in `..\..\packages`.


jcansdale:
Nope.

PR #2025, opened 2018-11-01
Remove repository responsibilities from TeamExplorerServiceHolder (repository refactor part 2)

jcansdale:
- Depends on #2008

### What this PR does

`TeamExplorerServiceHolder` has picked up responsibilities beyond holding services for the Team Explorer views. This PR removes its repository watching and event marshaling responsibilities and delegates them to `ITeamExplorerContext`. It also removes the need for refreshing a live `LocalRepositoryModel`, meaning the model can now be immutable after creation.

- Remove `ActiveRepo`, `Subscribe` and `Unsubscribe` from `ITeamExplorerServiceHolder`
- Remove `RefreshCloneUrl` from `IGitService`
- Expose `TeamExplorerContext` and `JoinableTaskFactory` from `ITeamExplorerServiceHolder` (for repo change events and marshaling)

### How to test

I've changed code that listens for repository change and refreshes the Team Explorer UI. We need to test moving between repositories and changing the clone URL.

1. Open a repository that it on GitHub
2. Open the `Team Explorer - Home` view
3. The repository URL should be visible
4. From the command line: `git remote rename _origin origin`
5. The repository URL and banner should no longer appear
6. From the command line: `git remote rename origin _origin`
7. The repository URL and banner should re-appear

### DIFF
--- a/src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
+++ b/src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
@@ -28,6 +28,8 @@ namespace GitHub.VisualStudio.TeamExplorer.Connect
     public class GitHubConnectSection : TeamExplorerSectionBase, IGitHubConnectSection
     {
         static readonly ILogger log = LogManager.ForContext<GitHubConnectSection>();
+        readonly ISimpleApiClientFactory apiFactory;
+        readonly ITeamExplorerServiceHolder holder;
         readonly IPackageSettings packageSettings;
         readonly ITeamExplorerServices teamExplorerServices;
         readonly int sectionIndex;
@@ -101,8 +103,6 @@ namespace GitHub.VisualStudio.TeamExplorer.Connect
 
         public ICommand Clone { get; }
 
-        internal ITeamExplorerServiceHolder Holder => holder;
-
         public GitHubConnectSection(IGitHubServiceProvider serviceProvider,
             ISimpleApiClientFactory apiFactory,
             ITeamExplorerServiceHolder holder,
@@ -127,6 +127,8 @@ namespace GitHub.VisualStudio.TeamExplorer.Connect
             IsVisible = false;
             sectionIndex = index;
 
+            this.apiFactory = apiFactory;
+            this.holder = holder;
             this.packageSettings = packageSettings;
             this.teamExplorerServices = teamExplorerServices;
             this.localRepositories = localRepositories;
@@ -325,7 +327,7 @@ namespace GitHub.VisualStudio.TeamExplorer.Connect
                     try
                     {
                         // TODO: Cache the icon state.
-                        var api = await ApiFactory.Create(newrepo.CloneUrl);
+                        var api = await apiFactory.Create(newrepo.CloneUrl);
                         var repo = await api.GetRepository();
                         newrepo.SetIcon(repo.Private, repo.Fork);
                     }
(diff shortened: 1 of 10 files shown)


grokys:
One thing I noticed: when I change the name of the `origin` remote I get an exception at this line:

https://github.com/github/VisualStudio/blob/2ce3e8480de4a38dbedd5146c27979163a7a4a4a/src/GitHub.InlineReviews/Services/PullRequestSessionManager.cs#L241

We should check to see if `repository.CloneUrl` is null there. Should that be done here or in a separate PR do you think?

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
```diff
@@ -344,13 +346,13 @@ async void UpdateRepositoryList(object sender, NotifyCollectionChangedEventArgs
                         .Cast<ILocalRepositoryModel>()
                         .ForEach(async r =>
                     {
-                        if (Equals(Holder.ActiveRepo, r))
+                        if (Equals(holder.TeamExplorerContext.ActiveRepository, r))
```
grokys: Is this comparison still valid? Previously the repository models overrode `Equals`, but now they don't. I think we need to decide what "equals" means in this case? Same `CloneUri` and `LocalPath`?

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
```diff
@@ -457,7 +459,7 @@ public void Retry()
 
         public bool OpenRepository()
         {
-            var old = Repositories.FirstOrDefault(x => x.Equals(Holder.ActiveRepo));
+            var old = Repositories.FirstOrDefault(x => x.Equals(holder.TeamExplorerContext.ActiveRepository));
```
grokys: Same here.

src/GitHub.VisualStudio.UI/Base/TeamExplorerItemBase.cs
```diff
@@ -75,21 +74,51 @@ public virtual void Invalidate()
...
+            {
+                await holder.JoinableTaskFactory.SwitchToMainThreadAsync();
+                UpdateRepo(repo);
+            }).Task.Forget();
         }
 
         void Unsubscribe()
         {
-            holder.Unsubscribe(this);
+            //holder.Unsubscribe(this);
```
grokys: Commented-out code.

src/GitHub.TeamFoundation.14/Connect/GitHubConnectSection.cs
grokys: Sorry, the comparison logic is still in there for now. Ignore this.

src/GitHub.VisualStudio.UI/Base/TeamExplorerItemBase.cs
jcansdale: Fixed in #2028 here https://github.com/github/VisualStudio/pull/2028/commits/bf4c2f9026d69b23986ea2a12660cf96746d5180.

jcansdale:
@grokys,

> We should check to see if `repository.CloneUrl` is null there. Should that be done here or in a separate PR do you think?

I'd be inclined to look at that in a separate PR. We need to handle the situation when there is no `origin ` remote better. I'm going to show a callout notification when there is no origin remote.  I'll look at this as the same time.

jcansdale:
Merged as part of #2028.

PR #2026, opened 2018-11-01
Warn when no origin remote

### LINKED ISSUE #1729: Repository not recognized as being from GitHub if there no remote named "origin"
- GitHub Extension for Visual Studio version: 2.5.3
- Visual Studio version: All

1. Rename the `origin` remote
```
git remote rename origin destination
```
2. Open solution from repository in Visual Studio

3. `GitHub` section doesn't appear on `Team Explorer - Home`
image

4. `GitHub` submenu doesn't appear on code context menu
image

jcansdale:
### What this PR does

If a repository doesn't have a remote named `origin`, the extension doesn't know which remote to use or even if it's dealing with a GitHub repository. This PR shows a call-out notification if a repository has remotes but none of them are named `origin`. If the used clicks on the notification title, the `GitHub` pane will open with an option to edit their remotes.

- Add `TippingService` for displaying call-out notifications on Visual Studio 2015, 2017 and 2019
- Display `NoRemoteOriginView` on `GitHub` pane when there are no remotes named `origin`
- Add `ShowRepositorySettingsRemotesAsync` to `ITeamExplorerServices`
- Prevent exception in `GetSessionInternal` when repository `CloneUrl` changes to null
- Add integration tests for `GitService` (that use a real repository)
- Add `HasRemotesButNoOrigin` property to `LocalRepositoryModel`
- Add `CantFindGitHubUrlForRepository` and `RepositoriesMustHaveRemoteOrigin` to `Resources`
   - These contain text for the notification and the pane view
- Add `TrackedRemoteName` to `BranchModel`
  - This property has tests but isn't currently used. I thought `NoRemoteOriginView` might suggest renaming the currently active remote (if any) to `origin`?

### Reviewers

- Is the wording of the call-out / pane view clear enough?

### What to expect

When a repository has at least one remote, but no remotes called "origin", the following call-out notification will appear.

image

Clicking the bonded title will open the `GitHub` pane, which will show the following message. Alternatively, clicking `Always ignore` will cause the call-out notification to _never_ show again.

image

Clicking the `Edit Remotes` button will open the `Repository Settings` page with the `Remotes` section visible. This lets the user rename one of their remotes to `origin`.

image

### To do

- [ ] Expand text on GitHub pane to guide user towards fix
- [ ] Add metrics

### How to test

1. Open solution from a repository on GitHub
2. From command line type
  `git remote rename origin porigin`
3. Expect call-out notification to appear
4. Click on title of call-out notification
5. Expect `GitHub` pane to appear with no "origin" message
6. Click on the `Edit Remotes` button
7. Expect `Repository Settings` page with the `Remotes` section visible
8. Edit the `porigin` rename and rename it to `origin`
9. Expect PR list to appear on `GitHub` pane

### Related

- #1727 Update the messaging in the GitHub panel when a solution is not open

Fixes #1729


jcansdale:
Exception thrown when renaming `origin` remote.
image

src/GitHub.App/ViewModels/GitHubPane/NoRemoteOriginViewModel.cs
```diff
@@ -0,0 +1,30 @@
...
+    [PartCreationPolicy(CreationPolicy.NonShared)]
+    public class NoRemoteOriginViewModel : PanePageViewModelBase, INoRemoteOriginViewModel
+    {
+        ITeamExplorerServices teamExplorerServices;
+
+        [ImportingConstructor]
+        public NoRemoteOriginViewModel(ITeamExplorerServices teamExplorerServices)
+        {
+            this.teamExplorerServices = teamExplorerServices;
+            EditRemotes = ReactiveCommand.CreateFromTask(OnEditRemotesAsync);
```
jnm2: Just curious, why not `ReactiveCommand.CreateFromTask(teamExplorerServices.ShowRepositorySettingsRemotesAsync)`?

src/GitHub.InlineReviews/Services/PullRequestStatusBarManager.cs
```diff
@@ -82,17 +86,42 @@ public void StartShowingStatus()
...
+            if (repository.HasRemotesButNoOrigin)
+            {
+                NoRemoteOriginCallout();
+            }
+
             var showStatus = await IsDotComOrEnterpriseRepository(repository);
             if (!showStatus)
             {
-                ShowStatus(null);
+                var view = ShowStatus(null);
```
jnm2: Unused variable?

src/GitHub.InlineReviews/Services/PullRequestStatusBarManager.cs
```diff
@@ -82,17 +86,42 @@ public void StartShowingStatus()
...
+                log.Warning("Couldn't find SccStatusBar");
+                return;
+            }
+
+            var calloutId = Guids.NoRemoteOriginCalloutId;
+            var title = Resources.CantFindGitHubUrlForRepository;
+            var message = Resources.RepositoriesMustHaveRemoteOrigin;
+            var isDismissable = true;
+            var commandSet = Guids.guidGitHubCmdSet;
+            var commandId = (uint)PkgCmdIDList.showGitHubPaneCommand;
```
jnm2: Would named args be easier?

src/GitHub.Services.Vssdk/Services/TippingService.cs
```diff
@@ -0,0 +1,96 @@
...
+using IServiceProvider = System.IServiceProvider;
+
+namespace GitHub.Services.Vssdk.Services
+{
+    /// <summary>
+    /// This service is a thin wrapper around <see cref="Microsoft.Internal.VisualStudio.Shell.Interop.IVsTippingService"/>.
+    /// </summary>
+    /// <remarks>
+    /// The <see cref="IVsTippingService"/> interface is public, but contained within the 'Microsoft.VisualStudio.Shell.UI.Internal' assembly.
+    /// To avoid a direct dependency on 'Microsoft.VisualStudio.Shell.UI.Internal', we use reflection to call this service.
```
jnm2: I'm curious about this. Since you have an *effective* dependency on `Microsoft.VisualStudio.Shell.UI.Internal`, what is the downside to expressing that via direct dependency?

src/GitHub.Services.Vssdk/Services/TippingService.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <inheritdoc/>
+        public void RequestCalloutDisplay(Guid calloutId, string title, string message,
+            bool isPermanentlyDismissible, FrameworkElement targetElement,
+            Guid vsCommandGroupId, uint vsCommandId)
+        {
+            var screenPoint = !Splat.ModeDetector.InUnitTestRunner() ?
+                    targetElement.PointToScreen(new Point(targetElement.ActualWidth / 2, 0)) : default;
+            var point = new Microsoft.VisualStudio.OLE.Interop.POINT { x = (int)screenPoint.X, y = (int)screenPoint.Y };
+            RequestCalloutDisplay(ClientId, calloutId, title, message, isPermanentlyDismissible,
+                point, vsCommandGroupId, vsCommandId);
```
jnm2: What does `RequestCalloutDisplay` do when `Splat.ModeDetector.InUnitTestRunner()` is true and `screenPoint` was (0, 0)?

src/GitHub.Services.Vssdk/Services/TippingService.cs
jnm2: Ah, I think I answered my own question by reading below:

> // The option to pass a command option is only available on Visual Studio 2017+.

src/GitHub.Services.Vssdk/Services/TippingService.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        // The option to pass a command option is only available on Visual Studio 2017+.
+        void RequestCalloutDisplay(Guid clientId, Guid calloutId, string title, string message,
+            bool isPermanentlyDismissible, FrameworkElement targetElement,
+            Guid vsCommandGroupId, uint vsCommandId, object commandOption = null)
+        {
+            var tippingService = serviceProvider.GetService(typeof(SVsTippingService));
+            Assumes.Present(tippingService);
+            var currentMethod = MethodBase.GetCurrentMethod();
```
jnm2: What happens if the runtime decides to tail-call or inline `RequestCalloutDisplay`?

src/GitHub.Services.Vssdk/Services/TippingService.cs
```diff
@@ -0,0 +1,96 @@
...
+            var parameterTypes = currentMethod.GetParameters().Select(p => p.ParameterType).ToArray();
+            var method = tippingService.GetType().GetMethod(currentMethod.Name, parameterTypes);
```
jnm2: I think I'd specify `BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly` ([REFL008](https://github.com/DotNetAnalyzers/ReflectionAnalyzers/blob/master/documentation/REFL008.md)).

jnm2: Ah, not `DeclaredOnly` since you're using `GetType()` instead of `typeof(SVsTippingService)`.

Would it possibly be safer to make an interface call (`typeof(IVsTippingService).GetMethod(...).Invoke(...)`) rather than duck-typing on whatever derived type is returned?

src/GitHub.Services.Vssdk/Services/TippingService.cs
```diff
@@ -0,0 +1,96 @@
...
+                point, vsCommandGroupId, vsCommandId);
+        }
+
+        // The option to pass a command option is only available on Visual Studio 2017+.
+        void RequestCalloutDisplay(Guid clientId, Guid calloutId, string title, string message,
+            bool isPermanentlyDismissible, FrameworkElement targetElement,
+            Guid vsCommandGroupId, uint vsCommandId, object commandOption = null)
+        {
+            var tippingService = serviceProvider.GetService(typeof(SVsTippingService));
+            Assumes.Present(tippingService);
```
jnm2: The 2015 method logs two errors that this method could log but doesn't.

src/GitHub.TeamFoundation.14/Services/TeamExplorerServices.cs
```diff
@@ -58,6 +63,14 @@ public void ShowPublishSection()
             publish?.Connect();
         }
 
+        public async Task ShowRepositorySettingsRemotesAsync()
+        {
+            var te = serviceProvider.TryGetService<ITeamExplorer>();
+            var page = await NavigateToPageAsync(te, new Guid("96903923-97e0-474a-9346-31a3ba28e6ff"));
+            var remotes = page?.GetSection(new Guid("2e31f317-7144-4316-8aae-a796e4be1fd4"));
```
jnm2: What's the significance of these GUIDs? Would you be interested in putting them in named variables to document this?

src/GitHub.TeamFoundation.14/Services/TeamExplorerServices.cs
```diff
@@ -110,5 +123,37 @@ void OpenFolder(string repositoryPath)
...
+            var control = section?.SectionContent as UserControl;
+            if (control != null)
+            {
+                if (control.IsLoaded)
+                {
+                    BringIntoView();
+                }
+                else
+                {
+                    control.Loaded += (s, e) => BringIntoView();
```
jnm2: Shouldn't the event handler deregister itself before calling `BringIntoView()`?

src/GitHub.TeamFoundation.14/Services/TeamExplorerServices.cs
```diff
@@ -110,5 +123,37 @@ void OpenFolder(string repositoryPath)
...
+                }
+            }
+
+            void BringIntoView()
+            {
+                var targetRectangle = new Rect(0, 0, 0, 1000);
+                control.BringIntoView(targetRectangle);
+            }
```
jnm2: Not sure if this is the best idea, but sometimes I create custom awaitables:

```cs
var control = section?.SectionContent as UserControl;
if (control is null) return;

await control.WhenLoaded(); // No-op if IsLoaded is true, continues inside the Loaded event handler if not
var targetRectangle = new Rect(0, 0, 0, 1000);
control.BringIntoView(targetRectangle);
```

jcansdale: Good idea. Actually, I could do something like this using Rx. Could you sanity check my implementation?

src/GitHub.TeamFoundation.14/Services/TeamExplorerServices.cs
jcansdale: I've changed this to use Rx, which should automatically handle the un-subscription.

src/GitHub.TeamFoundation.14/Services/TeamExplorerServices.cs
jcansdale: They're for the `RepositoryProperties` page and the `Remotes` section. I've put them in named fields.

src/GitHub.Services.Vssdk/Services/TippingService.cs
jcansdale: I never thought about that. I wonder if `MethodBase.GetCurrentMethod` has any magic to stop that from happening or whether this is a real risk?

I've changed the implementation to use a fixed set of type parameters / method name. 👍

jcansdale:
@jnm2 thanks for the review! I _think_ I've implemented all of your suggestions.

src/GitHub.Services.Vssdk/Services/TippingService.cs
jnm2: Okay. So the tests display the callout at (0, 0) on the screen?

src/GitHub.Services.Vssdk/Services/TippingService.cs
jnm2: Wasn't thinking; tail calling isn't a concern unless the statement is `return MethodBase.GetCurrentMethod();`. But it's good to make the code less dependent on assumptions about how inlining works.

src/GitHub.Services.Vssdk/Services/TippingService.cs
jnm2: Yes. What I meant by interface call was `typeof(IVsTippingService).GetMethod("RequestCalloutDisplay").Invoke(...)` which depends on a (hopefully stable) interface definition, instead of `tippingService.GetType().GetMethod("RequestCalloutDisplay").Invoke(...)` which does a duck-typing call on the (potentially internal and unstable) derived class.

jnm2: I think this is a new rule of thumb, that `.GetType().GetFoo(...)` really leads to pitfalls. I've seen this a few times now.  So far I can't think of a scenario when `typeof(AccessibleType)` or `assembly.GetType("InaccessibleType", throwOnError: ...)` wouldn't be better.

src/GitHub.Services.Vssdk/Services/TippingService.cs
jnm2: I'm wondering, shouldn't both `RequestCalloutDisplay` overloads call `log.Error` for the same things? (`tippingService == null` and `method == null`.)

jnm2: > TippingService

I'm reading this word so many times it's starting to put https://youtu.be/J4-7Kvhbthg?t=11 into me head 🙄

src/GitHub.Services.Vssdk/Services/TippingService.cs
```diff
@@ -0,0 +1,102 @@
...
+                log.Error("Can't find {ServiceType}", typeof(SVsTippingService));
+                return;
+            }
+
+            Assumes.Present(tippingService);
+            var parameterTypes = new Type[] { typeof(Guid), typeof(Guid), typeof(string), typeof(string), typeof(bool),
+                typeof(Microsoft.VisualStudio.OLE.Interop.POINT), typeof(Guid), typeof(uint) };
+            var tippingServiceType = tippingService.GetType();
+            var method = tippingService.GetType().GetInterfaces().Where(i => i.Name == "IVsTippingService")
+                .FirstOrDefault()?.GetMethod("RequestCalloutDisplay", parameterTypes);
```
jnm2: If you do `Type.GetType("Microsoft.Internal.VisualStudio.Shell.Interop.IVsTippingService, Microsoft.VisualStudio.Shell.UI.Internal, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a").GetMethod("RequestCalloutDisplay", ...)`, you can put it in a static readonly MethodInfo or delegate field.

You could also define `IVsTippingService` next to your `SVsTippingService` and do `typeof(IVsTippingService)`.

jnm2: Anyway; `Where(...).FirstOrDefault()` can be `FirstOrDefault(...)` and you can say `var parameterTypes = new[] {` if you like.

jnm2: I think I'd still do `BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly` ([REFL008](https://github.com/DotNetAnalyzers/ReflectionAnalyzers/blob/master/documentation/REFL008.md)).

src/GitHub.TeamFoundation.14/Services/TeamExplorerServices.cs
jnm2: I haven't used Rx, but everything about the code you have makes sense to me.

jcansdale:
```
2018-11-21 20:09:58.160 [05624] EROR [25] PullRequestStatusBarManager 
System.AggregateException: One or more errors occurred. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at GitHub.InlineReviews.Services.PullRequestStatusBarManager.NoRemoteOriginCallout() in C:\Source\github.com\github\VisualStudio\src\GitHub.InlineReviews\Services\PullRequestStatusBarManager.cs:line 115
   at GitHub.InlineReviews.Services.PullRequestStatusBarManager.<RefreshCurrentSession>d__10.MoveNext() in C:\Source\github.com\github\VisualStudio\src\GitHub.InlineReviews\Services\PullRequestStatusBarManager.cs:line 91
   --- End of inner exception stack trace ---
---> (Inner Exception #0) System.NullReferenceException: Object reference not set to an instance of an object.
   at GitHub.InlineReviews.Services.PullRequestStatusBarManager.NoRemoteOriginCallout() in C:\Source\github.com\github\VisualStudio\src\GitHub.InlineReviews\Services\PullRequestStatusBarManager.cs:line 115
   at GitHub.InlineReviews.Services.PullRequestStatusBarManager.<RefreshCurrentSession>d__10.MoveNext() in C:\Source\github.com\github\VisualStudio\src\GitHub.InlineReviews\Services\PullRequestStatusBarManager.cs:line 91<---
```

jcansdale:
```
2018-11-21 20:06:36.670 [05624] DBUG [29] TeamExplorerContext       ActiveRepository changed to https://github.com/jcansdale/EmojiVS @ C:\Source\github.com\jbevain\EmojiVS
2018-11-21 20:07:04.295 [05624] DBUG [09] VSGitExt                  IGitExt.ActiveRepositories (#18029106) returned ["C:\\Source\\github.com\\jbevain\\EmojiVS"]
2018-11-21 20:07:04.328 [05624] EROR [09] VSGitExt                  Error refreshing repositories
LibGit2Sharp.LockedFileException: Failed open - 'C:/Source/github.com/jbevain/EmojiVS/.git/config' is locked: The process cannot access the file because it is being used by another process.

   at LibGit2Sharp.Core.Ensure.HandleError(Int32 result)
   at LibGit2Sharp.Core.Proxy.git_remote_lookup(RepositoryHandle repo, String name, Boolean throwsIfNotFound)
   at LibGit2Sharp.RemoteCollection.RemoteForName(String name, Boolean shouldThrowIfNotFound)
   at LibGit2Sharp.RemoteCollection.get_Item(String name)
   at GitHub.Services.GitService.HasRemotesButNoOrigin(IRepository repo) in C:\Source\github.com\github\VisualStudio\src\GitHub.Exports\Services\GitService.cs:line 143
   at GitHub.Services.GitService.CreateLocalRepositoryModel(String localPath) in C:\Source\github.com\github\VisualStudio\src\GitHub.Exports\Services\GitService.cs:line 49
   at GitHub.VisualStudio.Base.VSGitExt.<RefreshActiveRepositories>b__9_1(IGitRepositoryInfo x) in C:\Source\github.com\github\VisualStudio\src\GitHub.TeamFoundation.14\Services\VSGitExt.cs:line 115
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at GitHub.VisualStudio.Base.VSGitExt.RefreshActiveRepositories() in C:\Source\github.com\github\VisualStudio\src\GitHub.TeamFoundation.14\Services\VSGitExt.cs:line 115
2018-11-21 20:07:04.375 [05624] DBUG [09] VSGitExt                  ActiveRepositories changed to []
```

src/GitHub.Services.Vssdk/Services/TippingService.cs
jcansdale: No, the tests don't actually display anything. They just check what would happen if the API changes.

I considered exposing `PointToScreen` as a service (so that it could be mocked). In the end I thought that would be overkill.

src/GitHub.Services.Vssdk/Services/TippingService.cs
jcansdale: Thanks for the warning. I'm now finding the interface and calling using that. 👍

src/GitHub.Services.Vssdk/Services/TippingService.cs
jcansdale: I'll remove the unused `RequestCalloutDisplay` and poorly logged method. 😉 

> I'm reading this word so many times it's starting to put https://youtu.be/J4-7Kvhbthg?t=11 into my head 🙄

Sounds like a good excuse to watch Cars with my 5yo daughter. 😆

src/GitHub.Services.Vssdk/Services/TippingService.cs
jcansdale: I tried defining my own `IVsTippingService`, but for some reason COM type equivalence wasn't working. 😕 

I've changed this method to look for the `IVsTippingService` type `GUID`. That should make sure it's unique, but avoids depending explicitly on the `Microsoft.VisualStudio.Shell.UI.Internal` assembly. The `IVsTippingService` interface might move in a future version.

> Anyway; Where(...).FirstOrDefault() can be FirstOrDefault(...) and you can say var parameterTypes = new[] { if you like. Also could be var method = tippingServiceType.GetInterfaces()....
Both good points. Done. 👍 

> I think I'd still do BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly (REFL008).
Also done. 🙂

src/GitHub.VisualStudio.UI/Views/GitHubPane/NoRemoteOriginView.xaml
```diff
@@ -0,0 +1,55 @@
...
+                Content="{x:Static ghfvs:Resources.CantFindGitHubUrlForRepository}" />
+
+            <TextBlock
+                Text="{x:Static ghfvs:Resources.RepositoriesMustHaveRemoteOrigin}"
+                TextWrapping="Wrap"
+                TextAlignment="Center"
+                HorizontalAlignment="Center" />
+
+            <TextBlock
+                Text="Please rename one of your existing remotes to 'origin' or add a new remote named 'origin' and fetch. This can be done from the command line or by clicking the button below."
```
grokys: This needs be be moved to `Resources` I think.

src/GitHub.VisualStudio/GitHubPackage.cs
```diff
@@ -150,7 +152,34 @@ public ServiceProviderExports([Import(typeof(SVsServiceProvider))] IServiceProvi
...
-        T GetService<T>() => (T)serviceProvider.GetService(typeof(T));
+        [ExportForVisualStudioProcess]
+        public ITippingService TippingService
+        {
+            get
+            {
+                var tippingService = GetService<ITippingService>();
+                if (tippingService == null)
+                {
+                    // GetService<TippingService>() was returning null on Visual Studio 2015, so fall back to using new TippingService(...)
```
grokys: Do you mean `GetService<ITippingService>()`? Also, any idea why this was failing on 2015?

jcansdale: Oops, yes I did mean `ITippingService`.

I have no idea why it was failing on Visual Studio 2015, it had me completely stumped. The service appears to be registered correctly (it works on 2017) and it didn't appear to be throwing any exceptions. It's pretty disconcerting that it wasn't working. Maybe I race-condition or Visual Studio bug?

We could register `GitHub.Services.Vssdk` as a MEF assembly and sidestep this issue. I thought it better to leave the workaround as a hint, incase any of the other services start returning null in 2015.

src/GitHub.VisualStudio.UI/Views/GitHubPane/NoRemoteOriginView.xaml
jcansdale: Added the resource. 👍

src/GitHub.VisualStudio/GitHubPackage.cs
grokys: Lets leave it like this for the moment 👍

PR #2027, opened 2018-11-02
Fix draft appearing after sucessful submit.

grokys:
In certain situations, the inline comment thread was refreshed before the `DeleteDraft(comment)` line in `PullRequestReviewCommentThreadViewModel.PostComment` was run, meaning that the draft would be displayed _after_ the comment was succesfully submitted, causing a situation like this:

image

Work around this by deleting the draft before submitting the comment, and if we get an error re-insert the draft.

Bonus fix: Don't save drafts of empty comments.


meaghanlewis:
This LGTM 👍

PR #2028, opened 2018-11-02
Repository model refactoring (part 3)

jcansdale:
Depends on https://github.com/github/VisualStudio/pull/2008
Depends on https://github.com/github/VisualStudio/pull/2025

The focus of this PR has been removing the model interfaces, `Octokit` dependencies in model constructors and fixing NSubstitute tests that were broken when an interface changed to a class. I've tried to avoid any functional changes.

### What this PR does

- Remove the `IGitService` dependency from `BranchModel`
- Remove `Octokit.Branch` dependency from `BranchModel`
- Remove `Octokit.Repository` dependency from `RemoteRepositoryModel`
- Remove `IBranch` interface from `BranchModel`
- Remove the `IRepositoryModel` interface from `RepositoryModel`
- Remove `IRemoteRepositoryModel` interface from `RemoteRepositoryModel`
- Remove `ILocalRepositoryModel` interface from `LocalRepositoryModel`
- Moved `RemoteRepositoryModel` from `GitHub.App` to `GitHub.Exports.Reactive`
- ~~Moved `Account` from `GitHub.App` to `GitHub.Exports.Reactive`~~
- Rename `IGitService.CreateCurrentBranchModel` to `GetBranch`
- Remove unused/buggy code for creating a `BranchModel` when `LibGit2Sharp.Branch.IsRemote=true`

### Still to do

- [x] Refactor `BranchModel` to make it an actual model
(@grokys)
- [x] Remove the model interfaces
https://github.com/github/VisualStudio/pull/2008#discussion_r228090302
- [x] `BranchModel` is being passed a repository constructed from a URL rather then a local path
https://github.com/github/VisualStudio/blob/e3f811400f73aa722664c388253b76b24cf079c9/src/GitHub.Exports/Models/BranchModel.cs#L36 (removed this code)


src/GitHub.App/ViewModels/GitHubPane/PullRequestCreationViewModel.cs
```diff
@@ -82,8 +82,11 @@ public class PullRequestCreationViewModel : PanePageViewModelBase, IPullRequestC
                 .Where(_ => TargetBranch != null)
                 .Subscribe(x =>
                 {
-                    if (!x.Any(t => t.Equals(TargetBranch)))
+                    // HACK: Why is `t` null?
+                    if (!x.Any(t => t != null && t.Equals(TargetBranch)))
```
jcansdale: I'm not sure why `t` started coming up `null` when unit testing or whether isn't important. 😕

jcansdale: I think this was caused by changing from an interface to a class. NSubstitute was automatically creating a substitute, so `t` wasn't `null`. I've fixed up the tests.

src/GitHub.App/ViewModels/GitHubPane/PullRequestCreationViewModel.cs
```diff
@@ -82,8 +82,12 @@ public class PullRequestCreationViewModel : PanePageViewModelBase, IPullRequestC
                 .Where(_ => TargetBranch != null)
                 .Subscribe(x =>
                 {
+                    //// HACK: Why is `t` null?
+                    //if (!x.Any(t => t != null && t.Equals(TargetBranch)))
```
grokys: Does this comment mean "Why was there a check for `t` being null here?"

jcansdale: I temporarily added the `t != null` to allow the tests to pass. For some reason changing from using an interface to a class meant this come up null when testing.

I believe it was because NSubstitute automatically generates substitutes rather than returning `null` by default. I've since fixed this in the tests and removed the null check/comment.

jcansdale:
@grokys,

> Why was Account moved into GitHub.Exports.Reactive? It's a class with logic so should really stay in GitHub.App I think.

After some later cleanup, the constructor of `RemoteRepositoryModel` no longer depends on `Account`. I think that was the original problem. I've moved `Account` back to `GitHub.App`.

src/GitHub.App/GlobalSuppressions.cs
```diff
@@ -6,6 +6,8 @@
 [assembly: SuppressMessage("Microsoft.Naming", "CA1703:ResourceStringsShouldBeSpelledCorrectly", MessageId = "Git", Scope = "resource", Target = "GitHub.App.Resources.resources")]
 [assembly: SuppressMessage("Microsoft.Globalization", "CA1305:SpecifyIFormatProvider", MessageId = "System.String.Format(System.String,System.Object,System.Object,System.Object)", Scope = "member", Target = "GitHub.Services.PullRequestService.#CreateTempFile(System.String,System.String,System.String)")]
 [assembly: SuppressMessage("Microsoft.Globalization", "CA1305:SpecifyIFormatProvider", MessageId = "System.String.Format(System.String,System.Object,System.Object,System.Object)", Scope = "member", Target = "GitHub.Services.PullRequestService.#CreateTempFile(System.String,System.String,System.String,System.Text.Encoding)")]
+[assembly: SuppressMessage("Reliability", "CA2007:Do not directly await a Task", Justification = "We always need to think about what thread we're running on")]
```
grokys: This justification seems wrong - if we need to think about what thread we're running on, then we _should_ be warning when awaiting a `Task` - if you don't add a call to `.ConfigureAwait()` then you're implicitly requesting `.ConfigureAwait(true)` which isn't correct in all cases.

I think my preference here would be to leave the warnings in, and go through and address each warning individually. That's a lot of work, but I think it's the only way we can choose the correct behavior for each case.

src/GitHub.App/ViewModels/GitHubPane/PullRequestCreationViewModel.cs
grokys: Could we remove the commented-out code here then, or is that fix in a separate PR?

jcansdale: Oops, in my mind I'd removed the comments! ;-)

jcansdale:
> Also, with this PR after cloning a new repository, the Pull Requests button appears for a moment on the TE home page and then disappears, we need to fix that.

Good catch! This has been fixed in 17716efff1f290a48269cd0a0d1abef3b2671f48.

PR #2038, opened 2018-11-05
bump version to 2.6.0

meaghanlewis:


src/common/SolutionInfo.cs
```diff
@@ -18,6 +18,6 @@
 namespace System
 {
     internal static class AssemblyVersionInformation {
-        internal const string Version = "2.5.9.0";
+        internal const string Version = "2.6.0.0";
```
sguthals: This looks good to me!

PR #2055, opened 2018-11-14
Remove trackingcollection

### LINKED ISSUE #2054: gitignore templates in "Create Repository" not sorted correctly
The .gitignore templates in the repository creation dialog are not sorted correctly, and in addition are sometimes duplicated:

image

The "Recommended" templates should appear at the top, and the remaining templates should be sorted. However this does not happen and they appear in essentially random order.

In addition the templates are read from Akavache, which when refreshing the list can produce the list of templates twice, causing duplicates to appear.

grokys:
`TrackingCollection` was originally written to display the Pull Request list without locking up the UI, however in #1737, we moved to using data virtualization, which meant we no longer use `TrackingCollection`.

The remaining place in which tracking was used was the PR creation dialog, however `TrackingCollection` was not acting correctly and causing #2054, in addition a few of the `TrackingCollection` tests frequently came up as flaky tests, failing our CI build.

The issue in the PR creation dialog was more easily fixed by transitioning to using simple LINQ, so this PR removes the class.

image

Fixes #2054


meaghanlewis:
This works well for me 👍

PR #2057, opened 2018-11-15
Polish the list of checks

donokuda:
Small PR targeting my branch that polishes the Annotations details list:

**Before**
image

**After**
image

For reference, this is what I was building towards:

image

EDIT: There are a couple deviations here so I'll explain the reasoning behind them

1. I included the annotation title to be the header instead of the line number. In my initial designs, I didn't realize that we could include titles as a potential thing to surface. 🤔I'm not sure if the line number should have a higher hierarchy than the title so I'm open to other opinions.
1. I made the line number a hyperlink to open up the corresponding annotation in the diff view. [My original design](https://user-images.githubusercontent.com/1174461/48582371-35070580-e8d9-11e8-8844-48edf7e9859c.png) used a list box where you could double click the annotation to open it, but I was having layout issues.


StanleyGoldman:
I made one tweak to add a `Line #:#` even when we can't link to the items...

PR #2059, opened 2018-11-16
[wip] Add SetActiveRepository to ITeamExplorerServices

jcansdale:
This is an alternative to the current hack which creates and then deletes a dummy solution in the target repository folder.

- Add `SetActiveRepository` implementation that uses `ISccUIService`


jcansdale:
So near, yet so far. 😉

PR #2070, opened 2018-11-20
Begin polish for conversation view

donokuda:
Hello! I wanted to start bringing in some polish work to the pull request conversation view ahead of time. (This branch target's @grokys' `feature/pr-conversation` branch)

**Before**

screen shot 2018-11-19 at 3 01 46 pm

**After**

screen shot 2018-11-19 at 7 32 55 pm

Since this pull request touches the `CommentView.xaml` file, which is shared between the conversation and the threaded comment view, I wanted to make sure things looked consistent between the two. There is one update that affects this view which is adding a separator line between comments:

screen shot 2018-11-19 at 3 31 18 pm

For reference, here's what it looked like before:

screen shot 2018-11-19 at 3 02 06 pm

Last, I've also made sure that the page worked in Dark and Light themes:

screen shot 2018-11-19 at 7 33 18 pm
screen shot 2018-11-19 at 7 32 30 pm

### DIFF
--- a/src/GitHub.VisualStudio.UI/Styles/ThemeDark.xaml
+++ b/src/GitHub.VisualStudio.UI/Styles/ThemeDark.xaml
@@ -12,8 +12,8 @@
 
   <Color x:Key="GitHubActionLinkItemColor">#FF0097FB</Color>
   <SolidColorBrush x:Key="GitHubActionLinkItemBrush" Color="{StaticResource GitHubActionLinkItemColor}" PresentationOptions:Freeze="true" />
-  
-  <Color x:Key="GitHubHeaderSeparatorColor">#FF2D2D30</Color>
+
+  <Color x:Key="GitHubHeaderSeparatorColor">#FF3f3f46</Color>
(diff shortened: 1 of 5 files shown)


meaghanlewis:
👋 @grokys in the team sync you talked about discussing whether to keep the header or not. 

Can you clarify the header that you were referencing and share your thoughts about pros/cons of keeping it?

grokys:
Thanks for the reminder @meaghanlewis! The header I was talking about is this:

image

Currently in this PR it just uses the login of the user that created the PR, however more than one user can add commits to a PR, for example:

image

From a quick look around, afaics it just always reports one of the logins followed by "and others" but we should probably check what the logic is if we want to display this header. Looking at atom and VSCode they don't display this header.

donokuda:
> From a quick look around, afaics it just always reports one of the logins followed by "and others" but we should probably check what the logic is if we want to display this header. Looking at atom and VSCode they don't display this header.

Would it be possible to port that logic in here? I find the header helpful in terms of describing timeline events.

grokys:
Yes, we certainly could. Am I ok to merge this PR? If so I'll add the logic tomorrow (if I can work out the .com behavior).

donokuda:
Yup, feel free to merge

PR #2074, opened 2018-11-20
WIP: Prefer text instead of icon in checks list

donokuda:
This pull request updates the checks list to use words instead of an icon. However I cannot figure out how to get the links to flush to the right of the Grid element 😭 

screen shot 2018-11-20 at 2 12 05 pm

If anyone has a clue why, I'd be super grateful for the help 🙏


StanleyGoldman:
@grokys to the rescue

PR #2083, opened 2018-11-26
Adding expanders to the annotations view

StanleyGoldman:
Looking at a pull request like this made @donokuda and I realize we should probably add expanders to this section...

https://github.com/justaprogrammer/TestConsoleApp1/pull/6

image

Yeesh...
And my attempt is not better...

image


donokuda:
Pushed a bunch of tweaks. Here's what it looks like currently:

screen shot 2018-11-27 at 11 35 58 am

I'm going to remove the `[WIP]` and add the "Ready for Review" label, but I'm open to any new feedback.

StanleyGoldman:
Looks good to me..

PR #2084, opened 2018-11-26
Using different inline comment & annotation markers

StanleyGoldman:
This adds functionality to display a different inline glyph in different situations.
1. When there is a comment
2. When there is an annotation
3. Both


donokuda:
Here's where I've landed on this one:

**When there is a comment**

screen shot 2018-11-28 at 11 34 16 am

**When there is an annotation**

screen shot 2018-11-28 at 11 33 32 am

**Both**

screen shot 2018-11-28 at 11 33 02 am

I landed on using rectangle since it's close to how inline annotations look like and being able to stack them vertically helped communicate that there are "more than one" types of annotations and comments. We could consider utilizing more colors to indicate the type of annotation (instead of me sticking with one color for all annotations.)

donokuda:
Pushed a few commits changing it back to my original design:

**When there is a comment**

screen shot 2018-11-28 at 3 11 13 pm


**When there is an annotation**

screen shot 2018-11-28 at 3 13 01 pm

**Both**

screen shot 2018-11-28 at 3 12 03 pm

donokuda:
Pushed a commit that updates the colors of the annotation markers

screen shot 2018-11-29 at 11 40 35 am
screen shot 2018-11-29 at 11 32 27 am
screen shot 2018-11-29 at 11 32 10 am

@StanleyGoldman I'm going to remove my assignment from this issue, but let me know if I should take a look at anything else.

PR #2091, opened 2018-11-29
Fixing CheckRun details markdown links

StanleyGoldman:
Fixing the links inside annotation bodies

### DIFF
--- a/src/GitHub.VisualStudio.UI/Views/GitHubPane/PullRequestAnnotationsView.xaml.cs
+++ b/src/GitHub.VisualStudio.UI/Views/GitHubPane/PullRequestAnnotationsView.xaml.cs
@@ -18,5 +18,21 @@ namespace GitHub.VisualStudio.Views.GitHubPane
         {
             InitializeComponent();
         }
+
+        IVisualStudioBrowser GetBrowser()
+        {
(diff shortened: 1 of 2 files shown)


src/GitHub.VisualStudio.UI/Views/GitHubPane/PullRequestAnnotationsView.xaml.cs
```diff
@@ -18,5 +17,21 @@ public PullRequestAnnotationsView()
         {
             InitializeComponent();
         }
+
+        IVisualStudioBrowser GetBrowser()
```
jcansdale: I think rather than using a global service, you could do the following:
```cs
[Import]
public IVisualStudioBrowser Browser { get; set; }
```

You can then use `Browser.OpenUrl(uri);`.

src/GitHub.VisualStudio.UI/Views/GitHubPane/PullRequestAnnotationsView.xaml.cs
```diff
@@ -18,5 +17,18 @@ public PullRequestAnnotationsView()
         {
             InitializeComponent();
         }
+
+        [Import]
+        public IVisualStudioBrowser Browser { get; set; }
```
jcansdale: Nit: properties should be below methods. 😉

PR #2093, opened 2018-11-30
Removing the experimental feature toggle for forking

StanleyGoldman:


meaghanlewis:
Tested this on VS2017 and 2019. Functionality wise this looks good. A couple things I noticed were:
- when there are less than 3 organizations/users available to fork the highlight around the box extends very long
- sometimes organizations/users available don't load or just take a really long time (>60 seconds)

PR #2094, opened 2018-11-30
Updated VSIX logo to 128x128 pixels

madskristensen:
Extensions look best in the VS Marketplace with icons in the size of 128x128px. Inside Visual Studio's Extensions and Updates... dialog, the icon is downscaled with high quality. It is the new recommendation from the Visual Studio extensibility team to use 128x128px icons.

I used the preview logo (200x200) and resized it to 128x128 and it looks crisp and clear - high res and quality


src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
```diff
@@ -166,13 +166,15 @@
...
   </ItemGroup>
   <ItemGroup>
+    <Content Include="Resources\logo_128x128.png">
+      <IncludeInVSIX>true</IncludeInVSIX>
+    </Content>
     <Content Include="Resources\preview_200x200.png">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
     <Content Include="Resources\logo_32x32%402x.png">
```
jcansdale: I think we can get rid of `logo_32x32@2x.png` and its associated `VSPackage.resx` entries. Am I correct in thinking that only the one from `source.extension.vsixmanifest` will be used?

https://github.com/github/VisualStudio/blob/3c2d45dfbd0c68ecbc856e2b2fa3658c86f211da/src/GitHub.VisualStudio/VSPackage.resx#L138
https://github.com/github/VisualStudio/blob/3c2d45dfbd0c68ecbc856e2b2fa3658c86f211da/src/GitHub.InlineReviews/VSPackage.resx#L128

madskristensen: I can definitely get rid of the 32x32 logo. The reason I didn't was because it might be referenced from somewhere else that I'm not aware of. Not in the extension, but perhaps its URL here on GitHub. 

The resx file can't use .png files to show the logo in Help -> About box. Only .ico files. I've stopped using it because it's not important for the logo to show in the Help  -> About box IMO.

I'll remove it and update the PR

madskristensen: Fixed

jcansdale:
I thought I'd take some before and after screen grabs of this:

Before

image

After

image

It's not a big deal, but there does appear to be some aliasing in the after. 😉 

I'll defer to @donokuda on this one. Maybe he could generate a native 128x128 pixels version of this image?

madskristensen:
@jcansdale It appears the Extensions & Updates... dialog isn't doing the best job scaling those images. We're working on a fix right now. So a fix will come but for VS2019 only. I'll leave it up to you if you wish to take this PR based on the scaling issue

jcansdale:
@madskristensen,

> It appears the Extensions & Updates... dialog isn't doing the best job scaling those images. We're working on a fix right now. So a fix will come but for VS2019 only. You can decide if it looks too bad to take the PR or not

Thanks for the explanation. It's smaller than it appears and looks fine. I was just curious where the aliasing was coming from. I'll ✔️ this PR.

PR #2098, opened 2018-12-03
Ensure x86|x64\e_sqlite3.dll are included in .vsix

jcansdale:
Ensure `e_sqlite.dll` is included in .vsix when built using Visual Studio. This file was only being included in the .vsix when built using MSBuild from the command line.

I don't like how brittle this is, but it was the only way I could find to ensure that `x86|x64\e_sqlite3.dll` are included in .vsix when built using Visual Studio.

### What this PR does

- Add a `<PackageReference Include="SQLitePCL.raw_basic">` to `GitHub.VisualStudio.csproj`
  - Referencing this package directly appears to ensure that native artifacts are included in the .vsix when built using Visual Studio. This was never an issue with command line builds.

### How to test

1. Build the solution using Visual Studio
2. Check that `GitHub.VisualStudio.vsix` contains `x86\e_sqlite3.dll` and `x64\e_sqlite3.dll` (you can view contents using 7-zip)

### What was happening

Here is a comparison between a .vsix build using Visual Studio and one build using MSBuild from the command line.

image

It was causing the following when attempting to view a PR file diff.

image

image

image

### Notes

I'm hoping it was also the reason some buttons weren't showing up. 🤞  We'll need to keep a look out for this one!

image


jcansdale:
@madskristensen i'm seeing a strange issue where artifacts are appearing in .vsix depending on whether is's build from the command line with MSBuild or from inside Visual Studio.  When built using Visual Studio the `e_sqlite3.dll` isn't getting included, so I've needed to resort to the hack in this PR. Any idea what's going on? 😕

madskristensen:
@dgriffen Do you know?

PR #2107, opened 2018-12-05
Pull requests button disappears when changing from a non-GitHub to a GitHub repository

jcansdale:
There can be delay before `IsAGitHubRepo` returns its result. If the repository changes before the call returns, the button can end up with the visibility for the wrong repository.

This fix only sets the button visibility if the repository hasn't changed. Button is invalidated multiple times, so one of them returns the visibility for the correct repository.

### What this PR does

- Don't change button visibility if repo changed after async call was made

### What this PR doesn't do

This is probably only one of a number of issues like this. This is a point fix for the pull requests button disappearing (this is likely to have the most impact). Unfortunately the Team Explorer integration code has become fix upon fix and is difficult to work with and test.

Ideally we would create a view-model that controls visibility of all repository related buttons and UI inside Team Explorer. This would be a large (but worthwhile) refactoring, but is out of scope for this fix.

### How to repro

1. Open a Git repository that isn't on GitHub (e.g. one from Azure DevOps)
2. Open a Git repository that is on GitHub
3. Expect the pull request button to appear and then disappear 

This is what it ends up looking like:

image

### How to test

1. Open a Git repository that isn't on GitHub (e.g. one from Azure DevOps)
2. Open a Git repository that is on GitHub
3. Expect the pull request button to appear and remain!

### DIFF
--- a/src/GitHub.TeamFoundation.14/Base/TeamExplorerNavigationItemBase.cs
+++ b/src/GitHub.TeamFoundation.14/Base/TeamExplorerNavigationItemBase.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Diagnostics;
 using System.Drawing;
+using System.Threading.Tasks;
 using GitHub.Api;
 using GitHub.Extensions;
 using GitHub.Logging;
@@ -39,10 +40,14 @@ namespace GitHub.VisualStudio.Base
             SubscribeToRepoChanges();
         }
 
-        public override async void Invalidate()
+        public override void Invalidate()
         {
             IsVisible = false;
+            InvalidateAsync().Forget(log);
+        }
 
+        async Task InvalidateAsync()
+        {
             var uri = ActiveRepoUri;
             var isVisible = await IsAGitHubRepo(uri);
             if (ActiveRepoUri != uri)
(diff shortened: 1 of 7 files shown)


src/GitHub.TeamFoundation.14/Base/TeamExplorerNavigationItemBase.cs
```diff
@@ -35,10 +40,23 @@ public class TeamExplorerNavigationItemBase : TeamExplorerItemBase, ITeamExplore
...
+        }
+
+        async Task InvalidateAsync()
+        {
+            var uri = ActiveRepoUri;
+            var isVisible = await IsAGitHubRepo(uri);
+            if (ActiveRepoUri != uri)
+            {
+                log.Information("Not setting button visibility because repository changed from {BeforeUrl} to {AfterUrl}", uri, ActiveRepoUri);
+                return;
```
StanleyGoldman: Can you explain what is happening to fall into this trap?

jcansdale: I believe what is happening is that `Invalidate` is being called twice, once by Team Explorer and once by us when we detect that the repository has changed.

The problem happens when Team Explorer calls `Invalidate` before we know the repository has actually changed. We call an `async` method to check is the original (non-GitHub) repository is a GitHub/GHE repository.

Before this async method returns we're notified that the repository has changed and fire off another async method to check if the new repository is visible. This returns immediately because it's a `github.com` domain so we don't need to probe for Enterprise. The first call would then return having established that the original repository wasn't a GitHub repo, stomping over the new/correct visible state of the PRs icon.

I hope that makes sense. 😉

StanleyGoldman: It took me a few reads but i got it.

PR #2119, opened 2018-12-11
Fix CA2007 ConfigureAwait warnings.

grokys:
Fixes all instances of the `CA2007: Do not directly await a Task without calling ConfigureAwait` warning in the codebase.

I've tried to use the correct parameter to `ConfigureAwait(()`, in that I pass `true` when the instructions following the await in the method require code to be run on the UI thread. If I've got this wrong anywhere, this could cause a crash so the features affected need testing.


grokys:
According to MS, [we should not be heeding CA2007 in VSSDK code](https://github.com/Microsoft/VSSDK-Analyzers/issues/95#issuecomment-447004367) so closing this. I'll disable the warning in #2134 and do the suggested fixes in another PR.

PR #2129, opened 2018-12-12
Deprecate using Visual Studio to build

donokuda:
I've been running into issues trying to build the extension using Visual Studio IDE and better luck using `cmd.exe` instead. Since building with Visual Studio hasn't been consistently successful, I propose we encourage folks to build from command prompt.

It is still possible to install the Debug build to the experimental instance with Visual Studio by going to `Debug > Start Debugging / Start without Debugging` in the menu bar.

/cc @meaghanlewis since you helped me troubleshoot my issue earlier


grokys:
Hmm, I think this would introduce a lot more friction for my workflow at least. You really shouldn't be having these problems! I think we should spend some time trying to work out why VS builds aren't working for you rather than just advice people to build outside VS...

donokuda:
> Hmm, I think this would introduce a lot more friction for my workflow at least. You really shouldn't be having these problems!

Good point. I'm going to close this out since I'd rather not potentially introduce more confusion and issue for folks other than me.

PR #2136, opened 2018-12-13
Fix string culture warnings.

grokys:
Fixes CA1304, CA1305, CA1307, CA1308 and CA1309.

~~Depends on #2135~~
~~Depends on #2134~~
~~Depends on #2132~~
~~Depends on #2119~~


src/GitHub.TeamFoundation.14/Base/EnsureLoggedInSection.cs
```diff
@@ -54,7 +54,7 @@ async Task CheckLogin()
             bool loggedIn = await ConnectionManager.IsLoggedIn(add);
             if (!loggedIn)
             {
-                var msg = string.Format(CultureInfo.CurrentUICulture, Resources.NotLoggedInMessage, add.Title, add.Title);
+                var msg = string.Format(CultureInfo.CurrentCulture, Resources.NotLoggedInMessage, add.Title, add.Title);
```
jcansdale: Curious why the change from `CurrentUICulture`?

PR #2137, opened 2018-12-13
Add script to help with troubleshooting some dev environment issues

donokuda:
A common thing that I've been running into often is my development environment not working if I haven't been keeping up to date with changes regularly. When I would pull a branch and try building, I'd often run into errors similar to this:

screen shot 2018-12-13 at 10 06 56 am

Some of the engineers suggested closing Visual Studio and running the following commands:

```bash
git clean -xdff
git submodule foreach git clean -xdff
```

Which ends up solving most of the issues I've been running into! 🎉 

I've added a script (`scripts/clean.ps1`) and took a crack at [documenting when to use this command](https://github.com/github/VisualStudio/tree/clean-me#troubleshooting) in the README, however there's a chance it could be improved a bit.

I've also added `git submodule update` to the top of the ~`clean.ps1`~ `clean.cmd` script so that submodules such as Octokit are sync'd before running the command.

/cc @grokys @StanleyGoldman since y'all have helped me this week regarding my dev issues.

### DIFF
--- a/clean.cmd
+++ b/clean.cmd
@@ -1,1 +1,3 @@
-powershell -ExecutionPolicy Unrestricted scripts\clean.ps1 
+git submodule update --init
+git clean -xdff
+git submodule foreach git clean -xdff
--- a/clean.cmd
+++ b/clean.cmd
@@ -1,0 +1,1 @@
+powershell -ExecutionPolicy Unrestricted scripts\clean.ps1 
--- a/README.md
+++ b/README.md
@@ -92,6 +92,16 @@ Logs can be viewed at the following location:
 
 `%LOCALAPPDATA%\GitHubVisualStudio\extension.log`
 
+## Troubleshooting
+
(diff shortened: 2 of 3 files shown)


clean.cmd
```diff
@@ -0,0 +1 @@
+powershell -ExecutionPolicy Unrestricted scripts\clean.ps1
```
grokys: Is there any reason why this script needs to call out to `powershell` to run the git commands? Couldn't it just run them here?

donokuda: You're right! I was following existing examples involving `.cmd` and `.ps1` files. I've moved things over to the `clean.cmd` flle and 🔥 the `clean.ps1`

donokuda:
🛀@grokys Ready for another review. Thanks for taking a look at this 😄

grokys:
Thanks @donokuda !

PR #2142, opened 2018-12-18
Update to use modern version of LibGit2Sharp and use indent heuristic when diffing

### LINKED ISSUE #1469: Inline comment appears on PR page but not in GitHub extension

### LINKED ISSUE #2144: Unable to add reply to comments
image

jcansdale:
Depends on:
- https://github.com/libgit2/libgit2sharp/pull/1651: Target net46 instead of net461
- https://github.com/libgit2/libgit2sharp/pull/1652: Add support for git diff --indent-heuristic option

Thanks @bording for the fast turnaround of the `net46` PR! 🙏 

### What this PR does

- Update to reference LibGit2Sharp `0.26.0`
- Stop using obsolete `LibGit2Sharp` methods
- Fix broken test caused by `Commit` and `Tree` no longer accepting `\` in paths
- Enable indent heuristic when comparing files
- Move `Compare` and `CompareWith` from `IGitClient` to `IGitService` (comparing files doesn't require the credentials of `GitClient`)
- Add tests for `Compare` and `CompareWith` methods of `GitClient`

### To do

- [x] Rebase and reference `LibGit2Sharp 0.26.*` from NuGet (not local version)
- [x] Enable the indent heuristic when diffing
- [x] Previously is was possible to use `\` as a separator when looking up `Tree` entries. We should maybe add checks for this in case we're using `\` instead of `/` anywhere.

### How to test

- Check that inline comment from #1469 appears.

Fixes #1469 
Fixes #2144


src/GitHub.App/Services/GitClient.cs
```diff
@@ -75,14 +79,11 @@ public Task Fetch(IRepository repository, string remoteName)
...
-            return Task.Factory.StartNew(() =>
+            return Task.Run(() =>
             {
                 try
                 {
-                    var remote = repository.Network.Remotes[remoteName];
-#pragma warning disable 0618 // TODO: Replace `Network.Fetch` with `Commands.Fetch`.
-                    repository.Network.Fetch(remote, fetchOptions);
-#pragma warning restore 0618
+                    repository.Network.Fetch(remoteName, new[] { "+refs/heads/*:refs/remotes/origin/*" }, fetchOptions);
```
StanleyGoldman: I guess things have changed since this warning was disabled...

PR #2145, opened 2018-12-19
Adding a Pull Request status circle User Control

StanleyGoldman:
My attempt to emulate this in Xaml

image

**TODO**:
- [x] Get love from @donokuda 
- [x] ~Remove the summary checks from the View Model, if we decide not to use it~

**IMAGE**:
image


jcansdale:
I'd love to see some indication of whether the pull request is in a merge-able state. At the moment we know that ✔️ is definitely and ❌ is definitely not merge-able (assuming any restrictions are in place).

Where we see a doughnut (💭 mmm [doughnuts](https://i.imgur.com/Mjoyxjf.mp4)), we have no way of knowing whether the pull request is in a merge-able state.

Crazy idea, could we color the center of the doughnut green or red, when we know the pull request is definitely merge-able or not merge-able? 😄

donokuda:
I pushed a small tweak adjusting the thickness of the donut chart so that it's closer to the line thickness of the octicons:

screen shot 2019-01-10 at 4 34 57 pm

This looks good to be merged in.

> I'd love to see some indication of whether the pull request is in a merge-able state. At the moment we know that ✔️ is definitely and ❌ is definitely not merge-able (assuming any restrictions are in place).
> 
> Crazy idea, could we color the center of the doughnut green or red, when we know the pull request is definitely merge-able or not merge-able? 

I'm hesitant on adding more information since we'll need to come up with an approach that's clear and can fit in the limited amount of space that we have. I struggled with this a bit with the annotation markers; and in the spirit of moving this pull request forward, I think it's best better to punt at this time.

Would you be open to showing this type of information in the pull request detail view / sidebar? That way, we can breakdown why a branch isn't mergeable (out of date, review needed, required checks, etc) and guide developers to fix it:

image

src/GitHub.VisualStudio.UI/UI/Controls/PullRequestStatusCircle.xaml.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        private int TotalCount => ErrorCount + SuccessCount + PendingCount;
+
+        public int ErrorCount
+        {
+            get => (int)GetValue(ErrorCountProperty);
+            set
+            {
+                SetValue(ErrorCountProperty, value);
+                GeneratePolygons();
```
grokys: Dependency property CLR setter wrappers shouldn't do anything other than call `SetValue` because when they're set from XAML the CLR setter isn't actually called (XAML will call `SetValue` directly). Instead of doing this, you should be adding a property changed handler on dependency property registration. See [here](https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/custom-dependency-properties#implementing-the-wrapper) for more details.

StanleyGoldman: Fixed

PR #2147, opened 2018-12-20
[wip] Find a way to improve the organization list load time

### LINKED ISSUE #2143: Regression: listing repositories is unusable for any significant amount of repos
As much as I hate vague issue titles, it's pretty much that. Any user who has access to a significant number of repos (even if they only own a few, they might have org access with large numbers of repos) will have a bad time when listing repositories.

This is me loading the clone repo dialog from the 2017 start page. It takes two minutes to get anything to show up on the list:

repo-list-2

There are two main problems here, which together make the UX poor:

1) No caching - every time the dialog opens, it runs the query again without showing previous results
2) No streaming - the list is only populated when all results come in

A side effect of no caching and re-running the full query every time is that it might exhaust API rate limits and lock users out temporarily, depending on how much many api calls are happening. One of my attempts at loading the list resulted in a 403 forbidden:

repo-list

jcansdale:
### What this PR does

Load organization repositories by including `OrganizationMember` in `ownerAffiliations` rather than iterating over Repositories for each of the viewer's `Organizations`.

### Related

- List org repos ordered by STARGAZERS not working
https://github.com/github/github/issues/100120

- Unable to fetch user’s repositories by organization membership
https://platform.github.community/t/unable-to-fetch-users-repositories-by-organization-membership/7557/6

Fixes #2143

### DIFF
--- a/src/GitHub.App/Services/RepositoryCloneService.cs
+++ b/src/GitHub.App/Services/RepositoryCloneService.cs
@@ -73,6 +73,8 @@ namespace GitHub.Services
                     Direction = OrderDirection.Asc
                 };
 
+                // Pass this for affiliations and ownerAffiliations. See:
+                // https://platform.github.community/t/unable-to-fetch-users-repositories-by-organization-membership/7557/6
                 var affiliation = new RepositoryAffiliation?[]
                 {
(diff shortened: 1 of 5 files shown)


jcansdale:
Closing in favor of #2148.

PR #2158, opened 2019-01-04
Modernizing and Implementing Autocompletebox

StanleyGoldman:
This pull request is the set of changes made to make AutoCompleteBox functional in our codebase

- Using modern structures like `Guard` instead of `Ensure`
- Removing the legacy cache system
- Using GraphQL for querying data

#### Depends On
- [x] #2159 being merged to this pull request


meaghanlewis:
@StanleyGoldman would it be possible to also show pull requests in the autocomplete list in addition to issues?

meaghanlewis:
@StanleyGoldman this is working well! One thing that differs from dotcom is that you only seem to filter on open issues/PRs (`filter = $"repo:{owner}/{name} is:open";`) where dotcom shows all.

PR #2167, opened 2019-01-09
Fix AvailableForComment calculation.

### LINKED ISSUE #2166: Can't leave comment on deleted line
## Versions
 - GitHub Extension for Visual Studio version: 2.7.0 (dev)
 - Visual Studio version: 2017

## What happened
1. I tried to leave a comment on the deleted line `134` of `RepositoryCreationViewModel.cs` PR #2165 
2. The inline comment UI appears but is empty

image

### Expected behavior
Expected inline comment box to appear like it does on the RHS.

image

grokys:
When showing a comment peek view on the LHS of the diff view, we need to look at `OldLineNumber` instead of `NewLineNumber` to calculate `AvailableForComment`.

Fixes #2166


StanleyGoldman:
Belated, but yea this makes sense.

PR #2181, opened 2019-01-16
GraphQL caching.

grokys:
This PR adds caching to GraphQL queries. It introduces a new `IGraphQLClient` which is used instead of (and wraps) Octokit.GraphQL's `IConnection`.

When a query is submitted, this class hashes the query and uses this as a cache key. Queries are cached in the filesystem using a slightly modified version of https://github.com/acarteas/FileCache.

The main complication here is that if you have say 3 pages worth of the PR list cached then pressing "Refresh" needs to clear _all_ pages of the PR list. We use the region name to identify such queries and when refresh is pressed, call `FileCache.ClearRegion`. This method isn't present in the mainline FileCache which is why we ship a slightly modified version of it.

Depends on https://github.com/octokit/octokit.graphql.net/pull/186


StanleyGoldman:
As per conversations had on slack graphql caching will be a double edged sword. We will need to review all of our usages of the graphql library to make sure caching will not result in any user experience bugs.

StanleyGoldman:
### Low Impact

#### Functionality to get a pull request node id
https://github.com/github/VisualStudio/blob/14eadbe90cb6ffabd655137d05a21f43346f04c5/src/GitHub.InlineReviews/Services/PullRequestSessionService.cs#L458-L472

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -218,7 +218,8 @@ public class PullRequestService : IssueishService, IPullRequestService, IStaticR
                 { nameof(states), states.Select(x => (Octokit.GraphQL.Model.PullRequestState)x).ToList() },
             };
 
-            var result = await graphql.Run(query, vars);
+            var region = owner + '/' + name + "/pr-list";
```
StanleyGoldman: Maybe centralize this.

src/GitHub.App/Services/PullRequestService.cs
```diff
@@ -293,6 +294,14 @@ public class PullRequestService : IssueishService, IPullRequestService, IStaticR
             return result;
         }
 
+        public async Task ClearPullRequestsCache(HostAddress address, string owner, string name)
+        {
+            var region = owner + '/' + name + "/pr-list";
```
StanleyGoldman: With that.

src/GitHub.Api/GraphQLClient.cs
```diff
@@ -0,0 +1,159 @@
...
+            FileCache cache)
+        {
+            this.connection = connection;
+            this.cache = cache;
+        }
+
+        public async Task ClearCache(string regionName)
+        {
+            // Switch to background thread because FileCache does not provide an async API.
+            await TaskScheduler.Default;
```
jcansdale: I wonder if we should avoid a `Microsoft.VisualStudio.Threading` dependency simply returning `Task.Run(() => cache.ClearRegion(GetFullRegionName(regionName)))`?

StanleyGoldman:
While I was prepping to present at the Visual Studio 2019 launch, my head had a thought about how the caching would affect the fork function and what routine is that using...

StanleyGoldman:
GraphQL is not used to fork repos

PR #2197, opened 2019-01-23
Unify the repository list and URL tabs

### LINKED ISSUE #2187: Unify the repository list / URL tabs of Open from GitHub dialog
### Summary

Users seem to rarely change the tab they're presented with on the open/clone dialog. This is despite that different tabs would be appropriate when cloning different repositories.

If the user is cloning a recently contributed to repository the repository list would make most sense. If they're cloning a public repository they don't own, the URL tab would be the only option.

By allowing users to enter a URL into the repository list search box, they won't need to change tabs in order to clone/open any repository.

### Implementation Ideas

- Add repository in search box to list of repositories
- Add repository in copy/paste buffer to list of repositories
- Allow user to search repositories by exact match URL

jcansdale:
It appears users rarely change the tab they're initially presented with. This is despite the fact that the different tabs are useful in different circumstances. If the user is cloning a repository they own or have contributed to, the GitHub/GitHub Enterprise tab makes sense. If they're cloning a public repository owned by a different user, it would be necessary to use the URL tab.

### What this PR does

This PR combines the repository list and URL tabs so that a user can clone any repository without needing to understand the difference between tabs.

- Allow filtering repositories by URL or part of URL
- Allow user to clone any URL in filter box
- Match repository URLs that include a trailing `/` or `.git`
- Change filter prompt to "Search or enter a URL"
- Remove the `URL` tab

### To do

- [x] Ignore trailing `/` or `.git` when searching

#### Finding a repository by its URL

1. Click `File > Open > Open from GitHub...`
2. Paste the URL for a repository you own into `Search or enter a URL`
3. Check repository appears under `Your repositories` section
4. Clone or open repository

image

#### Cloning a public repository by its URL

1. Click `File > Open > Open from GitHub...`
2. Paste the URL for a public repository that you don't own into `Search or enter a URL`
3. Check that `Clone` button is clickable
4. Clone or open repository

#### Re-cloning a repository

1. Click `File > Open StartPage`
2. Click on a `https://` link under `Recent`
3. Check clicked on URL is in `Search or enter a URL` box
4. Clone or open repository

image

#### Cloning a GitHub Enterprise repository by its URL

1. Copy the URL of a GitHub Enterprise repository
2. Click `File > Open > Open from GitHub...`
3. Check the Enterprise tab is active
4. Check the URL is in `Search or enter a URL` box
5. Clone or open repository

### What this PR doesn't do

- Clone a repository from a server (or user account) that is different to the one currently logged in

What if the user has already logged into a GitHub/Enterprise account, but not the one they they want to clone from? At the moment they would need to open `Team Explorer - Connect` and sign out/sign in to the correct account. This isn't very discoverable.

This is out of scope for this PR since it also wasn't possible with the previous implementation. Should we offer an explicit login/sign out option in another PR?

- Automatically confirm that a repository exists for the entered URL

It would be nice if we could automatically check that a repository exists when the user enters a URL. The repository could be automatically added to a section called something like `Public repositories`.

Alternatively, we could automatically fire off a search and supplement the repository list when the user starts entering a query. This would allow users to find public repositories they about without visiting github.com.

- Surface starred repositories

We can easily see if a repository has been starred, but we don't currently surface this information. This is likely a useful signal when scanning for a repository in a potentially long list.

Fixes #2187


meaghanlewis:
This looks good to me functionality wise. ✅

StanleyGoldman:
Looks good. What about that last todo?

meaghanlewis:
We will also need to update the documentation for this unified dialog here: https://github.com/github/VisualStudio/blob/master/docs/using/cloning-a-repository-to-visual-studio.md#clone-repositories.

I'm happy to do that.

PR #2201, opened 2019-01-28
Only show `Sign in with your browser` option for github.com

### LINKED ISSUE #2194: Only show `Sign in with your browser` option in sign in dialog
### Summary

We have had lots of conversations about our sign-in workflow over the last couple months.

One thing we have discussed is only showing the option to `Sign in with your browser` and removing additional fields. This will simplify sign in for users and save them time. 

- For signing in with `GitHub` this means removing the username or email and password field.
- For signing in with `GitHub Enterprise` this means keeping the server address field and removing the token field.

Example:
sign in with your browser

Perhaps we could make the dialog look similar to the SSO to GitHub dialog:
screen shot 2019-01-23 at 7 38 27 am

jcansdale:
In future only the `Sign in with your browser option will work` (see [here](https://developer.github.com/changes/2020-02-14-deprecating-oauth-auth-endpoint/)). We need to remove the user/pass/2FA UI.

Sign in with browser is by far the easiest authentication method and should be encouraged where possible. In order to do this, we're placing it first in the list of login options (see #2194).

### What this PR does

- Hide the username / password login for https://github.com
- Move the `Sign in with browser` option above other options

### What this PR doesn't do

We considered only showing the `Sign in with browser` option when available. We decided not to do this for the following reasons.

- A user might have multiple GitHub accounts and not want to login using the one currently active in their browser
- ~~There might be times where it is necessary to log in using a token (e.g to push a GitHub Action)~~ the `workflow` scope is now being requested, see #2462

### How to test

The GitHub and GitHub Enterprise logins should look like this:

- GitHub
image

- GitHub Enterprise
image


grokys:
Hmm, I'm not sure this really adds much :my gaze is still going straight to the user/pass textboxes as that's what I expect from a login dialog.

What about: when opening the dialog, show a button saying "Sign in with your browser", and below it a smaller link saying "Sign in with username and password" and only when you click that button the UI changes to display the user/pass boxes?

jcansdale:
> What about: when opening the dialog, show a button saying "Sign in with your browser", and below it a smaller link saying "Sign in with username and password" and only when you click that button the UI changes to display the user/pass boxes?

That sounds good. What to you think @donokuda ? Could you pick up where I left off on this PR? 🙏

jcansdale:
Obsoleted by #2469

PR #2202, opened 2019-01-28
Use modern clone dialog when re-cloning a repository

### LINKED ISSUE #2199: Re-clone uses old clone dialog
### Steps to Reproduce
1. Click on a repository link in `Recent`
2. Old re-clone dialog appears rather than `Open from GitHub`

_Note, I don't know exactly when/why these links get added. We should investigate and find out if these links are part of the `Get to Code` workflow._

image

### Expected behavior
- Show `Open from GitHub` dialog with URL field populated.

### Screenshots
image

jcansdale:
_Depends on #2198_

Now that the clone dialog can clone any arbitrary GitHub URL (not just ones owned by the user), we don't need a separate dialog when re-cloning a URL using the `Start Page`.

### What this PR does

- Use the same clone dialog when cloning a new repository or re-cloning one that has been cloned before
- Remove the obsolete `ShowReCloneDialog` API and related classes

### How to test

1. Open the `Start Page` (click `File > Start Page`)
2. Click on the `GitHub` button which is under `Open`
3. Clone a repository
4. Delete the repository folder (necessary to make the URL link appear)
5. Open the `Start Page` (click `File > Start Page`)
6. Click on the `https://` link to the repository you cloned then deleted
7. `Open from GitHub` dialog should appear

Fixes #2199


jcansdale:
@prasethu I'm trying to test this PR, but the cloned URLs don't appear on the `Start Page` history (step 5 above). Any idea what's going on or why they might not be appearing? 😕 

I was always unsure about when they were supposed to appear, so this isn't a feature I'm familiar with. I'd like to understand what's going on now so we can test it properly.

Thanks for any pointers!

PR #2208, opened 2019-02-01
Update solution so that it will compile on Visual Studio 2019

jcansdale:
The `GitHubVS` solution now contains projects that target multiple versions .NET and Visual Studio. This was causing errors in our `GitHub.VisualStudio` project, which needed to reference all projects in order to package them into a VSIX. Because this project contains source code that executes on all supported version of Visual Studio, it needs to target .NET 4.6 (to maintain compatibility with Visual Studio 2015 and 2017). This meant we couldn't reference `GitHub.TeamFoundation.16`, which targets Visual Studio 2019 and .NET 4.7.2.

### What this PR does

- Separate `GitHub.VisualStudio.csproj` into a project that is responsible for building `GitHub.VisualStudio.dll` and a project that is responsible for packaging `GitHub.VisualStudio.vsix`
- Update projects and submoduels to use `MSBuild.Sdk.Extras, v1.6.61` which is compatible with Visual Studio 2017 _and_ 2019 (`2.0.0-preview.14` only works with 2019)
- Use reflection to instantiate different versions of `VSGitExt` (depending on the executing version of Visual Studio) without referencing the containing projects
- Use `ReferenceOutputAssembly="false"` when referencing `GitHub.TeamFoundation.*` projects (which should avoid warnings)
- Remove VSIX packaging related crud from `GitHub.VisualStudio.csproj`
- Remove redundant project dependencies from `GitHub.VisualStudio.csproj` (this should make unit tests compile faster)

### How to test

This PR has caused a fair amount of churn. It would be a good idea to:

1. Run through the test manifest
2. Merge this PR at the beginning of a cycle


src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
```diff
@@ -166,6 +166,21 @@
...
+    </Content>
+    <Content Include="..\GitHub.TeamFoundation.15\obj\$(Configuration)\GitHub.TeamFoundation.15.dll">
+      <Link>GitHub.TeamFoundation.15.dll</Link>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <IncludeInVSIX>true</IncludeInVSIX>
+    </Content>
+    <Content Include="..\GitHub.TeamFoundation.16\obj\$(Configuration)\GitHub.TeamFoundation.16.dll">
+      <Link>GitHub.TeamFoundation.16.dll</Link>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <IncludeInVSIX>true</IncludeInVSIX>
```
AArnott: Keep these as ProjectReference items, but add `ReferenceOutputAssembly=false` metadata to each one. That *should* stop MSBuild from complaining that the target framework of the referenced project exceeds that of this project.

GitHubVS.sln
```diff
@@ -1,11 +1,14 @@
...
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27703.2035
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28516.95
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "GitHub.VisualStudio", "src\GitHub.VisualStudio\GitHub.VisualStudio.csproj", "{11569514-5AE5-4B5B-92A2-F10B0967DE5F}"
 	ProjectSection(ProjectDependencies) = postProject
+		{161DBF01-1DBF-4B00-8551-C5C00F26720D} = {161DBF01-1DBF-4B00-8551-C5C00F26720D}
+		{161DBF01-1DBF-4B00-8551-C5C00F26720E} = {161DBF01-1DBF-4B00-8551-C5C00F26720E}
```
AArnott: You can remove these solution dependencies if the `ReferenceOutputAssembly=false` metadata trick works.

src/GitHub.VisualStudio/GitHub.VisualStudio.csproj
jcansdale: Thanks for the suggestion! Unfortunately Visual Studio 2019 still shows errors, even though the build is actually completing.

image

I wonder if we should refactor the solution so that there is a project dedicated to packaging the VSIX that doesn't contain any actual code? This could target .NET 4.7.2 and therefore reference projects that target any other version.

AArnott: > I wonder if we should refactor the solution so that there is a project dedicated to packaging the VSIX that doesn't contain any actual code? This could target .NET 4.7.2 and therefore reference projects that target any other version.

I think _that's_ a great idea. I typically prefer such a pattern anyway so that packing doesn't have to happen to unit test that code.

Regrading VS error list entries, that *might* be tracked by this: https://github.com/Microsoft/msbuild/issues/2661. Can you review that issue and see if it covers the gap? There are workarounds disclosed there as well.

jcansdale: This is how I like to structure my personal VSIX projects as well. I've put off doing this with `GitHub.VisualStudio` because it contains quite a lot of code and dependencies, but this build error might be just the incentive we need! 🤞

jcansdale:
@AArnott Any idea what's going wrong on our Azure DevOps build?
image

It compiles locally and I think `GitHub.TeamFoundation.14` is referenced in exactly the same was as `GitHub.TeamFoundation.15` and `GitHub.TeamFoundation.16`. 😕

I tried `ReferenceOutputAssembly="false"` and `"true"` but that didn't make a difference.

AArnott:
@jcansdale I don't know why it builds locally for you, but the project names that show up in your extension.source.vsixmanifest file are supposed to match the `<Name>` metadata in a `ProjectReference` item, so since you don't have those any more, I wouldn't expect references to them in extension.source.vsixmanifest to work properly.

jcansdale:
@AArnott,

I am still referencing those projects from the `GitHub.VisualStudio.Vsix` project.

Here are extracts from the `.csproj` and `.vsixmanifest`. Maybe I'm missing something obvious, but I can't see any difference between `GitHub.TeamFoundation.14` and `GitHub.TeamFoundation.15/16' (which appear to work). 😕 

```xml
    <ProjectReference Include="..\GitHub.TeamFoundation.14\GitHub.TeamFoundation.14.csproj">
      <Project>{161dbf01-1dbf-4b00-8551-c5c00f26720d}</Project>
      <Name>GitHub.TeamFoundation.14</Name>
```

```xml
        <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project"
              d:ProjectName="GitHub.Services.Vssdk" Path="|GitHub.Services.Vssdk|" />
        <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project"
              d:ProjectName="GitHub.TeamFoundation.14" TargetVersion="[14.0,15.0)" Path="|GitHub.TeamFoundation.14|" />
        <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project"
              d:ProjectName="GitHub.TeamFoundation.15" TargetVersion="[15.0,16.0)" Path="|GitHub.TeamFoundation.15|" />
        <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project"
              d:ProjectName="GitHub.TeamFoundation.16" TargetVersion="[16.0,17.0)" Path="|GitHub.TeamFoundation.16|" />
```

AArnott:
I must have been looking at the wrong project then. It looks like you're doing it right. @michael-eng may be able to help more since his team owns this area.

jcansdale:
@AArnott

>  I don't know why it builds locally for you...

It was because the `GitHub.VisualStudio` project was still building a .vsix in `Release` configuration. 😊 

Now only the `GitHub.VisualStudio.Vsix` project is responsible for building a .vsix artifact, it's working much better! 😄

jcansdale:
There appears to be a bug in the VS SDK .targets that means the instance an extension is installed in can be different to the one that is opened. This is confusing because it causes extensions to not be updated/installed without any warning (they are installed, just not in the instance that opens).

Here are screenshots from the same .binlog file:

image

image

AArnott:
@Michael-Eng should see [this](https://github.com/github/VisualStudio/pull/2208#issuecomment-467464964).

AArnott:
In the meantime, @jcansdale, the best thing to do is probably use VS's Report a Problem command to report this.

PR #2241, opened 2019-02-26
Remove empty directory when clone fails

### LINKED ISSUE #2126: Don't leave behind empty directory when a clone fails
## Versions
 - GitHub Extension for Visual Studio version: 2.6.0
 - Visual Studio version: 2015 (probably all)

## What happened
1. Attempted to clone `grokys/PullRequestSandbox3` (which doesn't exist)
2. Attempted to clone `grokys/PullRequestSandbox` to the `grokys\PullRequestSandbox3` folder
3. Was block because there's an empty directory in the target destination

### Expected behavior
- User should be able to clone into empty directory
- When a clone fails, the empty directory created for the clone should be deleted.

### Screenshots
image

_Note: The logic that the new Visual Studio 2019 `Get to Code` feature uses is that a directory must be empty or not exist on order to clone. I think we should be consistent with this and allow cloning to empty directories._

image

YuPeiHenry:
Ready for review.
fixes #2126 . Added unit test to check that directory is removed when clone fails.

The fix deletes the directory when an exception in RepositoryCloneService:Clone is caught.


test/GitHub.App.UnitTests/Services/RepositoryCloneServiceTests.cs
```diff
@@ -122,6 +122,23 @@ public async Task UpdatesMetricsWhenDefaultClonePath(string targetPath, string d
...
 
+        [Test]
+        public async Task CleansDirectoryOnCloneFailed()
+        {
+            var serviceProvider = Substitutes.GetServiceProvider();
+            var operatingSystem = serviceProvider.GetOperatingSystem();
+            var vsGitServices = serviceProvider.GetVSGitServices();
+            var cloneService = CreateRepositoryCloneService(serviceProvider);
+
+            Assert.ThrowsAsync<Exception>(async () => {
```
jcansdale: This could be simplified to just:
```cs
Assert.ThrowsAsync<Exception>(() => cloneService.CloneRepository("https://github.com/failing/url", @"c:\dev\bar"));
```

test/GitHub.App.UnitTests/Services/RepositoryCloneServiceTests.cs
```diff
@@ -122,6 +122,23 @@ public async Task UpdatesMetricsWhenDefaultClonePath(string targetPath, string d
...
+                await cloneService.CloneRepository("https://github.com/failing/url", @"c:\dev\bar");
+            });
+
+            operatingSystem.Directory.Received().CreateDirectory(@"c:\dev\bar");
+            operatingSystem.Directory.Received().DeleteDirectory(@"c:\dev\bar");
+            await vsGitServices.Received().Clone("https://github.com/failing/url", @"c:\dev\bar", true);
```
jcansdale: Could you factor out "https://github.com/failing/url" and @"c:\dev\bar" into something like `cloneUrl` and `clonePath`?

src/GitHub.App/Services/RepositoryCloneService.cs
```diff
@@ -223,6 +223,7 @@ bool IsSolutionInRepository(string repositoryPath)
             }
             catch (Exception ex)
             {
+                operatingSystem.Directory.DeleteDirectory(repositoryPath);
                 log.Error(ex, "Could not clone {CloneUrl} to {Path}", cloneUrl, repositoryPath);
```
jcansdale: Maybe we should move the `log.Error(...)` to above `DeleteDirectory` in case it also throws?

meaghanlewis:
@YuPeiHenry thank you for your contribution!

YuPeiHenry:
@jcansdale I have updated the PR to adopt your suggestions.

test/GitHub.App.UnitTests/Services/RepositoryCloneServiceTests.cs
jcansdale: Thanks. 😄

PR #2248, opened 2019-02-27
[WIP] Resource cleanup

StanleyGoldman:
While reviewing resources it seems to me that these resources are never used.


jcansdale:
@StanleyGoldman The resources appear to be used here, but I wish they weren't!

https://github.com/github/VisualStudio/blob/72c146d0605489b38104729ebfdda1a0acb528da/src/GitHub.StartPage/StartPagePackage.cs#L22-L23

I don't see where they appear in the extension.  Thought it was an about page thing, but don't see the text there. 😕

StanleyGoldman:
Geez okay, i'm glad you found that.

StanleyGoldman:
https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.providecodecontainerproviderattribute?view=visualstudiosdk-2017

StanleyGoldman:
I have no idea what that attribute is for.

StanleyGoldman:
We actually have to figure out how to translate these strings...

Issue #2245, opened 2019-02-27
Honor the resolved/unresolved state of pull request comments


jcansdale:
At the moment pull request comments are only hidden inside the extension when the code near to them hasn't been changed (they're not marked as `Outdated` on github.com). The option to `Resolve conversation` has recently been added to github.com. We should continue showing conversations until they have been marked as resolved.

Here is a comment on the github.com conversation view that is marked as `Outdated` but that is still visible because it hasn't been resolved.

![image](https://user-images.githubusercontent.com/11719160/53482621-8cd80f80-3a77-11e9-98c9-e72453c4a5bb.png)

Here is a comment that is visible on the diff view that isn't `Outdated`, but that has been resolved.

![image](https://user-images.githubusercontent.com/11719160/53482808-feb05900-3a77-11e9-8116-9e605eaf20a8.png)

Issue #2101, opened 2018-12-04
Resolve comments in Visual Studio


meaghanlewis:
## Background
On github.com, there is an option to "resolve conversation" on all comments.

![resolve conversation diff view](https://user-images.githubusercontent.com/10103121/49408836-f19efa80-f712-11e8-93ff-f9fafcc3c542.png)
This option exists both when viewing a comment in the **conversation view** and **diff view**.

Once comments are resolved, they are displayed differently.

**In the conversation view:**
- the comments are no longer visible by default and must be expanded to see.
![resolved conversations diff view](https://user-images.githubusercontent.com/10103121/49409034-c8329e80-f713-11e8-8478-2a25f7a3357a.png)
- expanding the comments shows the option to `unresolve conversation` and who marked the comment as resolved
![unresolve from conversation view](https://user-images.githubusercontent.com/10103121/49409102-fa440080-f713-11e8-9f4e-7626a0a0d04a.png)

**In the diff view:**
- the comment is not visible by default and just shows who marked the comment as resolved
![resolved conversation diff view hidden](https://user-images.githubusercontent.com/10103121/49409156-32e3da00-f714-11e8-9ce3-79bdc78f11bf.png)
- expanding the comment will also show the option to `unresolve conversation`
![resolved conversation diff view](https://user-images.githubusercontent.com/10103121/49409220-73435800-f714-11e8-9a6b-fc49cdc766e0.png)

Once comments are unresolved, they go back to there original state and default display.

## Problem
While this functionality exists on github.com, it does not in GitHub for Visual Studio. There can be discrepancies between conversations that have been resolved on github.com and it can also cause confusion to someone reviewing pull requests with conversations that are no longer relevant.

## Solution
Add same option in Visual Studio to resolve and unresolve conversations from both the diff view and conversation view (once it is implemented).

donokuda:
I went ahead and created a few prototypes to get a feel of resolving conversations in Visual Studio. I followed the patterns that are established on dotcom and tweaked them so they look and feel at home in Visual Studio.

**Resolving in conversation view** ([Figma Prototype](https://www.figma.com/proto/Dr9g55XZtZqt22PxmMrnpniH/Visual-Studio-Conversation?node-id=563%3A330&viewport=1448%2C-270%2C1&scaling=min-zoom)):
![resolved-conversation](https://user-images.githubusercontent.com/1174461/52507092-696f2280-2ba5-11e9-8157-eecb5884813f.gif)

[...]

PR #2253, opened 2019-03-01
Hide resolved comments from changes tree and indicate resolved comments on margin

jcansdale:
Previously pull request comments would be implicitly resolved when when the text near to the comment was changed. GitHub now supports comments being explicitly resolved by a user _and_ comments becoming implicitly outdated when the text near to the comment is changed.

This pull request surfaces the `IsResolved` status of a review comment threads and hides resolved comments from the changes tree (in a similar way to how outdated comments are hidden). This makes it easy to find comments that still need to be addressed when responding to a pull request review.

### What this PR does

- Use `pullRequest/reviewThreads` to populate the `IsResolved` status of a pull request comment threads
- When using GitHub Enterprise fall back to using original model (`reviewThreads` might not be available)
- Hide resolved comments from the changes tree
- Show different in-line comment glyph for resolved comments (an empty circle)
- Show banner to clearly indicate when a thread has been resolved

### To do

- [x] Support fallback for older versions of GitHub Enterprise
- [x] Indicate outdated in-line comments in glyph
image
- [x] Show banner on resolved comment threads
image

### What this PR doesn't do

- Probe GitHub Enterprise schema to see if we could fetch resolved comment information
- Surface comments that are outdated but unresolved
- Allow resolving of comments using inline comment view

### How to test

1. Open `https://github.com/grokys/PullRequestSandbox` in Visual Studio
2. Open grokys/PullRequestSandbox#58 in `GitHub` pane
3. Check that only 1 comment appears on `Program.cs`
image
4. Click on `💬 1` comment
5. Check `Program.cs` appears in diff view and comment is visible
6. Check that empty circle glyph appears below indicating a resolved comment
7. Click on the resolved comment glyph
8. Check that `The conversation was marked as resolved` banner appears above comment

### Notes

Here is what is looks like in the `Dark` theme

image

Related #2245

### DIFF
--- a/src/GitHub.App/ViewModels/GitHubPane/PullRequestDetailViewModel.cs
+++ b/src/GitHub.App/ViewModels/GitHubPane/PullRequestDetailViewModel.cs
@@ -364,7 +364,8 @@ namespace GitHub.ViewModels.GitHubPane
 
                 Checks = (IReadOnlyList<IPullRequestCheckViewModel>)PullRequestCheckViewModel.Build(viewViewModelFactory, pullRequest)?.ToList() ?? Array.Empty<IPullRequestCheckViewModel>();
 
-                await Files.InitializeAsync(Session);
+                // Only show unresolved comments
+                await Files.InitializeAsync(Session, c => !c.IsResolved);
 
                 var localBranches = await pullRequestsService.GetLocalBranches(LocalRepository, pullRequest).ToList();
 
--- a/src/GitHub.App/ViewModels/PullRequestReviewCommentThreadViewModel.cs
+++ b/src/GitHub.App/ViewModels/PullRequestReviewCommentThreadViewModel.cs
@@ -70,6 +70,9 @@ namespace GitHub.ViewModels
         /// <inheritdoc/>
         public DiffSide Side { get; private set; }
 
+        /// <inheritdoc/>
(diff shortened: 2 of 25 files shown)


donokuda:
I went and did A Thing replacing the info icon with the one built into visual studio. The result is a look and feel closer to a native infobar:

**Blue Theme!**
screen shot 2019-03-05 at 3 47 52 pm

**Dark Theme!**
screen shot 2019-03-05 at 3 47 14 pm

At first I was concerned about how much it popped out in the Dark Theme even though I was using the background color that we get from Visual Studio. However, I think it's okay for the following reasons:

1. I tend to look over this information when the background was less intrusive.
1. It's what Microsoft does. 😆  

We've diverged a bit from what I've [prototyped](https://github.com/github/VisualStudio/issues/2101#issuecomment-461972466) but I think that's fine given that my prototype relied heavily on the interaction of collapsing resolved comments.

jcansdale:
@donokuda,

> At first I was concerned about how much it popped out in the Dark Theme even though I was using the background color that we get from Visual Studio. However, I think it's okay for the following reasons:
1. I tend to look over this information when the background was less intrusive.

I think this is a good call. 👍 We're trying to communicate that someone stuck a fork in this comment thread, you're probably done. Making this pop out is a good thing!

PR #2278, opened 2019-03-11
Modifying AppVeyor to build general packages for releases

StanleyGoldman:
Branches named "release/.*" without the "-vsinstaller" extension should create packages in order to deploy.


StanleyGoldman:
You are right about the comment, I meant it for releases. Thanks for the 👍.

PR #2312, opened 2019-04-01
Enable theme detection when running outside of Visual Studio

jcansdale:
Theme detection currently depends on Visual Studio's `VSColorTheme` class. This prevents the theme from being detected correctly when UI components are being previewed outside of Visual Studio.

This PR changes theme detection to use `Application.Current.TryFindResource` instead of `VSColorTheme.GetThemedColor`. This allows theme detection to work inside any WPF application.


StanleyGoldman:
It seems the changes in this pull request were already merged into master

PR #2316, opened 2019-04-04
Allow cloning of repositories into empty directories

### LINKED ISSUE #2250: Allow cloning of repositories into empty directories
## Versions
* GitHub Extension for Visual Studio version: 2.6.0
* Visual Studio version: 2015 (probably all)

## What happend
* Attempted to clone `grokys/PullRequestSandbox` to the `grokys\PullRequestSandbox3` folder (an empty folder)
* Was block because there's an empty directory in the target destination

### Expected behavior
* User should be able to clone into empty directory
* The logic that the new Visual Studio 2019 `Get to Code` feature uses is that a directory must be empty or not exist on order to clone. There should be consistency with this to allow cloning to empty directories.

### Screenshots
image

### Additional context
This issue is a continuation of #2126, to address cloning into empty directories separately.

YuPeiHenry:
Ready for review
fixes #2250 

The fix modifies `CanClone` and `CanOpen` on `RepositoryCloneViewModel`, and modified `RepositoryCloneService` CloneOrOpenRepository method to factor for empty directories.


src/GitHub.App/ViewModels/Dialog/Clone/RepositoryCloneViewModel.cs
```diff
@@ -240,6 +242,11 @@ string ValidatePathWarning(RepositoryModel repositoryModel, string path)
                 {
                     using (var repository = gitService.GetRepository(path))
                     {
+                        if (service.DestinationDirectoryEmpty(path))
```
jcansdale: I'd be inclined to make this part of the `if` above, e.g.

```cs
if (service.DestinationDirectoryExists(path)) && service.DestinationDirectoryEmpty(path))
```

That way we don't need to create a repository object if the directory is empty.

YuPeiHenry: Thanks for pointing this out. I have corrected this issue.

StanleyGoldman:
image

StanleyGoldman:
The more I look at it, it's just that one message that would no longer be accurate...
I don't even think it was right to begin with..

If the directory is not empty, and is not a git repo that could be scanned.
The error message probably should have been. "A directory exists at the destination path."
Similar to "A file exists at the destination path."

Based on the current logic in this Pull Request.
The new error message should be: "The directory at the destination path is not empty."
The chain of logic should be... (in pseudocode)

```
if file_exists_at_destination:
   return "A file exists at the destination path"

else:
   if directory_exists_at_destination AND directory_at_destination_not_empty:

      if directory_exists_at_destination_is_git_repo:
         if git_repo_has_no_origin:
            return "A repository already exists at this location, but it doesn't have a remote named `origin`."
         end if

         if git_repo_origin_is_different:
            return "A repository already exists at this location, but it doesn't have a remote named `origin`."
         end if

      else:
         return "The directory at the destination path is not empty."
      end if

   end if
end if
```

StanleyGoldman:
Let @meaghanlewis weigh in before you change anything...

meaghanlewis:
I agree with what @StanleyGoldman proposed: `The directory at the destination path is not empty.` Its straightforward and covers cases where there might be a git directory for the destination or not.

YuPeiHenry:
@StanleyGoldman I have updated the error message for `non-empty directory`. Should something be done about the translation to other languages in this PR?

StanleyGoldman:
Hey @YuPeiHenry I pushed some commits. It felt easier than nit picking at you. 😈 

So I copied the English text to all the language files, we have a 3rd party resource that translates things for us. So we just put the English in place until they get around to it.

I also realized I reordered the logic funny in my pseudocode. You changed the flow to match my ramblings. I undid the damage I caused. Sorry to put you through my spaghetti brain.

And I tweaked `DestinationDirectoryEmpty`. I everywhere that the function is used, we explicitly check if the directory exists. It makes sense to check if the directory exists inside `DestinationDirectoryEmpty` to prevent an error, but it's also redundant to all the code that uses `DestinationDirectoryEmpty`. The only correct choice is to assume the function with be used properly.

All nit picks. I didn't wanna bug you with it. 😅 

Since my hands have been all in this, I'm going to let @meaghanlewis test this and give us both the final approval.

meaghanlewis:
thank you @YuPeiHenry for your contribution! 🎉

PR #2333, opened 2019-04-29
Use JoinableTaskContext from MEF

jcansdale:
This PR updates the extension to use `JoinableTaskContext` from MEF rather than the Visual Studio version specific `ThreadHelper.JoinableTaskContext`. This allows the extension to be hosted outside of Visual Studio or as a zero-impact "essentials" extension inside Visual Studio 2019.

### What this PR does

- Remove the obsolete `ThreadingHelper` class
- Use `JoinableTaskContext` from MEF where possible (VS 2017 and 2019)
- Use `await TaskScheduler.Default` instead of `await ThreadingHelper.SwitchToPoolThreadAsync()`

### Notes

The `JoinableTaskContext` object is an used by an application to manage threads and for switching to the Main/UI thread. A single instance of this object exists for the application. Visual Studio makes this object available in a number of places.

- Visual Studio packages have a `JoinableTaskContext` property
- The static `ThreadHelper.JoinableTaskContext` property
- A `JoinableTaskContext` service is available via MEF in Visual Studio 2017 and 2019

Unfortunately the `ThreadHelper` class exists in the `Microsoft.VisualStudio.Shell.14.0`, `Microsoft.VisualStudio.Shell.15.0` and `Microsoft.VisualStudio.Shell.16.0`, and isn't guaranteed to have been initialized unless a package built against the same version has been loaded. This means it won't have been initialized when running outside of Visual Studio and might not have been initialized when the essentials extension is running inside Visual Studio 2019. Initializing it in these scenarios would involve a pretty nasty and fragile hack.

This PR changes the extension to use `JoinableTaskContext` from MEF and only fall back to using `ThreadHelper.JoinableTaskContext` when the MEF service isn't available (in Visual Studio 2015).

For example:
```cs
[ImportingConstructor]
public TwoFactorChallengeHandler([Import(AllowDefault = true)] JoinableTaskContext joinableTaskContext)
{
    JoinableTaskContext = joinableTaskContext ?? ThreadHelper.JoinableTaskContext;
}
```

`[Import(AllowDefault = true)]` means that `joinableTaskContext` will be null when running in Visual Studio 2015 and `joinableTaskContext ?? ThreadHelper.JoinableTaskContext` makes it fall back to use the property from `ThreadHelper`.

### How to test

Check that the extension is still working in Visual Studio 2015 and 2017/2019. The code path is slightly different in 2015 because of the fall back.

### Related

- Authoring a library with a JoinableTaskFactory dependency
https://github.com/Microsoft/vs-threading/blob/master/doc/library_with_jtf.md
- Testing Visual Studio extensions and packages
https://github.com/Microsoft/vs-threading/blob/master/doc/testing_vs.md
- Fixes ThreadHelper.JoinableTaskFactory throws when running test on Visual Studio 2017
https://github.com/josetr/VsixTesting/pull/6


src/GitHub.TeamFoundation.14/Base/TeamExplorerServiceHolder.cs
```diff
@@ -14,23 +14,17 @@ public class TeamExplorerServiceHolder : ITeamExplorerServiceHolder
...
-        public TeamExplorerServiceHolder(ITeamExplorerContext teamExplorerContext, JoinableTaskContext joinableTaskContext)
+        [ImportingConstructor]
+        public TeamExplorerServiceHolder(ITeamExplorerContext teamExplorerContext,
+            [Import(AllowDefault = true)] JoinableTaskContext joinableTaskContext)
         {
+            joinableTaskContext = joinableTaskContext ?? ThreadHelper.JoinableTaskContext;
+
             JoinableTaskCollection = joinableTaskContext.CreateCollection();
             JoinableTaskCollection.DisplayName = nameof(TeamExplorerServiceHolder);
             JoinableTaskFactory = joinableTaskContext.CreateFactory(JoinableTaskCollection);
```
StanleyGoldman: I know you didn't change it, but can you explain to me what is going on with this collection?

src/GitHub.VisualStudio/Services/UsageTracker.cs
```diff
@@ -31,11 +30,13 @@ public sealed class UsageTracker : IUsageTracker, IDisposable
         public UsageTracker(
             IGitHubServiceProvider gitHubServiceProvider,
             IUsageService service,
-            IPackageSettings settings)
+            IPackageSettings settings,
+            JoinableTaskContext joinableTaskContext)
```
StanleyGoldman: Do we not intend to use `[Import(AllowDefault = true)]` here?

src/GitHub.VisualStudio/Services/VSGitExtFactory.cs
```diff
@@ -13,12 +12,14 @@ public class VSGitExtFactory
         readonly int vsVersion;
         readonly IServiceProvider serviceProvider;
         readonly IGitService gitService;
+        readonly JoinableTaskContext joinableTaskContect;
 
-        public VSGitExtFactory(int vsVersion, IServiceProvider serviceProvider, IGitService gitService)
+        public VSGitExtFactory(int vsVersion, IServiceProvider serviceProvider, IGitService gitService, JoinableTaskContext joinableTaskContect)
```
StanleyGoldman: I noticed `[Import(AllowDefault = true)]` is not used here as well?

src/GitHub.TeamFoundation.14/Base/TeamExplorerServiceHolder.cs
jcansdale: It's creating a collection that tasks created by `JoinableTaskFactory` will be placed into. It's then possible to wait for all tasks in the collection to complete before disposing of the service when Visual Studio shuts down. This pattern was copied from another service.

In this case the service isn't disposable and the `JoinableTaskCollection` is never used. It could be simplified to simply store the `JoinableTaskContext` and use the default collection (which is associated with `JoinableTaskContext.Factory`).

I think it would make sense to be consistent and include this simplification as part of this PR.

src/GitHub.VisualStudio/Services/UsageTracker.cs
jcansdale: I thought this as well before realizing that `UsageTracker` is a Visual Studio service not a MEF service (so `[Import(...)` isn't supported). We simply pass it the `JoinableTaskContext` from the host package.

src/GitHub.VisualStudio/Services/VSGitExtFactory.cs
jcansdale: In a similar way to `UsageTracker`, `VSGitExtFactory` is instantiated by a Visual Studio package and not MEF. `JoinableTaskContext` is passed from the package and will never be null.

src/GitHub.TeamFoundation.14/Base/TeamExplorerServiceHolder.cs
StanleyGoldman: OK, cool.

src/GitHub.VisualStudio/Services/UsageTracker.cs
StanleyGoldman: Ok, I understand.

PR #2351, opened 2019-05-13
Ignore whitespace in Open from GitHub search box

### LINKED ISSUE #2350: Cannot auto detect invalid space in 'Open from GitHub' search box
VS Build: d16.2stg-28908.49
Repro Rate: 100%

Repro on previous VS # d16.1-28907.181: Not implemented

Operation System Information: Win10
 
Repro Steps:
1. Launch VS and click 'Clone or check out code' in Start Window
2. Click 'GitHub' under 'Browse a repository' and sign in
3. Input the repo name/location without any blank in searching box
4. Add a blank space and verify the filtered out repo

Expected: 
Corresponding repo can be listed out.

 
Actual:

Corresponding repo disappeared after inputting a blank space.
(Uploaded "Blank.PNG" and "ReproBlankIssue.gif" into attachment)

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/891264

jcansdale:
We should ignore whitespace when searching for a repository.

### What this PR does

- Ignore any whie-space that appears before or after the filter/URL
- Check that we ignore white-space before and after filter

### How to test

1. Launch VS and click 'Clone or check out code' in Start Window
2. Click 'GitHub' under 'Browse a repository' and sign in
3. Input the repo name/location without any blank in searching box
4. Add a blank space before or after the filter
5. Check that the filtered repository(s) still appear

Fixes #2350


meaghanlewis:
@jcansdale repository results are filtering as expected when there is a trailing or leading space. One thing that I noticed was that the notification disappears and the buttons get disabled when there is a space until a repo is selected from the list. 

Is it possible to keep the same state even after adding a space?
space when cloning

jcansdale:
> Is it possible to keep the same state even after adding a space?

I've just had a look and this would be possible. It would add a little complexity though.

I think the use-case we're likely to be fixing is when the user copy/pastes a repository name/URL which contains some trailing white-space. In this case they wouldn't be adding an extra space, it would simply be there as soon as they paste.

I'd be inclined to keep the code clean rather than add support adding extra spaces.

meaghanlewis:
That works for me then 👍.

PR #2353, opened 2019-05-14
Add a login option to Open from GitHub

### LINKED ISSUE #2352: No where for user to sign out GitHub account on VS 16.2stg with new GitHub essentials plugin
VS Build: d16.2stg-28908.49
Repro Rate: 100%

Repro on previous VS # d16.1-28907.181: Not implemented

Operation System Information: Win10

Repro Steps:
1. Install VS d16.2stg-28908.49 without optional GitHub extension, launch it
2. Click 'Clone or check out code' in Start Window
3. Click 'GitHub' under 'Browse a repository'
4. Sign in with a GitHub account
5. Try to Sign out that GitHub account   
 
Expected: 
The GitHub account can be signed out successfully.
 
Actual:

No where for user to sign out GitHub account.  
(Uploaded "AccountIssue.gif" into attachment)

- See https://devdiv.visualstudio.com/DevDiv/_workitems/edit/891245

jcansdale:
We need a way to allow users to log into a different GitHub or GitHub Enterprise from the Open from GitHub dialog. This PR adds a `Login` button so that a user can list and clone repositories from a different account.

### What this PR does

- Add a `Sign in` button to `Open from GitHub` dialog
- Don't disable tabs on the `Connect to GitHub` dialog when already logged in

### How it works

Click on the ~~`Login`~~ `Sign in` button.

image

The `Connect to GitHub` dialog appears. If the user is already logged into GitHub, the `GitHub Enterprise` tab will be selected by default. If the user wants to log in using a different GitHub account, they can select the `GitHub` tab and log in. Previously this tab would have been disabled and the user would only have been able to log into a GitHub Enterprise server.

image

When the user logs in successfully, the appropriate `GitHub.com` or `Enterprise` tab will be selected and the repository list will be refreshed. If the user closes the `Connect to GitHub` dialog, the previous `Open from GitHub` tab will remain active.

### Reviewers

I've used a simple button for the `Login` action so as not to immediately violate the Visual Studio UI guidelines. Is there a more appropriate placement or control that I should use?

### What we also tried

Put `Sign in` as button next to `Clone` and `Open`.

image

I found the `Sign in` button felt out of place next to `Clone` and `Open`.

`Sign in` has been moved to a link at the top right which is a convention on websites.

Fixes #2352

### DIFF
--- a/src/GitHub.App/ViewModels/Dialog/Clone/RepositoryCloneViewModel.cs
+++ b/src/GitHub.App/ViewModels/Dialog/Clone/RepositoryCloneViewModel.cs
@@ -26,6 +26,7 @@ namespace GitHub.ViewModels.Dialog.Clone
         readonly IRepositoryCloneService service;
         readonly IGitService gitService;
         readonly IUsageTracker usageTracker;
+        readonly IDialogService dialogService;
         readonly IReadOnlyList<IRepositoryCloneTabViewModel> tabs;
         string path;
         UriString url;
@@ -40,6 +41,7 @@ namespace GitHub.ViewModels.Dialog.Clone
             IRepositoryCloneService service,
             IGitService gitService,
             IUsageTracker usageTracker,
+            IDialogService dialogService,
             IRepositorySelectViewModel gitHubTab,
             IRepositorySelectViewModel enterpriseTab)
         {
@@ -48,6 +50,7 @@ namespace GitHub.ViewModels.Dialog.Clone
             this.service = service;
             this.gitService = gitService;
             this.usageTracker = usageTracker;
+            this.dialogService = dialogService;
 
             GitHubTab = gitHubTab;
             EnterpriseTab = enterpriseTab;
@@ -82,6 +85,8 @@ namespace GitHub.ViewModels.Dialog.Clone
             Open = ReactiveCommand.CreateFromObservable(
                 () => repository.Select(x => new CloneDialogResult(Path, x?.CloneUrl)),
                 canOpen);
+
+            LoginAsDifferentUser = ReactiveCommand.CreateFromTask(LoginAsDifferentUserAsync);
         }
 
         public IRepositorySelectViewModel GitHubTab { get; }
@@ -111,6 +116,8 @@ namespace GitHub.ViewModels.Dialog.Clone
 
         public IObservable<object> Done => Observable.Merge(Clone, Open);
 
+        public ReactiveCommand<Unit, Unit> LoginAsDifferentUser { get; }
+
         public ReactiveCommand<Unit, Unit> Browse { get; }
 
         public ReactiveCommand<Unit, CloneDialogResult> Clone { get; }
@@ -133,7 +140,11 @@ namespace GitHub.ViewModels.Dialog.Clone
                 EnterpriseTab.Initialize(enterpriseConnection);
             }
 
-            if (connection == enterpriseConnection)
+            if (connection == gitHubConnection)
+            {
+                SelectedTabIndex = 0;
+            }
+            else if (connection == enterpriseConnection)
             {
                 SelectedTabIndex = 1;
             }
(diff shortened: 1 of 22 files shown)


meaghanlewis:
Tested with 2 different GitHub accounts with and without and 2FA and also with the browser sign-in. Also different with 2 Enterprise accounts using a token, username/password and browser sign-in.

jcansdale:
@meaghanlewis Changing to `Sign in` instead of `Login` is a great idea. It means we can reuse translations from the connect to GitHub view. 😄 

image

meaghanlewis:
@jcansdale I really like the new placement of the `Sign in` link. That seems more natural.

Now im just considering the user experience and also taking into account the feedback from this work item: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/891245?src=WorkItemMention&src-action=artifact_link. I don't think we need to be as explicit to offer an option to `Sign out` here since multiple accounts can now be managed in the dialog and there is a `Sign out` option with the full extension from Team Explorer.

I do wonder if it's confusing to see a `Sign in` option even after being connected to an account or multiple accounts 🤔. Could something like `Manage accounts` or `Add account` be more appropriate?

jcansdale:
> I do wonder if it's confusing to see a Sign in option even after being connected to an account or multiple accounts 🤔. Could something like Manage accounts or Add account be more appropriate?

I agree it might be a little confusing. If/when we change it to a drop down list like Azure DevOps, we should definitely change it to something like `Add an account`.

image

I wonder if this would work for the current setup? My only concern at this point is the need to do translations. We might be able to scrape them from the Azure DevOps dialog somehow (though I'm not sure how).

src/GitHub.App/ViewModels/Dialog/Clone/RepositoryCloneViewModel.cs
```diff
@@ -133,7 +140,11 @@ public async Task InitializeAsync(IConnection connection)
                 EnterpriseTab.Initialize(enterpriseConnection);
             }
 
-            if (connection == enterpriseConnection)
+            if (connection == gitHubConnection)
+            {
+                SelectedTabIndex = 0;
```
StanleyGoldman: Curious, are you adding this for the sake of the reading programmer? or is this function called multiple times now?

src/GitHub.App/ViewModels/Dialog/Clone/RepositorySelectViewModel.cs
```diff
@@ -129,7 +129,7 @@ static string GroupName(KeyValuePair<string, IReadOnlyList<RepositoryListItemMod
 
         async Task LoadItems(bool refresh)
         {
-            if (connection == null && !IsLoading) return;
+            if (connection == null || IsLoading) return;
```
StanleyGoldman: Good catch. Thanks.

src/GitHub.App/ViewModels/Dialog/Clone/RepositoryCloneViewModel.cs
StanleyGoldman: After conversation, the function is not re-entrant, it's just left there to be explicit.

StanleyGoldman:
After some conversation we decided that the text `Sign In` should be changed to `Add/Change Account`.
Since `Sign In` is already translated, we are reluctant to change it immediately and are adding `Add/Change Account` to our resources files for translation. This will require a switch in the future.

meaghanlewis:
> Since Sign In is already translated, we are reluctant to change it immediately and are adding Add/Change Account to our resources files for translation. This will require a switch in the future.

I would rather change the text now and then wait for a translation later. We all agree it can be a bit confusing showing `Sign in` when you're already authenticated.

One last suggestion. What if we change it to say `Connect to GitHub` which is what we show in Team Explorer - Connect now. @jcansdale @StanleyGoldman thoughts?
Screen Shot 2019-05-21 at 2 47 37 PM

jcansdale:
@meaghanlewis,

> One last suggestion. What if we change it to say Connect to GitHub which is what we show in Team Explorer - Connect now. @jcansdale @StanleyGoldman thoughts?

Here is what it would look like. I'm reusing the `Resources.LoginTitle` string (from the login title).

image

I've just realized that the `Connect to GitHub` menu command you showed isn't actually translated. 😭  In fact none of our menu commands are currently translated (I'm not sure how best to do this). You can find there [here](https://github.com/github/VisualStudio/blob/5fb2d31673242e57e472a26dc7943e6144bf0034/src/GitHub.VisualStudio/GitHub.VisualStudio.vsct#L82).

Luckily we don't actually surface any menu commends in the essentials extension, but it's something to think about for the full extension/future essentials extension version.

Let me know what you think of the above and whether you think it's an improvement over `Log in`.  It could work as a stand in until we have `Add/Change Account` translated (which I think is most descriptive).

StanleyGoldman:
> I would rather change the text now and then wait for a translation later. We all agree it can be a bit confusing showing Sign in when you're already authenticated.

I'm not sure if my rambling made sense. I added the string "Add/Change Account" to the resources files so it can be translated I just didn't make the change to **_use_** "Add/Change Account".

> One last suggestion. What if we change it to say Connect to GitHub which is what we show in Team Explorer - Connect now. @jcansdale @StanleyGoldman thoughts?

My vote is for "Add/Change Account". I like it over "Connect to GitHub" because the user is already connected to GitHub or GHE and they are going to either add the other account type or change their current account.

meaghanlewis:
@jcansdale thanks for trying out that change! Perhaps it is not an improvement to `Sign in`. I think I agree with you and @StanleyGoldman that `Add/Change Account` makes the most sense.

PR #2360, opened 2019-05-22
Adding Publish to Essentials extension

StanleyGoldman:
Add `Publish to GitHub` section to `Team Explorer - Synchronization` page:

image

When `Publish to GitHub` button is clicked:

image

When user opens the `Team Explorer - Synchronization` page, only `GitHub.VisualStudio.16` will be loaded:

image

### How to test

1. Ensure that only the GitHub Essentials extension is installed
2. Create a new project
3. `Add to Source Control > Git`
4. Click `Publish to GitHub` button
image
5. Choose a server (GitHub or GitHub Enterprise), owner (user or organization) and repository name
6. Check `Private` (if you don't want your test repository to be publicly visible)
7. Click `Publish` button
image
8. Expect `Repository created successfully` message
image
9. Click on `Sync` button
10. Expect no publish sections to be visible
image


src/GitHub.VisualStudio.16/Sync/PublishSection.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// Constructor.
+        /// </summary>
+        public PublishSection()
+            : base()
+        {
+            this.Title = "Publish to GitHub";
+            this.IsExpanded = true;
+            this.IsBusy = false;
+            this.SectionContent = new PublishView();
+            this.View.ParentSection = this;
```
StanleyGoldman: Is this used for anything?

src/GitHub.VisualStudio.16/Sync/PublishView.xaml.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// Parent section.
+        /// </summary>
+        public PublishSection ParentSection
+        {
+            get { return (PublishSection)GetValue(ParentSectionProperty); }
+            set { SetValue(ParentSectionProperty, value); }
+        }
+        public static readonly DependencyProperty ParentSectionProperty =
+            DependencyProperty.Register("ParentSection", typeof(PublishSection), typeof(PublishView));
```
StanleyGoldman: It is a `DependencyProperty`

jcansdale: We can probably get rid of it since I don't think it's used for anything.

meaghanlewis:
I tested this out and it's mainly working as expected. Two things I experienced.

First, the GitHub icon isn't present on the Sync page.
Screen Shot 2019-06-10 at 10 42 28 AM

Second, the `Add/Change Accounts` option does allow me to sign in to a different account but after I signed in with my enterprise account and went back to select where to publish I didn't see the enterprise account as an option.
Screen Shot 2019-06-10 at 10 43 17 AM

jcansdale:
We seem to have an issue with the `Blue` theme.

`Dark` looks fine:

image

`Light` looks fine:

image

`Blue` (the default) looks washed out and grey:

image

I was hoping to avoid special casing the color themes by using the build in `SharpImage` component.

The image used on `Browse a repository` looks fine:

image

We should be using the same image! 😕

src/GitHub.VisualStudio.16/Sync/PublishView.xaml
```diff
@@ -0,0 +1,75 @@
...
+            </Grid.ColumnDefinitions>
+            <Grid.RowDefinitions>
+                <RowDefinition Height="Auto"/>
+                <RowDefinition Height="*"/>
+            </Grid.RowDefinitions>
+            <WrapPanel x:Name="GetStartedPanel" Grid.Row="0" Grid.ColumnSpan="2">
+                <!-- Description text -->
+                <TextBlock    x:Name="Description"
+                              x:Uid="Description"
+                              Text="Powerful collaboration, code review, and code management for open source and private projects."
```
jcansdale: Looks like we have some localization work to do. :-(

jcansdale:
@meaghanlewis,

> First, the GitHub icon isn't present on the Sync page.

This is surprising. I wonder if you somehow has two versions installed (this can happen after the extension name change). How did you install this version for testing?

> Second, the Add/Change Accounts option does allow me to sign in to a different account but after I signed in with my enterprise account and went back to select where to publish I didn't see the enterprise account as an option.

I've fixed it so that after you `Add/Change Accounts`, it will change the drop down list of hosts to the connection you just logged in using. Previously the new connection should have appeared in the list, but you would have had to select it manually.

I hope the unexpected behavior was it not being automatically selected, rather than it not appearing in the list at all.

image

meaghanlewis:
@jcansdale this looks good to me! I'm not experiencing the problems I was previously. Perhaps I didn't notice the account picker before.

One small thing that I noticed was that for GitHub repositories the repo name was autofilled, but for GHE it wasn't.

jcansdale:
> One small thing that I noticed was that for GitHub repositories the repo name was autofilled, but for GHE it wasn't.

I'm guessing the issue is here, but I don't understand exactly what's going wrong:
https://github.com/github/VisualStudio/blame/master/src/GitHub.App/ViewModels/TeamExplorer/RepositoryPublishViewModel.cs#L96

This will have been a long standing issue.

PR #2365, opened 2019-05-24
Fix switching back to Start Window after browser login

### LINKED ISSUE #2364: Sign in with browser doesn't re-activate Start Window on completion
## Versions
 - GitHub Extension for Visual Studio version: All
 - Visual Studio version: 2019

## What happened
1. Open Visual Studio 2019
2. Select `Clone or checkout code`
3. Select `GitHub` icon
4. Select `Sign in`
5. Select `GitHub` tab and `Sign in with your browser`
6. Web browser opened and sign in completed
7. Web browser remains on top (obscuring Start Window and Open from GitHub dialog)

### Expected behavior
After sign in completes Visual Studio should be re-activated and appear on top.

### Additional context
It appears the Main window is collapsed when the Start Window appears. I wasn't able to get the `Activate()` method to work (even calling it on the Start Window). It seems we need to find another way.

jcansdale:
Switching back to Visual Studio wasn't working when logging in using browser from the Start Window. This PR uses the `SetForegroundWindow` native method instead of calling `Activate()` on Visual Studio's main window (which was hidden).

### What this PR does

- Use `SetForegroundWindow` native method to re-activate Visual Studio's window
- Stop possible `InvalidOperationException` when hidden Visual Studio window was activated

### How to test

1. Open Visual Studio 2019
2. Select `Clone or checkout code`
3. Select `GitHub` icon
4. Select `Sign in`
5. Ensure than any browser windows are using the same monitor as Visual Studio
6. Select `GitHub` tab and then `Sign in with your browser`
7. Web browser should open (complete sign in if necessary)
8. The Start Window and `Open from GitHub` should appear on top of the browser (if it doesn't appear on top, its task bar icon should flash)

### Notes

This PR uses `SetForegroundWindow` instead of `SwitchToThisWindow`. `SwitchToThisWindow` is [marked](https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-switchtothiswindow) with the following:
> [This function is not intended for general use. It may be altered or unavailable in subsequent versions of Windows.]

`SetForegroundWindow` is also used internally by the WPF `Window.Activate()` method.

The previous implementation that used `Window.Activate()` would sometimes crash with the following:

```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.InvalidOperationException
   at System.Windows.Window.VerifyHwndCreateShowState()
   at System.Windows.Window.Activate()
   at GitHub.VisualStudio.Views.Dialog.LoginCredentialsView+<>c.<.ctor>b__0_4(GitHub.Models.IConnection)
```

Fixes #2364

### DIFF
--- a/src/GitHub.VisualStudio.UI/Views/Dialog/LoginCredentialsView.xaml.cs
+++ b/src/GitHub.VisualStudio.UI/Views/Dialog/LoginCredentialsView.xaml.cs
@@ -48,7 +48,7 @@ namespace GitHub.VisualStudio.Views.Dialog
             this.WhenAnyObservable(
                 x => x.ViewModel.GitHubLogin.LoginViaOAuth,
                 x => x.ViewModel.EnterpriseLogin.LoginViaOAuth)
-                .Subscribe(_ => SwitchToThisWindow());
+                .Subscribe(_ => SetForegroundWindow());
 
             hostTabControl.SelectionChanged += (s, e) =>
             {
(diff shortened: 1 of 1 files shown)


meaghanlewis:
I tried this out but step 8 doesn't happen for me. The browser still stays on top of the Start Window and Open from GitHub dialog.

PR #2371, opened 2019-05-31
Make the essentials extension a system component

jcansdale:
The essentials extension is currently the only extension that is installed as part of the Core Editor that appears on the extension manager. Marking is as a `SystemComponent` will prevent it from appearing on the extension manager.

image

- Add `SystemComponent=true` setting to `.vsixmanafest`
- Change `DisplayName` to "GitHub Essentials"

### How to test

Testing before insertion

1. Install essentials extension
2. Check that `GitHub` icon appears on `Start Window > Clone or checkout code`
3. Check the essentials extension doesn't appear on `Extensions > Manage Extensions` (it should be empty)

Testing after insertion

1. Install Visual Studio without selecting _any_ workloads (e.g. .NET desktop development)
2. Check that `GitHub` icon appears on `Start Window > Clone or checkout code`
3. Check the essentials extension doesn't appear on `Extensions > Manage Extensions` (it should be empty)


meaghanlewis:
This looks good to me! I followed the `Testing before insertion` steps and this works as expected.

PR #2386, opened 2019-06-13
Be consistent about how we use relative paths and Git paths

### LINKED ISSUE #2380: Inline comment margin doesn't appear when viewing file
## Versions
 - GitHub Extension for Visual Studio version: `2.10.2`
 - Visual Studio version: All

## What happened
1. Open pull request detail view
2. Right-click on PR file and `View File`
3. File opens but comments aren't visible on margin

### Expected behavior
- Comment margin should be visible

### Screenshots
image

### Logs
```
2019-06-10 14:55:30.939 [11504] EROR [01] InlineCommentTagger       Exception caught while executing background task
System.ArgumentException: The value 'C:/Source/github.com/github/VisualStudio/src/GitHub.VisualStudio.16/Sync/PublishView.xaml.cs' must not be rooted
Parameter name: path
   at GitHub.Extensions.Guard.ArgumentIsGitPath(String value, String name) in C:\Source\github.com\github\VisualStudio\src\GitHub.Extensions\Guard.cs:line 21
   at GitHub.Services.GitService.Compare(IRepository repository, String sha1, String sha2, String path) in C:\Source\github.com\github\VisualStudio\src\GitHub.Exports\Services\GitService.cs:line 242
   at GitHub.InlineReviews.Services.DiffService.<Diff>d__2.MoveNext() in C:\Source\github.com\github\VisualStudio\src\GitHub.InlineReviews\Services\DiffService.cs:line 34
```

jcansdale:
In versions of `LibGit2Sharp` we were using prior to #2142, `LibGit2Sharp` didn't care whether paths passed contained `/` or `\` and could cope with absolute as well as working directory relative paths. In recent versions of `LibGit2Sharp`, Git paths must be working directory relative and use `/` as their directory separator. If they're not in this format the operation will silently fail (as if the file simply doesn't exist).

Because GitHub for Visual Studio was developed when `LibGit2Sharp` didn't care about directory separators, we often mixed up relative paths and Git paths (which use '/'). This has created the potential for a whole category of bugs and has made the code difficult to reason about. This PR is attempt to bring a bit of sanity to the situation.

### What this PR does

- Use `relativePath` when a parameter contains a Windows style relative path
- Use `gitPath` when a parameter contains a Git style relative path
- Use `Guard.ArgumentIsRelativePath` to check we're not passing absolute paths
- Automatically convert Windows paths to Git style relative paths before using them
- Factor out conversions into `Paths.ToGitPath` and `Paths.ToWindowsPath`
- Fix #2380 which was caused by passing `AddBufferTag` an absolute path
- Fix issue where `PullRequestSessionFile` used a Git relative path and `PullRequestSessionLiveFile` a Windows relative path
- Fix issue where `GetOldFileName` was using a Windows style relative path

### How to test

Here is how to test one of the bugs it fixes

1. Open a solution that contains pull requests
2. Open the pull request detail view without checking out the PR
3. Right-click on a PR file that has comments and `View File`
4. Check that comments appear on comment margin

Fixes #2380

### DIFF
--- a/src/GitHub.App/Services/PullRequestEditorService.cs
+++ b/src/GitHub.App/Services/PullRequestEditorService.cs
@@ -573,15 +573,13 @@ namespace GitHub.Services
         void AddBufferTag(
             ITextBuffer buffer,
             IPullRequestSession session,
-            string gitPath,
+            string relativePath,
             string commitSha,
             DiffSide? side)
         {
-            Guard.ArgumentIsGitPath(gitPath, nameof(gitPath));
-
             buffer.Properties.GetOrCreateSingletonProperty(
                 typeof(PullRequestTextBufferInfo),
-                () => new PullRequestTextBufferInfo(session, gitPath, commitSha, side));
+                () => new PullRequestTextBufferInfo(session, relativePath, commitSha, side));
 
             var projection = buffer as IProjectionBuffer;
 
@@ -589,7 +587,7 @@ namespace GitHub.Services
             {
                 foreach (var source in projection.SourceBuffers)
                 {
-                    AddBufferTag(source, session, gitPath, commitSha, side);
+                    AddBufferTag(source, session, relativePath, commitSha, side);
                 }
             }
         }
@@ -656,9 +654,10 @@ namespace GitHub.Services
                 session.LocalRepository,
                 session.PullRequest))
             {
-                var fileChange = changes.FirstOrDefault(x => x.Path == file.RelativePath);
+                var gitPath = file.RelativePath.Replace(Path.DirectorySeparatorChar, '/');
+                var fileChange = changes.FirstOrDefault(x => x.Path == gitPath);
                 return fileChange?.Status == LibGit2Sharp.ChangeKind.Renamed ?
-                    fileChange.OldPath : file.RelativePath;
+                    fileChange.OldPath.Replace('/', Path.DirectorySeparatorChar) : file.RelativePath;
             }
         }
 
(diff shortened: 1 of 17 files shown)


src/GitHub.App/Services/PullRequestEditorService.cs
```diff
@@ -119,7 +120,7 @@ public class PullRequestEditorService : IPullRequestEditorService
 
                     if (!workingDirectory)
                     {
-                        AddBufferTag(wpfTextView.TextBuffer, session, fullPath, commitSha, null);
+                        AddBufferTag(wpfTextView.TextBuffer, session, relativePath, commitSha, null);
```
jcansdale: This is the fix for #2380. `AddBufferTag` expects a relative path.

src/GitHub.Exports/Primitives/Paths.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            return relativePath.Replace(Path.DirectorySeparatorChar, GitDirectorySeparatorChar);
+        }
+
+        /// <summary>
+        /// Convert from a Git path to a relative path.
+        /// </summary>
+        /// <param name="gitPath">A working directory relative path which uses the '/' directory separator.</param>
+        /// <returns>A relative paht that uses the <see cref="Path.DirectorySeparatorChar"/> directory separator.</returns>
+        public static string ToRelativePath(string gitPath)
```
jcansdale: `ToWindowsPath` would be clearer.

PR #2388, opened 2019-06-14
Re-enabling BuildCrossCheck

StanleyGoldman:


jcansdale:
We're no longer using AppVeyor.

PR #2395, opened 2019-06-27
Refactoring RepositoryForm Validators

StanleyGoldman:


src/GitHub.App/ViewModels/TeamExplorer/RepositoryPublishViewModel.cs
```diff
@@ -93,17 +93,17 @@ public class RepositoryPublishViewModel : RepositoryFormViewModel, IRepositoryPu
...
-                .Where(x => RepositoryNameValidator.ValidationResult != null && SafeRepositoryNameWarningValidator.ValidationResult != null)
-                .Subscribe(async _ =>
-                {
-                    var name = RepositoryName;
-                    RepositoryName = null;
-                    await RepositoryNameValidator.ResetAsync();
-                    await SafeRepositoryNameWarningValidator.ResetAsync();
-                    RepositoryName = name;
-                });
+//            this.WhenAny(x => x.SelectedConnection, x => x.SelectedAccount,
```
jcansdale: Comment-tastic. ;-)

src/GitHub.App/ViewModels/TeamExplorer/RepositoryPublishViewModel.cs
```diff
@@ -173,20 +173,22 @@ IObservable<ProgressState> OnPublishRepository()
...
-                .Add(repoName =>
+            var nameValidationConditions = this.WhenAny(model => model.RepositoryName,
+                    model => model.SelectedConnection,
+                    model => model.SelectedAccount,
+                    (repositoryName, connection, account) => (repositoryName: repositoryName.Value,
+                        connection: connection.Value, account: account.Value))
+                .Where(tuple => tuple.repositoryName != null);
+
+            RepositoryNameValidator = ReactivePropertyValidator.ForObservable(nameValidationConditions)
+                .IfTrue(tuple => string.IsNullOrEmpty(tuple.repositoryName), Resources.RepositoryNameValidatorEmpty)
```
jcansdale: Is this redundant given the `.Where(tuple => tuple.repositoryName != null);` above?

src/GitHub.Exports.Reactive/ViewModels/IRepositoryForm.cs
```diff
@@ -25,8 +25,8 @@ public interface IRepositoryForm : IViewModel
         /// dashes.
         /// </summary>
         string SafeRepositoryName { get; }
-        ReactivePropertyValidator<string> RepositoryNameValidator { get; }
-        ReactivePropertyValidator<string> SafeRepositoryNameWarningValidator { get; }
+        ReactivePropertyValidator<(string repositoryName, IConnection connection, IAccount account)> RepositoryNameValidator { get; }
```
jcansdale: Looking at the code, I can't see where we're using the `connection` or `account` for validation. I wonder if once upon a time we were checking for an existing repository with the same name as part of the validation? It looks like this check is now done and the error surfaced when the user attempts to create the repository.

I think we might be able to simplify this to use just the `repositoryName`. I'm wondering if the buggy code was actually completely obsolete. ;-)

src/GitHub.App/ViewModels/TeamExplorer/RepositoryPublishViewModel.cs
jcansdale: done 😄

jcansdale: done!

jcansdale: done!

RobbyLena (approved):

PR #2426, opened 2019-10-03
Use only VS Telemetry in Essentials extension

jcansdale:
The Essentials extension doesn't include an options pane, so there is no way to opt out of GitHub telemetry. Change to use only the built in VS Telemetry which has a global opt out via `Help > Send Feedback >  Settings...`.

### What this PR does

- Use `VisualStudioUsageTracker` instead of `UsageTracker` for telemetry
- Pass a `Lazy<IConnectionManager>` to `VisualStudioUsageTracker` so we don't need to create a live object when initializing MEF

### Testing

I've checked that the `Open from GitHub` and `Publish to GitHub` functionality still works.

### DIFF
--- a/src/GitHub.VisualStudio.16/CompositionServices.cs
+++ b/src/GitHub.VisualStudio.16/CompositionServices.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.ComponentModel.Composition;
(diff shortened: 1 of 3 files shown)


jcansdale:
I'll link back to this from the VS institution PR. 😄

jcansdale:
I'll link back to this from the VS institution PR. 😄

PR #2436, opened 2019-10-23
Make GitHub.OpenFromUrl command use active repository when possible

jcansdale:
At the moment `GitHub.OpenFromUrl` always shows the open/clone dialog, even when the target repository is already active. This disrupts the workflow when opening a file in the active workspace.

## What this PR does

- Check if the target repository host/owner/repositoryName matches the active repository
- Attempt to navigate inside the active workspace when the repositories match
- Only show the open/clone when the active repository is different to the target repository

## How to test

1. Open the command window
2. Enter
```
> GitHub.OpenFromUrl https://github.com/grokys/PullRequestSandbox
```
3. Open/clone dialog should appear
4. Clone the repository
5. Enter
```
> GitHub.OpenFromUrl https://github.com/grokys/PullRequestSandbox/blob/master/PullRequestSandbox/Program.cs#L11-L14
```
6. The `Program.cs` file should open (without showing the Open/Clone dialog)
7. Enter
```
> GitHub.OpenFromUrl https://github.com/github/VisualStudio
```
8. The Open/Clone dialog should appear

### DIFF
--- a/src/GitHub.VisualStudio/Commands/OpenFromUrlCommand.cs
+++ b/src/GitHub.VisualStudio/Commands/OpenFromUrlCommand.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.ComponentModel.Composition;
 using GitHub.Commands;
+using GitHub.Models;
+using GitHub.Primitives;
 using GitHub.Services;
 using GitHub.Services.Vssdk.Commands;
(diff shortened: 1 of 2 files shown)


jcansdale:
Ran through the `How to test` steps and it seems to work. 🎈  LGTM

PR #2493, opened 2020-04-07
Update publishing-an-existing-project-to-github.md

BulldogCTISolutions:
Hindsight is asking me to include a step before starting to clean out and remove unwanted sub-directories.

### DIFF
--- a/docs/using/publishing-an-existing-project-to-github.md
+++ b/docs/using/publishing-an-existing-project-to-github.md
@@ -1,5 +1,6 @@
 # Publishing an existing project to GitHub
 
+0. Clean solution directories, bin, obj, etc. or else these will be uploaded to Github
 1. Open a solution in Visual Studio.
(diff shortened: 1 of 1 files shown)


docs/using/publishing-an-existing-project-to-github.md
```diff
@@ -1,5 +1,6 @@
 # Publishing an existing project to GitHub
 
+0. Clean solution directories, bin, obj, etc. or else these will be uploaded to Github
```
Kryptos-FR, line 3: That should be covered with a proper `.gitignore`. I think it would be better to mention that instead of a step that just state WHAT but doesn't explain WHY, nor HOW.

So maybe a note or warning "make sure to have properly set your `.gitignore` before adding the project" witha  link to GitHub's page about `.gitignore` (i.e. https://docs.github.com/en/github/using-git/ignoring-files).

PR #2512, opened 2020-05-07
Prevent error when filenames have braces in them

### LINKED ISSUE #2510: Can't open files containing curly braces
## Versions
 - GitHub Extension for Visual Studio version: 2.11.96.24460
 - Visual Studio version: 16.5.4

## What happened
Trying to open files from the GitHub PR explorer for [this PR](https://github.com/windows-toolkit/WindowsCommunityToolkit/pull/3128). It contains files with curly braces `{}` which won't open in VS. Getting a `Error opening file: Input string was not in a correct format.` error.

### Steps to Reproduce
1. Open [PR](https://github.com/windows-toolkit/WindowsCommunityToolkit/pull/3128) in GitHub extension.
2. Try to open any of the generic classes like `ArrayPoolBufferWriter{T}.cs`
3. See error in status bar
4. Note, if you pick another file (like the first one `AllocationMode.cs` it works fine, just fails on any file with curly brace).

### Expected behavior
File opens.

### Screenshots
image

davidwengier:
Fixes #2510

As per [the documentation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.interop.ivsdifferenceservice.opencomparisonwindow2?view=visualstudiosdk-2019) the OpenComparisonWindow2 method passes captions and tooltips through `string.Format` so braces break it.

> The caption, tooltip, and label parameters use C#-style string formatting where {0} corresponds to the name of the left file and {1} corresponds to the name of the right file.

The existing captions have more detail than using the inbuilt `{0}` and `{1}` can provide, particularly as temporary files are used, so there is no option but to escape the braces. I reasoned that since this code is trigger from a UI gesture by the user the perf wasn't super critical so am just doing some lazy replaces. Let me know if thats an issue.


jcansdale:
Hey @michael-hawker 👋 Have you had a chance to try this build? 😄

jcansdale:
Confirmed fix!

image

Thanks @davidwengier 🙏

PR #2515, opened 2020-05-19
Change GitHub.TeamFoundation.16 version of TeamExplorerServices

daschult:
Change GitHub.TeamFoundation.16 version of TeamExplorerServices to use IGitActionsExt2.InitializeOrPushRepositoryToGitService().

This allows the GitHub Extension to take advantage of the Create Repository dialog available in the new Git experience in Visual Studio 16.6.


src/GitHub.TeamFoundation.14/Services/TeamExplorerServices.cs
```diff
@@ -66,10 +67,8 @@ public void ShowHomePage()
 
         public void ShowPublishSection()
         {
-            var te = serviceProvider.TryGetService<ITeamExplorer>();
-            var page = te.NavigateToPage(new Guid(TeamExplorerPageIds.GitCommits), null);
-            var publish = page?.GetSection(new Guid(GitHubPublishSection.GitHubPublishSectionId)) as GitHubPublishSection;
-            publish?.Connect();
+            IGitActionsExt2 gitActionsExt = serviceProvider.TryGetService<IGitActionsExt2>();
+            gitActionsExt?.InitializeOrPushRepositoryToGitService();
```
daschult, line 71: I only want these two lines to be executed in the 16.0 scenario since this type won't be available in 14.0 or 15.0.

src/GitHub.TeamFoundation.14/Services/TeamExplorerServices.cs
```diff
@@ -66,10 +67,15 @@ public void ShowHomePage()
 
         public void ShowPublishSection()
         {
+#if TEAMEXPLORER14 || TEAMEXPLORER15
             var te = serviceProvider.TryGetService<ITeamExplorer>();
             var page = te.NavigateToPage(new Guid(TeamExplorerPageIds.GitCommits), null);
             var publish = page?.GetSection(new Guid(GitHubPublishSection.GitHubPublishSectionId)) as GitHubPublishSection;
             publish?.Connect();
+#else
+            IGitActionsExt2 gitActionsExt = serviceProvider.TryGetService<IGitActionsExt2>();
```
jcansdale, line 76: Will we need to avoid attempting to load the `IGitActionsExt2` on pre-Git experience versions of VS 2019?

We could do something like this:

```cs
        public void ShowPublishSection()
        {
#if TEAMEXPLORER16
            // Only call InitializeOrPushRepositoryToGitService when IGitActionsExt2 exists
            if (FindIGitActionsExt2() is object)
            {
                InitializeOrPushRepositoryToGitService();
                return;
            }
#endif

            var te = serviceProvider.TryGetService<ITeamExplorer>();
            var page = te.NavigateToPage(new Guid(TeamExplorerPageIds.GitCommits), null);
            var publish = page?.GetSection(new Guid(GitHubPublishSection.GitHubPublishSectionId)) as GitHubPublishSection;
            publish?.Connect();
        }

#if TEAMEXPLORER16
        static Type FindIGitActionsExt2()
        {
            var type = typeof(IGitActionsExt);
            var name = $"{type.FullName}2";
            return type.Assembly.GetType(name, false);
        }

        void InitializeOrPushRepositoryToGitService()
        {
            IGitActionsExt2 gitActionsExt = serviceProvider.TryGetService<IGitActionsExt2>();
            gitActionsExt?.InitializeOrPushRepositoryToGitService();
        }
#endif
```

daschult: I'm fine with this. I'll make the change.

daschult: Done.

daschult:
For what it's worth, I just verified that this works as expected in the latest preview version of Visual Studio 16.

yannduran:
@jcansdale 

Before I update to this version, what will happen for those who choose to NOT use the so-called *new Git experience*? I for one find it incomplete and impossible to even consider using so far.

daschult:
If you choose not to use the new Git experience, the GitHub Visual Studio extension will direct you to Team Explorer's Sync page, just like it used to.

yannduran:
@daschult Excellent, thank you!

daschult:
@yannduran Also, I should've asked before, but what features do you find incomplete in the new Git experience? We certainly have our own list of features that we're working on, but if you could provide your own list of missing features that make the new Git experience unusable, then that would help us to prioritize. 😄 Thanks!

yannduran:
@daschult I find it incomplete compared to *Team Explorer*. Although *Team Explorer* could benefit from some improvements, many experienced developers, myself included, like the way it works on the whole. 

I've only tried the *new Git Experience* a couple of times, because it was so hard to get anything done that I immediately had to switch back to using *Team Explorer*. Some features were even ***removed*** from *Team Explorer* which really annoyed me. The features in *Team Explorer* should not have been removed or changed in any way. 

How are we expected to test your new window out when ***switching over to it*** means that we're expected to have to hunt all over Visual Studio to find "other ways" to do **basic** things that were all together in *Team Explorer*. Why does it always have to be so **binary**, one or the other? It would have been far easier if you just made the new window available, without crippling *Team Explorer*, and see how developers either used it or didn't use it.

I actually hope that *Team Explorer* will **remain as an option** and not be replaced by something that is dumbed down and makes us less productive. The recent trend of **replacing** features that *just work* with new "prettier" but **less productive** "new experiences" is annoying and completely flawed. 

I guess I'm just getting sick of these "new experiences" being forced on us when they actually make **professional** developers less productive. Visual Studio is slowly being crippled by them (an **unmodifiable** new *Start Window*, a new **search-based** *New Project* dialog. and now the new "*Git* window").

JBell-Tech:
I do not have any issues mainly because I am just here tinkering. I have no earthly idea what I am doing. So far I am in full support of the community.


Respectfully,
Bell Capital Won<https://www.bellcapitalwon.com/>
James Bell
Questions Email me!<mailto:info@bellcapitalwon.com>
7738 New Sandy Hill Church Road
Bailey, NC 27807
(833) 977-7733

PR #2642, opened 2021-10-23
Updating VS Market place content to support VS2022 launch

Taysser-Gherfal:
This PR proposes an updated content that supports VS2022 launch and clarifies that the GitHub extension is not needed for newer versions of Visual Studio.


3nth:
Unless I'm mistaken, this extension still has at least one feature that is missing from 2022, PR reviews. Very handy to be able to review a PR in visual studio, make comments, etc.

For me, that was probably the best feature of this extension, as I typically just use the git command line for creating branches, etc.

janecosham:
I'm also looking for the PR review functionality - where do I find that in VS 2022?

ctolkien:
Likewise.. came looking for the PR Reviews feature... which is no longer available.

PR #2709, opened 2022-12-29
Remove reference to deprecated support email

djdefi:
Support should be initiated via the GitHub Support Portal.


ProKashif (commented):
Why did you change this?

ProKashif (approved):

